PR #17, opened 2024-12-12
Add Mutating Webhooks for PodGangSet

ritikasrivastava:
Defaulting webhook to add defaults to PodGangSet CR is added in this PR.

### DIFF
--- a/operator/internal/webhook/pgs/mutation/podgangset.go
+++ b/operator/internal/webhook/pgs/mutation/podgangset.go
@@ -79,6 +79,10 @@ func defaultPodCliqueTemplateSpecs(cliqueSpecs []v1alpha1.PodCliqueTemplateSpec)
 		// default PodSpec
 		defaultPodSpec(&cliqueSpec.Spec.Spec)
 
+		if cliqueSpec.Spec.Replicas < 1 {
+			cliqueSpec.Spec.Replicas = 1
+		}
+
 		if cliqueSpec.Spec.ScaleConfig != nil {
 			*cliqueSpec.Spec.ScaleConfig.MinReplicas = 1
 		}


operator/internal/webhook/pgs/mutation/handler.go
```diff
@@ -0,0 +1,78 @@
...
+func NewHandler(mgr manager.Manager) (*MutatingHandler, error) {
+	return &MutatingHandler{
+		client:  mgr.GetClient(),
+		decoder: admission.NewDecoder(mgr.GetScheme()),
+		logger:  mgr.GetLogger(),
+	}, nil
+}
+
+// Handle adds defaults to PodGangSet resource.
+func (h *MutatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
```
unmarshall: There is a very simple alternative. We should just leverage `CustomDefaulter`
https://github.com/kubernetes-sigs/controller-runtime/blob/1ec7c1b76e98efd4f177eb8cd96050596b0fa3d3/pkg/webhook/admission/defaulter_custom.go#L35

Once you do that you can then do the following:

- You can remove the need to have a `decoder`
- Just need to implement `func Default(ctx context.Context, obj runtime.Object)` where you already get the parsed runtime.Object.

operator/internal/webhook/pgs/mutation/handler.go
```diff
@@ -0,0 +1,78 @@
...
+	v1 "k8s.io/api/admission/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/NVIDIA/grove/operator/api/core/v1alpha1"
+)
+
+// MutatingHandler is a handler for mutating PodGangSet resources.
+type MutatingHandler struct {
```
unmarshall: Just rename this to `Handler`. If we ever need another mutating webhook then we can rename this to `DefaultingHandler`

operator/internal/webhook/pgs/mutation/handler.go
```diff
@@ -0,0 +1,78 @@
...
+// Handle adds defaults to PodGangSet resource.
+func (h *MutatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	log := h.logger.WithValues("name", req.Name, "namespace", req.Namespace, "resource", fmt.Sprintf("%s/%s", req.Kind.Group, req.Kind.Kind), "operation", req.Operation, "user", req.UserInfo.Username)
+	log.V(1).Info("Grove mutation webhook invoked")
+
+	return h.mutate(req)
+}
+
+func (h *MutatingHandler) mutate(req admission.Request) admission.Response {
+	switch req.Operation {
```
unmarshall: you do not need to switch on `req.Operation` once you start to use `CustomDefaulter`

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,73 @@
...
+
+package mutation
+
+import (
+	"github.com/NVIDIA/grove/operator/api/core/v1alpha1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// mutatePodGangSet adds defaults to a PodGangSetSpec.
+func mutatePodGangSet(pgs *v1alpha1.PodGangSet) {
```
unmarshall: instead of using a prefix of `mutate` a semantically more appropriate name would be `default`

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// default NetworkPackStrategy
+	if spec.NetworkPackStrategy == nil {
+		*spec.NetworkPackStrategy = v1alpha1.BestEffort
+	}
+}
+
+func mutateUpdateStrategy(obj *v1alpha1.GangUpdateStrategy) {
+	//default UpdateStrategy
+	if obj.Type == "" {
```
unmarshall: can you use `utils.IsEmptyStringType` instead?

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,73 @@
...
+		obj.Type = v1alpha1.GangUpdateStrategyRecreate
+	}
+
+	//default RollingUpdateConfig
+	if obj.RollingUpdateConfig.MaxSurge == nil {
```
unmarshall: This should be defaulted only when update strategy is `RollingUpdate` right?

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,73 @@
...
+		*spec.RestartPolicy = v1alpha1.GangRestartPolicyAlways
+	}
+
+	// default NetworkPackStrategy
+	if spec.NetworkPackStrategy == nil {
+		*spec.NetworkPackStrategy = v1alpha1.BestEffort
+	}
+}
+
+func mutateUpdateStrategy(obj *v1alpha1.GangUpdateStrategy) {
```
unmarshall: can you choose better parameter names instead of using `obj`? This applies elsewhere as well

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,73 @@
...
+	if obj.Type == "" {
+		obj.Type = v1alpha1.GangUpdateStrategyRecreate
+	}
+
+	//default RollingUpdateConfig
+	if obj.RollingUpdateConfig.MaxSurge == nil {
+		*obj.RollingUpdateConfig.MaxSurge = intstr.FromInt(1)
+	}
+
+	if obj.RollingUpdateConfig.MaxUnavailable == nil {
```
unmarshall: This should be defaulted only when update strategy is RollingUpdate right?

operator/internal/webhook/pgs/mutation/podClique.go
```diff
@@ -0,0 +1,29 @@
...
+// */
+
+package mutation
+
+import (
+	"github.com/NVIDIA/grove/operator/api/core/v1alpha1"
+)
+
+// mutatePodCliqueTemplateSpec adds defaults to PodTemplateSpec.
+func mutatePodCliqueTemplateSpec(cliqueSpecs []v1alpha1.PodCliqueTemplateSpec) {
```
unmarshall: do you really need a separate file for this?

operator/internal/webhook/pgs/mutation/pod.go
```diff
@@ -0,0 +1,33 @@
...
+// */
+
+package mutation
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// mutatePodTemplateSpec adds defaults to PodSpec.
+func mutatePodTemplateSpec(spec *corev1.PodTemplateSpec) {
```
unmarshall: do you really need a separate file for this?

operator/internal/webhook/pgs/mutation/pod.go
```diff
@@ -0,0 +1,33 @@
...
+	mutatePodSpec(spec.Spec)
+}
+
+func mutatePodSpec(spec corev1.PodSpec) {
```
unmarshall: merge this function with `mutatePodTemplateSpec` if and when it grows then you can refactor.

operator/internal/webhook/pgs/mutation/pod.go
```diff
@@ -0,0 +1,33 @@
...
+	// default TerminationGracePeriodSeconds
+	if spec.TerminationGracePeriodSeconds == nil {
```
unmarshall: if the RestartPolicy is not set then that should be set to `Never`

ritikasrivastava: As per the table in the [doc](https://docs.google.com/document/d/1lX0dc5ZwVgAq3uVblmmtYQTK8TiyZjme6wEkAG9ihw0/edit?usp=sharing), defaulting RestartPolicy in PodSpec should not be allowed. Should we reconsider it? 
If we do default, should it be Always (as [indicated](https://github.com/kubernetes/api/blob/bc0564c95e7494dc33c8f4c149f4daad616066eb/core/v1/types.go#L3826)) or Never. If Never, then it would defer from the default PodGangTemplateSpec.RestartPolicy which is always.

operator/internal/webhook/pgs/mutation/handler.go
ritikasrivastava: Done.

operator/internal/webhook/pgs/mutation/handler.go
ritikasrivastava: Done.

operator/internal/webhook/pgs/mutation/podgangset.go
ritikasrivastava: Done.

operator/internal/webhook/pgs/mutation/podClique.go
ritikasrivastava: Removed podClique and pod.go. As we expand, can refactor at that point.

operator/internal/webhook/pgs/mutation/pod.go
ritikasrivastava: Done

operator/internal/webhook/pgs/mutation/podgangset.go
ritikasrivastava: Done

operator/internal/webhook/pgs/mutation/podgangset.go
ritikasrivastava: Done

operator/internal/webhook/pgs/mutation/handler.go
```diff
@@ -0,0 +1,52 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// Handler struct sets default values on PodGangSet CR
+type Handler struct {
+	logger logr.Logger
+}
+
+func NewHandler() *admission.Webhook {
```
unmarshall: Since you no longer have other fields in the handler perhaps you can remove this function altogether.

operator/internal/webhook/pgs/mutation/register.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const (
+	// HandlerName is the name of the default webhook handler for PodGangSet.
+	HandlerName = "podgangset-default-webhook"
+	webhookPath = "/webhooks/default-podgangset"
+)
+
+// RegisterWithManager registers the webhook with the manager.
+func RegisterWithManager(mgr manager.Manager) error {
```
unmarshall, line 32: You can instead have the following:
```go
func RegisterWithManager(mgr manager.Manager) error {
  webhook := admission.
     WithCustomDefaulter(mgr.GetScheme(), 
     &v1alpha1.PodGangSet{},
     Handler { logger: mgr.GetLogger().WithName("webhook").WithName(HandlerName) }).
    WithRecoverPanic(true)
  
  mgr.GetWebhookServer().Register(webhookPath, webhook)
  return nil
 }
```

operator/internal/webhook/pgs/mutation/handler.go
ritikasrivastava: Done

operator/internal/webhook/pgs/mutation/register.go
ritikasrivastava: Done

operator/internal/webhook/pgs/mutation/pod.go
ritikasrivastava: Set to Never.

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,96 @@
...
+	defaultPodGangSetSpec(&pgs.Spec)
+}
+
+// defaultPodGangSetSpec adds defaults to the specification of a PodGangSet.
+func defaultPodGangSetSpec(spec *v1alpha1.PodGangSetSpec) {
+	// default PodGangTemplateSpec
+	defaultPodGangTemplateSpec(&spec.Template)
+
+	// default UpdateStrategy
+	defaultUpdateStrategy(spec.UpdateStrategy)
```
unmarshall, line 41: The update strategy has been removed now, can you adapt?

operator/internal/webhook/pgs/mutation/handler.go
```diff
@@ -0,0 +1,47 @@
...
+
+	pgs, ok := obj.(*v1alpha1.PodGangSet)
+	if !ok {
+		return fmt.Errorf("expected an PodGangSet object but got %T", obj)
+	}
+	d.addDefaults(pgs)
+	return nil
+}
+
+func (Handler) addDefaults(pgs *v1alpha1.PodGangSet) {
```
unmarshall: no need of `addDefaults` you can just directly call `defaultPodGangSet` from `Default` method on the `Handler`

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,96 @@
...
+		*spec.RestartPolicy = v1alpha1.GangRestartPolicyAlways
+	}
+
+	// default NetworkPackStrategy
+	if spec.NetworkPackStrategy == nil {
+		*spec.NetworkPackStrategy = v1alpha1.BestEffort
+	}
+}
+
+func defaultUpdateStrategy(updateStrategy *v1alpha1.GangUpdateStrategy) {
```
unmarshall, line 64: This function is no longer required

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,96 @@
...
+			*updateStrategy.RollingUpdateConfig.MaxSurge = intstr.FromInt(1)
+		}
+
+		if updateStrategy.RollingUpdateConfig.MaxUnavailable == nil {
+			*updateStrategy.RollingUpdateConfig.MaxUnavailable = intstr.FromInt(1)
+		}
+	}
+}
+
+func defaultPodCliqueTemplateSpec(cliqueSpecs []v1alpha1.PodCliqueTemplateSpec) {
```
unmarshall: since it takes a slice of cliques shouldn't it be called `defaultPodCliqueTemplateSpecs`?

operator/internal/webhook/pgs/mutation/podgangset.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, cliqueSpec := range cliqueSpecs {
```
unmarshall, line 78: For every PodClique, can replicas ever be 0? This is only possible if we need to support scale down to 0. Lets discuss this point in the next sync.

operator/internal/webhook/pgs/mutation/podgangset.go
ritikasrivastava: GangUpdateStrategy struct still exists. Only the Type was removed.

operator/internal/webhook/pgs/mutation/podgangset.go
ritikasrivastava: Same. The RollingUpdateConfig still exists.

operator/internal/webhook/pgs/mutation/handler.go
ritikasrivastava: Done

operator/internal/webhook/pgs/mutation/podgangset.go
ritikasrivastava: Done

operator/internal/webhook/pgs/mutation/podgangset.go
unmarshall: Ah you are right, my bad.

operator/internal/webhook/pgs/mutation/podgangset.go
unmarshall: My bad.

PR #27, opened 2025-01-22
implement validation for update operation

dmitsh:


operator/internal/webhook/pgs/validation/podgangset.go
```diff
@@ -312,3 +320,85 @@ func validateScaleConfig(scaleConfig *v1alpha1.AutoScalingConfig, fldPath *field
...
+	//  `spec.activeDeadlineSeconds`,
+	//  `spec.tolerations` (only additions to existing tolerations),
+	//  `spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)
+
+	if len(newSpec.Tolerations) < len(oldSpec.Tolerations) || !reflect.DeepEqual(oldSpec.Tolerations, newSpec.Tolerations[:len(oldSpec.Tolerations)]) {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("tolerations"), "not allowed to change immutable pod fields"))
+	}
+
+	if !(*oldSpec.TerminationGracePeriodSeconds < 0 && *newSpec.TerminationGracePeriodSeconds == 1) {
+		allErrs = append(allErrs, field.Forbidden(fldPath.Child("terminationGracePeriodSeconds"), "not allowed to change immutable pod fields"))
```
unmarshall: can we change the error msg?
For this field its not really immutable. It allows changing the value (setting it to 1 when previous value was negative). The error msg should reflect that.

operator/internal/webhook/pgs/validation/podgangset.go
```diff
@@ -312,3 +320,85 @@ func validateScaleConfig(scaleConfig *v1alpha1.AutoScalingConfig, fldPath *field
...
+	}
+
+	// hide mutable fields
```
unmarshall, line 390: Is this correct?
You can change the image for any container/init-container but if other fields in Containers or InitContainers are changed then that is not really allowed. So emptying out the entire Containers and InitContainers is not correct.

PR #28, opened 2025-01-22
update podgangset crd

dmitsh:


operator/api/core/v1alpha1/podgangset.go
```diff
@@ -145,8 +145,6 @@ const (
 
 // GangUpdateStrategy defines the strategy to be used when updating a PodGang.
 type GangUpdateStrategy struct {
-	// Type is the type of update strategy.
-	Type GangUpdateStrategyType `json:"type"`
```
ritikasrivastava, line 149: Why are we removing the type? Would it always be rolling then?

dmitsh: correct, only one type of update

PR #48, opened 2025-03-13
implement pclq status update

dmitsh:


operator/internal/controller/podclique/reconcilespec.go
```diff
@@ -37,6 +44,167 @@ func (r *Reconciler) ensureFinalizer(ctx context.Context, logger logr.Logger, pc
 }
 
 func (r *Reconciler) syncPodCliqueResources(ctx context.Context, logger logr.Logger, pclq *v1alpha1.PodClique) ctrlcommon.ReconcileStepResult {
```
ritikasrivastava, line 55: Should PodClique reconcile also follow the pattern operator.Sync/Delete?

dmitsh: correct, implemented

PR #50, opened 2025-03-20
Ritika/headlessservice

ritikasrivastava:


operator/internal/component/pgs/service/service.go
```diff
@@ -19,8 +25,76 @@ func New(client client.Client) component.Operator[v1alpha1.PodGangSet] {
...
 
-func (_ _resource) Sync(ctx context.Context, logger logr.Logger, pgs *v1alpha1.PodGangSet) error {
-	//TODO implement me
+const (
+	min_port   = 30000
+	max_port   = 50000
+	port_range = 20001
+)
+
+func randomPortGenerator() (int32, error) {
```
dmitsh: This is a shorter version of a function with the same functionality:
```
func getAvailablePort() (int, error) {
	listener, err := net.Listen("tcp", ":0")
	if err != nil {
		return 0, err
	}
	defer listener.Close()

	return listener.Addr().(*net.TCPAddr).Port, nil
}
```

unmarshall: This is no longer required as we are not going to set Port and TargetPort for the headless service.

operator/internal/component/pgs/service/service.go
```diff
@@ -19,8 +25,76 @@ func New(client client.Client) component.Operator[v1alpha1.PodGangSet] {
...
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%d-service", pgs.Name, replicaID),
+			Namespace: pgs.Namespace,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				v1alpha1.LabelPartOfKey: pgs.Name,
+			},
+			Ports: []corev1.ServicePort{{
```
unmarshall: you can remove the Port from the Spec.
Can you also add `PublishNotReadyAddresses` based on the changes in the API that we discussed?

operator/internal/component/pgs/service/service.go
```diff
@@ -19,8 +25,76 @@ func New(client client.Client) component.Operator[v1alpha1.PodGangSet] {
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r _resource) Sync(ctx context.Context, logger logr.Logger, pgs *v1alpha1.PodGangSet) error {
+	for replicaID := range pgs.Spec.Replicas {
+		err := r.createHeadlessService(ctx, pgs, replicaID)
```
unmarshall: In sync you should use `CreateOrPatch` so that it takes care of both create and update scenarios. You can see other components as reference.

ritikasrivastava: Done

PR #61, opened 2025-04-11
Renamed PodClique to PodGroup in scheduler-api

sanjaychatterjee:


scheduler-api/api/v1alpha1/podgang.go
```diff
@@ -34,13 +34,16 @@ type PodGangList struct {
...
+	// TerminationDelay is a delay timer that activates gang termination of a running PodGang
+	// whenever the number of running pods violates the minReplicas constraint of any PodGroup.
+	TerminationDelay metav1.Duration `json:"terminationDelay"`
 }
 
-// MemberClique defines a set of members in a PodGang that share the same PodTemplateSpec.
-type MemberClique struct {
-	// PodReferences is a list of references to the Pods that are part of this member set.
+// PodGroup defines a set of pods in a PodGang that share the same PodTemplateSpec.
+type PodGroup struct {
```
unmarshall, line 45: There is term is now so common, everyone seems to use this term.

scheduler-api/api/v1alpha1/podgang.go
```diff
@@ -34,13 +34,16 @@ type PodGangList struct {
 
 // PodGangSpec defines the specification of a PodGang.
 type PodGangSpec struct {
-	// MemberCliques is a list of member cliques in the PodGang.
-	MemberCliques []MemberClique `json:"memberSet"`
+	// Groups is a list of member pod groups in the PodGang.
+	Groups []PodGroup `json:"groups"`
```
unmarshall: If you want to rename PodGroups instead of Groups if you wish to use `PodGroup` type.

scheduler-api/api/v1alpha1/podgang.go
sanjaychatterjee: Yes, it is kind of getting standardized.

scheduler-api/api/v1alpha1/podgang.go
sanjaychatterjee: Done.

PR #78, opened 2025-05-13
Introduced PodCliqueScalingGroup and refactored API modules

unmarshall:
This PR introduces the following changes:

- Introduces `PodCliqueScalingGroup` custom resource.
- Adds `PodCliqueScalingGroupConfig` to `PodGangSet.Spec.TemplateSpec.PodCliqueScalingGroupConfigs`
- Moves Operator API to its own go module.
- Adjusts the Makefile targets.
- Adapts the Operator code to use the new Operator API module.
- Changed the structure of scheduler-api go module.
- Regenerated code.


operator/internal/controller/podgangset/register.go
```diff
@@ -37,7 +37,7 @@ func (r *Reconciler) RegisterWithManager(mgr manager.Manager) error {
 	return builder.ControllerManagedBy(mgr).
 		Named(controllerName).
 		WithOptions(controller.Options{
-			MaxConcurrentReconciles: *r.config.ConcurrentSyncs,
+			MaxConcurrentReconciles: *r.config,
 		}).
```
renormalize: `r.config` can not be indirected here, and is not the right type either.
```suggestion
			MaxConcurrentReconciles: *r.config.ConcurrentSyncs,
```

operator/go.mod
```diff
@@ -1,10 +1,9 @@
 module github.com/NVIDIA/grove/operator
 
-go 1.24.0
-
-toolchain go1.24.1
+go 1.24.3
```
renormalize: Do we plan on following the convention kubernetes follows by only specifying the `1.xx.0` go version?

operator/internal/controller/podgangset/register.go
```diff
@@ -61,7 +61,7 @@ func podCliquePredicate() predicate.Predicate {
 		},
 		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
 			// Only allow update event if the PodClique is managed by Grove and the spec has not changed.
```
renormalize: This comment is not valid anymore.
```suggestion
```

renormalize (commented):
The build is broken. Will push changes to fix that and address these comments.

operator/go.mod
unmarshall: Sure we can change that.

renormalize (approved):

PR #90, opened 2025-06-17
Misc fixes and partial implementation of Pod component

unmarshall:
* Partially implements Pod component. With this commit pods without scheduling gates can be created.
* Enables Pod component in the PodClique reconciler.
* Fixed HPA component which now correctly sets the target resource ref.
* Fixed service component, which now creates a headless service.
* Introduced some convenient functions.
* Changed the pgs-replica-index label key as it was not as per allowed conventions.


operator/internal/component/pclq/pod/pod.go
```diff
@@ -16,272 +16,200 @@
...
+			component.OperationSync,
+			fmt.Sprintf("failed to get existing pods for PodClique %v", k8sutils.GetObjectKeyFromObjectMeta(pclq.ObjectMeta)),
+		)
+	}
+	diff := len(existingPodNames) - int(pclq.Spec.Replicas)
+	if diff < 0 {
+		diff *= -1 // we only care about the absolute value of the difference
+		logger.Info("Too few replicas for PodClique, creating new pods", "need", pclq.Spec.Replicas, "existing", len(existingPodNames), "creating", diff)
+		return r.doCreatePods(ctx, logger, pclq, diff)
+	}
```
renormalize, line 104: Scale in will be brought in, in a following PR once we start creating `PodGang`s.

renormalize (approved):
thanks for bringing in these changes!

operator/internal/component/pclq/pod/pod.go
unmarshall: Yes, that is correct.

PR #98, opened 2025-07-02
Integrate `grove-initc` with the Pod component.

renormalize:
* Massively simplify the `grove-initc` component. `grove-initc` watches `Pod`s  for the ready condition, instead of watching `PodClique.Status.ReadyReplicas`. This removes `grove-initc`'s dependency on `grove-operator` to be up and running, and update `PodClique.Status` to enforce correct startup ordering.

* `grove-initc` is now specified as the final init container in the `Pod.Spec` of the pods created as a part of `PodClique`s which have the `startsAfter` field specified. It is not included if the field is left empty.

* `grove-initc` is always deployed with the same tag as that of `grove-operator`.

* Specify the `Pod.Spec.ServiceAccountName` as the `ServiceAccount` created for pods by `grove-operator`.

* Enhance the operator `ClusterRole` to include `pods/status`, as thisis required to grant Pod `ServiceAccounts` privilege to `list` and `get` `Pod`s in `grove-initc`.

* Use `downwardAPI` to fetch the namespace of the `Pod` the `grove-initc`, is a part of, and the name of the `PodGang` the pod belongs to. The former is used while creating the watcher, and the latter is used to filter events based on this label. Only pods that belong to a specific `PodGang` are relevant to the start-up ordering enforced by `grove-initc`.

* Move to `github.com/spf13/plag` to support the new multi-flag syntax being used for passing information about each parent PodClique to the `grove-initc`. Both `grove-operator` and `grove-initc` both use the `pflag` package.

* `buildResource` of the Pod component uses the `minAvailable` specified in the spec to pass it as a value of the flag for each `PodClique`. To perform this, the `PodGangSet` resource is now fetched in the `PodClique` reconciliation.

* Enhance `skaffold.yaml` to be able to correctly tag and publish `grove-initc`.


nvrohanv:
Question: Are init containers that the user specifies typically pod-local setup? If so makes sense to have grove-initc be last init container to run.

renormalize:
> Are init containers that the user specifies typically pod-local setup?

@nvrohanv I'm not sure I understand your question to be honest, what do you mean by "pod-local setup"?

operator/initc/internal/wait.go
```diff
@@ -18,163 +18,238 @@ package internal
...
+			operationWaitForParentPodClique,
+			"failed to convert labels required for the PodGang to selector",
+		)
+	}
 
 	eventHandlerContext, cancel := context.WithCancel(ctx)
 
-	defer close(c.readyCh) // close the channel the informers write to after the context they use is cancelled
-	defer cancel()         // cancel the context used by the informers if the wait is successful, or an err occurs
+	defer close(c.allReadyCh) // Close the channel the informers write to after the context they use is cancelled.
```
julienmancuso: move this up . if an error occurs, the channel will never be closed

operator/initc/internal/wait.go
```diff
@@ -18,163 +18,238 @@ package internal
...
+			"failed to register the Pod event handler",
+		)
 	}
 
-	wg.Wait()
+	factory.WaitForCacheSync(eventHandlerContext.Done())
+	factory.Start(eventHandlerContext.Done())
+
+	<-c.allReadyCh
 	return nil
```
julienmancuso: shouldn't we be doing something like 
```
select {
    case <-c.allReadyCh:
        return nil
    case <-ctx.Done():
        return ctx.Err()
    }
```

operator/initc/internal/wait.go
julienmancuso: Also, more importantly are we sure  that no other go-routine will try to write after we close this channel ?

operator/initc/internal/wait.go
```diff
@@ -18,163 +18,238 @@ package internal
...
+	}
+
+	return state, nil
 }
 
 // WaitForReady waits for all upstream start-up dependencies to be ready.
-func (c *CliqueState) WaitForReady(ctx context.Context) error {
-	c.log.Info("The clique names are:", "cliques", c.cliqueFQNs)
-	c.log.Info("The clique namespace is:", "cliquesNamespace", c.namespace)
+func (c *PodCliqueState) WaitForReady(ctx context.Context) error {
```
julienmancuso: we should unit test this function

operator/internal/component/podclique/pod/pod.go
```diff
@@ -145,6 +162,67 @@ func (r _resource) buildResource(pclq *grovecorev1alpha1.PodClique, podGangName
...
+		args := make([]string, 0)
+		for _, parentClique := range pclq.Spec.StartsAfter {
+			parentCliqueTemplateSpec, ok := lo.Find(pgs.Spec.Template.Cliques, func(templateSpec *grovecorev1alpha1.PodCliqueTemplateSpec) bool {
+				return strings.HasSuffix(parentClique, templateSpec.Name)
+			})
+			if !ok {
+				return groveerr.New(
+					errCodeMissingPodCliqueTemplate,
+					component.OperationSync,
+					fmt.Sprintf("no PodClique %s specifiec in startsAfter present in the templates", parentClique),
```
julienmancuso: specifiec -> specified

operator/initc/internal/wait.go
renormalize: Definitely, I am working on it currently!

operator/initc/internal/wait.go
renormalize: > move this up . if an error occurs, the channel will never be closed

Will move `c.allReadyCh` to the top to ensure the channel is closed in all cases. Thanks for finding this!

> Also, more importantly are we sure that no other go-routine will try to write after we close this channel ?

Yes. Only the functions registered in the event handler write to this channel. They are registered in `registerEventHandler()`.

In all cases, we ensure that these event handlers are shut down first, through the deferred `cancel()` for `eventHandlerContext` in the line below; since the event handlers use the `eventHandlerContext.Done()` channel to be shut down.

The deferred close of the context always shuts down the goroutines that write to this channel, after which the channel is closed.

operator/initc/internal/wait.go
renormalize: Thanks for catching this @julienmancuso, this was a big miss from my side!

operator/internal/component/podclique/pod/pod.go
renormalize: Addressed in [45ca4ef](https://github.com/NVIDIA/grove/pull/98/commits/45ca4ef0fe2478fb71b5721d71d0743cb290a58e).

operator/initc/internal/wait.go
renormalize: Addressed in [45ca4ef](https://github.com/NVIDIA/grove/pull/98/commits/45ca4ef0fe2478fb71b5721d71d0743cb290a58e).

operator/initc/internal/wait.go
renormalize: Addressed in [45ca4ef](https://github.com/NVIDIA/grove/pull/98/commits/45ca4ef0fe2478fb71b5721d71d0743cb290a58e).

operator/initc/internal/wait.go
```diff
@@ -18,163 +18,242 @@ package internal
...
 			if !ok {
 				return
 			}
-			if !slices.Contains(c.cliqueFQNs, pclq.Name) {
-				// pclq is not a parent
-				return
+
+			c.refreshReadyPodsOfPodClique(pod, false)
+			if c.checkAllParentsReady() {
+				c.allReadyCh <- struct{}{}
```
julienmancuso, line 186: I think there is still a chance this send (or 2 others in different event handlers) could be ran after the channel is closed : This is possible if informers just before the context is cancelled and after WaitForReady exits.
we might want to do something like :
```
type PodCliqueState struct {
    ...
    allReadyCh chan struct{}
    signaled   bool
}

func (c *PodCliqueState) signalAllReady() {
    if !c.signaled {
        c.signaled = true
        c.allReadyCh <- struct{}{}
    }
}
```

Then in event handlers (inside the mutex):
```
if c.checkAllParentsReady() {
    c.signalAllReady()
}
```

wdyt ?

renormalize: > This is possible if informers just before the context is cancelled and after WaitForReady exits.

I'm still not sure how this would happen. 

```go
	factory.Start(eventHandlerContext.Done())
```

where `Done()` returns a channel that is closed when the `eventHandlerContext` is `close()`d. Looking at `Start()`'s documentation:

```go
	// Start initializes all requested informers. They are handled in goroutines
	// which run until the stop channel gets closed.
	// Warning: Start does not block. When run in a go-routine, it will race with a later WaitForCacheSync.
	Start(stopCh <-chan struct{})
```

where the documentation specifies that all informer goroutines are shut down when `stopCh`, which is the `eventHandlerContext.Done()` channel in this case, is closed.

This very channel is deferred a close:

```go
	defer cancel() // Cancel the context used by the informers if the wait is successful, or an err occurs.
```

right after creation of the `eventHandlerContext`, which guarantees that this channel is closed **before** `WaitForReady` exits.

Only after the this channel is closed, and the informers are shut down, does the channel being written to, `c.allReadyCh` is closed, as it is the first statement, (which is deferred) in the function.

****

If the parent context is cancelled, then the child context `eventHandlerContext` is also closed, only after which the `c.allReadyCh` is closed.

renormalize: The concern was clarified, and the essence of the problem is that two informers could be currently waiting to process events, where the first is performing the computation when the context is closed, while the second is awaiting the mutex unlock before it performs its computation.

The first writes to the channel and exits successfully, and the second starts processing since the mutex is now unlocked. Now, when the `stopCh` is closed, the second receiver will only exit once the processing is done, during which it could potentially write to the channel again, which is unsafe since the `c.allReadyCh` can already be closed by that time.

It is therefore a better idea to use a flag like was suggested in a previous comment to only write to the channel once.

operator/initc/internal/wait.go
renormalize: I will be adding unit tests in a following PR for all changes made in operator in this PR, and for everything in initc.

This feature is being waited on by multiple interested parties, therefore I would like to bring this feature in ASAP.

unmarshall (approved):

operator/initc/internal/wait.go
renormalize: @julienmancuso 

During a review with @unmarshall, we found that the event handler only runs one goroutine per informer. Since we only have one informer, there would be no case where a write to a closed channel would happen. Therefore, I reverted this change as it is unnecessary.

PR #105, opened 2025-07-10
Pod name validation pgs

Ronkahn21:
## Summary
  - Add comprehensive pod name constraints validation for all PodClique templates during PodGangSet admission
  - Implement selective validation logic that validates PodCliques differently based on whether they belong to scaling groups
  - Improve error messages with clearer, more actionable feedback for constraint violations

  ## Changes
  - Added pod name constraints validation to `validatePodCliqueTemplates` for PodCliques NOT in scaling groups (45-char limit: pgs-name + pclq-name)
  - Added pod name constraints validation to `validateScalingGroupPodCliqueNames` for PodCliques IN scaling groups (45-char limit: pgs-name + pcsg-name + pclq-name)
  - Updated error messages to be more user-friendly while maintaining technical accuracy

  ## Technical Details
  Pod names follow different formats:
  - **PCSG pods**: `<pgs-name>-<pgs-index>-<pcsg-name>-<pcsg-index>-<pclq-name>-<random>`
  - **Non-PCSG pods**: `<pgs-name>-<pgs-index>-<pclq-name>-<random>`

  The validation ensures:
  - **PodCliques NOT in scaling groups**: validated with 45-character limit for pgs-name + pclq-name
  - **PodCliques IN scaling groups**: validated with 45-character limit for pgs-name + pcsg-name + pclq-name
  - Total pod names stay under 64 characters for annotation support

### DIFF
--- a/operator/internal/webhook/admission/pgs/validation/podgangset.go
+++ b/operator/internal/webhook/admission/pgs/validation/podgangset.go
@@ -24,6 +24,7 @@ import (
 
 	grovecorev1alpha1 "github.com/NVIDIA/grove/operator/api/core/v1alpha1"
 	"github.com/NVIDIA/grove/operator/internal/utils"
+
 	"github.com/samber/lo"
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
--- a/operator/internal/webhook/admission/pgs/validation/podgangset.go
+++ b/operator/internal/webhook/admission/pgs/validation/podgangset.go
@@ -113,30 +113,15 @@ func (v *pgsValidator) validatePodCliqueTemplates(fldPath *field.Path) ([]string
 	cliqueRoles := make([]string, 0, len(cliqueTemplateSpecs))
 	schedulerNames := make([]string, 0, len(cliqueTemplateSpecs))
 	for _, cliqueTemplateSpec := range cliqueTemplateSpecs {
-		if err := apivalidation.NameIsDNSSubdomain(cliqueTemplateSpec.Name, false); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("name"), cliqueTemplateSpec.Name,
-				"invalid PodCliqueTemplateSpec name, must be a valid DNS subdomain"))
-		}
-
-		// Only validate pod name constraints for PodCliques that are NOT part of any scaling group
-		// any pod clique that is part of scaling groups will be checked as part of scaling group pod name constraints.
-		if !scalingGroupCliqueNames.Has(cliqueTemplateSpec.Name) {
-			if err := validatePodNameConstraints(v.pgs.Name, "", cliqueTemplateSpec.Name); err != nil {
-				// add error to each of filed paths that compose the podName in case of a PodCliqueTemplateSpec
-				allErrs = append(allErrs, field.Invalid(fldPath.Child("name"), cliqueTemplateSpec.Name, err.Error()))
-				allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("name"), v.pgs.Name, err.Error()))
-			}
-		}
-
-		cliqueNames = append(cliqueNames, cliqueTemplateSpec.Name)
-		cliqueRoles = append(cliqueRoles, cliqueTemplateSpec.Spec.RoleName)
-		warns, errs := v.validatePodCliqueTemplateSpec(cliqueTemplateSpec, fldPath)
+		warns, errs := v.validatePodCliqueTemplateSpec(cliqueTemplateSpec, fldPath, scalingGroupCliqueNames)
 		if len(errs) != 0 {
 			allErrs = append(allErrs, errs...)
 		}
 		if len(warns) != 0 {
 			warnings = append(warnings, warns...)
 		}
+		cliqueNames = append(cliqueNames, cliqueTemplateSpec.Name)
+		cliqueRoles = append(cliqueRoles, cliqueTemplateSpec.Spec.RoleName)
 		schedulerNames = append(schedulerNames, cliqueTemplateSpec.Spec.PodSpec.SchedulerName)
 	}
 
@@ -158,6 +143,31 @@ func (v *pgsValidator) validatePodCliqueTemplates(fldPath *field.Path) ([]string
 	}
 
 	return warnings, allErrs
+}
+
+func (v *pgsValidator) validatePodCliqueNameConstrains(fldPath *field.Path, cliqueTemplateSpec *grovecorev1alpha1.PodCliqueTemplateSpec, scalingGroupCliqueNames sets.Set[string]) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if err := apivalidation.NameIsDNSSubdomain(cliqueTemplateSpec.Name, false); err != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("name"), cliqueTemplateSpec.Name,
+			"invalid PodCliqueTemplateSpec name, must be a valid DNS subdomain"))
+	}
+
+	// Only validate pod name constraints for PodCliques that are NOT part of any scaling group
+	// any pod clique that is part of scaling groups will be checked as part of scaling group pod name constraints.
+	if !scalingGroupCliqueNames.Has(cliqueTemplateSpec.Name) {
+		allErrs = append(allErrs, validateStandalonePodClique(fldPath, v, cliqueTemplateSpec)...)
+	}
+	return allErrs
+}
+
+func validateStandalonePodClique(fldPath *field.Path, v *pgsValidator, cliqueTemplateSpec *grovecorev1alpha1.PodCliqueTemplateSpec) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if err := validatePodNameConstraints(v.pgs.Name, "", cliqueTemplateSpec.Name); err != nil {
+		// add error to each of filed paths that compose the podName in case of a PodCliqueTemplateSpec
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("name"), cliqueTemplateSpec.Name, err.Error()))
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("name"), v.pgs.Name, err.Error()))
+	}
+	return allErrs
 }
 
 func (v *pgsValidator) validatePodGangSchedulingPolicyConfig(schedulingPolicyConfig *grovecorev1alpha1.SchedulingPolicyConfig, fldPath *field.Path) field.ErrorList {
@@ -191,7 +201,6 @@ func (v *pgsValidator) validatePodCliqueScalingGroupConfigs(fldPath *field.Path)
 		// validate that scaling groups only contains clique names that are defined in the PodGangSet.
 		allErrs = append(allErrs, v.validateScalingGroupPodCliqueNames(scalingGroupConfig.Name, allPodGangSetCliqueNames,
 			scalingGroupConfig.CliqueNames, fldPath.Child("cliqueNames"), groupNameFiledPath)...)
-
 	}
 
 	// validate that the scaling group names are unique
@@ -225,7 +234,8 @@ func (v *pgsValidator) validateTerminationDelay(fldPath *field.Path) field.Error
 	return allErrs
 }
 
-func (v *pgsValidator) validatePodCliqueTemplateSpec(cliqueTemplateSpec *grovecorev1alpha1.PodCliqueTemplateSpec, fldPath *field.Path) ([]string, field.ErrorList) {
+func (v *pgsValidator) validatePodCliqueTemplateSpec(cliqueTemplateSpec *grovecorev1alpha1.PodCliqueTemplateSpec,
+	fldPath *field.Path, scalingGroupCliqueNames sets.Set[string]) ([]string, field.ErrorList) {
 	allErrs := field.ErrorList{}
 
 	allErrs = append(allErrs, validateNonEmptyStringField(cliqueTemplateSpec.Name, fldPath.Child("name"))...)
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -46,3 +46,4 @@ scheduler/bin/*
 # generated/copied chart resources
 operator/charts/pki-resources/*
 operator/charts/crds/*
+/CLAUDE.md
```
unmarshall: can you please explain why is this required in .gitignore?

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -0,0 +1,492 @@
...
+	startupType := grovecorev1alpha1.CliqueStartupTypeAnyOrder
+	return &grovecorev1alpha1.PodGangSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: "default",
+		},
+		Spec: grovecorev1alpha1.PodGangSetSpec{
+			Replicas: 1,
+			Template: grovecorev1alpha1.PodGangSetTemplateSpec{
+				StartupType: &startupType,
```
renormalize: We typically tend to use the `k8s.io/utils/ptr.To` function when pointer fields are to be set. Helps readability and avoids too many local variables.

```suggestion
				StartupType: ptr.To(grovecorev1alpha1.CliqueStartupTypeAnyOrder),
```

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+// Helper function to create a PodCliqueTemplateSpec
+func createPodCliqueTemplate(name string) *grovecorev1alpha1.PodCliqueTemplateSpec {
+	minAvailable := int32(1)
+	return &grovecorev1alpha1.PodCliqueTemplateSpec{
+		Name: name,
+		Spec: grovecorev1alpha1.PodCliqueSpec{
+			Replicas:     1,
+			MinAvailable: &minAvailable,
```
renormalize: Same as https://github.com/NVIDIA/grove/pull/105#discussion_r2198420128.

operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -278,18 +308,39 @@ func (v *pgsValidator) checkNetworkPackGroupConfigsForPartialPCSGInclusions(fldP
...
+// getScalingGroupCliqueNames returns a set of all clique names that belong to scaling groups
+func (v *pgsValidator) getScalingGroupCliqueNames() sets.Set[string] {
+	scalingGroupCliqueNames := sets.New[string]()
+	for _, scalingGroupConfig := range v.pgs.Spec.Template.PodCliqueScalingGroupConfigs {
+		for _, cliqueName := range scalingGroupConfig.CliqueNames {
+			scalingGroupCliqueNames.Insert(cliqueName)
+		}
+	}
+	return scalingGroupCliqueNames
+}
```
renormalize: I am unsure if we really need to generate a set of unique values. In `validatePodCliqueScalingGroupConfigs()`, we already ensure that clique names are unique across all `PodCliqueScalingGroup`s in

https://github.com/Ronkahn21/grove/blob/61a984d5d75f41eb4ee30c9a4ba71822307ead37/operator/internal/webhook/admission/pgs/validation/podgangset.go#L206-L210.

So I'd like to go about it this way:
  - In `validatePodGangTemplateSpec()`, we can first call the validation `validatePodCliqueScalingGroupConfigs()`, then `validatePodGangSchedulingPolicyConfig()`, and finally `validatePodCliqueTemplates()`. The uniqueness is checked in `validatePodCliqueScalingGroupConfigs()`.
  - We simply return the clique names as a slice here.


```suggestion
// getScalingGroupCliqueNames returns all clique names that belong to scaling groups
func (v *pgsValidator) getScalingGroupCliqueNames() []string {
	scalingGroupCliqueNames := []string{}
	for _, pcsgConfig := range v.pgs.Spec.Template.PodCliqueScalingGroupConfigs {
		scalingGroupCliqueNames = append(scalingGroupCliqueNames, pcsgConfig.CliqueNames...)
	}
	return scalingGroupCliqueNames
}
```

operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -278,18 +308,39 @@ func (v *pgsValidator) checkNetworkPackGroupConfigsForPartialPCSGInclusions(fldP
...
+			// add error to each of filed paths that compose the podName
+			allErrs = append(allErrs, field.Invalid(fldPath.Child(nameFiledName), pclqName, err.Error()))
+			allErrs = append(allErrs, field.Invalid(pcsgNameFieldPath, pclqName, err.Error()))
+			allErrs = append(allErrs, field.Invalid(field.NewPath(medataFieldName).Child(nameFiledName), v.pgs.Name, err.Error()))
+		}
+	}
 	return allErrs
 }
 
+// kube
```
renormalize: Could this comment be expanded on or removed?

operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -100,9 +106,26 @@ func (v *pgsValidator) validatePodCliqueTemplates(fldPath *field.Path) ([]string
...
+		}
+
+		// Only validate pod name constraints for PodCliques that are NOT part of any scaling group
+		if !scalingGroupCliqueNames.Has(cliqueTemplateSpec.Name) {
+			if err := validatePodNameConstraints(v.pgs.Name, "", cliqueTemplateSpec.Name); err != nil {
+				// add error to each of filed paths that compose the podName in case of a PodCliqueTemplateSpec
+				allErrs = append(allErrs, field.Invalid(fldPath.Child(nameFiledName), cliqueTemplateSpec.Name, err.Error()))
+				allErrs = append(allErrs, field.Invalid(field.NewPath(medataFieldName).Child(nameFiledName), v.pgs.Name, err.Error()))
+			}
+		}
```
renormalize: Nit:

I prefer not having the main logic not too nested, and use early return/continue constructs to skip special cases. This helps readability in my opinion.

So we could do the following:


```suggestion
		// Only validate pod name constraints for PodCliques that are NOT part of any scaling group
		if scalingGroupCliqueNames.Contains(cliqueTemplateSpec.Name) {
            continue
		}
        if err := validatePodNameConstraints(v.pgs.Name, "", cliqueTemplateSpec.Name); err != nil {
            // add error to each of filed paths that compose the podName in case of a PodCliqueTemplateSpec
            allErrs = append(allErrs, field.Invalid(fldPath.Child(nameFiledName), cliqueTemplateSpec.Name, err.Error()))
            allErrs = append(allErrs, field.Invalid(field.NewPath(medataFieldName).Child(nameFiledName), v.pgs.Name, err.Error()))
        }
```

and move the name validation (along with the `NameIsDNSSubdomain` validation) to the end of this `for` loop. This way we validate the spec for all `PodClique`s, and only validate names of `PodClique`s that are not part of any scaling group.

operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -471,3 +522,34 @@ func clearContainerImages(containers []corev1.Container) {
...
+// - Random string + hyphens: 10 chars for PCSG pods, 8 chars for non-PCSG pods
+// - Max sum of all resource name characters: 45 chars
+func validatePodNameConstraints(pgsName, pcsgName, pclqName string) error {
+	// Check resource name constraints
+	resourceNameLength := len(pgsName) + len(pclqName)
+	if pcsgName != "" {
+		resourceNameLength += len(pcsgName)
+	}
+
+	if resourceNameLength > 45 {
```
renormalize: Could a constant be declared for this, like `maxResourceNameLength`, given the possibility of this constraint changing in the future?
It would also help readibility `if resourceNameLength > maxResourceNameLength`.

operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -33,6 +33,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
+const (
+	medataFieldName    = "metadata"
+	cliqueNameFieldRef = "cliqueNames"
+	nameFiledName      = "name"
+)
```
renormalize: I don't particularly think this it is necessary to declare constants for fields names that are a part of the schema of the resource. We don't particularly follow this convention anywhere else in webhooks in general. Even kubernetes does not follow that convention as far as I know.

renormalize: @unmarshall would you like to chime in?

unmarshall: defining constants for JSON field names is unnecessary as they are never going to change for a given version of the API. They do not add more context as the JSON names are quite clear themselves.

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -0,0 +1,492 @@
...
+	"github.com/stretchr/testify/assert"
+	admissionv1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// Helper function to create a basic valid PodGangSet
+func createBasePodGangSet(name string) *grovecorev1alpha1.PodGangSet {
```
unmarshall: What is the semantics of a `Base` PodGangSet? Can you elaborate on what does it mean?

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -0,0 +1,492 @@
...
+					if tc.expectedErrPath != "" {
+						// Check if this is an aggregate error with field errors
+						if aggErr, ok := err.(utilerrors.Aggregate); ok && len(aggErr.Errors()) > 0 {
+							if fieldErr, ok := aggErr.Errors()[0].(*field.Error); ok {
+								assert.Contains(t, fieldErr.Field, tc.expectedErrPath, "Error field path should match expected")
+							}
+						}
+					}
+				}
+			} else {
```
unmarshall: Make this the if part as that makes code more readable.
```go
if !tc.expectError {
   assert.NoError(t, err, "Expected no validation error for test case: %s", tc.description)
} else {
  // the case where you expect an error.
}
```

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -0,0 +1,492 @@
...
+				pgs.Spec.Template.Cliques = append(pgs.Spec.Template.Cliques, createPodCliqueTemplate(cliqueName))
+			}
+
+			validator := newPGSValidator(pgs, admissionv1.Create)
+			warnings, err := validator.validate()
+
+			if tc.expectError {
+				assert.Error(t, err, "Expected validation error for test case: %s", tc.description)
+				if err != nil {
+					if tc.expectedErrMsg != "" {
```
unmarshall: In the above test cases there is currently no case where the error is expected but there is no error message. Is there a real need to check `if tc.expectedErrMsg != ""`?

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -0,0 +1,492 @@
...
+						assert.Contains(t, err.Error(), tc.expectedErrMsg, "Error message should contain expected text")
+					}
+					if tc.expectedErrPath != "" {
```
unmarshall: If there is an error there is always going to be a error path. Is there a need to check if it is not empty?
If you make it conditional then you hide a coding error.

unmarshall (changes requested):

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
Ronkahn21: Base on the context skeleton to create a pogGangSet for testing the functionlty

operator/internal/webhook/admission/pgs/validation/podgangset.go
Ronkahn21: You are correct, added the const !

operator/internal/webhook/admission/pgs/validation/podgangset.go
Ronkahn21: NP, fixed. I will follow the convention; it was more about not having multiple strings without consts and less about the JSON field.

.gitignore
Ronkahn21: It was by mistake,  removed

operator/internal/webhook/admission/pgs/validation/podgangset.go
Ronkahn21: removed

operator/internal/webhook/admission/pgs/validation/podgangset.go
Ronkahn21: The validation for pod clique is different from if it is already  part of scaling group or not, it allow to create loigger neams for standealon pc, i am using this not for checking uniq for checking if the pc is standalone or not

operator/internal/webhook/admission/pgs/validation/podgangset.go
Ronkahn21: I moved the entire validation to smaller functions under the validatePodCliqueTemplateSpec main validation fuction

operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -101,19 +107,22 @@ func (v *pgsValidator) validatePodCliqueTemplates(fldPath *field.Path) ([]string
 		allErrs = append(allErrs, field.Required(fldPath, "at least one PodClique must be defined"))
 	}
 
+	// Get all clique names that belong to scaling groups
+	scalingGroupCliqueNames := v.getScalingGroupCliqueNames()
```
renormalize, line 111: Can this be moved into `v.validatePodCliqueTemplateSpec` or `validatePodCliqueNameConstraints`, since the latter is the only place that this information is used? It would also make the function signature of `validatePodCliqueTemplateSpec` cleaner if it is.

renormalize (approved):
Just one nit that I've added. Will merge the PR into main once it is addressed. Thanks!

PR #119, opened 2025-07-29
Pod discovery env vars

Ronkahn21:


renormalize (approved):
Verified that the changes are identical to the old PR. Thanks!

Issue #123, opened 2025-07-31
Make deploy-local failed


kangclzjc:
When I tried to deploy grove in my k8s cluster, it failed.

log like below:
```
make deploy-local
Copying CRDs to helm charts...
Creating /root/grove/operator/charts/crds to copy the CRDs if not present...
Copying grove-operator CRDS...
Copying CRD grove.io_podgangsets.yaml to /root/grove/operator/charts/crds
Copying CRD grove.io_podcliques.yaml to /root/grove/operator/charts/crds
Copying CRD grove.io_podcliquescalinggroups.yaml to /root/grove/operator/charts/crds
Copying scheduler CRDS...
Copying CRD scheduler.grove.io_podgangs.yaml to /root/grove/operator/charts/crds
Generating PKI resources if not present or expired...
E0730 23:47:23.427496  898799 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: invalid character '<' looking for beginning of value"
E0730 23:47:23.431030  898799 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: invalid character '<' looking for beginning of value"
E0730 23:47:23.434575  898799 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: invalid character '<' looking for beginning of value"
E0730 23:47:23.438117  898799 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: invalid character '<' looking for beginning of value"
E0730 23:47:23.441401  898799 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: invalid character '<' looking for beginning of value"
error: invalid character '<' looking for beginning of value
make: *** [Makefile:120: deploy-local] Error 1
```

kubectl version
```
Client Version: v1.32.7
Kustomize Version: v5.5.0
Server Version: v1.32.7
```

renormalize:
@kangclzjc Are you using the supported kind cluster that can be started using `make kind-up`, or are you using your own Kubernetes cluster for this?

`make deploy-local` only works with the local cluster that is started with `make kind-up`.

unmarshall:
Also please check https://signoz.io/guides/kubectl-error-memcache-go-265-couldnt-get-current-server-api-group-list-get/
The link uses minikube but you can easily replace that with kind.

kangclzjc:
I test it in my local k8s cluster, set repo and kubeconfig then it will deploy succesfully. See my PR #129

PR #129, opened 2025-08-04
feat: add local k8s deploy

kangclzjc:
Add local k8s deployment guide here. 

Related issue #123


docs/getting-started.md
```diff
@@ -26,6 +26,24 @@ All grove operator Make targets are located in [Operator Makefile][./operator/Ma
   export KUBECONFIG=./hack/kind/kubeconfig
   ```
 
+### Local Kubernetes cluster set-up
```
renormalize: Not sure if I'd call it a "Local Kubernetes cluster set-up" any more, since it might as well point to a remote cluster hosted in the cloud, which is not "local".

We should think of renaming the make targets from `deploy-local` to `deploy` so it would be semantically correct.

Currently skaffold.yaml also (unfortunately) specifies the registry where the init container is pushed to as `localhost:5001` as an env variable.

Do you have any suggestions on how you'd like this to work with remote clusters? We would need to overrride the value we are setting here based on whether the cluster is remote or local (kind). This value could possibly be passed to the skaffold command as an env var which it could read and populate.

kangclzjc: Yes, we'd better put it as env var now, and for dev now, it's convenient. However maybe we should not hardcode localhost:5001 here. Can we pass it from cmd line or pass it from make instead of editting it directly in skaffold ?

I will add a make deploy and try to pass it from make command. What do you think of this solution ?

renormalize: Yes, not hardcoding `localhost:5001` is exactly my point as well.

We could make the env var's value `localhost:5001` if no explicit KUBECONFIG is declared by the user. The user should only have to declare the remote registry in one location, and the makefile should be able to pass this override to all places it is being used.

Please go ahead and give this a try, and I will review it if you are willing to make this change.

kangclzjc:
> @kangclzjc are you still working on this or do you want me to pick this up?

yes，I'm doing this. Will update this PR today

operator/Makefile
```diff
@@ -115,6 +115,18 @@ ifndef NAMESPACE
...
+	@PACKAGE_PATH=${LD_PACKAGE_PATH} PROGRAM_NAME=${OPERATOR_NAME} $(MODULE_HACK_DIR)/deploy-local.sh run -m grove-operator -n $(NAMESPACE)
+
+# Undeploys grove-operator from k8s cluster using skaffold.
+.PHONY: undeploy
+undeploy: $(SKAFFOLD)
+	@$(SKAFFOLD) delete -m grove-operator -n $(NAMESPACE)
+
+# Deploys grove-operator to a local k8s cluster using skaffold.
 .PHONY: deploy-local
 deploy-local: $(SKAFFOLD)
```
renormalize: What I meant from my previous comments is that we can simply merge `deploy` and `deploy-local` into one target as each target supports both remote *and* local clusters. There isn't any reason to have two separate targets that are running the same exact thing. The same applies to `deploy-local-dev`, and `deploy-local-debug`.

All of these can be renamed to `deploy`, `deploy-dev`, and `deploy-debug`.

All the other targets are also to be modified accordingly.

Similarly, we rename the `deploy-local.sh` to `deploy.sh`.

operator/hack/deploy-local.sh
```diff
@@ -45,7 +45,14 @@ function check_prereq() {
...
   export LD_FLAGS="${ld_flags}"
-  skaffold "$@"
+  
+  # Set default registry if not provided
+  if [[ -z "${DEFAULT_REPO:-}" ]]; then
+    export DEFAULT_REPO="localhost:5001"
+  fi
+  
+  local skaffold_cmd="skaffold --default-repo=${DEFAULT_REPO}"
+  $skaffold_cmd "$@"
```
renormalize: Do we need to specify the `local skaffold_cmd`?

```suggestion
  # Set default registry if not provided
  if [[ -z "${DEFAULT_REGISTRY:-}" ]]; then
    export DEFAULT_REGISTRY="localhost:5001"
  fi
  skaffold --default-repo=${DEFAULT_REGISTRY} "$@"
```

docs/getting-started.md
```diff
@@ -26,6 +26,24 @@ All grove operator Make targets are located in [Operator Makefile][./operator/Ma
...
+If you want to use your own Kubernetes cluster instead of the KIND cluster, follow these steps:
+
+- **Set the KUBECONFIG environment variable** to point to your Kubernetes cluster configuration:
+
+   ```bash
+   # Set KUBECONFIG to use your Kubernetes cluster kubeconfig
+   export KUBECONFIG=/path/to/your/kubernetes/kubeconfig
+   ```
+
+- **Set the DEFAULT_REPO environment variable** to specify your Docker image repository:
```
renormalize: We actually need to specify the registry where the image is pushed, not the repository itself, right?

If I were to push my image to `ghcr.io/renormalize/grove-operator`, then I would have to specify the `--default-repo` flag of skaffold to `ghcr.io/renormalize`, right?

Perhaps we can rename this to `DEFAULT_REGISTRY`. The skaffold flag name is confusing in my opinion.

operator/skaffold.yaml
```diff
@@ -46,7 +46,7 @@ deploy:
           image.tag: "{{.VERSION}}@{{.IMAGE_DIGEST}}"
           deployment.env:
             - name: GROVE_INIT_CONTAINER_IMAGE
-              value: localhost:5001/grove-initc
+              value: "{{.DEFAULT_REPO}}/grove-initc"
```
renormalize:
```suggestion
              value: "{{.DEFAULT_REGISTRY}}/grove-initc"
```

renormalize (changes requested):
The idea looks good to me overall. Just a few opinions on the implementation and naming.

Will merge once these minor comments are addressed.

Thanks @kangclzjc!

docs/getting-started.md
kangclzjc: yes, you're right

docs/getting-started.md
```diff
@@ -26,6 +26,24 @@ All grove operator Make targets are located in [Operator Makefile][./operator/Ma
...
+If you want to use your own Kubernetes cluster instead of the KIND cluster, follow these steps:
+
+- **Set the KUBECONFIG environment variable** to point to your Kubernetes cluster configuration:
+
+   ```bash
+   # Set KUBECONFIG to use your Kubernetes cluster kubeconfig
+   export KUBECONFIG=/path/to/your/kubernetes/kubeconfig
+   ```
+
+- **Set the DEFAULT_REGISTRY environment variable** to specify your Docker image registry:
```
unmarshall: Ideally we should call this `CONTAINER_REGISTRY` and if not specified then we should fallback to the `DEFAULT_CONTAINER_REGISTRY`. Using `DEFAULT` gives an impression that there is a provision to overwrite that default as well which is not the case.

unmarshall (changes requested):

docs/getting-started.md
kangclzjc: ok, CONTAINER_REGISTRY could be better understanding.

unmarshall (approved):
Thanks for addressing the review comments. Looks good!

PR #132, opened 2025-08-06
Publish Docker images for Grove operator and init container to GitHub Container Registry.

renormalize:
* Publish Docker images for Grove operator and init container to GitHub Container Registry.
* Fix broken `check` target.


.github/workflows/build-and-test.yaml
```diff
@@ -1,3 +1,6 @@
+# Copyright 2025 NVIDIA CORPORATION
```
unmarshall: this header is different from license headers on go files, do we need to streamline that?
The same comment is applicable to other files under .github

PR #135, opened 2025-08-07
Feat/available replicas and status

Ronkahn21:
https://github.com/NVIDIA/grove/issues/112#issuecomment-3155448512

### DIFF
--- a/operator/internal/utils/kubernetes/pod.go
+++ b/operator/internal/utils/kubernetes/pod.go
@@ -144,12 +144,12 @@ func GetContainerStatusIfTerminatedErroneously(containerStatuses []corev1.Contai
 }
 
 func logTerminatedErroneouslyPodContainerStatus(logger logr.Logger, podObjKey client.ObjectKey, containerStatus *corev1.ContainerStatus) {
-	if containerStatus != nil {
+	if containerStatus != nil && containerStatus.LastTerminationState.Terminated != nil {
 		logger.Info("container exited with a non-zero exit code",
 			"pod", podObjKey,
 			"container", containerStatus.Name,
-			"exitCode", containerStatus.State.Terminated.ExitCode,
-			"exitReason", containerStatus.State.Terminated.Reason)
+			"exitCode", containerStatus.LastTerminationState.Terminated.ExitCode,
+			"exitReason", containerStatus.LastTerminationState.Terminated.Reason)
 	}
 }
 


operator/internal/component/utils/podcliquescalinggroup.go
```diff
@@ -101,3 +118,21 @@ func GenerateDependencyNamesForBasePodGang(pgs *grovecorev1alpha1.PodGangSet, pg
...
 	return parentPCLQNames
 }
+
+// GroupPCSGsByPGSReplicaIndex filters PCSGs that have a PodGangSetReplicaIndex label and groups them by the PGS replica.
+func GroupPCSGsByPGSReplicaIndex(pcsgs []grovecorev1alpha1.PodCliqueScalingGroup) map[string][]grovecorev1alpha1.PodCliqueScalingGroup {
+	return groupPCSGsByLabel(pcsgs, grovecorev1alpha1.LabelPodGangSetReplicaIndex)
+}
+
+// groupPCSGsByLabel groups PCSGs by the specified label key.
+func groupPCSGsByLabel(pcsgs []grovecorev1alpha1.PodCliqueScalingGroup, labelKey string) map[string][]grovecorev1alpha1.PodCliqueScalingGroup {
```
julienmancuso: nit: I see we are using `lo` library in some other places in Grove for this kind of groupBy functions

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if !r.checkPCSGsAvailability(logger, pgs, expectedPCSGs, replicaIndex, replicaPCSGs) {
+		return false
+	}
+
+	logger.Info("PGS replica is available", "pgs", client.ObjectKeyFromObject(pgs), "replicaIndex", replicaIndex)
+	return true
+}
+
+func (r *Reconciler) isPGSReplicaPCLAsAvailable(logger logr.Logger, pgs *grovecorev1alpha1.PodGangSet, expectedStandalonePCLQs, replicaIndex int, pclqs []grovecorev1alpha1.PodClique) bool {
```
julienmancuso: can't we factorize code with https://github.com/NVIDIA/grove/pull/135/files#diff-318a49cd15c8d0ad56db94128c4073430fbf903c1087d620f725f22332904374R67 ?

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,162 @@
...
+	isAvailable := lo.Reduce(nonTerminatedPCLQs, func(agg bool, pclq grovecorev1alpha1.PodClique, _ int) bool {
+		return agg && k8sutils.IsConditionFalse(pclq.Status.Conditions, grovecorev1alpha1.ConditionTypeMinAvailableBreached)
+	}, true)
+
+	logger.Info("PGS replica availability status", "pgs",
+		client.ObjectKeyFromObject(pgs), "replicaIndex", replicaIndex, "isAvailable", isAvailable)
+	return isAvailable
+}
+
+func (r *Reconciler) checkPCSGsAvailability(logger logr.Logger, pgs *grovecorev1alpha1.PodGangSet, expectedPCSGs, replicaIndex int, pcsgs []grovecorev1alpha1.PodCliqueScalingGroup) bool {
```
julienmancuso: nit : checkPCSGsAvailability and isPGSReplicaPCLAsAvailable can be merged into :

type Resource interface {
   GetObjectMeta() 
   GetStatusConditions()
} 

checkAvailability(logger logr.Logger, pgs *grovecorev1alpha1.PodGangSet, expected int, resources []Resource)

operator/internal/common/constants.go
```diff
@@ -23,4 +23,9 @@ const (
 	PodNamespaceFileName = "namespace"
 	// VolumeMountPathPodInfo contains the file path at which the downward API volume is attached.
 	VolumeMountPathPodInfo = "/var/grove/pod-info"
+	// NamePGSPodClique is the component name for a PodClique resource handled by a PodGangSet.
+	NamePGSPodClique = "pgs-podclique"
```
unmarshall: All constants for component name label key are defined in `internal/component/types.go`. Do you see a reason why only these 2 constants should be defined here?

operator/internal/component/podclique/pod/syncflow.go
```diff
@@ -177,7 +177,7 @@ func getTerminatingAndNonTerminatingPodUIDs(existingPCLQPods []*corev1.Pod) (ter
 	nonTerminatingUIDs = make([]types.UID, 0, len(existingPCLQPods))
 	terminatingUIDs = make([]types.UID, 0, len(existingPCLQPods))
 	for _, pod := range existingPCLQPods {
-		if k8sutils.IsResourceTerminating(pod.ObjectMeta) {
+		if k8sutils.IsResourceTerminating(&pod.ObjectMeta) {
```
unmarshall: can you revert this change and instead change the parameter type to `IsResourceTerminating` to take `metav1.ObjectMeta` instead?

operator/internal/component/types.go
```diff
@@ -41,10 +41,6 @@ const (
 // These component names will be set against grovecorev1alpha1.LabelComponentKey label key on
 // respective components.
 const (
-	// NamePGSPodClique is the component name for a PodClique resource handled by a PodGangSet.
```
unmarshall, line 44: any reason for this change?

operator/internal/component/utils/generic.go
```diff
@@ -0,0 +1,39 @@
...
+package utils
+
+import (
+	"github.com/NVIDIA/grove/operator/internal/component"
+
+	"github.com/samber/lo"
+)
+
+// groupByLabel groups Kubernetes objects by the specified label key using lo utilities.
+func groupByLabel[T component.GroveCustomResourceType, P interface {
```
unmarshall: * Is this function necessary?
* Signature of this function is a bit complex -> using embedded interface P is not at all legible and hard to follow.
* Doc string should not mention utility library specifically.
* file name `generic.go` is way to generic and can be easy dumping ground for anything that uses generics.
* You are iterating twice once to filter and then again to group. This is a bit inefficient.

In all the code readability is compromised by this function. Can we keep things a bit simple? There are only 2 places this is called and this complexity seems quite unnecessary.

unmarshall: If you really want to have a single function that can work with all `component.GroveCustomResourceType` you can just enhance the interface to:
```go
type GroveCustomResourceType interface {
	grovecorev1alpha1.PodGangSet | grovecorev1alpha1.PodClique | grovecorev1alpha1.PodCliqueScalingGroup
	GetLabels() map[string]string
}
```

Then your function looks a LOT simpler and easier to comprehend:
```go
func GroupByLabelValue[T component.GroveCustomResourceType](objs []T, labelKey string) map[string][]T {
	groups := make(map[string][]T)
	for _, obj := range objs {
		labelVal, ok := obj.GetLabels()[labelKey]
		if !ok {
			continue
		}
		groups[labelVal] = append(groups[labelVal], obj)
	}
	return groups
}
```

unmarshall: Also move this function to `component/utils/labels.go` instead.

operator/internal/component/utils/podclique.go
```diff
@@ -117,3 +112,22 @@ func GetMinAvailableBreachedPCLQInfo(pclqs []grovecorev1alpha1.PodClique, termin
 	slices.Sort(waitForDurations)
 	return pclqCandidateNames, waitForDurations[0]
 }
+
+// GetPodCliqueForPGSNotInPCSG retrieves PodClique objects that are not part of any PodCliqueScalingGroup for the given PodGangSet.
+func GetPodCliqueForPGSNotInPCSG(ctx context.Context, cl client.Client, pgsObjKey client.ObjectKey) ([]grovecorev1alpha1.PodClique, error) {
```
unmarshall: * The function should be renamed to `GetPodCliquesForPGSNotInPCSG`.
* The function does not really filter what it says. It will get all PodCliques in a PGS

So the caller:
```go
	standalonePCLQs, err := componentutils.GetPodCliqueForPGSNotInPCSG(ctx, r.client, client.ObjectKeyFromObject(pgs))
```
assumes that these are standalone when as per the code they should not be, right?

operator/internal/component/utils/podcliquescalinggroup.go
```diff
@@ -47,19 +47,36 @@ func FindScalingGroupConfigForClique(scalingGroupConfigs []grovecorev1alpha1.Pod
...
+	pcsgList, err := getPcsgForPGS(ctx, cl, pgsObjKey, map[string]string{
+		grovecorev1alpha1.LabelPodGangSetReplicaIndex: strconv.Itoa(pgsReplicaIndex),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return pcsgList.Items, nil
+}
+
+func getPcsgForPGS(ctx context.Context, cl client.Client, pgsObjKey client.ObjectKey, extraLabels map[string]string) (*grovecorev1alpha1.PodCliqueScalingGroupList, error) {
```
unmarshall: * Rename this function to `doGetPCSGsForPGS`
* Rename the `extraLabels` to just `matchingLabels`

operator/internal/status/replica.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/NVIDIA/grove/operator/internal/component"
+	"github.com/NVIDIA/grove/operator/internal/utils/kubernetes"
+
+	"github.com/go-logr/logr"
+	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ValidateReplicaAvailable validates that replica resources meet availability requirements
+func ValidateReplicaAvailable[T component.GroveCustomResourceType, P interface {
```
unmarshall: * The function says its validating something when it is not really validating anything. It is merely checking if it is available, therefore the return value is bool.
* There is no need for 2 types [P, T]
* The complexity of the function stems from the fact that we are attempting to make things generic when it is not really required.

There are currently 2 reconcilers that call this function - PGS status reconciler and PCSG status reconciler.

IMHO it is far more simpler if these are specific functions that are present in each reconciler itself. Yes there is some duplication of code but that is ok. Can we keep things simple for now and only complicate them when absolutely necessary?

operator/internal/status/replica.go
```diff
@@ -0,0 +1,99 @@
...
+) bool {
+	return validateReplicaState[T, P](
+		logger, resources, expectedCount,
+		getConditions, IsAvailable,
+		resourceType, replicaIndex,
+	)
+}
+
+// ValidateReplicaScheduled validates that replica resources are scheduled
+func ValidateReplicaScheduled[T component.GroveCustomResourceType, P interface {
```
unmarshall: The definition of `ScheduledReplicas` as per the API states the following:
```go
	// ScheduledReplicas is the number of replicas that are scheduled for the PodCliqueScalingGroup.
	// A replica of PodCliqueScalingGroup is considered "scheduled" when at least MinAvailable number
	// of pods in each constituent PodClique has been scheduled.
	ScheduledReplicas int32 `json:"scheduledReplicas,omitempty"`
```
Are we checking/passing minAvailable anywhere?

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -62,30 +63,55 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	return ctrlcommon.ContinueReconcile()
 }
 
+// computeReplicaStatus processes a single PodCliqueScalingGroup replica and returns whether it is scheduled and available.
+func computeReplicaStatus(logger logr.Logger, expectedPCSGReplicaPCLQSize int, pcsgReplicaIndex string, pclqs []grovecorev1alpha1.PodClique) (bool, bool) {
```
unmarshall: We follow a convention where the method that invokes precedes the method that gets called. Thus `computeReplicaStatus` should come after `mutateReplicas`

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -62,30 +63,55 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
+	replicaIndex, _ := strconv.Atoi(pcsgReplicaIndex)
+
+	isScheduled := status.ValidateReplicaScheduled(
```
unmarshall: Instead of using `Validate*` functions just keep it simple. Example:

```go
func mutateReplicas(logger logr.Logger, pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pclqsPerPCSGReplica map[string][]grovecorev1alpha1.PodClique) {
	pcsg.Status.Replicas = pcsg.Spec.Replicas
	var (
		scheduledReplicas int32
		availableReplicas int32
		minAvailable      = int(*pcsg.Spec.MinAvailable)
	)

	for pcsgReplicaIndex, pclqs := range pclqsPerPCSGReplica {
		nonTerminatingPCLQs := lo.Filter(pclqs, func(pclq grovecorev1alpha1.PodClique, _ int) bool {
			return !k8sutils.IsResourceTerminating(&pclq)
		})
		if isPCSGReplicaScheduled(logger, pcsgReplicaIndex, minAvailable, pclqs) {
			scheduledReplicas++
		}
		if isPCSGReplicaAvailable(logger, nonTerminatingPCLQs) {
			availableReplicas++
		}
	}
	logger.Info("Mutating PodCliqueScalingGroup replicas",
		"pcsg", client.ObjectKeyFromObject(pcsg),
		"scheduledReplicas", scheduledReplicas, "availableReplicas", availableReplicas)
	pcsg.Status.ScheduledReplicas = scheduledReplicas
	pcsg.Status.AvailableReplicas = availableReplicas
}

func isPCSGReplicaScheduled(logger logr.Logger, pcsgReplicaIndex string, numExpectedPCLQs int, pclqs []grovecorev1alpha1.PodClique) bool {
	if len(pclqs) < numExpectedPCLQs{
		logger.V(4).Info("PCSG replica is missing some PCLQs, this PCSG replica cannot be considered as scheduled", "pcsgReplicaIndex", pcsgReplicaIndex, "want", numExpectedPCLQs, "got", len(pclqs))
		return false
	}
	return lo.EveryBy(pclqs, func(pclq grovecorev1alpha1.PodClique) bool {
		return k8sutils.IsConditionTrue(pclq.Status.Conditions, grovecorev1alpha1.ConditionTypePodCliqueScheduled)
	})
}

func isPCSGReplicaAvailable(logger logr.Logger, pclqs []grovecorev1alpha1.PodClique) bool {
	// TODO similary this function can be coded.
}

```

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,130 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pgs *grovecorev1alpha1.PodGangSet) ctrlcommon.ReconcileStepResult {
+	if pgs.Status.ObservedGeneration == nil {
+		return ctrlcommon.ContinueReconcile()
+	}
+	// Set basic replica count
+	pgs.Status.Replicas = pgs.Spec.Replicas
+	// Calculate available replicas using PCSG-inspired approach
```
unmarshall, line 35: see comments for PCSG reconciler.

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,130 @@
...
+	// Update the PodGangSet status
+	if err = r.client.Status().Update(ctx, pgs); err != nil {
+		logger.Error(err, "failed to update PodGangSet status")
+		return ctrlcommon.ReconcileWithErrors("failed to update PodGangSet status", err)
+	}
+	return ctrlcommon.ContinueReconcile()
+}
+
+func (r *Reconciler) computePGSAvailableReplicas(ctx context.Context, logger logr.Logger, pgs *grovecorev1alpha1.PodGangSet) (int32, error) {
+	availableReplicas := int32(0)
```
unmarshall: use `var availableReplicas int32` instead. This is more idiomatic go.

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// Calculate expected resource counts per replica (same for all replicas)
+	expectedStandalonePCLQs, expectedPCSGs := r.computeExpectedResourceCounts(pgs)
+
+	// Fetch all PCSGs for this PGS
+	pcsgs, err := componentutils.GetPCSGsForPGS(ctx, r.client, client.ObjectKeyFromObject(pgs))
+	if err != nil {
+		return 0, err
```
unmarshall: use -1 as the value in case of error. Its only symbolic but it clearly represents that this value should never be considered. So in case the caller forgets to handle the error you will immediately see that the value is negative indicating that there is an issue with the code.

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// Fetch all standalone PodCliques for this PGS
+	standalonePCLQs, err := componentutils.GetPodCliqueForPGSNotInPCSG(ctx, r.client, client.ObjectKeyFromObject(pgs))
+	if err != nil {
```
unmarshall, line 75: see the comment above.

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,130 @@
...
+		return 0, err
+	}
+
+	// Group both resources by PGS replica index
+	standalonePCLQsByReplica := componentutils.GroupPCLQsByPGSReplicaIndex(standalonePCLQs)
+	pcsgsByReplica := componentutils.GroupPCSGsByPGSReplicaIndex(pcsgs)
+
+	for replicaIndex := 0; replicaIndex < int(pgs.Spec.Replicas); replicaIndex++ {
+		replicaIndexStr := strconv.Itoa(replicaIndex)
+
```
unmarshall: create a function that takes in the input per replica and gives a bool result.
Call that function `isPGSReplicaAvailable`

operator/internal/component/types.go
Ronkahn21: I moved it due to a cycle import issue with the file from the contoller packge

operator/internal/common/constants.go
Ronkahn21: https://github.com/NVIDIA/grove/pull/135#discussion_r2280974767

operator/internal/component/podclique/pod/syncflow.go
Ronkahn21: It was changed so I could work with client.Object interface

Ronkahn21: I allow passing objects as interfaces and check them

operator/internal/component/utils/generic.go
Ronkahn21: The solution you suggested won't work because the GetLabels function uses a pointer receiver, but you're passing it as a value. My approach isn't the most elegant, but since it's only used internally, it's acceptable to keep it hidden from the public interface.

operator/internal/component/utils/podclique.go
Ronkahn21: It will get all the pod cliques that are not part of any PCSG, not matter on PGS replica

operator/internal/common/constants.go
unmarshall: The issue is that we now have a dependency of `operator/internal/common` in `operator/test/utils` - this should never be done.
All internal packages can depend upon `operator/test/utils` and not vice versa.
Can you please remove any such dependency and then re-evaluate if you still need to move these constants?

operator/internal/component/utils/podclique.go
```diff
@@ -117,3 +112,34 @@ func GetMinAvailableBreachedPCLQInfo(pclqs []grovecorev1alpha1.PodClique, termin
 	slices.Sort(waitForDurations)
 	return pclqCandidateNames, waitForDurations[0]
 }
+
+// GetPodCliquesForPGSNotInPCSG retrieves PodClique objects that are not part of any PodCliqueScalingGroup for the given PodGangSet.
+func GetPodCliquesForPGSNotInPCSG(ctx context.Context, cl client.Client, pgsObjKey client.ObjectKey) ([]grovecorev1alpha1.PodClique, error) {
```
sanjaychatterjee: nit: rename function name to `GetPodCliquesWithParentPGS`

julienmancuso:
Hey @Ronkahn21 , I deployed your changes and tried to deploy a simple PodGangSet . What I'm seeing is that as soon as the pod are created ,the podgangset status is as follows :
Screenshot 2025-08-20 at 3 31 19 PM


no matter what the status of the pods (error, crashloopbackof, running but not ready, ...)

podcliques status are following (I have only 2 podcliques in my example, no PCSG):
Screenshot 2025-08-20 at 3 35 37 PM

julienmancuso:
I tried upgrading an existing install as well as brand new deployment
here is a PodGangSet that allows to reproduce the issue :
```
apiVersion: grove.io/v1alpha1
kind: PodGangSet
metadata:
  name: simple1
  labels:
    app: simple1
spec:
  replicas: 1
  template:
    cliques:
      - name: pca
        spec:
          roleName: rolea
          replicas: 1
          podSpec:
            containers:
              - name: pca
                image: nginx:latest
                resources:
                  requests:
                    cpu: 10m
                readinessProbe: 
                  httpGet:
                    path: /invalid
                    port: 80
                  initialDelaySeconds: 3
                  periodSeconds: 5
                  timeoutSeconds: 1
                  successThreshold: 1
                  failureThreshold: 100
          autoScalingConfig:
            maxReplicas: 5
            metrics:
              - type: Resource
                resource:
                  name: cpu
                  target:
                    type: Utilization
                    averageUtilization: 80
      - name: pcb
        spec:
          roleName: roleb
          replicas: 1
          podSpec:
            containers:
              - name: pcb
                image: nginx:latest
                resources:
                  requests:
                    cpu: 10m
```

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -95,44 +123,29 @@ func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1a
 			Message: fmt.Sprintf("Insufficient scheduled replicas. expected at least: %d, found: %d", minAvailable, scheduledReplicas),
 		}
 	}
-	minAvailableBreachedReplicas := computeMinAvailableBreachedReplicas(logger, pclqsPerPCSGReplica)
-	readyReplicas := scheduledReplicas - minAvailableBreachedReplicas
-	if readyReplicas < minAvailable {
+	availableReplicas := int(pcsg.Status.AvailableReplicas)
+	if availableReplicas < minAvailable {
```
sanjaychatterjee, line 130: A log message here would have been very helpful

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -62,30 +61,59 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	return ctrlcommon.ContinueReconcile()
 }
 
+// computeReplicaStatus processes a single PodCliqueScalingGroup replica and returns whether it is scheduled and available.
+func computeReplicaStatus(logger logr.Logger, expectedPCSGReplicaPCLQSize int, pcsgReplicaIndex string, pclqs []grovecorev1alpha1.PodClique) (bool, bool) {
+	var isAvailable, isScheduled bool
+	NonTermintedPCSGReplicaIndex := lo.Filter(pclqs, func(pclq grovecorev1alpha1.PodClique, _ int) bool {
```
sanjaychatterjee: Is this exported variable? If not, we can make it lower-case.. also suggest renaming to `nonTermintedPCSGPodCliques`

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -62,30 +61,59 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
+			"actualPCSGReplicaPCLQSize", len(NonTermintedPCSGReplicaIndex))
+		return false, false
+	}
+	isScheduled = lo.EveryBy(NonTermintedPCSGReplicaIndex, func(pclq grovecorev1alpha1.PodClique) bool {
+		return k8sutils.IsConditionTrue(pclq.Status.Conditions, grovecorev1alpha1.ConditionTypePodCliqueScheduled)
+	})
+
+	isAvailable = lo.EveryBy(NonTermintedPCSGReplicaIndex, func(pclq grovecorev1alpha1.PodClique) bool {
+		return k8sutils.IsConditionFalse(pclq.Status.Conditions, grovecorev1alpha1.ConditionTypeMinAvailableBreached)
+	})
```
sanjaychatterjee: This may return `isAvailable` as `true` even if `isScheduled` is `false`. Not sure if we want this behavior. Let me also update the issue.

sanjaychatterjee: See https://github.com/NVIDIA/grove/issues/112#issuecomment-3208752674

Ronkahn21: I agree, it should not be available if not scheduled, but this means we have a bug in the underlying component, I will be explicit about it

operator/internal/component/utils/podclique.go
Ronkahn21: I will do it :D, but we need a better name for it 👍

operator/internal/component/utils/podcliquescalinggroup.go
```diff
@@ -47,19 +47,36 @@ func FindScalingGroupConfigForClique(scalingGroupConfigs []grovecorev1alpha1.Pod
...
+	pcsgList, err := doGetPCSGsForPGS(ctx, cl, pgsObjKey, map[string]string{
+		grovecorev1alpha1.LabelPodGangSetReplicaIndex: strconv.Itoa(pgsReplicaIndex),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return pcsgList.Items, nil
+}
+
+func doGetPCSGsForPGS(ctx context.Context, cl client.Client, pgsObjKey client.ObjectKey, matchingLabels map[string]string) (*grovecorev1alpha1.PodCliqueScalingGroupList, error) {
```
unmarshall, line 68: please move this function below `GetPCSGsForPGS`

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -62,30 +61,64 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	return ctrlcommon.ContinueReconcile()
 }
 
+// computeReplicaStatus processes a single PodCliqueScalingGroup replica and returns whether it is scheduled and available.
+func computeReplicaStatus(logger logr.Logger, expectedPCSGReplicaPCLQSize int, pcsgReplicaIndex string, pclqs []grovecorev1alpha1.PodClique, minAvailable int32) (bool, bool) {
```
unmarshall, line 81: can you please move this function below `mutateReplicas`?
The calling function comes first and then the function being called.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
unmarshall: >but this means we have a bug in the underlying component, I will be explicit about it

Can you elaborate on the bug you mention and in which underlying component?

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -62,30 +61,64 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
 		meta.SetStatusCondition(&pcsg.Status.Conditions, newCondition)
 	}
 }
 
-func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pclqsPerPCSGReplica map[string][]grovecorev1alpha1.PodClique) metav1.Condition {
-	minAvailable := int(*pcsg.Spec.MinAvailable)
+func computeMinAvailableBreachedCondition(pcsg *grovecorev1alpha1.PodCliqueScalingGroup) metav1.Condition {
+	// its should not be possible to have a PCSG without MinAvailable defined, but if it happens then we assume that the min available is 1
+	// to prevent case of passing nil to int conversion
+	minAvailable := int(ptr.Deref(pcsg.Spec.MinAvailable, pcsg.Spec.Replicas))
```
unmarshall: I think we should not have this. If the defaulting has not happened then there will be larger set of issues. So it is fine to have a nil pointer dereference instead of hiding that coding error here.

operator/internal/controller/podgangset/reconciler.go
```diff
@@ -65,7 +65,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return result.Result()
 	}
 
-	return r.reconcileSpec(ctx, logger, pgs).Result()
+	if result := r.reconcileSpec(ctx, logger, pgs); ctrlcommon.ShortCircuitReconcileFlow(result) {
```
unmarshall: this will cause the status to never be updated if there is an issue during `reconcileSpec`, do you want that behavior?

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,182 @@
...
+	k8sutils "github.com/NVIDIA/grove/operator/internal/utils/kubernetes"
+
+	"github.com/go-logr/logr"
+	"github.com/samber/lo"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pgs *grovecorev1alpha1.PodGangSet) ctrlcommon.ReconcileStepResult {
+	if pgs.Status.ObservedGeneration == nil {
```
unmarshall: consider the following case:
PGS event arrived - all components are reconciled but when setting `ObservedGeneration` as the final step it fails. So while the resources get created, the status would not reflect that.

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		return -1, err
+	}
+
+	// Group both resources by PGS replica index
+	standalonePCLQsByReplica := componentutils.GroupPCLQsByPGSReplicaIndex(standalonePCLQs)
+	pcsgsByReplica := componentutils.GroupPCSGsByPGSReplicaIndex(pcsgs)
+
+	for replicaIndex := 0; replicaIndex < int(pgs.Spec.Replicas); replicaIndex++ {
+		replicaStandalonePCLQs, replicaPCSGs := r.getPCSGAndPCQLByPGSReplica(replicaIndex, standalonePCLQsByReplica, pcsgsByReplica)
```
unmarshall: do you really need `getPCSGAndPCLQByPGSReplica` -  all it does is get the corresponding value from the maps.

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,182 @@
...
+			// This clique is not part of any scaling group, so it's standalone
+			expectedStandalonePCLQs++
+		}
+	}
+
+	return expectedStandalonePCLQs, expectedPCSGs
+}
+
+func (r *Reconciler) isPGSReplicaAvailable(logger logr.Logger, pgs *grovecorev1alpha1.PodGangSet, replicaIndex int, replicaPCSGs []grovecorev1alpha1.PodCliqueScalingGroup, standalonePCLQs []grovecorev1alpha1.PodClique, expectedPCSGs int, expectedStandalonePCLQs int) bool {
+	if !r.isPGSReplicaPCLAsAvailable(logger, pgs, expectedStandalonePCLQs, replicaIndex, standalonePCLQs) {
```
unmarshall: There is a typo in the name. I think you meant `isPGSReplicaPCLQsAvailable`, right?

operator/internal/controller/podgangset/reconcilestatus.go
```diff
@@ -0,0 +1,182 @@
...
+		logger.Info("PGS replica is not available due to insufficient standalone PodCliques",
+			"pgs", client.ObjectKeyFromObject(pgs), "replicaIndex", replicaIndex,
+			"expectedStandalonePCLQs", expectedStandalonePCLQs, "actualPCLQsCount", len(standalonePCLQs))
+		return false
+	}
+
+	if !r.checkPCSGsAvailability(logger, pgs, expectedPCSGs, replicaIndex, replicaPCSGs) {
```
unmarshall: can we harmonize on the method names?
For PCSGs you have used - `checkPCSGsAvailability` and for standalone PCLQs you have used `isPGSReplicaPCLQsAvailable`

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
sanjaychatterjee: I think in a previous commit, `isAvailable` calculation was dependent on `MinAvailableBreached` condition only. With the latest change, the check `pclq.Status.ReadyReplicas >= minAvailable` is correct. So, now there should never be a case when `isScheduled` is `false` and `isAvailable` is `true`. However, maybe it is better to move the `isAvailable` calculation to be done only when `isScheduled` is true. E.g.
```
if isScheduled {
    // A PodClique is considered available if it has at least minAvailable replicas ready.
	isAvailable = lo.EveryBy(nonTerminatedPCSGPodCliques, func(pclq grovecorev1alpha1.PodClique) bool {
		return pclq.Status.ReadyReplicas >= minAvailable
	})
}
```

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -40,14 +40,13 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
-	}
-
 	if pcsg.Status.ObservedGeneration != nil {
+		pclqsPerPCSGReplica, err := r.getPodCliquesPerPCSGReplica(ctx, pgs.Name, client.ObjectKeyFromObject(pcsg))
+		if err != nil {
+			return ctrlcommon.ReconcileWithErrors(fmt.Sprintf("failed to list PodCliques for PodCliqueScalingGroup: %q", client.ObjectKeyFromObject(pcsg)), err)
+		}
 		mutateReplicas(logger, pcsg, pclqsPerPCSGReplica)
-		mutateMinAvailableBreachedCondition(logger, pcsg, pclqsPerPCSGReplica)
+		mutateMinAvailableBreachedCondition(logger, pcsg)
```
sanjaychatterjee: I don't think this PR needs to touch anything related to MinAvailableBreached condition. We should revert these changes.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -62,30 +61,64 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
+			"conditionType", newCondition.Type,
+			"conditionStatus", newCondition.Status,
+			"reason", newCondition.Reason)
 		meta.SetStatusCondition(&pcsg.Status.Conditions, newCondition)
 	}
 }
 
-func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pclqsPerPCSGReplica map[string][]grovecorev1alpha1.PodClique) metav1.Condition {
-	minAvailable := int(*pcsg.Spec.MinAvailable)
+func computeMinAvailableBreachedCondition(pcsg *grovecorev1alpha1.PodCliqueScalingGroup) metav1.Condition {
```
sanjaychatterjee: We should revert all changes to this function as it does not impact calculation of `AvailableReplicas`.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
Ronkahn21: Yes, as @sanjaychatterjee said, I was referring to it when it used the MinAvailableBreached, but there is a possibility that there is a bug where the PCLQ is ready but not scheduled

operator/internal/controller/podgangset/reconciler.go
Ronkahn21: I agree it should reconcile even if there is an error in the reconciliation of the spec

operator/internal/controller/podgangset/reconcilestatus.go
Ronkahn21: I was aligned with the PCSG reconcile status, but you are correct, I will change the logic as well

Ronkahn21:
@julienmancuso I fixed the issue you found, thanks for letting me know

unmarshall (approved):

Issue #136, opened 2025-08-08
Pod Index/Hostname Changes After Pod Restart


julienmancuso:
I’ve noticed that when a pod (either as from a single PodClique or in a PodScaling group) gets killed and restarted, its index and hostname change. This disrupts endpoint stability, since other components expect the hostname to remain consistent as long as the replica count is 1.

**How to reproduce:**
- deploy the [simple1.example](https://github.com/NVIDIA/grove/blob/main/operator/samples/simple/simple1.yaml)
- list the endpoints : `kubectl get endpoints`
- kill the pca pod associated with hostname `simple1-0-pca-0`

**Actual Behavior:**
- a new pod is created and the hostname `simple1-0-pca-0` is deleted and replaced by `simple1-0-pca-3`

**Expected Behavior:** 
- the enpoint entry for `simple1-0-pca-0` should be kept and pointing to the newly created pod

unmarshall:
_Small correction in the issue description above:_
`simple1.example` spec has 3 replicas for `pca` PodClique but you have mentioned replicas=1 above

@renormalize and I would like to understand the issue a bit more. 
Consider the following case:
```
PCLQ-A:  initial replicas=2 {p0 (hostname index:0), p1 (hostname index:1)}
scaled-out to replicas=5 {p0 (hostname index:0), p1(hostname index:1), p2(hostname index:2), p3(hostname index:3), p4(hostname index:4)}
scale-in to replicas=2 {p0(hostname index:0), p4(hostname index:4)}  
```
Holes can be created since we choose the pods to delete based on unassigned/pending/unknown followed by pods that are not-ready followed by creation timestamp.

So you already have a case where the hostnames are dynamic due to scale-out for non-PCSG PCLQs. With that context we would like to understand if you now get a hostname of `simple1-0-pca-3` then what is the issue?

Additionally:
Since k8s 1.26 pods that are marked for termination (but their containers have not exited yet) can continue to receive traffic see: (https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#terminating) and also see https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1672-tracking-terminating-endpoints#summary

In nutshell:
If pods are marked for termination (have DeletionTimestamp set) but the pod containers have not exited then the `EndpointSlice` will continue to have an entry for such pods with an additional condition set. This ensures that till the containers have exited or the pod reports ready=false the traffic can continue to be routed to the pod which has been marked for termination.

[...]

PR #137, opened 2025-08-08
fix: reuse pod index when pod is replaced

julienmancuso:
Fixing #136


operator/internal/index/tracker.go
```diff
@@ -24,10 +24,32 @@ import (
...
+		return false
+	}
+
+	// Pod is active and should keep its index
+	return true
+}
+
 // GetAvailableIndices returns the `requiredIndicesCount` available indices for pods.
-// Extracts indices from existing pod hostnames and returns the available indices,
+// Extracts indices from existing active pods hostnames and returns the available indices,
```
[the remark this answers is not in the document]
julienmancuso: fixed

operator/internal/index/tracker.go
```diff
@@ -24,10 +24,32 @@ import (
...
+		return false
+	}
+
+	if pod.Status.Phase == corev1.PodFailed && pod.Spec.RestartPolicy == corev1.RestartPolicyNever {
+		return false
+	}
+
+	if pod.Status.Phase == corev1.PodSucceeded {
+		return false
+	}
```
Ronkahn21: This functionality should be in a different package as it is related to pod logic more than index logic 
IsPodInFiniteState

operator/internal/index/tracker.go
```diff
@@ -42,6 +64,11 @@ func extractUsedIndices(existingPods []*corev1.Pod) (sets.Set[int], error) {
 	usedIndices := sets.New[int]()
 
 	for _, pod := range existingPods {
+		// Only consider active pods as occupying their indices
+		if !isPodActiveForIndexAssignment(pod) {
+			continue
```
Ronkahn21, line 53: Add log here

Ronkahn21 (commented):
Currently, we don't allow a finite state for pods; this might change in the future, so it would be better to support both options: one that ignores pods with finite state and the other that does not.

operator/internal/index/tracker.go
julienmancuso: fixed

operator/internal/index/tracker.go
julienmancuso: done

operator/internal/index/tracker.go
```diff
@@ -22,26 +22,37 @@ import (
...
 }
 
 // extractUsedIndices extracts and validates indices from existing pods.
-func extractUsedIndices(existingPods []*corev1.Pod) (sets.Set[int], error) {
+func extractUsedIndices(logger logr.Logger, existingPods []*corev1.Pod) (sets.Set[int], error) {
 	usedIndices := sets.New[int]()
 
 	for _, pod := range existingPods {
+		// Only consider active pods as occupying their indices
+		if !k8sutils.IsPodActive(pod) {
```
sanjaychatterjee, line 51: I am not sure if this is the right place to add this check. Instead, why can't we avoid such pods being added to the existingPCLQPods? Also, rather than checking phase and deletetion timestamps, why can't we leverage the expectations framework? 
cc @unmarshall to comment.

renormalize, line 51: I am not sure if this is correct. In the case where the pod has `terminationGracePeriodSeconds` set to some value, that Pod will get filtered out by this.

But as long as the readiness probe is true, it is possible that the Pod will continue to serve traffic even when it has a deletion timestamp on it. See https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#terminating.

Therefore there will be *2* Pods with identical hostnames. Would it be correct for two pods to have identical hostnames?

We need a different approach to solving this in my opinion.

further information: https://github.com/NVIDIA/grove/issues/136#issuecomment-3183060733.

Ronkahn21: Here's the comment with GitHub-style formatting:

---

@renormalize  You're right about the hostname collision issue during graceful termination. Here's an alternative approach that addresses this specific problem:

## Label-based Service exclusion to prevent hostname conflicts

Instead of filtering at the index allocation level, we can exclude terminating pods from Service endpoints:

### Option 1: preStop hook
- Add `preStop` hook to remove `active=true` label before termination
- Service selector targets only `active=true` pods  
- Terminating pod immediately excluded from DNS/endpoints

### Option 2: Controller-managed (cleaner)
- Controller removes `active=true` label when detecting `pod.DeletionTimestamp != nil`
- No RBAC complexity, centralized within Grove controller

### Result:
- ✅ Terminating pod excluded from Service immediately upon deletion
- ✅ New pod can safely use same hostname without collision
- ✅ Graceful shutdown continues normally for existing connections
- ✅ Keeps current index allocation logic intact

This solves the **"2 Pods with identical hostnames"** issue you identified while preserving the index reuse functionality.

What do you think on this workarround

unmarshall: please see https://github.com/NVIDIA/grove/issues/136#issuecomment-3183060733

unmarshall: see https://github.com/NVIDIA/grove/issues/136#issuecomment-3183060733

unmarshall: `preStop` hook can be used to overwrite the readiness endpoint behavior to force the removal of the endpoint from `EndpointSlice`. This is not correct.
Also, we are currently assuming that the user wishes to stop all ingress to the pod which has `DeletionTimestamp` set. This was the default behavior prior to k8s 1.26 but post that this behavior was changed for a reason. 

Therefore before we jump into the solution space we really need to discuss the problem and its impact and what should be the expected behavior.

Ronkahn21: I am not suggesting to delete endpoint from EndpoinSlice, I am suggesting the use label selectror on the serivce
and remove the active label from the pod itself so it wont be seen by the headless service

unmarshall: @sanjaychatterjee  it is fine to filter the pods here.

unmarshall: I think the PR in its current phase is fine. See the revised https://github.com/NVIDIA/grove/issues/136#issuecomment-3183060733

unmarshall (approved):

operator/internal/index/tracker.go
renormalize: After a bit more discussion, we have realized that this is not actually an issue for us.

The potential problem of traffic being forwarded to a Pod that is both `terminating` and `serving` only happens when all the other Pods that are present in the endpoint slices are also in `terminating`.

> Service proxies will normally ignore endpoints that are `terminating`, but they may route traffic to endpoints that are both `serving` and `terminating` *if all available endpoints are `terminating`*.

We create one headless service for the *entire* `PodGangSet`, which implies Pods from multiple `PodClique`s are specified in the same endpointslice.
The case when all available endpoints are in `terminating` is when all the Pods of the `PodGangSet` replica are also in `terminating`, which happens either in the case of gang termination, or deletion of the replica of the `PodGangSet`.

Therefore, it is safe to conclude that it would not be the case that there would be two Pods that would have the same hostname, where both are `serving`.

renormalize (approved):

unmarshall (approved):

operator/internal/index/tracker.go
renormalize: @Ronkahn21 

We found that the DNS entry does get updated correctly when a Pod has its deletion timestamp set. There has not been an instance where the hostname is present on two different pods in the endpointslices in our experiments.

What we did was the following:

- Set a finalizer on a pod.
- Delete that pod.
- Peform an `nslookup` using the hostname of the pod.

We saw that the address returned by the lookup was actually the address of the pod that was newly created, with the hostname set as the hostname of the pod with a deletion timestamp set on it.

Issue #134, opened 2025-08-07
Default publishNotReadyAddresses to True


nvrohanv:
Multinode startup of inference workers requires worker pods to access the address of the leader pods. We expose this functionality through injected env vars and a headless service. However, when readiness probes are specified, the DNS entries for the pods are not ready until the pods are ready. However in order for the pod to be ready the inference worker needs to start which requires the leader pod's DNS entry to be populated. This is a cyclical dependency and the way to avoid this is by setting publishNotReadyAddresses to true. Since launching mutlinode inference workers is a key capability in Grove, we should default this value to be true.

PR #139, opened 2025-08-11
feat: set default HeadlessServiceConfig to publish not ready addresses

Ronkahn21:
https://github.com/NVIDIA/grove/issues/134
This pull request introduces a defaulting mechanism for the `HeadlessServiceConfig` field within the `PodGangSetTemplateSpec` struct. The main change ensures that if `HeadlessServiceConfig` is not set, it is automatically initialized with sensible defaults. Additionally, the related test has been updated to reflect this new defaulting behavior.

Defaulting logic improvements:

* Added a `defaultHeadlessServiceConfig` function to initialize `HeadlessServiceConfig` with `PublishNotReadyAddresses: true` if it is not already set, and integrated this logic into the `defaultPodGangSetTemplateSpec` function (`operator/internal/webhook/admission/pgs/defaulting/podgangset.go`).

Test updates:

* Updated the `TestDefaultPodGangSet` test case to remove the explicit setting of `HeadlessServiceConfig`, as it is now automatically defaulted (`operator/internal/webhook/admission/pgs/defaulting/podgangset_test.go`).


Ronkahn21:
Fixes #134

Issue #141, opened 2025-08-12
serviceAccountName field in podSpec not taken into account


julienmancuso:
When creating a podGangSet and specifying a serviceAccountName for podSpec(s) , this field is ignored
e.g.

```
apiVersion: grove.io/v1alpha1
kind: PodGangSet
metadata:
  name: sglang-agg
spec:
  replicas: 1
  template:
    cliqueStartupType: CliqueStartupTypeAnyOrder
    cliques:
    - labels:
        nvidia.com/dynamo-component-type: main
        nvidia.com/metrics-enabled: "true"
        nvidia.com/selector: sglang-agg-frontend
      name: frontend
      spec:
        minAvailable: 1
        podSpec:
          ......
          serviceAccountName: planner-serviceaccount
        replicas: 1
        roleName: frontend
    headlessServiceConfig:
      publishNotReadyAddresses: true
```

the associated created pod has this serviceAccountName :

```
serviceAccount: sglang-agg                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          serviceAccountName: sglang-agg
```

PR #142, opened 2025-08-12
fix: allow serviceAccountName override

julienmancuso:
fix for #141

### DIFF
--- a/operator/internal/component/podclique/pod/pod.go
+++ b/operator/internal/component/podclique/pod/pod.go
@@ -168,13 +168,20 @@ func (r _resource) buildResource(pgs *grovecorev1alpha1.PodGangSet, pclq *grovec
 	}
 	pod.Spec = *pclq.Spec.PodSpec.DeepCopy()
 	pod.Spec.SchedulingGates = []corev1.PodSchedulingGate{{Name: podGangSchedulingGate}}
-	pod.Spec.ServiceAccountName = grovecorev1alpha1.GeneratePodServiceAccountName(pgs.Name)
+	setServiceAccountName(pod, pgs)
 	// Add GROVE specific Pod environment variables
 	addEnvironmentVariables(pod, pclq, pgsName, pgsReplicaIndex, podIndex)
 	// Configure hostname and subdomain for service discovery
 	configurePodHostname(pgsName, pgsReplicaIndex, pclq.Name, pod, podIndex)
 	// Conditionally add init container to the Pod to ensure start-up ordering.
 	return appendGroveInitContainerIfNeeded(pgs, pclq, pod)
+}
+
+func setServiceAccountName(pod *corev1.Pod, pgs *grovecorev1alpha1.PodGangSet) {
+	if pod.Spec.ServiceAccountName == "" {
+		// only set the service account name if it is not already set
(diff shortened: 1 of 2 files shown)


Ronkahn21 (dismissed):

renormalize:
Hold merge due to https://github.com/NVIDIA/grove/issues/141#issuecomment-3182141589.

operator/internal/component/podclique/pod/pod.go
```diff
@@ -177,6 +177,13 @@ func (r _resource) buildResource(pgs *grovecorev1alpha1.PodGangSet, pclq *grovec
 	return appendGroveInitContainerIfNeeded(pgs, pclq, pod)
 }
 
+func setServiceAccountName(pod *corev1.Pod, pgs *grovecorev1alpha1.PodGangSet) {
```
unmarshall, line 180: There was an issue in the original code:
```
pod.Spec.ServiceAccountName = grovecorev1alpha1.GeneratePodServiceAccountName(pgs.Name)
```
This assumed that the `ServiceAccount` that is generated by grove for the `init` container is also the service account that must be used by the main container. This is incorrect, thanks for catching that issue.

There can just be one `ServiceAccount` that can be associated to all the containers in a Pod via the Pod specification. Therefore if the user explicitly specifies a `serviceAccountName` then it must be honored and should in no way get overridden. This also means that if no `serviceAccountName` is set then we should continue to respect that and not set any `ServiceAccountName` in the final PodSpec.

In that light this function is not required and should be removed.

operator/internal/component/podclique/pod/pod.go
```diff
@@ -168,7 +168,7 @@ func (r _resource) buildResource(pgs *grovecorev1alpha1.PodGangSet, pclq *grovec
 	}
 	pod.Spec = *pclq.Spec.PodSpec.DeepCopy()
 	pod.Spec.SchedulingGates = []corev1.PodSchedulingGate{{Name: podGangSchedulingGate}}
-	pod.Spec.ServiceAccountName = grovecorev1alpha1.GeneratePodServiceAccountName(pgs.Name)
+	setServiceAccountName(pod, pgs)
```
unmarshall, line 171: you can remove this line (new code) and the line above (existing code) and then you will always respect the `serviceAccountName` that the user has explicitly set.

unmarshall (changes requested):
There is a larger problem that needs to be solved and therefore this PR in its current form will not really work.

The init container requires a dedicated set of permissions which might or might not overlap with the RBAC that the main container needs and is also specified via the service-account that the user creates and sets in the `PodSpec`.

We are evaluating options to ensure that we do not mix RBAC for init container to that of the main container.

operator/internal/component/podclique/pod/pod.go
julienmancuso: @unmarshall , Would it be ok if we add the serviceAccount the user provide in the list of subjects of the default rolebinding that grove operator is maintaining ?

unmarshall: we would like to keep the roles that are assigned to the init container separate than the ones that are assigned to main container. If the user has explicitly specified a service account then there is a backing role/clusterrole and role-binding/clusterrole-binding present which has added the user specified service account as subject. In future if we evolve the RBAC roles for the init container SA then those should not be shared with the main container.

PR #144, opened 2025-08-13
Introduce `cert-controller` to manage TLS artifacts in the operator, and publish charts as OCI images.

renormalize:
Introduce cert-controller to manage TLS artifacts used by webhooks, and publish Helm charts.

* Use `github.com/open-policy-agent/cert-controller` to generate TLS artifacts in the operator, instead of relying on shell scripts to generate these artifacts. This adds a controller to the grove-operator, which can both generate TLS artifacts, and rotate them on expiry.

* Package and publish Helm charts as an OCI artifact.

* `cert-controller` is enabled by default in the operator, and can be configured through `OperatorConfiguration`.

* The operator now has privileges to update `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration` since they need to be patched with the CA Bundle that the cert-controller generates.

* The operator now has privileges to modify `Secret`s since the `cert-controller` updates the webhook server certificate secret with the generated TLS artifacts.

* `ValidatingWebhookConfiguration` name is changed to `podgangset-validating-webhook` from `pgs-validating-webhook` to be consitent with the `MutatingWebhookConfiguration`, and have the same name as the handler.

* Scripts which used to generate TLS artifacts are now removed as they are unnecessary.

* Charts are modified according to the changes needed from the above.


.github/workflows/push-artifacts.yaml
```diff
@@ -71,3 +71,33 @@ jobs:
...
           make --directory=operator docker-build-amd64 VERSION=$PACKAGE_VERSION REGISTRY=$DOCKER_REGISTRY DOCKER_BUILD_ADDITIONAL_ARGS=--push
           make --directory=operator docker-build-arm64 VERSION=$PACKAGE_VERSION REGISTRY=$DOCKER_REGISTRY DOCKER_BUILD_ADDITIONAL_ARGS=--push
+
+      - name: Copy CRDs to charts
+        run: |
+          ./operator/hack/prepare-charts.sh
+
+      - name: Build amd64 Helm chart
+        run: |
+          sed -i "s#name: grove#name: grove-amd64#" operator/charts/Chart.yaml
```
unmarshall: why do we need all these sed expressions?
This requires you to do in-place update of existing chart files and then later revert it. This is a complex way to achieve what you wish to.
Is there a need to have architecture specific helm charts?

docs/api-reference/operator-api.md
```diff
@@ -561,6 +561,22 @@ _Appears in:_
 | `enableProfiling` _boolean_ | EnableProfiling enables profiling via host:port/debug/pprof/ endpoints. |  |  |
 
 
+#### InternalCertificateManagement
```
unmarshall: I would not introduce this in this PR. We defined the scope of this PR to not make any API changes. The scope was defined as - replace the existing TLS artifact generation with Cert-Controller. Lets remove this from this PR as we can improve on the design.

operator/hack/prepare-charts.sh
```diff
@@ -64,63 +62,9 @@ function copy_crds() {
...
-}
-
 function prepare_local_deploy() {
-  if [[ $# -ne 2 ]]; then
-    echo -e "${FUNCNAME[0]} requires 2 arguments: namespace and cert-expiry"
-    exit 1
-  fi
   echo "Copying CRDs to helm charts..."
   copy_crds
-  echo "Generating PKI resources if not present or expired..."
```
unmarshall, line 114: you can remove the function `prepare_local_deploy` - that is no longer required.

operator/api/config/v1alpha1/defaults.go
```diff
@@ -92,6 +92,13 @@ func SetDefaults_ServerConfiguration(serverConfig *ServerConfiguration) {
 	if serverConfig.Metrics.Port == 0 {
 		serverConfig.Metrics.Port = 2752
 	}
+
+	if serverConfig.InternalCertificateManagement == nil {
```
unmarshall: remove all this code from this PR.

operator/api/config/v1alpha1/types.go
```diff
@@ -145,6 +147,13 @@ type Server struct {
 	Port int `json:"port"`
 }
 
+// InternalCertificateManagement defines the configuration for certificates managed by Grove
+type InternalCertificateManagement struct {
```
unmarshall: Remove this code from this PR.

operator/internal/controller/cert.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/NVIDIA/grove/operator/internal/webhook/admission/pgs/defaulting"
+	"github.com/NVIDIA/grove/operator/internal/webhook/admission/pgs/validation"
+
+	cert "github.com/open-policy-agent/cert-controller/pkg/rotator"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	serviceName                      = "grove-operator"
```
unmarshall: change the name to `grove` since you renamed the chart to `grove`

operator/internal/controller/cert.go
```diff
@@ -0,0 +1,66 @@
...
+	certificateAuthorityName         = "grove-operator-ca"
```
unmarshall: change the name to `grove-ca`

operator/internal/controller/cert.go
```diff
@@ -0,0 +1,66 @@
...
+	certificateAuthorityOrganization = "grove-operator"
```
unmarshall: change the Org to `grove` - check @sanjaychatterjee  if this needs to be NVIDIA

operator/internal/controller/cert.go
```diff
@@ -0,0 +1,66 @@
...
+			{
+				Type: cert.Mutating,
+				Name: defaulting.HandlerName,
+			},
+			{
+				Type: cert.Validating,
+				Name: validation.HandlerName,
+			},
+		},
+		RestartOnSecretRefresh: true,
```
unmarshall: as discussed restarting the grove container once whenever the secret is refreshed is not very nice. We will have to relook at this.

operator/internal/controller/manager.go
```diff
@@ -29,27 +32,55 @@ import (
...
-		return nil, err
+// CreateManager creates the manager.
+func CreateManager(operatorCfg *configv1alpha1.OperatorConfiguration) (ctrl.Manager, error) {
+	return ctrl.NewManager(getRestConfig(operatorCfg), createManagerOptions(operatorCfg))
+}
+
+// InitializeManager adds all the controllers and webhooks to the controller-manager using the passed in Config.
+func InitializeManager(mgr ctrl.Manager, operatorCfg *configv1alpha1.OperatorConfiguration) error {
+	if operatorCfg.Server.InternalCertificateManagement != nil && *operatorCfg.Server.InternalCertificateManagement.Enabled {
+		doneCh := make(chan struct{})
```
unmarshall: please rename this channel to `webhookCertsReady` to reflect what it semantically does.
Also create a separate function that will call AddRotator and waits for it to finish.

docs/api-reference/operator-api.md
renormalize: Will remove all API changes 👍.

.github/workflows/push-artifacts.yaml
renormalize: Addressed in [dd36fa2](https://github.com/NVIDIA/grove/pull/144/commits/dd36fa2f9a5c7c9c3f155be9e3b5b742845fed0d).

renormalize: Multi-arch images are now used in the charts. Architecture is not specified anymore.

PR #145, opened 2025-08-14
[WIP] Add cert-manager support as an option

kangclzjc:
Currently we do the certificate generation ourselves. If they already have integrated cert-manager into their cluster, then Grove should give them an option to depend on that, instead


kangclzjc:
> It's a good idea to add some flexibility around certificate generation, however I'd recommend we simplify this. Rather than have flags to enable certmanager, I'd recommend allowing the operator to provide a secret which contains a the certficate, the associated CA bundle. Basically the approach described her in the Elastic Cloud helm chart [here](https://www.elastic.co/guide/en/cloud-on-k8s/2.12/k8s-webhook.html).
> 
> Rationale:
> 
> it's even more flexible, perhaps someone wants to use a different cert manager that build themselves for... reasons any operators who want to provide certmanager, they generally know what they're trying to accomplish and how. All they need is a way to point grove in the right direction 99.9% of deployments won't ever provide this and want to involved minimally in its internal operations, so no need to complicate everything for the small edge cases who do

It's a good idea, maybe it doesn't have conflict with current implementation. I just give a internal example here. For externel mode, we can allow the operator to provide a secret which contains a certficate. Because for some cases, customer may want to easily deploy it in develop mode for internal use/developement. In internal mode, we can hide the complexity of certificate generation for developer.

unmarshall:
@gflarity 
>The github.com/open-policy-agent/cert-controller/pkg/rotator solution have will meet 99% of users needs and seems robust to me.

Well, for short term we will be using the cert-controller but it has several issues which i shared on the shared slack. I will add those points here as well.

- https://github.com/open-policy-agent/cert-controller/issues/50 (Saketh has already pointed this out) and https://github.com/open-policy-agent/cert-controller/issues/35. We both have observed that it can take upto to 100 seconds for the certificates to be ready and injected into the webhook configurations. [@Saketh Kalaga](https://kubernetes.slack.com/team/U0705K2A18X) shared [this](https://ahmet.im/blog/kubernetes-secret-volumes-delay/) with  and this explains this behavior clearly.
- CA rotation is not without downtime. They do not have the capability to have the old and the new CA at the same time so that the switch is without any downtime.
- Possible race conditions: https://github.com/open-policy-agent/cert-controller/issues/13 (open since 2020)
- Problems when running pods in HA with leader election (see https://github.com/open-policy-agent/cert-controller/issues/53)

While we see dev activity on this project but these are minor fixes. Tickets that highlight a bigger gap are open for more than 2-3 years. This is not a very good sign for a project that we wish to have dependency on. Others like `kubewarden` also evaluated cert-controller and ended up having their own custom solution.

gflarity:
I definitely think cert-manager support is a good idea. But let's just keep it simple. Optional CA/Secet helm values should do the trick.

It's shame that cert-controller isn't seeing much investment... Looks like another approach is to use [kube-webhook-certgen
](https://github.com/jet/kube-webhook-certgen) in [helmchart preinstall/postinstall scripts](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#prometheusrules-admission-webhooks).  Seems to work well for the Prometheus community.

renormalize:
Totally agree with @gflarity's suggestion. We should simply allow for external management of these secrets. How the user achieves that, should not be our concern. The approach suggested in https://github.com/NVIDIA/grove/pull/145#issuecomment-3368549831 looks good to me. We should emulate that.

PR #161, opened 2025-08-21
Update validation webhooks

### LINKED ISSUE #154: Enhance validating webook for PodGangSet with additional checks
Following validations have been identified as missing in the current validating webhook:

- [ ] Remove immutable check for `PodGangSetTemplateSpec.PriorityClassName`
- [ ] Add immutable checks for the following fields during updates:
  - [ ] `PodCliqueTemplateSpec.Spec.RoleName`
  - [ ] `PodCliqueTemplateSpec.Spec.MinAvailable`
  - [ ] `PodCliqueTemplateSpec.Spec.StartsAfter` 
  - [ ] `PodGangSetTemplateSpec.StartupType`
  - [ ] If the `StartupType == InOrder || StartupType == Explicit` then the order in which the PodCliques have been specified cannot change.
  - [ ] If the `StartupType == AnyOrder` then the order in which the PodCliques are specified does not matter. We should therefore skip doing this check.
  - [ ] `PodGangSetTemplateSpec.PodCliqueScalingGroupConfigs` - in this slice the constituent PCSG names cannot change. Ensure that an update does not introduce new `PodCliqueScalingGroupConfigs` or remove an existing `PodCliqueScalingGroupConfigs`
  - [ ] `PodCliqueScalingGroupConfig.CliqueNames` - slice of Clique names cannot be changed.
  - [ ] `PodCliqueScalingGroupConfig.MinAvailable`
- [ ] Change the validations for PodSpec. The current set of validations are valid if one updates the PodSpec. However, `PodSpec` is actually a template PodSpec and therefore it should not be constrained to the same set of validations that are applicable if `Pod` is getting updated. In our case we always recreate Pods. Check the validations that are there for `Deployment` and we apply the same set of validations.

Ronkahn21:
This pull request introduces significant improvements to the validation logic for `PodGangSet` updates and expands the builder utilities used in testing. The main focus is on enforcing immutability and correct ordering for clique and scaling group configurations, as well as enhancing test code flexibility.

### Validation Logic Improvements

* Added new validation for `PodCliqueScalingGroupConfigs` to ensure that scaling group composition and key fields (`cliqueNames`, `minAvailable`) cannot be changed during updates.
* Updated clique validation to enforce immutable fields (`roleName`, `minAvailable`, `startsAfter`) and, for certain startup types, to ensure clique order is not changed. The validation now takes `StartupType` into account for order enforcement.
* Removed the validation that enforced immutability of `PriorityClassName` in the top-level update function, allowing this field to be updated.

### Test Builder Enhancements

* Added new builder methods to `PodGangSetBuilder` for setting `PriorityClassName` and `TerminationDelay`, improving test setup flexibility.
* Added new builder methods to `PodCliqueTemplateSpecBuilder` for setting `RoleName` and custom `PodSpec`, allowing more granular control in tests.
* Added a builder method to `PodSpecBuilder` for setting the `RestartPolicy` field.

### Miscellaneous

* Removed the default setting of `RestartPolicy` in test pod specs, making it optional and configurable via the builder.
* Added missing imports in test utility files for completeness. [[1]](diffhunk://#diff-eecef43b34e56f68c37bf6cdbcc8a1352a612f8f6d7bbcd6d9e853d9344a0e81R21-R22) [[2]](diffhunk://#diff-a1d68e1aabf67a33e527d183ede17351b6d94272df5756724289c9febfe09beeR20-R21)

Fixes #154

### DIFF
--- a/operator/internal/webhook/admission/pgs/validation/podgangset.go
+++ b/operator/internal/webhook/admission/pgs/validation/podgangset.go
@@ -477,7 +477,6 @@ func (v *pgsValidator) validateUpdate(oldPgs *grovecorev1alpha1.PodGangSet) erro
 	allErrs := field.ErrorList{}
 	fldPath := field.NewPath("spec")
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(v.pgs.Spec.ReplicaSpreadConstraints, oldPgs.Spec.ReplicaSpreadConstraints, fldPath.Child("replicaSpreadConstraints"))...)
-	allErrs = append(allErrs, apivalidation.ValidateImmutableField(v.pgs.Spec.Template.PriorityClassName, oldPgs.Spec.Template.PriorityClassName, fldPath.Child("priorityClassName"))...)
 	allErrs = append(allErrs, validatePodGangSetSpecUpdate(&v.pgs.Spec, &oldPgs.Spec, fldPath)...)
 	return allErrs.ToAggregate()
 }
@@ -491,9 +490,10 @@ func validatePodGangSetSpecUpdate(newSpec, oldSpec *grovecorev1alpha1.PodGangSet
 func validatePodGangTemplateSpecUpdate(newSpec, oldSpec *grovecorev1alpha1.PodGangSetTemplateSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	allErrs = append(allErrs, validatePodCliqueUpdate(newSpec.Cliques, oldSpec.Cliques, fldPath.Child("cliques"))...)
+	allErrs = append(allErrs, validatePodCliqueUpdate(newSpec.Cliques, oldSpec.Cliques, newSpec.StartupType, fldPath.Child("cliques"))...)
 	allErrs = append(allErrs, apivalidation.ValidateImmutableField(newSpec.StartupType, oldSpec.StartupType, fldPath.Child("cliqueStartupType"))...)
 	allErrs = append(allErrs, validatePodGangSchedulingPolicyConfigUpdate(newSpec.SchedulingPolicyConfig, newSpec.SchedulingPolicyConfig, fldPath.Child("schedulingPolicyConfig"))...)
+	allErrs = append(allErrs, validatePodCliqueScalingGroupConfigsUpdate(newSpec.PodCliqueScalingGroupConfigs, oldSpec.PodCliqueScalingGroupConfigs, fldPath.Child("podCliqueScalingGroups"))...)
 
 	return allErrs
 }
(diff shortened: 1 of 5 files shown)


operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -504,26 +504,83 @@ func validatePodGangSchedulingPolicyConfigUpdate(newConfig, oldConfig *grovecore
...
 	allErrs := field.ErrorList{}
 
 	if len(newCliques) != len(oldCliques) {
 		allErrs = append(allErrs, field.Forbidden(fldPath, "not allowed to change clique composition"))
 	}
 
 	// Create a map of old cliques by name for efficient lookup
-	oldCliqueMap := lo.SliceToMap(oldCliques, func(clique *grovecorev1alpha1.PodCliqueTemplateSpec) (string, *grovecorev1alpha1.PodCliqueTemplateSpec) {
-		return clique.Name, clique
+	// this allows to check the type and order without deal with non exits indexes in a slice of the old cliques
```
gflarity:
```suggestion
	// this allows checking the type and order without dealing with non-existent indexes in a slice of the old cliques
```

operator/internal/webhook/admission/pgs/validation/podgangset.go
```diff
@@ -504,26 +504,83 @@ func validatePodGangSchedulingPolicyConfigUpdate(newConfig, oldConfig *grovecore
...
+	// if the length the old cliques and new cliques is different is an error but we dont return it immediately so farther validation can be done
```
gflarity:
```suggestion
	// if the length the old cliques and new cliques is different, this is an error but we don't return it immediately so that further validation can be done
```

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -23,488 +23,157 @@ import (
...
-		cliqueNames     []string
-		expectError     bool
-		expectedErrMsg  string
-		expectedErrPath string
+		description      string
+		pgsName          string
+		cliqueNames      []string
+		scalingGroups    []grovecorev1alpha1.PodCliqueScalingGroupConfig
+		expectError      bool
+		expectedErrMsg   string
```
gflarity, line 73: Outside the scope of this PR as it wasn't introduced in it, but I'd recommend replacing string-based error testing with error codes and constants to decouple test logic from error message formatting, ensuring tests remain valid when messages are updated while providing a stable contract for error identification.

gflarity (changes requested):
Generally LGTM, left a few suggestions.  Also asked cursor to check for test case coverage and this came up as some missing suggestions. I'm still getting acquainted with the code base, but they seem like good suggestions?
```
   - Container image changes (should pass)
   - Tolerations additions (should pass)
   - Tolerations removals (should fail)
   - Other immutable pod field changes (should fail)
 ```

File: operator/internal/webhook/admission/pgs/validation/podgangset_test.go [hunk 4]
```diff
@@ -807,11 +418,304 @@ func TestValidatePodCliqueUpdate(t *testing.T) {
...
+				return pgs
+			},
+			setupNewPGS: func() *grovecorev1alpha1.PodGangSet {
+				pgs := createDummyPodGangSet("test")
+				pgs.Spec.Template.Cliques[0].Spec.MinAvailable = ptr.To(int32(1))
+				return pgs
+			},
+			expectError:    true,
+			expectedErrMsg: "field is immutable",
+		},
```
renormalize: `minAvailable` fields are defaulted, right? So when would they be `nil`?

operator/internal/webhook/admission/pgs/validation/podgangset_test.go
```diff
@@ -743,61 +326,89 @@ func TestValidatePodCliqueUpdate(t *testing.T) {
...
+				createDummyPodCliqueTemplate("a"),
+				createDummyPodCliqueTemplate("b"),
 			},
 			newCliques: []*grovecorev1alpha1.PodCliqueTemplateSpec{
-				createDummyPodCliqueTemplate("worker"),
+				createDummyPodCliqueTemplate("b"),
+				createDummyPodCliqueTemplate("a"),
 			},
 			expectError: false,
 		},
```
renormalize: `StartupType` is defaulted by the kube-apiserver due the schema of the CRD. Not sure when it would be `nil`.

renormalize (dismissed):

unmarshall (approved):

PR #215, opened 2025-10-07
Remove deadlock when deploying PCS with ComputeDomain

### LINKED ISSUE #209: Grove Schedule Gating Deadlocks with DRA on Blackwell
### What happened?

On Blackwell, ComputeDomains require all pods that claim them to be scheduled before the domain itself can become ready. Since ComputeDomains are fixed in size and cannot autoscale, if I launch a 3p1d DeepSeek-R1 model with Dynamo on a full NVL72, I need to define the ComputeDomain across all 18 nodes.

However, Grove handles 3p1d differently: it assigns one prefill and one decode pod to the base PodGang, while the remaining two prefill workers are gated until the base PodGang reaches the Ready state. This creates a deadlock. The base PodGang (1p1d) cannot transition to Ready because the ComputeDomain itself is not ready until all pods for the full 3p1d configuration are scheduled. Meanwhile, the additional prefill workers cannot be scheduled until the base PodGang is marked Ready.

Instead of requiring non-BasePodGang pods to wait until the BasePodGang is ready, the scheduling gate should lift as soon as the BasePodGang is scheduled. The risk is that if the additional prefill workers start too early, they might consume resources needed for the BasePodGang (the MinAvailable set) to succeed. For example, the BasePodGang could be scheduled onto a node that then fails during the lengthy startup process (often 30–60 minutes for large models). If other pods have already occupied the remaining capacity, there may not be enough resources to reschedule the BasePodGang. Even so, this seems like an acceptable tradeoff, since node health and recovery should ideally be managed by Kubernetes.

unmarshall:
#### What type of PR is this?

/kind bug

When using DRA [ComputeDomain](https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/a8e6de09e1e2a51fd445445d1be6055911c741d6/api/nvidia.com/resource/v1beta1/computedomain.go#L38) on NVL72 then it will enter into a deadlock for a PCS which has `minAvailable` < `replicas` for PCSG.

Consider the following scenario:
* Create a PCS for disaggregated inference with prefills and decodes being modeled as separate PCSG.
* Create a `ComputeDomain` with numNodes = 18
* Ensure that your `minAvailable` < `replicas` for the prefill or/and decode PCSGs.

Lets assume that you have a total of 12 pods (across prefill and decode PCSGs) that are part of the base pod gang. These have a reference to compute domain. Grove operator will first try and schedule and start the base pod gang. When the based pod gang is ready, only then it will lift the `scheduling gates` for the pods belonging to the scaled pod gangs.

When based pod gang pods are started, they will not come up as `ComputeDomain.Status.Nodes` only lists 12 nodes and `ComputeDomain.Status.Status` is set to `NotReady` as it expects 18 nodes and currently there are only 12. Due to scheduling gates on the remaining 6 pods (belong to the scaled pod gang) kube-scheduler (or equivalent) never picks these pods to be scheduled.

As a result you end up into a deadlock situation where the base pod gang pods will never start as the ComputeDomain resource on which they have a dependency never gets `Ready` and since based pod gang pods do not start the scaled pod gang pods will continue to be schedule gated.

This PR resolves this deadlock by changing the predicate to remove the scheduling gates for pods from scaled pod gangs.
Instead of waiting for the pods to be ready (num readyPods >= minAvailable) it now only waits for  (num scheduledPods >= minAvailable).

Fixes #209 

```release-note
NONE
```

```docs
NONE
```


renormalize (approved):

operator/internal/controller/podclique/components/pod/syncflow.go
```diff
@@ -331,11 +329,11 @@ func (r _resource) isBasePodGangReady(ctx context.Context, logger logr.Logger, n
 
 		// CRITICAL READINESS CHECK: Compare actual ready pods vs required minimum
```
sanjaychatterjee: Please update this comment.

unmarshall: removed the comment as it was not really required.

sanjaychatterjee (approved):

julienmancuso (approved):

PR #217, opened 2025-10-15
Add Core Concepts Tutorial

nvrohanv:
- Adding tutorial for introducing core Grove Primitives. Examples can be run on local kind cluster
- Allowing make kind-up to create arbitrary number of fake nodes

### DIFF
--- a/operator/e2e/setup/k8s_clusters.go
+++ b/operator/e2e/setup/k8s_clusters.go
@@ -18,6 +18,7 @@ package setup
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
@@ -26,6 +27,7 @@ import (
 
 	"github.com/ai-dynamo/grove/operator/e2e/utils"
 	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/filters"
 	dockerclient "github.com/docker/docker/client"
 	"github.com/k3d-io/k3d/v5/pkg/client"
 	"github.com/k3d-io/k3d/v5/pkg/config"
@@ -66,32 +68,90 @@ type NodeLabel struct {
 
 // ClusterConfig holds configuration for creating a k3d cluster
 type ClusterConfig struct {
-	Name             string      // Name of the k3d cluster
-	ServerNodes      int         // Number of server (non-worker) nodes
-	WorkerNodes      int         // Number of worker nodes (called agents in k3s terminology)
-	Image            string      // Docker image to use for k3d cluster (e.g., "rancher/k3s:v1.28.8-k3s1")
-	HostPort         string      // Port on host to expose Kubernetes API (e.g., "6550")
-	LoadBalancerPort string      // Load balancer port mapping in format "hostPort:containerPort" (e.g., "8080:80")
-	NodeLabels       []NodeLabel // Kubernetes labels to apply with specific node filters
-	WorkerNodeTaints []NodeTaint // Taints to apply to worker nodes
-	WorkerMemory     string      // Memory allocation for worker/agent nodes (e.g., "150m")
-	EnableRegistry   bool        // Enable built-in Docker registry
-	RegistryPort     string      // Port for the Docker registry (e.g., "5001")
+	Name              string      // Name of the k3d cluster
+	ControlPlaneNodes int         // Number of Control Plane Nodes (k3s calls these server nodes)
+	WorkerNodes       int         // Number of worker nodes (called agents in k3s terminology)
+	Image             string      // Docker image to use for k3d cluster (e.g., "rancher/k3s:v1.28.8-k3s1")
+	HostPort          string      // Port on host to expose Kubernetes API (e.g., "6550")
+	LoadBalancerPort  string      // Load balancer port mapping in format "hostPort:containerPort" (e.g., "8080:80")
+	NodeLabels        []NodeLabel // Kubernetes labels to apply with specific node filters
+	WorkerNodeTaints  []NodeTaint // Taints to apply to worker nodes
+	WorkerMemory      string      // Memory allocation for worker/agent nodes (e.g., "150m")
+	EnableRegistry    bool        // Enable built-in Docker registry
+	RegistryPort      string      // Port for the Docker registry (e.g., "5001")
 }
 
 // DefaultClusterConfig returns a sensible default cluster configuration
 func DefaultClusterConfig() ClusterConfig {
 	return ClusterConfig{
-		Name:             "test-k3d-cluster",
-		ServerNodes:      1,
-		WorkerNodes:      2,
-		Image:            "rancher/k3s:v1.28.8-k3s1",
-		HostPort:         "6550",
-		LoadBalancerPort: "8080:80",
-		WorkerMemory:     "150m",
-		EnableRegistry:   false,
-		RegistryPort:     "5001",
+		Name:              "test-k3d-cluster",
+		ControlPlaneNodes: 1,
+		WorkerNodes:       2,
+		Image:             "rancher/k3s:v1.28.8-k3s1",
+		HostPort:          "6550",
+		LoadBalancerPort:  "8080:80",
+		WorkerMemory:      "150m",
+		EnableRegistry:    false,
+		RegistryPort:      "5001",
 	}
+}
+
+// ensureClusterDoesNotExist removes any stale k3d cluster with the same name from previous runs.
+func ensureClusterDoesNotExist(ctx context.Context, clusterName string, logger *utils.Logger) error {
+	cluster := &k3d.Cluster{Name: clusterName}
+
+	existingCluster, err := client.ClusterGet(ctx, runtimes.Docker, cluster)
+	if err != nil {
+		if errors.Is(err, client.ClusterGetNoNodesFoundError) {
+			return nil
+		}
+		return fmt.Errorf("failed to inspect existing k3d cluster %s: %w", clusterName, err)
+	}
+
+	logger.Warnf("🧹 Removing stale k3d cluster '%s' before setup", clusterName)
+	if err := client.ClusterDelete(ctx, runtimes.Docker, existingCluster, k3d.ClusterDeleteOpts{}); err != nil {
+		return fmt.Errorf("failed to delete existing k3d cluster %s: %w", clusterName, err)
+	}
+
+	return nil
+}
+
+// ensureRegistryDoesNotExist removes any stale k3d registry container from previous runs.
+func ensureRegistryDoesNotExist(ctx context.Context, clusterName string, logger *utils.Logger) error {
+	registryContainerName := fmt.Sprintf("k3d-%s-registry", clusterName)
+
+	dockerClient, err := dockerclient.NewClientWithOpts(dockerclient.FromEnv, dockerclient.WithAPIVersionNegotiation())
+	if err != nil {
+		return fmt.Errorf("failed to create Docker client: %w", err)
+	}
+	defer dockerClient.Close()
+
+	filterArgs := filters.NewArgs()
+	filterArgs.Add("name", registryContainerName)
+
+	containers, err := dockerClient.ContainerList(ctx, container.ListOptions{All: true, Filters: filterArgs})
+	if err != nil {
+		return fmt.Errorf("failed to list Docker containers: %w", err)
+	}
+
+	if len(containers) == 0 {
+		return nil
+	}
+
+	for _, c := range containers {
+		displayName := registryContainerName
+		if len(c.Names) > 0 {
+			displayName = strings.TrimPrefix(c.Names[0], "/")
+		}
+
+		logger.Warnf("🧹 Removing stale k3d registry container %s (%s) before cluster setup", displayName, c.ID[:12])
+
+		if err := dockerClient.ContainerRemove(ctx, c.ID, container.RemoveOptions{Force: true, RemoveVolumes: true}); err != nil {
+			return fmt.Errorf("failed to remove existing registry container %s: %w", displayName, err)
+		}
+	}
+
+	return nil
 }
 
 // SetupCompleteK3DCluster creates a complete k3d cluster with Grove, Kai Scheduler, and NVIDIA GPU Operator
@@ -230,7 +290,7 @@ func SetupK3DCluster(ctx context.Context, cfg ClusterConfig, logger *utils.Logge
 		ObjectMeta: types.ObjectMeta{
 			Name: cfg.Name,
 		},
-		Servers: cfg.ServerNodes,
+		Servers: cfg.ControlPlaneNodes,
 		// k3s calls these agents, but we call them worker nodes
 		Agents: cfg.WorkerNodes,
 		Image:  cfg.Image,
@@ -292,6 +352,16 @@ func SetupK3DCluster(ctx context.Context, cfg ClusterConfig, logger *utils.Logge
 		return nil, nil, fmt.Errorf("failed to transform config: %w", err)
 	}
 
+	if err := ensureClusterDoesNotExist(ctx, k3dConfig.Name, logger); err != nil {
+		return nil, nil, err
+	}
+
+	if cfg.EnableRegistry {
+		if err := ensureRegistryDoesNotExist(ctx, cfg.Name, logger); err != nil {
+			return nil, nil, err
+		}
+	}
+
 	// this is the cleanup function, we always return it now so the caller can decide to use it or not
 	cleanup := func() {
 		logger.Debug("🗑️ Deleting cluster...")
@@ -304,7 +374,7 @@ func SetupK3DCluster(ctx context.Context, cfg ClusterConfig, logger *utils.Logge
 
 	// Create cluster
 	logger.Debugf("🚀 Creating cluster '%s' with %d server(s) and %d worker node(s)...",
-		k3dConfig.Name, cfg.ServerNodes, cfg.WorkerNodes)
+		k3dConfig.Name, cfg.ControlPlaneNodes, cfg.WorkerNodes)
 
 	if err := client.ClusterRun(ctx, runtimes.Docker, k3dConfig); err != nil {
 		return nil, cleanup, fmt.Errorf("failed to create cluster: %w", err)
--- a/operator/e2e/setup/shared_cluster.go
+++ b/operator/e2e/setup/shared_cluster.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ai-dynamo/grove/operator/api/common"
 	"github.com/ai-dynamo/grove/operator/e2e/utils"
 	"github.com/docker/docker/api/types/image"
 	dockerclient "github.com/docker/docker/client"
@@ -40,6 +41,33 @@ const (
 
 	defaulPollInterval = 1 * time.Second
 )
+
+// resourceType represents a Kubernetes resource type for cleanup operations
+type resourceType struct {
+	group    string
+	version  string
+	resource string
+	name     string
+}
+
+// groveManagedResourceTypes defines all resource types managed by Grove operator that need to be tracked for cleanup
+var groveManagedResourceTypes = []resourceType{
+	// Grove CRDs
+	{"grove.io", "v1alpha1", "podcliquesets", "PodCliqueSets"},
+	{"grove.io", "v1alpha1", "podcliquescalinggroups", "PodCliqueScalingGroups"},
+	{"grove.io", "v1alpha1", "podgangsets", "PodGangSets"},
+	{"scheduler.grove.io", "v1alpha1", "podgangs", "PodGangs"},
+	{"grove.io", "v1alpha1", "podcliques", "PodCliques"},
+	// Kubernetes core resources
+	{"", "v1", "services", "Services"},
+	{"", "v1", "serviceaccounts", "ServiceAccounts"},
+	{"", "v1", "secrets", "Secrets"},
+	// RBAC resources
+	{"rbac.authorization.k8s.io", "v1", "roles", "Roles"},
+	{"rbac.authorization.k8s.io", "v1", "rolebindings", "RoleBindings"},
+	// Autoscaling resources
+	{"autoscaling", "v2", "horizontalpodautoscalers", "HorizontalPodAutoscalers"},
+}
 
 // SharedClusterManager manages a shared (singleton) k3d cluster for E2E tests
 type SharedClusterManager struct {
@@ -74,17 +102,20 @@ func (scm *SharedClusterManager) Setup(ctx context.Context, testImages []string)
 		return nil
 	}
 
+	// Track whether setup completed successfully
+	var setupSuccessful bool
+
 	// Configuration for maximum cluster size needed (28 worker nodes + 3 server nodes)
 	customCfg := ClusterConfig{
-		Name:             "shared-e2e-test-cluster",
-		ServerNodes:      3,
-		WorkerNodes:      28,     // Maximum needed across all tests
-		WorkerMemory:     "150m", // 150m memory per agent node to fit one workload pod
-		Image:            "rancher/k3s:v1.33.5-k3s1",
-		HostPort:         "6560", // Use a different port to avoid conflicts
-		LoadBalancerPort: "8090:80",
-		EnableRegistry:   true,
-		RegistryPort:     "5001",
+		Name:              "shared-e2e-test-cluster",
+		ControlPlaneNodes: 3,
+		WorkerNodes:       28,     // Maximum needed across all tests
+		WorkerMemory:      "150m", // 150m memory per agent node to fit one workload pod
+		Image:             "rancher/k3s:v1.33.5-k3s1",
+		HostPort:          "6560", // Use a different port to avoid conflicts
+		LoadBalancerPort:  "8090:80",
+		EnableRegistry:    true,
+		RegistryPort:      "5001",
 		NodeLabels: []NodeLabel{
 			{
 				Key: "node_role.e2e.grove.nvidia.com",
@@ -115,6 +146,12 @@ func (scm *SharedClusterManager) Setup(ctx context.Context, testImages []string)
 	scm.logger.Info("🚀 Setting up shared k3d cluster for all e2e tests...")
 
 	restConfig, cleanup, err := SetupCompleteK3DCluster(ctx, customCfg, relativeSkaffoldYAMLPath, scm.logger)
+	// Defer cleanup on error - only call if setup was not successful and we have a cleanup function
+	defer func() {
+		if !setupSuccessful && cleanup != nil {
+			cleanup()
+		}
+	}()
 	if err != nil {
 		return fmt.Errorf("failed to setup shared k3d cluster: %w", err)
 	}
@@ -125,7 +162,6 @@ func (scm *SharedClusterManager) Setup(ctx context.Context, testImages []string)
 	// Create clientset from restConfig
 	clientset, err := kubernetes.NewForConfig(restConfig)
 	if err != nil {
-		cleanup()
 		return fmt.Errorf("failed to create clientset: %w", err)
 	}
 	scm.clientset = clientset
@@ -133,21 +169,18 @@ func (scm *SharedClusterManager) Setup(ctx context.Context, testImages []string)
 	// Create dynamic client from restConfig
 	dynamicClient, err := dynamic.NewForConfig(restConfig)
 	if err != nil {
-		cleanup()
 		return fmt.Errorf("failed to create dynamic client: %w", err)
 	}
 	scm.dynamicClient = dynamicClient
 
 	// Setup test images in registry
 	if err := setupRegistryTestImages(scm.registryPort, testImages); err != nil {
-		cleanup()
 		return fmt.Errorf("failed to setup registry test images: %w", err)
 	}
 
 	// Get list of worker nodes for cordoning management
 	nodes, err := clientset.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
 	if err != nil {
-		cleanup()
 		return fmt.Errorf("failed to list nodes: %w", err)
 	}
 
@@ -160,6 +193,7 @@ func (scm *SharedClusterManager) Setup(ctx context.Context, testImages []string)
 
 	scm.logger.Infof("✅ Shared cluster setup complete with %d worker nodes", len(scm.workerNodes))
 	scm.isSetup = true
+	setupSuccessful = true
 	return nil
 }
 
@@ -169,15 +203,10 @@ func (scm *SharedClusterManager) PrepareForTest(ctx context.Context, requiredWor
 		return fmt.Errorf("shared cluster not setup")
 	}
 
-	// First, uncordon all nodes to reset state
-	for _, nodeName := range scm.workerNodes {
-		if err := utils.CordonNode(ctx, scm.clientset, nodeName, false); err != nil {
-			return fmt.Errorf("failed to uncordon node %s: %w", nodeName, err)
-		}
-	}
-
-	// Cordon nodes that are not needed for this test
-	if requiredWorkerNodes < len(scm.workerNodes) {
+	if requiredWorkerNodes > len(scm.workerNodes) {
+		return fmt.Errorf("required worker nodes (%d) is greater than the number of worker nodes in the cluster (%d)", requiredWorkerNodes, len(scm.workerNodes))
+	} else if requiredWorkerNodes < len(scm.workerNodes) {
+		// Cordon nodes that are not needed for this test
 		nodesToCordon := scm.workerNodes[requiredWorkerNodes:]
 		for _, nodeName := range nodesToCordon {
 			if err := utils.CordonNode(ctx, scm.clientset, nodeName, true); err != nil {
@@ -203,10 +232,10 @@ func (scm *SharedClusterManager) CleanupWorkloads(ctx context.Context) error {
 	}
 
 	// Step 2: Poll for all resources and pods to be cleaned up
-	if err := scm.waitForAllResourcesAndPodsDeleted(ctx, defaulPollInterval); err != nil {
+	if err := scm.waitForAllGroveManagedResourcesAndPodsDeleted(ctx, defaulPollInterval); err != nil {
 		scm.logger.Warnf("timeout waiting for resources and pods to be deleted: %v", err)
 		// List remaining resources and pods for debugging
-		scm.listRemainingResources(ctx)
+		scm.listRemainingGroveManagedResources(ctx)
 		scm.listRemainingPods(ctx, "default")
 	}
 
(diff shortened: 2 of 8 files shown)


docs/installation.md
athreesh, line 1: nitpick on smth that you didn't necessarily add but 

"Let's try scaling the PodCliqueScalingGroup from 1 to 2 replicas:
kubectl scale pcsg simple1-0-pcsg --replicas=2"

didn't work for me. I had to run kubectl scale pcsg simple1-0-sga --replicas=2

athreesh: I also had to cd into /operator before the make targets worked. probably worth adding that step to make it "just work"

  - Add "Navigate to the operator directory: `cd operator`" before this step
  - Or change the command to: `cd operator && make kind-up`

athreesh: I think the kind-up script currently has a bug and doesn't create the kubeconfig file properly. I had to manually
 create it.

would love a gutcheck on this. i had to run:

  ```bash
  # Create the kubeconfig file in the expected location
  kind get kubeconfig --name grove-test-cluster > hack/kind/kubeconfig

  # Set the KUBECONFIG environment variable (from the operator/ directory)
  export KUBECONFIG=$(pwd)/hack/kind/kubeconfig

  Also add a note that users need to keep the terminal open or re-export KUBECONFIG in new sessions.

docs/user_guide/overview.md
```diff
@@ -0,0 +1,24 @@
+# Grove Core Concepts Tutorial
+
+This tutorial provides a comprehensive overview of Grove's core concepts: **PodClique**, **PodCliqueSet**, and **PodCliqueScalingGroup**. Through practical examples, you'll learn how to deploy and scale inference workloads from simple single-node setups to complex multi-node distributed systems. Since Grove's creation was motivated by inference the examples are tailored to inference but the core idea is to demonstrate how Grove's primitives allow you to express a collection of single node and multinode components that require tighter coupling from a scheduling (and in future releases network topology) aspect.
+
+## Prerequisites
+
+Before starting this tutorial, ensure you have:
+- [A Grove demo cluster running.](../installation.md#developing-grove) Make sure to run `make kind-up FAKE_NODES=40`, set `KUBECONFIG` env variable as directed in the instructions, and run `make deploy`
+- [Or an actual k8s cluster with Grove installed.](../installation.md#deploying-grove) If you choose this path make sure to adjust the tolerations in the example to fit your cluster
```
gflarity:
```suggestion
- [A Kubernetes cluster with Grove installed.](../installation.md#deploying-grove) If you choose this path make sure to adjust the tolerations in the example to fit your cluster
```

docs/user_guide/overview.md
```diff
@@ -0,0 +1,24 @@
...
+- Basic understanding of Kubernetes concepts
```
gflarity:
```suggestion
-  A basic understanding of Kubernetes concepts, [this is a good place to start](https://kubernetes.io/docs/tutorials/kubernetes-basics/). 
```

docs/user_guide/overview.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+## Core Concepts Overview
+
+### PodClique: The Fundamental Unit
+A **PodClique** is the core building block in Grove. It represents a group of pods with the same exact configuration (similar to a Deployment) that can be used in a standalone manner to represent single-node components of your inference system, or can represent roles within a multi-node component such as leader and worker.
```
gflarity: Node has a very specific Kubernetes meaning, and it's also one of those terms that  is super overloaded generally.  So I would remove it, but if there's a more very specific term we should use that instead. 
```suggestion
A **PodClique** is the core building block in Grove. It represents a group of pods with the same exact configuration (similar to a Deployment) that can be used in a standalone manner to represent single components of your inference system, or can represent roles within a multi-node component such as leader and worker.
```

docs/user-guide/core-concepts/overview.md
```diff
@@ -0,0 +1,24 @@
+# Grove Core Concepts Tutorial
```
gflarity, line 1: Tgis is an overview, I'd recommend Core Concepts and Tutorial get moved into docs/user_guide/pcs_and_pclq_intro.md as we reference back anyways.  I'd also rename that into tutorial.

docs/user_guide/overview.md
```diff
@@ -0,0 +1,24 @@
+# Grove Core Concepts Tutorial
+
+This tutorial provides a comprehensive overview of Grove's core concepts: **PodClique**, **PodCliqueSet**, and **PodCliqueScalingGroup**. Through practical examples, you'll learn how to deploy and scale inference workloads from simple single-node setups to complex multi-node distributed systems. Since Grove's creation was motivated by inference the examples are tailored to inference but the core idea is to demonstrate how Grove's primitives allow you to express a collection of single node and multinode components that require tighter coupling from a scheduling (and in future releases network topology) aspect.
+
+## Prerequisites
+
+Before starting this tutorial, ensure you have:
+- [A Grove demo cluster running.](../installation.md#developing-grove) Make sure to run `make kind-up FAKE_NODES=40`, set `KUBECONFIG` env variable as directed in the instructions, and run `make deploy`
```
gflarity: I'd swap the ordering as unless we make a separate quick start guide, the tutorial is where folks will go to get this up and running in a real cluster for their POC. Might as well prioritize that. Just my 0.02.

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
```diff
@@ -0,0 +1,217 @@
+# PodCliqueSet and PodClique
+
+In this guide we go over some hands-on examples showcasing how to use PodCliqueSet and PodClique
+
+Refer to [Overview](./overview.md) for instructions on how to run the examples in this guide.
```
gflarity, line 5: Referencing the overview is helpful, but it seems like the prerequiests should go in this file as they are directly related. This file should probably be renamed to tutorial. We might also want to turn this into a quick start as well minus the local/fake references as folks will want that.

docs/user_guide/pcs_and_pclq_intro.md
```diff
@@ -0,0 +1,217 @@
+# PodCliqueSet and PodClique
+
+In this guide we go over some hands-on examples showcasing how to use PodCliqueSet and PodClique
+
+Refer to [Overview](./overview.md) for instructions on how to run the examples in this guide.
+
+## Example 1: Single-Node Aggregated Inference
+
+In this simplest scenario, each pod is a complete model instance that can service requests. This is mapped to a single standalone PodClique within the PodCliqueSet. The PodClique provides horizontal scaling capabilities at the model replica level similar to a Deployment, and the PodCliqueSet provides horizontal scaling capabilities at the system level (useful for things such as blue-green deployments and spreading across availability zones).
```
gflarity:
```suggestion
In this simplest scenario, each pod is a complete model instance that can service requests. This is mapped to a single standalone PodClique within the PodCliqueSet. Under this scenario, a PodCliqueSet is essentially just a Deployment. The PodClique provides horizontal scaling capabilities at the model replica level. The PodCliqueSet provides horizontal scaling capabilities at the system level (useful for things such as blue-green deployments and spreading across availability zones).
```

docs/user_guide/pcs_and_pclq_intro.md
```diff
@@ -0,0 +1,217 @@
...
+  name: single-node-aggregated
+  namespace: default
+spec:
+  replicas: 1
+  template:
+    cliques:
+    - name: model-worker
+      spec:
+        replicas: 2
+        podSpec:
```
gflarity:
```suggestion
        podSpec: # This is a standard Kubernetes PodSpec
```

docs/user-guide/core-concepts/pcsg_intro.md
```diff
@@ -0,0 +1,319 @@
+# PodCliqueScalingGroup
```
gflarity, line 1: I'd just put these all into a single tutorial file rather than split them up.

docs/user-guide/core-concepts/pcsg_intro.md
```diff
@@ -0,0 +1,319 @@
...
+          containers:
+          - name: model-worker
+            image: nginx:latest
+            command: ["/bin/sh"]
+            args: ["-c", "echo 'Model Worker (Aggregated) on node:' && hostname && sleep 3600"]
+            resources:
+              requests:
+                cpu: "4"
+                memory: "8Gi"
+    podCliqueScalingGroups:
```
gflarity, line 63: After reading through the examples I think we should call out when you'd increase the PCS replicas vs when you would increase the PSG replicas, because this first example seems equivalent?

docs/user-guide/core-concepts/takeaways.md
```diff
@@ -0,0 +1,203 @@
+# Takeaways
+
+Refer to [Overview](./overview.md) for instructions on how to run the examples in this guide.
```
gflarity, line 3: Again, I think this should go into one big file. I think you can add a TOC with markdown.

operator/hack/kind-up.sh
```diff
@@ -135,7 +141,16 @@ function kind::create_cluster() {
   mkdir -p "${KIND_CONFIG_DIR}"
   echo "Creating kind cluster ${CLUSTER_NAME}..."
   kind::generate_config
+  
+  # If KUBECONFIG is not already set (e.g., by the Makefile), set it to our default location
```
gflarity, line 145: Just fyi, ~/.kube/config is the defacto default without KUBECONFIG. New clusters get added there which can be good or bad. But you don't absolutely need to have KUBECONFIG set all the time.

gflarity:
Oh, one more thing. I think a quickstart would also be useful (that doesn't involve the fakes). It's the first thing I look for a POC.

docs/user-guide/core-concepts/pcsg_intro.md
nvrohanv: I initially had it like that but my only worry was that it was too long so i decided to break it up into the concepts it actually exposes. What are your thoughts on that? I feel like pcs and pclq are one set of concepts and then pcsg is different so splitting them out makes the whole thing digestible.

docs/user-guide/core-concepts/pcsg_intro.md
nvrohanv: Sure can add a line about that

operator/hack/kind-up.sh
nvrohanv: Right the way it was set up the kind cluster relies on a kubeconfig different than the default, are you saying when i make a new cluster its auto added to the default config and we just need to instruct the user to set the context?

docs/installation.md
nvrohanv: @athreesh was the issue that you were in the root directory? Also ya it creates the kubeconfig in grove directory so you have to re-export because you have to use that kubeconfig instead of default (I think), I wasnt sure if we wanted to mess with the user's default one so our options are either

1. make it be in default and just tell user to select the context (@gflarity below was mentioning something about default kube_config if its true that its added there anyways then this might be a good option)
2. just explicitly call out that you have to make sure to set kubeconfig from the operator directory and re-export

which do you prefer?

Regarding the first two items i'll add that in

README.md
```diff
@@ -1,69 +1,84 @@
+[![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/NVIDIA/grove)
+
```
renormalize: Badges have already been added.

```suggestion
```

README.md
```diff
@@ -1,69 +1,84 @@
...
+Get Grove running in 5 minutes:
 
-Additionally, as workloads scale in size and complexity, achieving efficient resource utilization and optimal performance depends on capabilities such as all-or-nothing (“gang”) scheduling, topology-aware placement, prescriptive startup ordering, and independent scaling of components. Grove is designed with these needs as first-class citizens — providing native abstractions for expressing scheduling intent, topology constraints, startup dependencies, and per-component scaling behaviors that can be directly interpreted by underlying schedulers.
+```bash
+# 1. Create a local kind cluster
+cd operator && make kind-up
 
-## Core Concepts
+# 2. Deploy Grove
+kind get kubeconfig --name grove-test-cluster > hack/kind/kubeconfig
```
renormalize: This is not needed? The output of `make kind-up` is the following:

```sh
❯ make kind-up
...
Creating kind cluster grove-test-cluster...
Generating kind cluster config...
...
You can now use your cluster with:

kubectl cluster-info --context kind-grove-test-cluster

...
📌 NOTE: To target the newly created kind cluster, please run the following command:

 export KUBECONFIG=/Users/renormalize/code/grove/operator/hack/kind/kubeconfig
```

The necessary `KUBECONFIG` that is to be exported, is printed out as the part of the output of the make target.

It will also *always* be written to `$(pwd)/hack/kind/kubeconfig` if they're creating a kind cluster (unless users intentionally set their KUBECONFIG path as something else).

We can therefore get rid of the `kind get kube...` command here, and only keep the `export KUBECONFIG...` below.

```suggestion
```

docs/installation.md
renormalize: If you already have a `KUBECONFIG` env var exported in your session, the kind-up.sh script will use the path as specified in that env var. This path however is still printed out as a part of the script.

Shell sessions with an already set KUBECONFIG is not expected for most people getting started, since they would obviously not want other KUBECONFIG files overwritten. If they happen to, the script's output notfies where the kind cluster's KUBECONFIG is, which is the path they'd exported.

It is also a bad idea to overwrite the default KUBECONFIG at `~/.kube/config`.

I'd like to know in what cases the script is going wrong so we can fix it, instead of making the quick start have one more step by including the `kind get kubeconfig....` step.

renormalize (commented):
1/n as I've not gotten a chance to look through the entire PR yet.

README.md
```diff
@@ -1,69 +1,84 @@
...
-| [PodCliqueSet](operator/api/core/v1alpha1/podcliqueset.go)          | The top-level Grove object that defines a group of components managed and colocated together. Also supports autoscaling with topology aware spread of PodCliqueSet replicas for availability.            |
-| [PodClique](operator/api/core/v1alpha1/podclique.go)                | A group of pods representing a specific role (e.g., leader, worker, frontend). Each clique has an independent configuration and supports custom scaling logic.                                           |
-| [PodCliqueScalingGroup](operator/api/core/v1alpha1/scalinggroup.go) | A set of PodCliques that scale and are scheduled together as a gang. Ideal for tightly coupled roles like prefill leader and worker.                                                                               |
-| [PodGang](scheduler/api/core/v1alpha1/podgang.go)                   | The scheduler API that defines a unit of gang-scheduling. A PodGang is a collection of groups of similar pods, where each pod group defines a minimum number of replicas guaranteed for gang-scheduling. |
+# 4. Watch it scale
+kubectl get pcs,pclq,pcsg,pg,pod -owide
```
renormalize: - There's no `kubectl scale`, or load being put on the pods. How would it scale?
- The command here only gets the resources once. Even if resources were scaling, the user would not notice it happening in real time, since there is no watch on this command.

Was this the intended comment?

```suggestion
# 4. Fetch the resources created by grove
kubectl get pcs,pclq,pcsg,pg,pod -owide
```

docs/user-guide/core-concepts/overview.md
renormalize, line 1: Nit: The convention the repository uses for directories, and other repositories in the Kubernetes ecosystem in general are hyphens as sepeartors, instead of underscores. Can this be changed to `docs/user-guide/overview.md`?

All directories/files introduced in this PR can be hyphenated instead of underscored.

docs/user_guide/overview.md
renormalize: Agree with @gflarity. Someone more familiar with Kubernetes than inference would immediately jump to thinking this is talking about workloads that run on one host machine vs multiple, which is not necessarily the case when Grove is being used.

docs/user_guide/overview.md
```diff
@@ -0,0 +1,24 @@
...
+## Core Concepts Overview
+
+### PodClique: The Fundamental Unit
+A **PodClique** is the core building block in Grove. It represents a group of pods with the same exact configuration (similar to a Deployment) that can be used in a standalone manner to represent single-node components of your inference system, or can represent roles within a multi-node component such as leader and worker.
+
+### PodCliqueScalingGroup: Multi-Node Coordination
+A **PodCliqueScalingGroup** coordinates multiple PodCliques that must scale together, preserving specified replica ratios across roles (e.g. leader/worker) in multi-node components.
+
+### PodCliqueSet: The Inference Service Container
+A **PodCliqueSet** contains all the inference components for a complete service. It manages one or more PodCliques or PodCliqueScalingGroups that work together to provide inference capabilities. Can be replicated in order to provide blue-green deployment and spread across availability zones.
```
renormalize: Are we sure we want to talk about the "blue-green" deployment here?

Also, a PCS with multiple replicas can be spread across any toplogy, not just an availability zone.

Are we mentioning this here because grove does not support killing an entire PCS replica and recreating it in one go at the moment, as it is known that some frameworks have components that don't really play well with each other during upgrades.

I'm not really sure what all we want to mention here as this is an overview only.

docs/user_guide/pcs_and_pclq_intro.md
renormalize: I wouldn't be very comfortable with "essentially just a Deployment", since we have behavior like gang termination, which would never happen in a Deployment. "similar to a Deployment" is fine, in my opinion.

I understand that this is only an analogy, but I wouldn't go too far with it either. Also, PodCliques are closer to ReplicaSets, than Deployments.

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
```diff
@@ -0,0 +1,217 @@
...
+- `replicas: 2` creates 2 instances for horizontal scaling
+- Each replica handles complete inference pipeline
+- Tolerations allow scheduling on fake nodes for demo, remove if you are trying to deploy on a real cluster
+
+### **Deploy:**
+```bash
+# actual single-node-aggregated.yaml file is in samples/user_guide/concept_overview, change path accordingly
+kubectl apply -f [single-node-aggregated.yaml](../../operator/samples/user_guide/concept_overview/single-node-aggregated.yaml)
+kubectl get pods -l app.kubernetes.io/part-of=single-node-aggregated -o wide
+```
```
renormalize, line 52: The user will be in `grove/operator` anyway while setting up the kind cluster, before they reach here. Even if they're not, I think it's trivial to find the file.

```suggestion
```bash
kubectl apply -f samples/user_guide/concept_overview/single-node-aggregated.yaml
kubectl get pods -l app.kubernetes.io/part-of=single-node-aggregated -o wide
```
```

File: docs/user-guide/core-concepts/pcs_and_pclq_intro.md [hunk 18]
```diff
@@ -0,0 +1,217 @@
...
+After running you will observe there are now 4 `model-worker` pods belonging to the `single-node-aggregated-0` PodCliqueSet
+```
+rohanv@rohanv-mlt operator % kubectl get pods -l app.kubernetes.io/part-of=single-node-aggregated -o wide
+NAME                                          READY   STATUS    RESTARTS   AGE   IP            NODE            NOMINATED NODE   READINESS GATES
+single-node-aggregated-0-model-worker-jvgdd   1/1     Running   0          22s   10.244.11.0   fake-node-011   <none>           <none>
+single-node-aggregated-0-model-worker-n9gcq   1/1     Running   0          44m   10.244.7.0    fake-node-007   <none>           <none>
+single-node-aggregated-0-model-worker-tjb78   1/1     Running   0          22s   10.244.8.0    fake-node-008   <none>           <none>
+single-node-aggregated-0-model-worker-zfhbb   1/1     Running   0          44m   10.244.15.0   fake-node-015   <none>           <none>
+```
+You can also scale the entire PodCliqueSet. For instance run the following command to increase the replicas on `single-node-aggregated` to 3. `pcs` is short for PodCliqueSet and can be used to reference PodCliqueSet as a resource in kubectl commands.
```
renormalize, line 77:
```suggestion
You can also scale the entire PodCliqueSet (short name `pcs`). For instance run the following command to increase the replicas on `single-node-aggregated` to 3.
```

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
```diff
@@ -0,0 +1,217 @@
...
+single-node-aggregated-1-model-worker-kkmsq   1/1     Running   0          74s    10.244.10.0   fake-node-010   <none>           <none>
+single-node-aggregated-1-model-worker-pn5cm   1/1     Running   0          74s    10.244.15.0   fake-node-015   <none>           <none>
+single-node-aggregated-2-model-worker-h5xqk   1/1     Running   0          74s    10.244.3.0    fake-node-003   <none>           <none>
+single-node-aggregated-2-model-worker-p4kjj   1/1     Running   0          74s    10.244.16.0   fake-node-016   <none>           <none>
+```
+
+Note how there are pods belonging to `single-node-aggregated-0`, `single-node-aggregated-1`, and `single-node-aggregated-2`, representing 3 different PodCliqueSets. Also note how `single-node-aggregated-1` and `single-node-aggregated-2` only have two replicas in their `model-worker` PodClique. This is in line with k8s patterns and occurs because the template that was applied (single-node-aggregated.yaml) specified the number of replicas on `model-worker` as 2. To scale them up you would have to apply `kubectl scale pclq` commands like previously done for `single-node-aggregated-0-model-worker` above.
```
renormalize: Only one PodCliqueSet, and 3 replicas of the PodCliqueSet. Not 3 PodCliqueSets. 3 PodCliqueSets means 3 different PodCliqueSet resources.

```suggestion
Note how there are pods belonging to `single-node-aggregated-0`, `single-node-aggregated-1`, and `single-node-aggregated-2`, which are 3 different PodCliques, belonging to different PodCliqueSet replicas. Also note how `single-node-aggregated-1` and `single-node-aggregated-2` only have two replicas in their `model-worker` PodClique. This is because these resources are created based on the template defined for the PodCliques in the PodCliqueSet, which specifies 2 replicas per PodClique.
```

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
```diff
@@ -0,0 +1,217 @@
...
+- Two separate PodCliques: `prefill` and `decode`
+- Independent scaling: We start with 3 prefill workers, 2 decode workers and can scale them independently based on workload characteristics
+- Different resource requirements for each component are supported (in the example prefill requests 2 cpu and decode only 1)
+
+### **Deploy**
+```bash
+# actual single-node-disaggregated.yaml file is in samples/user_guide/concept_overview, change path accordingly
+kubectl apply -f [single-node-disaggregated.yaml](../../operator/samples/user_guide/concept_overview/single-node-disaggregated.yaml)
+kubectl get pods -l app.kubernetes.io/part-of=single-node-disaggregated -o wide
+```
```
renormalize, line 171:
```suggestion
```bash
kubectl apply -f samples/user_guide/concept_overview/single-node-disaggregated.yaml
kubectl get pods -l app.kubernetes.io/part-of=single-node-disaggregated -o wide
```
```

File: docs/user_guide/pcs_and_pclq_intro.md [hunk 21]
```diff
@@ -0,0 +1,217 @@
...
+single-node-disaggregated-0-prefill-tnw26   1/1     Running   0          29s   10.244.17.0   fake-node-017   <none>           <none>
+single-node-disaggregated-0-prefill-xvvtk   1/1     Running   0          29s   10.244.11.0   fake-node-011   <none>           <none>
+single-node-disaggregated-0-prefill-zglvn   1/1     Running   0          29s   10.244.9.0    fake-node-009   <none>           <none>
+```
+Note how within the `single-node-disaggregated-0` PodCliqueSet replica there are pods from the `prefill` PodClique and `decode` PodClique
+
+### **Scaling**
+You can scale the `prefill` and `decode` PodCliques the same way the [`model-worker` PodClique was scaled](#scaling) in the previous example. 
+
+Additionally, the `single-node-disaggregated` PodCliqueSet can be scaled the same way the `single-node-aggregated` PodCliqueSet was scaled in the previous example. We show an exampleto demonstrate how when PodCliqueSets are scaled, all constituent PodCliques are replicated, underscoring why scaling PodCliqueSets should be treated as scaling the entire system (usually for blue-green deployment or high availability across zones).
```
renormalize:
```suggestion
Additionally, the `single-node-disaggregated` PodCliqueSet can be scaled the same way the `single-node-aggregated` PodCliqueSet was scaled in the previous example. We show an example to demonstrate how when PodCliqueSets are scaled, all constituent PodCliques are replicated, underscoring why scaling PodCliqueSets should be treated as scaling the entire system (usually for blue-green deployment or high availability across zones).
```

docs/user-guide/core-concepts/pcsg_intro.md
renormalize: Let's split across files, like it is being done right now. A single file will be too large to consume.

docs/user_guide/pcsg_intro.md
```diff
@@ -0,0 +1,319 @@
...
+- Creates 2 model isntances, each with 1 leader + 3 workers
+- Total pods: 2 × (1 leader + 3 workers) = 8 pods
+- Scaling the group preserves the 1:3 leader-to-worker ratio
+
+### **Deploy:**
+```bash
+# actual multi-node-aggregated.yaml file is in samples/user_guide/concept_overview, change path accordingly
+kubectl apply -f [multi-node-aggregated.yaml](../../operator/samples/user_guide/concept_overview/multi-node-aggregated.yaml)
+kubectl get pods -l app.kubernetes.io/part-of=multinode-aggregated -o wide
+```
```
renormalize:
```suggestion
```bash
kubectl apply -f samples/user_guide/concept_overview/multi-node-aggregated.yaml
kubectl get pods -l app.kubernetes.io/part-of=multinode-aggregated -o wide
```
```

File: docs/user_guide/pcsg_intro.md [hunk 23]
```diff
@@ -0,0 +1,319 @@
...
+- Decode PCSG: 1 replica × (1 leader + 2 workers) = 3 pods
+- Each PCSG can scale independently based on workload demands
+- Each PCSG can have different resource allocations
+
+### **Deploy**
+```bash
+# actual multi-node-disaggregated.yaml is under /operator/samples/user_guide/concept_overview. Adjust paths accordingly
+kubectl apply -f [multi-node-disaggregated.yaml](../../operator/samples/user_guide/concept_overview/multi-node-disaggregated.yaml)
+kubectl get pods -l app.kubernetes.io/part-of=multinode-disaggregated -o wide
+```
```
renormalize:
```suggestion
```bash
kubectl apply -f samples/user_guide/concept_overview/multi-node-disaggregated.yaml
kubectl get pods -l app.kubernetes.io/part-of=multinode-disaggregated -o wide
```
```

File: docs/user_guide/takeaways.md [hunk 24]
```diff
@@ -0,0 +1,203 @@
...
+- **Multi-node prefill**: 1 replica × (1 leader + 4 workers) = 5 pods
+- **Multi-node decode**: 1 replica × (1 leader + 2 workers) = 3 pods
+- **Total**: 11 pods providing a complete inference pipeline
+
+**Deploy and explore:**
+```bash
+# Actual complete-inference-pipeline.yaml is under /operator/samples/user_guide/concept_overview, adjust path accordingly
+kubectl apply -f [complete-inference-pipeline.yaml](../../operator/samples/user_guide/concept_overview/complete-inference-pipeline.yaml)
+kubectl get pods -l app.kubernetes.io/part-of=comp-inf-ppln -o wide
+```
```
renormalize:
```suggestion
```bash
kubectl apply -f samples/user_guide/concept_overview/complete-inference-pipeline.yaml
kubectl get pods -l app.kubernetes.io/part-of=comp-inf-ppln -o wide
```
```

File: operator/hack/kind-up.sh [hunk 25]
```diff
@@ -135,7 +141,16 @@ function kind::create_cluster() {
...
+  # This ensures kubectl commands target the correct cluster
+  if [ -z "${KUBECONFIG:-}" ]; then
+    export KUBECONFIG="${KIND_CONFIG_DIR}/kubeconfig"
+    echo "Setting KUBECONFIG to ${KUBECONFIG}"
+  fi
```
renormalize, line 150: Not needed. See https://github.com/ai-dynamo/grove/blob/7252d2c9febc1a6ae055f054b7399155442eebcf/operator/Makefile#L30

README.md
athreesh: agreed. It was a ref to increasing # of replicas in prev command but wording could be better.

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
nvrohanv: the yaml files are in a very different place from this doc, that said i guess its fair to assume that user starts from the root so maybe i can put a command cd  operator and then have this?

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
nvrohanv: will update to say PodCliqueSetReplicas but I don't want to mention the different PodCliques because I believe the PodCliqueName would be `single-node-aggregated-1-model-worker`.  Also I prefer the more verbose explanation of why there is only two since it calls out the yamls that were applied

operator/hack/kind-up.sh
nvrohanv: will it handle setting it to `${KIND_CONFIG_DIR}/kubeconfig` if its not set. basically is $(KUBECONFIG_PATH) automatically configured?

docs/user_guide/pcs_and_pclq_intro.md
nvrohanv: agree on keeping it as Deployment. Discussions with others also said its closer to ReplicaSet, should we be saying that or is Deployment better conceptually because its what people associate with what you would create to deploy pods that you want to replicate horizontally?

docs/user_guide/overview.md
nvrohanv: I think the point is more to motivate some ways you would use a PCS otherwise people might wonder why you need the top level. Currently the way its phrased makes it sound like thats all you can use it for so I'll adjust it to make clear blue-green and availability zone spread are some uses. I think it is good to mention them so people's mental model for what kind of thing PCS is used for is aligned with how we want them to think about it.

docs/user_guide/overview.md
nvrohanv: I think we should make a separate quick start guid, will address that in a separate PR

docs/user-guide/core-concepts/overview.md
nvrohanv: sure

nvrohanv: done

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
nvrohanv: I think inference people that don't know k8s will not understand what is meant by short name so I wanted to explicitly call out that you need to run that with kubectl (ofc the command shows it but in this case i think its. better to be explicit)

docs/user_guide/pcs_and_pclq_intro.md
nvrohanv: done

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
nvrohanv: reworded to address the confusion between pcs replica and pclw

operator/hack/kind-up.sh
nvrohanv: following up on this @renormalize

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
nvrohanv: addressed.

docs/user-guide/core-concepts/pcs_and_pclq_intro.md
nvrohanv: addressed

README.md
```diff
@@ -11,21 +7,24 @@ Modern AI inference workloads need capabilities that Kubernetes doesn't provide
...
 - **Topology-aware placement** - NVLink-connected GPUs or workloads shouldn't be scattered across nodes
 
-Grove is a Kubernetes API that provides a single declarative interface for orchestrating any AI inference workload — from simple, single-pod deployments to complex multi-node, disaggregated systems.
+Grove is a Kubernetes API that provides a single declarative interface for orchestrating any AI inference workload — from simple, single-pod deployments to complex multi-node, disaggregated systems. Grove lets you scale your multinode inference deployment from a single replica to data center scale, supporting tens of thousands of GPUs. It allows you to describe your whole inference serving system in Kubernetes - e.g. prefill, decode, routing or any other component - as a single Custom Resource Definition (CRD). From that one spec, the platform coordinates hierarchical gang scheduling, topology‑aware placement, multi-level autoscaling and explicit startup ordering. You get precise control of how the system behaves without stitching together scripts, YAML files, or custom controllers.
```
renormalize: It's a CR, not a CRD.

```suggestion
Grove is a Kubernetes API that provides a single declarative interface for orchestrating any AI inference workload — from simple, single-pod deployments to complex multi-node, disaggregated systems. Grove lets you scale your multinode inference deployment from a single replica to data center scale, supporting tens of thousands of GPUs. It allows you to describe your whole inference serving system in Kubernetes - e.g. prefill, decode, routing or any other component - as a single Custom Resource (CR). From that one spec, the platform coordinates hierarchical gang scheduling, topology‑aware placement, multi-level autoscaling and explicit startup ordering. You get precise control of how the system behaves without stitching together scripts, YAML files, or custom controllers.
```

README.md
```diff
@@ -11,21 +7,24 @@ Modern AI inference workloads need capabilities that Kubernetes doesn't provide
...
 **One API. Any inference architecture.**
 
 ## Quick Start
 
 Get Grove running in 5 minutes:
+[![Go Report Card](https://goreportcard.com/badge/github.com/ai-dynamo/grove/operator)](https://goreportcard.com/report/github.com/NVIDIA/grove/operator)
+[![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
+[![GitHub Release](https://img.shields.io/github/v/release/ai-dynamo/grove)](https://github.com/ai-dynamo/grove/releases/latest)
+[![Discord](https://dcbadge.limes.pink/api/server/D92uqZRjCZ?style=flat)](https://discord.gg/GF45xZAX)
+[![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/ai-dynamo/grove)
```
renormalize: @unmarshall please provide your opinions on the badges here.

docs/installation.md
```diff
@@ -252,6 +266,94 @@ As specified in the [README.md](../README.md) and the [docs](../docs), there are
...
+   ```
+3. For kind clusters, you may need to install metrics-server separately.
+
+### Getting Help
+
+If you encounter issues not covered here:
+1. Check the [GitHub Issues](https://github.com/NVIDIA/grove/issues) for similar problems
+2. Join the [Grove mailing list](https://groups.google.com/g/grove-k8s)
+3. Start a [discussion thread](https://github.com/NVIDIA/grove/discussions)
+
```
renormalize, line 338: Discord channel for grove can be added here as well?

```suggestion
4. Ask a question on the [ai-dynamo discord](https://discord.gg/GF45xZAX)

File: operator/hack/kind-up.sh (hunk 25 above)
Comment (renormalize) [reply operator/hack/kind-up.sh:150]: It is. See: https://github.com/ai-dynamo/grove/blob/141155e714cb5df99f1a291a548aa9fe911062f5/operator/Makefile#L21.

**Review by renormalize — APPROVED** [review]

**Comment by renormalize** [comment]
@nvrohanv reminder to clean up the git commit message while merging this PR, as the PR has been merged with main multiple times, and there are a significant number of commmits. It would be nice to keep the commit message short.

File: operator/hack/kind-up.sh (hunk 12 above)
Comment (gflarity) [reply operator/hack/kind-up.sh:145]: Sorry, to clarify, if the environment variable KUBECONFIG is not set, the default is always ~/.kube/config. There's no need to set a default, this is the default. So the wording should be more like:


```suggestion
  # If  KUBECONFIG isn't set, the default location ($HOME/.kube/config) will be used, which might not be desirable. 
```

README.md
```diff
@@ -1,74 +1,80 @@
-> [!NOTE]
->
-> :construction_worker: `This project site is currently under active construction, keep watching for announcements!`
-
 # Grove
 
+Modern AI inference workloads need capabilities that Kubernetes doesn't provide out-of-the-box:
```
sanjaychatterjee:
```suggestion
Modern AI inference workloads need capabilities that Kubernetes natively doesn't provide out-of-the-box:
```

README.md
```diff
@@ -1,74 +1,80 @@
...
+
+- **Gang scheduling** - Prefill and decode pods must start together or not at all
```
sanjaychatterjee:
```suggestion
- **Gang scheduling** - Disaggregated inference requires all-or-nothing semantics for prefill and decode component pods
```

README.md
```diff
@@ -1,74 +1,80 @@
...
+- **Grouped scaling** - Tightly-coupled components that need to scale as a unit
```
sanjaychatterjee:
```suggestion
- **Scaling groups** - Multiple tightly-coupled component groups that need to scale as individual units
```

README.md
```diff
@@ -1,74 +1,80 @@
...
+- **Startup ordering** - Different components in a workload which must start in an explicit ordering
```
sanjaychatterjee:
```suggestion
- **Startup ordering** - Different components in a workload which must be scheduled together but start in an explicit ordering
```

README.md
```diff
@@ -1,74 +1,80 @@
...
+- **Topology-aware placement** - NVLink-connected GPUs or workloads shouldn't be scattered across nodes
```
sanjaychatterjee:
```suggestion
- **Topology-aware placement** - Network optimized placement, e.g. NVLink domains, will minimize communication costs and maximize inference performance
```

sanjaychatterjee (approved):

PR #224, opened 2025-10-21
Grove proposal/topology

Ronkahn21:
#### What type of PR is this?

  /kind documentation
  /kind feature

  This PR adds a comprehensive design document for topology-aware scheduling in the Grove operator. The design introduces a flexible topology system that enables optimal placement of multinode inference
  workloads based on cluster network topology.

  Key components:
  - **TopologyDomain CRD**: Admin-configured cluster topology hierarchy mapping friendly names to node labels
  - **Operator Configuration**: Selects active topology via `--topology-domain-name` argument
  - **TopologyConstraint API**: User-specified packing requirements in workloads (PodCliqueSet, PodCliqueScalingGroup, PodClique)
  - **Automatic Optimization**: Out-of-box topology optimization via auto-generated preferred constraints
  - **KAI Scheduler Integration**: Automatic Kueue Topology generation and three-level constraint translation

  The design addresses critical requirements for multinode inference workloads including network locality, coordinated placement, and latency optimization.

  Fixes #

  - This is a design document only - no implementation changes
  - Design includes detailed API specifications, validation rules, and scheduler contract
  - Addresses review feedback including RBAC, runtime behavior, and edge cases
  - Immutability enforced at multiple levels to ensure scheduling consistency
  - Explicit failure modes prevent silent degradation of topology features

  suggest change

### DIFF
--- a/docs/designs/topology.md
+++ b/docs/designs/topology.md
@@ -200,12 +200,20 @@ spec:
 
 - Level names must be from predefined set: region, zone, datacenter, block, rack, host, numa (enum validation)
 - Each level `domain` and `key` must be unique
-- Mutation webhook automatically reorders levels to match predefined ordering (Region > Zone > DataCenter > Block >
-  Rack > Host > Numa)
 - Admins can skip intermediate levels (e.g., define only region, rack, host)
 - Partially immutable after creation (see Characteristics above for details)
 - Deletion protection via controller finalizer (blocks deletion while PodCliqueSet resources reference this topology OR
   topology is enabled)
+
+**Mutation Webhook:**
+
+The mutation webhook automatically modifies ClusterTopology resources on CREATE operations:
+
+- **Level Reordering**: Automatically reorders levels to match predefined ordering (Region > Zone > DataCenter > Block >
+  Rack > Host > Numa)
+- **Subset Support**: Admins can define any subset of levels; webhook reorders only the levels provided
+- **Consistency**: Ensures all ClusterTopology resources follow the same ordering convention across the cluster
+- **Example**: If admin defines levels as `[host, rack, region]`, webhook reorders to `[region, rack, host]`
 
 #### ClusterTopology Controller
 
@@ -276,7 +284,7 @@ topology:
     - Workloads without topology constraints: no impact (checked via label presence)
 6. For new workloads:
     - Topology constraints validated against ClusterTopology CR
-    - Mutation webhook adds `grove.io/topology-name` label to PodCliqueSet
+   - Mutation webhook adds `grove.io/topology-name` label to PodCliqueSet (see Mutation Webhook section)
 
 **Disabling Topology (topology.enabled: true → false):**
 
@@ -299,7 +307,7 @@ topology:
 5. Restart operator
 6. Operator reconciles existing workloads with updated topology configuration
 
-*note: in future, we may support dynamic updates to ClusterTopology without disabling topology first.*
+*note: in the future, we may support dynamic updates to ClusterTopology without disabling topology first.*
 
 ### 2. Operator API Changes (Grove CRDs)
 
@@ -371,6 +379,17 @@ TopologyConstraint *TopologyConstraint `json:"topologyConstraint,omitempty"`
 }
 ```
 
+#### Mutation Webhook
+
+The mutation webhook automatically modifies PodCliqueSet resources on CREATE operations:
+
+- **Label Addition**: Automatically adds `grove.io/topology-name` label to PodCliqueSet
+- **Label Value**: Set to the configured ClusterTopology name from operator config (e.g., "grove-topology")
+- **Condition**: Label only added when `topology.enabled: true` in operator configuration
+- **Purpose**: Enables workload-to-topology mapping for controller lifecycle management and deletion protection
+- **Immutability**: Once set, label cannot be changed (enforced by validation webhook)
+- **No Topology**: If topology disabled (`topology.enabled: false`), no label is added
+
 #### Validation Webhook
 
 **Hierarchy Constraints:**
@@ -389,7 +408,7 @@ TopologyConstraint *TopologyConstraint `json:"topologyConstraint,omitempty"`
 **Label Immutability:**
 
 - Webhook rejects changes to `grove.io/topology-name` label on PodCliqueSet after creation
-- Label can only be set during PodCliqueSet creation (via mutation webhook when topology enabled)
+- Label can only be set during PodCliqueSet creation (see Mutation Webhook section above)
 - Ensures topology reference remains consistent throughout resource lifecycle
 
 ### 3. Scheduler API Changes (Contract with KAI)
@@ -606,17 +625,6 @@ spec:
 - Changes only affect new or unscheduled pods (already scheduled pods retain placement)
 - Operator re-translates constraints to PodGang on each reconciliation
 
-## Component Architecture
-
-### End-to-End Flow
-
-1. **Admin Setup**: Create ClusterTopology "grove-topology", configure operator with `EnableTopology: true`, create
-   aligned Kueue Topology
-2. **User Creates Workload**: PodCliqueSet with optional topology constraints
-3. **Validation**: Webhook validates against ClusterTopology
-4. **Translation**: Operator builds PodGang with three-level constraints (required + preferred)
-5. **Scheduling**: KAI scheduler reads annotation, applies topology constraints with fallback
-
 ## Security and RBAC
 
 Grove operator requires read access to ClusterTopology and permission to manage finalizers:
@@ -624,6 +632,6 @@ Grove operator requires read access to ClusterTopology and permission to manage 
 ```yaml
 rules:
    - apiGroups: [ "grove.io" ]
-    resources: [ "topologydomains", "topologydomains/finalizers" ]
+   resources: [ "clustertopologies", "clustertopologies/finalizers" ]
     verbs: [ "get", "list", "watch", "update" ]
 ```


docs/designs/topology.md
```diff
@@ -0,0 +1,915 @@
+# Topology-Aware Scheduling - Grove Operator Design
+
+## Overview
+
+This document defines the design for implementing topology-aware scheduling in the Grove operator..
```
romanbaron:
```suggestion
This document defines the design for implementing topology-aware scheduling in the Grove operator.
```

docs/designs/topology.md
```diff
@@ -0,0 +1,915 @@
...
+
+**Motivation**: Topology-aware scheduling is critical for Grove's multinode inference workloads because these
+applications require:
+
+- **Network Locality**: High-bandwidth communication between prefill and decode workers benefits from proximity
```
romanbaron: I was under the impression we mainly aim to pack different roles of the same inference pipeline phase: workers and leader. At the moment, we were not requested to specifically pack prefill and decode workers.

docs/designs/topology.md
```diff
@@ -0,0 +1,915 @@
...
+
+**Design Approach**: This design introduces a flexible topology system with three main components:
+
+1. **TopologyDomain CRD**: Admin-configured cluster topology hierarchy mapping friendly names to node labels
+    2. **Operator Configuration**: Selects active topology via `--topology-domain-name` argument
+    3. **TopologyConstraint**: User-specified packing requirements in workloads (PodCliqueSet, PodCliqueScalingGroup,
+       PodClique)
+
+**Key Feature**: Grove provides automatic out-of-box topology optimization by generating preferred packing constraints
+at all levels, even without user configuration. Users can optionally specify required constraints for strict placement
```
romanbaron: Shouldn't it say something like "at the global/workload level"?

Ronkahn21: we are adding preferred to each level, so it will as close it can be, and not only at workload level

docs/designs/topology.md
Ronkahn21: its was left over from the support of group affinities, before we decided to postpone this required

docs/designs/topology.md
```diff
@@ -0,0 +1,915 @@
...
+PodCliqueSet ────────┘
+  (packDomain: "rack")
+```
+
+### Automatic Optimization
+
+**Out-of-Box Optimization:**
+
+- Operator automatically generates **preferred** constraints using strictest topology level (e.g., "host")
+    - Applied at all three levels (PodGang, NetworkPackGroup, PodGroup) during translation to scheduler API
```
romanbaron: Why not only at the PodGang level?

docs/designs/topology.md
Ronkahn21: fixed description

docs/designs/topology.md
Ronkahn21: The intention is to make the scheduler prioritize keeping pods with a shared rule or relationship (like a 'leader' and its 'workers') together.

To illustrate: Imagine a host has capacity for one more pod. The scheduler has two choices:

1. An unrelated pod from the same workload.

2. A 'worker' pod that belongs to a service already running on that host.

With this change, the scheduler should strongly prefer option 2.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+  prefill and decode, etc)
+    - **Coordinated Placement**: Related components (e.g., model shards) perform better when co-located within the same
+      topology domain
+  - **Latency Optimization**: Minimizing network hops between interdependent inference components improves end-to-ends
+      performance
+
+**Design Approach**: This design introduces a flexible topology system with three main components:
+
+1. **TopologyDomain CRD**: Admin-configured cluster topology hierarchy mapping friendly names to node labels
+    2. **Operator Configuration**: Selects active topology via `--topology-domain-name` argument
```
sanjaychatterjee: Can we fix a default topology CR name for the controller, e.g. `grove-topology`?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+    3. **Workload API (TopologyConstraint)**: Users specify packing requirements
+        - PodCliqueSet, PodCliqueScalingGroup, and PodClique each have TopologyConstraint field
+        - Users reference level names from TopologyDomain (e.g., `packDomain: "rack"`)
+        - No direct TopologyDomain reference needed in workloads
+
+### Component Interactions
+
+```
+TopologyDomain CRD ──┐
```
sanjaychatterjee: CRD -> CR

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+- Users can specify **required** constraints via `packDomain` for strict placement requirements
+    - Required constraints validated and must be satisfied
+    - Preferred constraints enable best-effort optimization with graceful fallback
+
+### Controller Responsibilities
+
+The TopologyDomain controller manages:
+
+- **Kueue Topology Generation**: Auto-creates Kueue Topology CRD for KAI scheduler integration
```
sanjaychatterjee: 1. CRD -> CR
2. I don't think this should be Grove operator responsibility. Instead, KAI PodGrouper's Grove plugin can do this.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+```
+┌─────────────────────────────────────────────────────────────────────────┐
+│                        Topology Architecture                            │
+├─────────────────────────────────────────────────────────────────────────┤
+│                                                                         │
+│  Admin Layer:                                                           │
+│  ┌──────────────────┐              ┌────────────────────┐               │
+│  │ TopologyDomain   │─────────────▶│ TopologyDomain     │               │
+│  │ CRD              │              │ Controller         │               │
```
sanjaychatterjee: CRD -> CR

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+// +kubebuilder:validation:MaxLength=1024
+// +optional
+Description string `json:"description,omitempty"`
+}
+```
+
+**Example TopologyDomain:**
+
+```yaml
+apiVersion: grove.run.ai/v1alpha1
```
sanjaychatterjee: `grove.run.ai/v1alpha1` -> `grove.io/v1alpha1`

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+    - **Immutable**: Once created, cannot be modified
+    - **List-ordered hierarchy**: Index 0 = broadest (e.g., region), last = narrowest (e.g., host)
+
+**API Structure:**
+
+```go
+// TopologyDomain defines the topology hierarchy for the cluster
+// This resource is immutable after creation
+// Multiple TopologyDomain resources can exist; Grove operator references one via --topology-domain-name argument
+type TopologyDomain struct {
```
sanjaychatterjee: Can we rename this to just `Topology` ?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+metav1.TypeMeta   `json:",inline"`
+metav1.ObjectMeta `json:"metadata,omitempty"`
+
+Spec TopologyDomainSpec `json:"spec,omitempty"`
+}
+
+type TopologyDomainSpec struct {
```
sanjaychatterjee: Can we rename this to `TopologySpec`

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+  cluster nodes
+    - TopologyDomain creation succeeds even if labels don't exist yet (allows pre-configuration)
+    - Workloads may fail to schedule if referenced topology labels are missing from nodes
+    - Administrators should verify node labels match TopologyDomain configuration before creating workloads
+
+#### TopologyDomain Controller
+
+The TopologyDomain controller manages the TopologyDomain resource lifecycle with two primary responsibilities:
+
+**1. Kueue Topology Generation**
```
sanjaychatterjee: This responsibility should be with KAI PodGrouper.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+```go
+type PodGangSpec struct {
+// PodGroups is a list of member pod groups in the PodGang
+PodGroups []PodGroup `json:"podgroups"`
+
+// TopologyRef references the Kueue Topology resource
+// Points to Kueue Topology CRD auto-generated by TopologyDomain controller
+// +optional
+TopologyRef *string `json:"topologyRef,omitempty"`
```
sanjaychatterjee: We won't need this if we move the Kueue topology generation to KAI PodGrouper.

docs/designs/topology.md
Ronkahn21: we could but what will the value of this? the admin create the CR, He could choose any name he want,
if we were the one to create the crd I will agree with you

docs/designs/topology.md
Ronkahn21: As we need to use both our topology and kueue topology and kueue already taken the name Topology,
I was thinking it would make confusion and will annoying to add the full name (include api group and version) of each topology.
what do you think?

docs/designs/topology.md
Ronkahn21: its depend who you ask, some the kai team didnt agree about this 
(we could do it for kubecon, but overall I believe we need to integrate with other and not expect other to do it for us), some day vlolcano might support this api, They will never create the topology they using for us

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+## Overview
+
+This document defines the design for implementing topology-aware scheduling in the Grove operator.
+
+**Motivation**: Topology-aware scheduling is critical for Grove's multinode inference workloads because these
+applications require:
+
+- **Network Locality**: Proximity improves high-bandwidth communication between leaders and their respective workers (
+  prefill and decode, etc)
+    - **Coordinated Placement**: Related components (e.g., model shards) perform better when co-located within the same
```
shayasoolin: All the indentations in the file should be fixed, they're not formatted properly. Both bullet and numbered ones.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+at all levels, even without user configuration. Users can optionally specify required constraints for strict placement
+requirements.
+
+## Goals
+
+- Provide flexible, cluster-agnostic topology hierarchy definition via TopologyDomain CRD
+    - Enable packing constraints for network locality across all Grove scalable resources
+    - Support multiple topology configurations for different environments
+    - Automatic Kueue Topology generation for KAI scheduler integration
+    - Immutable topology configuration ensuring scheduling consistency
```
shayasoolin: That means that in order to fix or edit an existing configuration, the admin will have to create a new one, switch to it and delete the previous one.
Can we consider allowing edits if no PCS is currently using the topology? Similarly to how deletions are handled.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+- **Network Locality**: Proximity improves high-bandwidth communication between leaders and their respective workers (
+  prefill and decode, etc)
+    - **Coordinated Placement**: Related components (e.g., model shards) perform better when co-located within the same
+      topology domain
+  - **Latency Optimization**: Minimizing network hops between interdependent inference components improves end-to-ends
+      performance
+
+**Design Approach**: This design introduces a flexible topology system with three main components:
+
+1. **TopologyDomain CRD**: Admin-configured cluster topology hierarchy mapping friendly names to node labels
```
unmarshall: Is the TopologyDomain CRD required?

docs/designs/topology.md
unmarshall: what is `active topology`?

docs/designs/topology.md
unmarshall: what is the `out-of-box` topology optimization that is generated by Grove especially when no pack configuration is defined by the user? Ideally if no pack configuration is specified then should it not mean that there is no strict requirement or in other words there is no packing constraints that is observed by the scheduler while taking scheduling decisions?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+**Key Feature**: Grove provides automatic out-of-box topology optimization by generating preferred packing constraints
+at all levels, even without user configuration. Users can optionally specify required constraints for strict placement
+requirements.
+
+## Goals
+
+- Provide flexible, cluster-agnostic topology hierarchy definition via TopologyDomain CRD
+    - Enable packing constraints for network locality across all Grove scalable resources
+    - Support multiple topology configurations for different environments
```
unmarshall: what does `environment` mean here?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+    - Automatic Kueue Topology generation for KAI scheduler integration
```
unmarshall: who generates the `keueu` topology CR. It should not be done in Grove IMHO.

docs/designs/topology.md
unmarshall: Any reason to make it `immutable`?
Since we support scale-out it is quite possible that the scale-out now does not fit the resource pack constraints w.r.t available capacity. In this case would you not allow the user to relax the pack constraints?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+    - Immutable topology configuration ensuring scheduling consistency
+    - Hierarchical constraint validation (child stricter than parent)
+
+## Non-Goals
+
+- Spread constraints across topology domains (ReplicaSpreadDomain)
+    - Root domain constraints for entire resource (RootDomain)
+    - Ratio-based affinity groups between scaling groups (AffinityGroups with PackRatio)
+    - Dynamic topology reconfiguration after creation
+    - Per-workload topology domain selection
```
unmarshall: not clear, can you elaborate on what this means?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+    - Automatic topology inference from workload characteristics
```
unmarshall: is this point and the next point the same? What is the difference?

docs/designs/topology.md
Ronkahn21: we still need it, if the name of the topology isn't fixed, as it will need to know the name the user topology and create kueue topology out of it

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+                     │
+PodCliqueSet ────────┘
+  (packDomain: "rack")
+```
+
+### Automatic Optimization
+
+**Out-of-Box Optimization:**
+
+- Operator automatically generates **preferred** constraints using strictest topology level (e.g., "host")
```
unmarshall: If the user is provided an explicit API to define network pack constraints then shouldn't it be wise to let the user take the call w.r.t pack constraints? Why assume `host`? When you mention `Out of the box optimization` then assuming `host` might be completely wrong as we are not looking at the resource requirements upfront and the feasibility to schedule such a workload given the available capacity and where the capacity is available. 

We are trying to do in Grove what the scheduler should ideally do. On top of that assuming Host for a workload that will never fit in a single Node cannot be termed as `out of the box optimization`.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+1. **TopologyDomain CRD**: Cluster-scoped resource defining topology hierarchy
+    - Admin creates TopologyDomain with ordered list of topology levels
+    - Each level maps friendly name (e.g., "rack") to node label key (e.g., "topology.kubernetes.io/rack")
+    - Multiple TopologyDomains supported for different environments
+
+    2. **Operator Configuration**: References TopologyDomain by name
+        - Operator argument `--topology-domain-name=default` selects which TopologyDomain to use
+        - All workload validation performed against configured TopologyDomain
+        - Enables switching between topologies without changing workloads
```
shayasoolin: What do you mean here? Maybe "without changing the workload specs"? That's also not always true, it depends on the nature of the change in topology configuration (e.g. whether level names were changed.)

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+- Each level `name` must be unique within the `levels` array of a single TopologyDomain
+    - Each `topologyKey` must be unique within the `levels` array of a single TopologyDomain
+    - Cannot modify any field after creation
+    - Deletion protection via controller finalizer
+
+**Node Label Responsibility:**
+
+- Cluster administrators are responsible for ensuring that node labels specified in `topologyKey` fields exist on
+  cluster nodes
+    - TopologyDomain creation succeeds even if labels don't exist yet (allows pre-configuration)
```
shayasoolin: I understand the requirement for pre-configuration, but I believe we should at least offer the admin to provide validation for the TopologyDomain resource. Leaving misconfiguration errors to the user, upon workload creation, is (a) too late and (b) not the right persona to handle.
I suggest to allow some annotation for example (either opt-in or out) that instructs the operator to validate the topology resource against the nodes in the cluster.
Sample basic validation we can apply there: each of the node labels must exist on at least one node in the cluster. Perhaps we can think of a stronger validation, but that's a start and will catch simple typos.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+      containers:
+        - name: operator
+          args:
+            - --topology-domain-name=default  # References TopologyDomain by name
+```
+
+Configuration:
+
+- `--topology-domain-name`: Specifies TopologyDomain resource name for validation
+    - Operator loads referenced TopologyDomain at startup
```
shayasoolin: Please mention here that this is an optional argument, the operator can run without it.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+**Hierarchy Constraints:**
+
+- Child resource PackDomain must be equal to or stricter than parent
+    - PodCliqueSet → PodCliqueScalingGroup → PodClique hierarchy
+    - Stricter = higher index (narrower scope) in TopologyDomain.Spec.Levels
+    - Example: If parent uses "zone" (index 1), child can use "zone", "rack" (index 4), or "host" (index 5)
+
+**Immutability:**
+
+- All TopologyConstraint fields immutable after resource creation
```
shayasoolin: Why is that? Do we plan to lift this limitation in the future, and describing here the contents of the first phase?
I mean, we can allow changing the topology constraint of a component in two cases:
1. It doesn't conflict with the overall topology spec of the PCS as a whole.
2. The rolling update is for the whole PCS (rather than individual components), a case in which a complete PCS replica is going to be created, and it can use the new topology constraints.

But again, if this note is mentioned here as "out of scope for now" I definitely agree we better not deal with this complication at this stage.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+**Required Constraints:**
+
+- If user specifies `packDomain: "rack"` → becomes `TopologyConstraint.Required.PackDomain = "rack"`
+    - If user doesn't specify packDomain → `Required` is nil
+    - Applied at the appropriate level (PodGang, NetworkPackGroup, or PodGroup)
+
+**Preferred Constraints (Auto-Generated):**
+
+- Operator ALWAYS generates preferred constraint at all three levels
```
shayasoolin, line 565: Nitpick, but this is true unless the user provided a "Required" constraint for the same (strictest) level. Or we don't care and set preferred in addition, as it's harmless?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+Grove implements topology-aware scheduling through three key components:
+
+1. **TopologyDomain CRD**: Cluster-scoped resource defining topology hierarchy
+    - Admin creates TopologyDomain with ordered list of topology levels
+    - Each level maps friendly name (e.g., "rack") to node label key (e.g., "topology.kubernetes.io/rack")
+    - Multiple TopologyDomains supported for different environments
+
+    2. **Operator Configuration**: References TopologyDomain by name
+        - Operator argument `--topology-domain-name=default` selects which TopologyDomain to use
```
omer-dayan: Another option is to have a annotation in the `TopologyDomain` CR "is-default: true" that only one topology has, and grove knows which one is the default.

A benefit of that, is that you can change the used topology without restarting the grove controller.
The bad side is that you need to handle what happen if there are 2 instances or more with that annotation.

Thats how default StorageClass is implemented in K8S: https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/

docs/designs/topology.md
omer-dayan: Does the user has the ability to override that? Or will have in the future?

docs/designs/topology.md
unmarshall: Typically prefill/decode worker/leader will be realized as a PCSG. Each replica of PCSG will be translated to a `PodGang`, so what you want @romanbaron will already happen for the case that @Ronkahn21 is describing above.

However this cannot be a generic rule, you would need flexibility to define pack constraints at every level.

docs/designs/topology.md
omer-dayan: I have to say I am a bit perplexed why Grove and KAI wouldnt use same CRD

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+  (packDomain: "rack")
+```
+
+### Automatic Optimization
+
+**Out-of-Box Optimization:**
+
+- Operator automatically generates **preferred** constraints using strictest topology level (e.g., "host")
+    - Applied at all three levels (PodGang, NetworkPackGroup, PodGroup) during translation to scheduler API
+    - Users get optimal packing without configuration
```
unmarshall: I am not sure this is the correct as described above. Grove does not understand the workload semantics and therefore is in no position to determine the optimal packing configuration for it.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+**User Control:**
+
+- Users can specify **required** constraints via `packDomain` for strict placement requirements
```
unmarshall: Maybe you can remove the word `strict` here as users can also define relaxed pack constraints as well.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+// +kubebuilder:validation:Required
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=316
+// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
+TopologyKey string `json:"topologyKey"`
+
+// Description provides human-readable information about this level
+// +kubebuilder:validation:MaxLength=1024
+// +optional
+Description string `json:"description,omitempty"`
```
unmarshall: why is this required? who will use it?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+### Architecture Overview
+
+```
+┌─────────────────────────────────────────────────────────────────────────┐
+│                        Topology Architecture                            │
+├─────────────────────────────────────────────────────────────────────────┤
+│                                                                         │
+│  Admin Layer:                                                           │
+│  ┌──────────────────┐              ┌────────────────────┐               │
```
unmarshall: why do you recommend a different controller for it when it is part of the creation flow of a PCS?

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+**User Control:**
+
+- Users can specify **required** constraints via `packDomain` for strict placement requirements
+    - Required constraints validated and must be satisfied
+    - Preferred constraints enable best-effort optimization with graceful fallback
+
+### Controller Responsibilities
+
+The TopologyDomain controller manages:
```
unmarshall: Why should Grove manage any Kueue CRs? Today KAI scheduler decides to have dependency on Kueue but that is the choice that KAI makes. So while Grove today uses KAI scheduler its responsibility should not be tied to a single scheduler (in this case KAI scheduler). Therefore creation of CRs that are required by KAI should be done in KAI using the enhancements to the `PodGang` scheduler API that Grove defines.

docs/designs/topology.md
Ronkahn21: you are correct, it a bit wired, I will change this section and clearfy that deletion fo topology is allowed only if no one ref them, it will be easier  and more logical

docs/designs/topology.md
Ronkahn21: the default topology configured by admin, and applied to all the pcs submitted

docs/designs/topology.md
Ronkahn21: I'm proposing a CRD for cluster topology to support complex scenarios, like a single cluster with multiple network layouts (e.g., for hybrid on-prem/cloud environments).

This lets admins dynamically define multi-level topologies with user-friendly names (`rack`, `spine`) and is far more extensible than a hard-coded list or ConfigMap, allowing for a future `status` field.

docs/designs/topology.md
Ronkahn21: cluster with  hybrid on-prem/cloud environments.

docs/designs/topology.md
Ronkahn21: Some of KAI maintains prefer that grove will integrate with KAI and not the other way around, I belive we are the one that support to integrate with other schedulers today it's KAI but tomorrow it might be Volcano, so if KAI work in a cerain way we are the one that should create it IMHO, (also KAI need to agree for this)

docs/designs/topology.md
Ronkahn21: No particular, I agree with you I will change it +1

docs/designs/topology.md
Ronkahn21: I meant workload that submitted before the enabled of topology, I will clarify it

docs/designs/topology.md
Ronkahn21: The current API enforces strict pod placement, but we need a **preferred grouping mechanism** for optimization, as KAI (our scheduler) does not perform this automatically.

This  designed so that **Grove** can guide KAI (it does not happend if not requested as preferred) to make better placement decisions. Grove will instruct KAI to pack related pods as closely as possible using a **topology-aware fallback logic**:

1.  Attempt placement on the same `node`.
2.  If not possible, attempt placement in the same `rack` (the next level up).
3.  Continue up the hierarchy until a valid placement is found.

This provides an opportunistic performance boost by improving pod locality, without the risk of pods failing to schedule if the ideal placement isn't available.

docs/designs/topology.md
Ronkahn21: I describe it to optimal, it better to say it provides an opportunistic performance boost

docs/designs/topology.md
Ronkahn21: I thought i will helpful for users who want to understand better the levels but it only suggestion

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+      topology domain
+  - **Latency Optimization**: Minimizing network hops between interdependent inference components improves end-to-ends
+      performance
+
+**Design Approach**: This design introduces a flexible topology system with three main components:
+
+1. **TopologyDomain CRD**: Admin-configured cluster topology hierarchy mapping friendly names to node labels
+    2. **Operator Configuration**: Selects active topology via `--topology-domain-name` argument
+    3. **TopologyConstraint**: User-specified packing requirements in workloads (PodCliqueSet, PodCliqueScalingGroup,
+       PodClique)
```
renormalize: Each numbered point is not being rendered in its own line because of the indentation. Can this be corrected? There are also multiple other places in the file where there are unnecessary newlines/indentation which causes the rendered markdown to look weird.

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
+# Topology-Aware Scheduling - Grove Operator Design
+
+## Overview
+
+This document defines the design for implementing topology-aware scheduling in the Grove operator.
+
+**Motivation**: Topology-aware scheduling is critical for Grove's multinode inference workloads because these
```
renormalize: Nit: we hyphenate `multi-node` instead of writing it as one word in all documentation in the repository.

```suggestion
**Motivation**: Topology-aware scheduling is critical for Grove's multi-node inference workloads because these
```

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
+# Topology-Aware Scheduling - Grove Operator Design
+
+## Overview
+
+This document defines the design for implementing topology-aware scheduling in the Grove operator.
```
renormalize: nit:

```suggestion
This document defines the design for supporting topology-aware scheduling in the Grove operator.
```

docs/designs/topology.md
```diff
@@ -0,0 +1,916 @@
...
+
+**Motivation**: Topology-aware scheduling is critical for Grove's multinode inference workloads because these
+applications require:
+
+- **Network Locality**: Proximity improves high-bandwidth communication between leaders and their respective workers (
+  prefill and decode, etc)
```
renormalize: nit:

A period is typically added after the abbreviation `etc.`. Also, unnecessary new line.

```suggestion
  prefill and decode, etc.)
```

docs/designs/topology.md
renormalize: Can `friendly names` be expanded on here? Perhaps examples like `rack`, `spine`, and `host` can be added in brackets so the sentence is clearer in a first pass?

Or we can instead write `cluster topology hierarchy mapping Grove defined cluster topologies to node labels`?

docs/designs/topology.md
renormalize: What would happen if the user creates multiple instances of this CustomResource? How will Grove decide which to pick?

During runtime, there is only one CR that Grove must use for specifying topologies.

If we allow any name, we will then have to implement a validating webhook that ensures that only one instance of this CustomResource exists in the entire cluster. This seems more complicated the enforcing the name of the CustomResource that is to be created, which Grove will always look for.

If the consumer chooses to change the topology, then they will edit this configuration CR, which Grove will make use of by fetching it and making use of the newly specified mappings before creating any resources.

renormalize: We do not provide configuration flags like `--topology-domain-name` in Grove. The operator can only be configured through `OperatorConfiguration`, where each configuration is a field/subfield in the manfiest file provided. So this could be `OperatorConfiguration.TopologyDomainName`.

docs/designs/topology.md
Ronkahn21: I don't agree with you about this,KAI are doing us a favor here (they would not agree it on them), if you want to integrate with them not the other way around, in the feature we would want to integrate Volcano/unicron/ or even the default scheduler    you need to work with the API they agree upon

docs/designs/topology.md
renormalize: If no configuration is provided, and the user has no real constraints that they need to follow, how would they be able to opt out if Grove always tries for the optimal by default? 

If by default Grove tries for the optimal placement, and the scheduler is unable to schedule pods with these constraints, then the workload will never fully be up even though the user has no requirement for it.

Perhaps we can provide a field in the API through which the user indicates that they want out-of-box topology optimization. Or, we can provide a way for user to opt out of these out-of-box optimizations if they do not really care?

docs/designs/topology.md
renormalize: If the configuration is changed, and the PCS is mutated, will Grove then go ahead and recreate all the children PCSG and PCLQs because the previous configurations were less strict, and now Grove enforces the more strict configuration?

renormalize: If a PCS does end up using this CR, then we are stuck with it forever/until that PCS is deleted.

An easier way to tackle this is to specify a reference to an instance of the configuration CR? Newly created PCS-s can rely on a different instance of the CR, which has the correct configuration?

docs/designs/topology.md
Ronkahn21: I would like to allow this

docs/designs/topology.md
```diff
@@ -0,0 +1,887 @@
...
+constraints at all levels, even without user configuration. This opportunistic packing may improve performance when
+cluster resources allow, but users should specify required constraints when strict placement is critical for their
+workload.
+
+## Goals
+
+- Provide flexible, cluster-agnostic topology hierarchy definition via TopologyDomain CRD
+- Enable packing constraints for network locality across all Grove scalable resources
+- Support multiple topology configurations for different environments
+- Automatic Kueue Topology generation for KAI scheduler integration
```
shayasoolin: AFAIU this is not a goal anymore, so please remove it. Also appears in other parts of the document, search for "kueue" to find them.

docs/designs/topology.md
```diff
@@ -0,0 +1,887 @@
...
+- Immutable topology configuration ensuring scheduling consistency
```
shayasoolin, line 20: Please be more explicit here and clarify that the immutable topology is of the global configuration, not the require topology of a certain workload (as there it's now mutable.)

docs/designs/topology.md
```diff
@@ -0,0 +1,887 @@
...
+3. **Workload API (TopologyConstraint)**: Users specify packing requirements
+    - PodCliqueSet, PodCliqueScalingGroup, and PodClique each have TopologyConstraint field
+    - Users reference level names from TopologyDomain (e.g., `packDomain: "rack"`)
+    - No direct TopologyDomain reference needed in workloads
+
+### Automatic Optimization
+
+**Out-of-Box Optimization:**
+
+- Operator automatically generates **preferred** constraints using strictest topology level (e.g., "host")
```
shayasoolin: Is this agreed or still under discussion? @unmarshall commented about it in a different place.

docs/designs/topology.md
Ronkahn21: we decide to allow topology constrains to be mutable

docs/designs/topology.md
```diff
@@ -0,0 +1,556 @@
...
+      description: "Individual compute host"
+    - name: numa
+      topologyKey: "topology.kubernetes.io/numa"
+      description: "NUMA node within host"
+```
+
+**Creating TopologyDomain:**
+
+1. Customize example above with your cluster's actual `topologyKey` values
+2. Create resource: `kubectl apply -f topologydomain.yaml` (name MUST be "grove-topology")
```
kangclzjc: Shouldn't we make it configurable?

docs/designs/topology.md
sanjaychatterjee: Right, but the group would be different between Grove and Kueue. Anyway, if we want to keep the name of this type as `TopologyDomain` then we need to change the `PackDomain` string to something like `PackLevel` in the `TopologyConstraint` API. Right now, there term Domain conflicts in its usage in the CRD vs the TopologyConstraint API.

docs/designs/topology.md
sanjaychatterjee: Let's make it optional

docs/designs/topology.md
sanjaychatterjee: Yes, we can make it configurable. `grove-topology` can be default, but, the Grove operator can have an optional flag that accepts the name of the resource as configured by the admin.

docs/designs/topology.md
```diff
@@ -0,0 +1,556 @@
...
+- Since topology is singleton, any PodCliqueSet potentially uses it
+- Controller continuously reconciles deletion requests
+- Prevents orphaned workloads with invalid topology configuration
+
+#### Operator Configuration
+
+Operator enables/disables topology features via OperatorConfiguration manifest:
+
+```yaml
+apiVersion: grove.run.ai/v1alpha1
```
sanjaychatterjee: `grove.run.ai/v1alpha1`-> `grove.io/v1alpha1`

docs/designs/topology.md
```diff
@@ -0,0 +1,556 @@
...
+### 2. Operator API Changes (Grove CRDs)
+
+#### TopologyConstraint Model
+
+```go
+type TopologyConstraint struct {
+// PackDomain references a level name from TopologyDomain.Spec.Levels
+// Defines required topology packing constraint for replicas
+// Replicas packed together within specified topology level for network locality
+PackDomain *string `json:"packDomain,omitempty"`
```
sanjaychatterjee: If the CRD kind is `TopologyDomain`, then we should rename this to `PackLevel` or something. The term domain is conflicting between its usage here vs the CRD.

Also, can you please clarify whether this constraint means to pack the entire resource including all the replicas within a specific instance of a `PackDomain`? For example, when a PCSG resource scales and the `PackDomain` == `rack`, then which one of A or B is true:
A) all the replicas of that PCSG will be constrained within one specific rack?
B) each replica be constrained in a rack but individual replicas can be placed in different racks?

docs/designs/topology.md
```diff
@@ -0,0 +1,556 @@
...
+```
+
+This annotation allows the scheduler to locate the Kueue Topology resource without requiring a spec field, providing
+flexibility for future API changes.
+
+**NetworkPackGroupConfig:**
+
+```go
+// NetworkPackGroupConfig indicates PodGroups should be optimally placed w.r.t cluster's network topology
+type NetworkPackGroupConfig struct {
```
sanjaychatterjee: Maybe we should rename the type from `NetworkPackGroupConfig` to `TopologyConstraintGroupConfig`

docs/designs/topology.md
```diff
@@ -0,0 +1,556 @@
...
+// Translated from PodCliqueSet.TopologyConstraint
+// Updated by operator on each reconciliation when PodCliqueSet topology constraints change
+// +optional
+TopologyConstraint *TopologyConstraint `json:"topologyConstraint,omitempty"`
+
+// NetworkPackGroupConfigs defines groups of PodGroups for network optimization
+// Enhanced with topology constraints for PCSG-level packing
+// Updated by operator on each reconciliation when PCSG topology constraints change
+// +optional
+NetworkPackGroupConfigs []NetworkPackGroupConfig `json:"networkPackGroupConfigs,omitempty"`
```
sanjaychatterjee: Maybe we can rename the variable and type as follows:

`TopologyConstraintGroups []TopologyConstraintGroupConfig`

docs/designs/topology.md
Ronkahn21: A we agree, for now it only control B option, we will alllow to control A in the futere 
I will make it clear

docs/designs/topology.md
gflarity: Agree re renaming to Topology... Seems ok to for the names to be the same as it this is really an extension of Kueue's Topology with a friendly name and descriptions added.

docs/designs/topology.md
```diff
@@ -0,0 +1,578 @@
...
+- If `topology.enabled: true`:
+   - `topologyDomainName` not specified → defaults to "grove-topology"
+   - Operator looks for TopologyDomain with configured name (defaults to "grove-topology")
+   - If TopologyDomain with that name doesn't exist → operator fails to start
+- If `topology.enabled: false`: topology features disabled
+- Admin must create TopologyDomain with matching name OR disable topology
+
+**Admin Responsibilities:**
+
+- Manually create Kueue Topology with same name as Grove TopologyDomain for KAI scheduler
```
gflarity: I believe this was discussed during the call but I arrived a bit late.  Can we call out why this is necessary? 

I'm guessing because KAI already supports this is will help us meet deadlines?

I'm wondering if should make this even simpler and just skip the Grove TopologyDomain and just require a Kueue Topology instead (for a V1/ minimum viable product).  Seems like the Grove TopologyDomain only provides a human friendly name and optional description right now... Seeing as the Admin has to create the Kueue one anyways, what's the value?

gflarity (approved):
Thanks @Ronkahn21!  IMO we should  simplify things and just go with the Kueue CR since it's required *regardless*. Having both just seems like more work for the Admin for very little benefit.  I don't want to waste any time with a debate though if I can't convince you.  🚢 it :)

docs/designs/topology.md
gflarity: We don't own grove.io do we? 

Should probably be grove.nvidia.com/* unless we have/want a separate domain.  Generally it's a domain, or a plausible domain.

docs/designs/topology.md
gflarity: After chatting with a colleague, I think we can hide this complexity entirely in the Helm chart. There's no real need even spend time creating the the two CRs, the values can just be specified in the helm chart values and populated with templates. In the future we can have a preinstall script that determines the topology automatically if the user doesn't supply the values: https://github.com/ai-dynamo/grove/issues/236

docs/designs/topology.md
```diff
@@ -0,0 +1,603 @@
...
```
gflarity: We should manage both the Kueue Topology and Grove Topology in the helm chart so the Admins don't need to actually create these manually.  In the future a pre-install script can determine the topology values automagically.  I created a ticket to track this: https://github.com/ai-dynamo/grove/issues/236

Ronkahn21: sound like a good option

docs/designs/topology.md
```diff
@@ -0,0 +1,599 @@
...
+- Controller checks if any PodCliqueSet exists (no need to check specific references)
+- Since topology is singleton, any PodCliqueSet potentially uses it
+- Controller continuously reconciles deletion requests
+- Prevents orphaned workloads with invalid topology configuration
+
+#### Operator Configuration
+
+Operator enables/disables topology features via OperatorConfiguration manifest:
+
+```yaml
```
sanjaychatterjee, line 252: `OperatorConfiguration` is not a CRD. Instead grove operator config options come in through a [config map](https://github.com/ai-dynamo/grove/blob/main/operator/charts/templates/configmap-operator.yaml). Let's use that instead.

sanjaychatterjee (approved):
Looks good to me. Thanks!

docs/designs/topology.md
unmarshall: It has been clarified and accepted that Grove is not going to generate Kueue Topology CR in the first iteration. When we introduce the scheduler backend framework to support KAI, Volcano and other schedulers then we can take the responsibility to create dependent CRs as required by any specific scheduler.

docs/designs/topology.md
unmarshall: We do not necessarily need a dedicated controller for Grove Topology CR since it is immutable (first iteration). Once we allow updates to it then perhaps we can have a dedicate controller.

shayasoolin: @unmarshall I think the controller is there for the deletion prevention. The controller is maintaining a finalizer as long as any PCS exists.

docs/designs/topology.md
unmarshall: Currently all our types use `grove.io` (See https://github.com/ai-dynamo/grove/blob/main/operator/api/core/v1alpha1/register.go#L27). It is not required to have a domain registered to use it as a group. Can you please point to any document where it mentions otherwise?

docs/designs/topology.md
unmarshall: isn't the cluster role or role with which the operator is deployed defining if it has sufficient RBAC to delete a resource? Also the lifecycly of Topology CR is not tied to a lifecycle of PCS. It is tied to a cluster's lifecycle. As long as the cluster exists and continues to be a target for deploying PCS workloads Topology CR must exist. This is outside the responsibility of Grove as it is does not create this CR.

docs/designs/topology.md
unmarshall: @Ronkahn21  Can you update this comment based on latest changes?

docs/designs/topology.md
gflarity: Yes, see  the [kubernetes label documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/). Specifically:

> Labels are key/value pairs. Valid label keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).Labels are key/value pairs. Valid label keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).


I don't think we need to have a domain registered per say, but it should be a domain that could be  registered and one  that's controlled by Nvidia or the AI Dynamo project etc. My concerned is that grove.io is currently a hosted IRC service, but the domain could change hands and  end up being a porn site or something more nefarious in the future. There's no reason not use a dns entry we control. Note kubernetes.io  labels are common, and that's the official site.

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+│                                                                         │
+└─────────────────────────────────────────────────────────────────────────┘
+```
+
+### 1. TopologyDomain Infrastructure
+
+#### TopologyDomain CR
+
+TopologyDomain is a cluster-scoped CR that defines the topology hierarchy for scheduling. It maps friendly level names
+to Kubernetes node labels and establishes ordering from broadest to narrowest scope.
```
unmarshall: The intent is to define consistent naming for cluster topology hierarchy to be used by workload designers. You should perhaps replace the reason of providing `friendly level names`

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+
+*note: this CR is independent of Kueue Topology CRD, which must be manually created by admin to align with Grove's
```
unmarshall: This proposal should not really mention anything about Kueue Topology CRD.

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+TopologyDomain for KAI scheduler usage.* (also named "grove-topology")
+**Characteristics:**
+
+- **Cluster-scoped singleton**: Only one TopologyDomain allowed cluster-wide, user chooses name
```
unmarshall: There is no need to explicitly state that `user chooses name`. Every Custom resource has a `ObjectMeta` which will always have a name which must be specified. Nothing can be assumed.

docs/designs/topology.md
sanjaychatterjee: Operator config will contain the name of the CR that Grove operator will use. Also, in phase1, we don't allow editing of topology CR when there are active PCS in the cluster.

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+- **Default name**: "grove-topology" used when topologyDomainName not specified in operator config
```
unmarshall: The default name is only w.r.t operator configuration where the `topology` CR name is defaulted if not specified. This does not come out clearly from the sentence.

docs/designs/topology.md
sanjaychatterjee: We discussed this in our design meeting. There are a few reasons:

1. First, Grove's design for Topology Aware scheduling cannot be satisfied by the Kueue topology CRD. We require the admin to define a mapping between Grove defined constants of topology levels to cluster specific topology keys. This cannot be supported in the Kueue CRD.
2. Second, the Kueue topology CRD is part of Kueue API and to use that we would have had to import the full Kueue dependency. We don't want to do that. So, we had raised an [issue](https://github.com/kubernetes-sigs/kueue/issues/7032) in Kueue to separate out the topology API into a separate go module. Everyone agrees it is a good idea but will be prioritized later. We need to have a plan until they do it.
3. Third, the cluster admin needs to create a Kueue CR to satisfy the requirement of KAI scheduler. In future, eventually KAI may also move out of the Kueue topology dependency, but for now this is a requirement for KAI.
4. Finally, we could have allowed Grove to generate the Kueue CR but then not only would Grove need to take on the Kueue dependency but also that Grove should not be responsible for this. In general, Grove should be independent of specific scheduler requirements until we come up with a good framework in Grove to have scheduler specific backends.

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+- **Immutable**: Once created, cannot be modified
+- **List-ordered hierarchy**: Index 0 = broadest (e.g., region), last = narrowest (e.g., host)
```
unmarshall:
```suggestion
- **List-ordered hierarchy**: Index 0 represents the broadest category (e.g., region), and the final index represents the narrowest (e.g., host).
```

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+- **Predefined ordering**: Region > Zone > DataCenter > Block > Rack > Host > Numa (broadest to narrowest)
```
unmarshall: Instead of calling it `Predefined ordering` call it `Supported topology levels`

docs/designs/topology.md
sanjaychatterjee: For simplicity, in phase 1 we will only allow one topology CR. We can discuss the need to update it but I don't see a reason to handle it in phase1.

docs/designs/topology.md
sanjaychatterjee: Yes, let's keep this as out-of-scope for phase1.

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+- **Webhook-validated**: Webhook enforces singleton constraint (any name allowed)
```
unmarshall: Do you need this? As long as a topology CR reference is passed as part of `OperatorConfiguration` Grove does not need to bother about how many Topology CRs exist, right?

docs/designs/topology.md
sanjaychatterjee: We decided to support only one topology CR in phase 1.

docs/designs/topology.md
sanjaychatterjee: We are not supporting multiple topology configurations.

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+TopologyLevelNuma       TopologyLevelName = "numa"
+)
+
+// Topology ordering (broadest to narrowest):
+// Region > Zone > DataCenter > Block > Rack > Host > Numa
+
+// TopologyDomain defines the topology hierarchy for the cluster
+// This resource is immutable after creation
+// Only one TopologyDomain can exist cluster-wide with enforced name "grove-topology"
+type TopologyDomain struct {
```
unmarshall: Rename this to `Topology` - It does not represent a `Topology Domain` rather it represents a topology for a cluster.

docs/designs/topology.md
sanjaychatterjee: We are doing automatic Kueue Topology generation.

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+metav1.TypeMeta   `json:",inline"`
+metav1.ObjectMeta `json:"metadata,omitempty"`
+
+Spec TopologyDomainSpec `json:"spec,omitempty"`
```
unmarshall: Rename this to `TopologySpec`

docs/designs/topology.md
sanjaychatterjee: Not addressing in phase 1

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+Name TopologyLevelName `json:"name"`
+
+// TopologyKey is the node label key that identifies this topology domain
+// Must be a valid Kubernetes label key (qualified name)
+// Examples: "topology.kubernetes.io/zone", "kubernetes.io/hostname"
+// +kubebuilder:validation:Required
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=316
+// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
+TopologyKey string `json:"topologyKey"`
```
unmarshall: Should this instead be just called `labels`. If you insist in calling it `TopologyKey` then you should change the field name to just `Key` as it is within the struct which is already named `TopologyLevel`

Should it be a `[]string` which allows mapping of a name with more than one labels. Case in point: 
The following labels represent a region:
- failure-domain.beta.kubernetes.io/region
- topology.kubernetes.io/region

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+- **Block**: Network local to a switching block unit within a data-center
+- **Rack**: First-level network grouping of compute hosts (includes NVLink domains as logical racks)
+- **Host**: Individual compute host
+- **Numa**: NUMA node (processor and memory locality domain) within a compute host
+
+**API Structure:**
+
+```go
+// TopologyLevelName represents a predefined topology level in the hierarchy
+type TopologyLevelName string
```
unmarshall: Rename this to `TopologyDomain`

docs/designs/topology.md
```diff
@@ -0,0 +1,594 @@
...
+// +kubebuilder:validation:MaxItems=8
+Levels []TopologyLevel `json:"levels"`
+}
+
+type TopologyLevel struct {
+// Name is the predefined level identifier used in TopologyConstraint references
+// Must be one of: region, zone, datacenter, block, rack, host, numa
+// +kubebuilder:validation:Required
+// +kubebuilder:validation:Enum=region;zone;datacenter;block;rack;host;numa
+Name TopologyLevelName `json:"name"`
```
unmarshall: Rename this to `Domain` and its type is `TopologyDomain`

unmarshall (changes requested):

docs/designs/topology.md
sanjaychatterjee: Will update

docs/designs/topology.md
sanjaychatterjee: Will do

docs/designs/topology.md
sanjaychatterjee: We decided to name this as `Key` and also keep mapping 1:1

docs/designs/topology.md
sanjaychatterjee: We decided to name this `ClusterTopology`

docs/designs/topology.md
sanjaychatterjee: We decided to name this `ClusterTopologySpec`

docs/designs/topology.md
sanjaychatterjee: We addressed this in another [comment](https://github.com/ai-dynamo/grove/pull/224#discussion_r2478513833).

docs/designs/topology.md
sanjaychatterjee: Thanks @gflarity

docs/designs/topology.md
Ronkahn21: We should not allow the default topology to be changed while workloads are running, if there more then one, as this can cause them to get stuck. To make our lives easier and prevent bugs, let's enforce a 'one default topology at a time' rule. This also means we don't have to immediately fix the way we check for topology references. We can explore allowing multiple topologies in the future, but for now, let's keep it simple.

Ronkahn21 (commented):
.

docs/designs/topology.md
sanjaychatterjee: I think this is fine for phase 1. Will need to discuss on how to support topology updates for next phase.

sanjaychatterjee (approved):

PR #225, opened 2025-10-22
Changes for migration from @NVIDIA to @ai-dynamo

renormalize:
#### What type of PR is this?

/kind cleanup

The repository is planned to be migrated from @NVIDIA to @ai-dynamo.
All corresponding changes in source which correspond to this migration are brought in by this PR:
- Change Go module paths from `^github\.com/NVIDIA/grove(?:/.*)?$` to `^github\.com/ai-dynamo/grove(?:/.*)?$`.
- Change all references of these modules to the new format.
- Change documentation.
- Change organization references in GitHub workflows.

I've made changes module wise, and committed them individually to make it easy to review.

The changes are identical when the following is run on main at c2171c7cdeb0d84e2f281fe83d57a014694960aa (other than miscellaneous changes in 601d17b6373f2f0389a41c4aa3c4d7e4de73f650):

```
find . -type f | xargs sed -i "s_NVIDIA/grove_ai-dynamo/grove_gI"
```

Fixes #

**DO NOT MERGE BEFORE MIGRATION.**

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/.github/workflows/push-artifacts.yaml
+++ b/.github/workflows/push-artifacts.yaml
@@ -27,9 +27,6 @@ permissions:
   contents: read
   packages: write
 
-env:
-  DOCKER_REGISTRY: "ghcr.io/ai-dynamo/grove"
-
 jobs:
   build-and-push:
     name: Build & Push
@@ -67,6 +64,11 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
+      - name: Set Docker Registry
+        run: |
+          DOCKER_REGISTRY=$(echo 'ghcr.io/${{ github.repository }}' | tr '[:upper:]' '[:lower:]')
+          echo "DOCKER_REGISTRY=$DOCKER_REGISTRY" >> $GITHUB_ENV
+
       - name: Docker build & push
         run: |
           make --directory=operator docker-build PLATFORM=linux/amd64,linux/arm64 VERSION=$PACKAGE_VERSION REGISTRY=$DOCKER_REGISTRY DOCKER_BUILD_ADDITIONAL_ARGS=--push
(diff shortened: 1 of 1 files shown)


unmarshall (approved):

.github/workflows/push-artifacts.yaml
```diff
@@ -28,7 +28,7 @@ permissions:
   packages: write
 
 env:
-  DOCKER_REGISTRY: "ghcr.io/nvidia/grove"
+  DOCKER_REGISTRY: "ghcr.io/ai-dynamo/grove"
```
gflarity: Let's make this generic by relying on the contexts variable available to all workflows:
```suggestion
  DOCKER_REGISTRY: "ghcr.io/${{ github.repository }}"
```

I ran into this recently as I forked the repo and wanted to test the development image on a remote cluster and thus needed an image that was publicly accessable.  Here's the [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context).

.github/workflows/push-artifacts.yaml
```diff
@@ -77,9 +77,9 @@ jobs:
 
       - name: Build Helm chart
         run: |
-          sed -i "s#repository: grove-operator#repository: ghcr.io/nvidia/grove/grove-operator#" operator/charts/values.yaml
+          sed -i "s#repository: grove-operator#repository: ghcr.io/ai-dynamo/grove/grove-operator#" operator/charts/values.yaml
           sed -i "s#tag:.*#tag: $PACKAGE_VERSION#" operator/charts/values.yaml
-          sed -i "s#value: grove-initc#value: ghcr.io/nvidia/grove/grove-initc#" operator/charts/values.yaml
+          sed -i "s#value: grove-initc#value: ghcr.io/ai-dynamo/grove/grove-initc#" operator/charts/values.yaml
```
gflarity:
```suggestion
          sed -i "s#value: grove-initc#value: ${{ env.DOCKER_REGISTRY }}/grove-initc#" operator/charts/values.yaml
```

.github/workflows/push-artifacts.yaml
renormalize: @gflarity thanks for the suggestion. But unfortunately we ran into an issue because of this very change. See #222.

If the repository owner's username happens to have capital letters (like @NVIDIA), then the suggested change will not work. The workflow will fail since ghcr.io will not accept capital letters in the path; where the username is added to the path. GitHub allows usernames to have capital letters, so before we generalize, I'd like to make sure the change that will be bring in will work for all potential usernames.

Perhaps we could write some workflow steps that convert the DOCKER_REGISTRY to all lower case before being used?

gflarity: @renormalize  Take a look now (sorry, I wanted it to be a suggestion but it appended the commit).  It's a bit ugly but it should do the trick.

gflarity (dismissed):

renormalize:
@gflarity could you sign-off your commit so the DCO passes? thanks.

.github/workflows/push-artifacts.yaml
renormalize: I think this can be made slightly cleaner by performing the lowercase conversion in a step, and appending it to the env variables that are shared across multiple steps, thereby avoiding this repetition.

I will push that in the next commit.

PR #230, opened 2025-10-23
test coverage for internal/webhooks

gflarity:
#### What type of PR is this?

Testing

Increased test coverage for internal/webhooks

Prompt:

```release-note
None
```

### DIFF
--- a/operator/internal/webhook/admission/pcs/validation/util_test.go
+++ b/operator/internal/webhook/admission/pcs/validation/util_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/ptr"
 )
 
 // TestValidateEnumType tests validation of enum values against allowed sets.
@@ -49,26 +50,26 @@ func TestValidateEnumType(t *testing.T) {
 		},
 		{
 			name:          "valid value passes validation",
-			value:         stringPtr("option1"),
+			value:         ptr.To("option1"),
 			allowedValues: sets.New("option1", "option2"),
 			expectError:   false,
 		},
 		{
 			name:          "invalid value returns Invalid error",
-			value:         stringPtr("option3"),
+			value:         ptr.To("option3"),
 			allowedValues: sets.New("option1", "option2"),
 			expectError:   true,
 			errorType:     field.ErrorTypeInvalid,
 		},
 		{
 			name:          "empty string value in allowed set passes",
-			value:         stringPtr(""),
+			value:         ptr.To(""),
 			allowedValues: sets.New("", "option1"),
 			expectError:   false,
 		},
 		{
 			name:          "empty string value not in allowed set fails",
-			value:         stringPtr(""),
+			value:         ptr.To(""),
 			allowedValues: sets.New("option1", "option2"),
 			expectError:   true,
 			errorType:     field.ErrorTypeInvalid,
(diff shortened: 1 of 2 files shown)


operator/internal/webhook/register_test.go
```diff
@@ -0,0 +1,236 @@
...
+			namespace:          "",
+			serviceAccountName: "test-sa",
+			expected:           "system:serviceaccount::test-sa",
+		},
+		{
+			name:               "empty service account",
+			namespace:          "default",
+			serviceAccountName: "",
+			expected:           "system:serviceaccount:default:",
+		},
```
renormalize: Not sure if these are meaningful. We can just get rid of these.
```suggestion
```

operator/internal/webhook/register_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// Ensure env var is not set
+	originalEnv := os.Getenv(constants.EnvVarServiceAccountName)
+	err := os.Unsetenv(constants.EnvVarServiceAccountName)
+	require.NoError(t, err)
+	defer func() {
+		if originalEnv != "" {
+			os.Setenv(constants.EnvVarServiceAccountName, originalEnv)
+		}
+	}()
```
renormalize: I feel this handling is not required, because this code only runs in tests on a developer's machine/CI.
```suggestion
	// Ensure env var is not set
	err := os.Unsetenv(constants.EnvVarServiceAccountName)
	require.NoError(t, err)
```

operator/internal/webhook/register_test.go
```diff
@@ -0,0 +1,236 @@
...
+	originalEnv := os.Getenv(constants.EnvVarServiceAccountName)
+	os.Setenv(constants.EnvVarServiceAccountName, "test-sa")
+	defer func() {
+		if originalEnv != "" {
+			os.Setenv(constants.EnvVarServiceAccountName, originalEnv)
+		} else {
+			err := os.Unsetenv(constants.EnvVarServiceAccountName)
+			require.NoError(t, err)
+		}
+	}()
```
renormalize: https://github.com/ai-dynamo/grove/pull/230#discussion_r2499312836

operator/internal/webhook/register_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// Create a real webhook server
+	server := webhook.NewServer(webhook.Options{
+		Port: 9443,
+	})
+	mgr.WebhookServer = server
+
+	// Set env var
+	originalEnv := os.Getenv(constants.EnvVarServiceAccountName)
+	os.Setenv(constants.EnvVarServiceAccountName, "test-sa")
```
renormalize: We can let the testing framework handle the setting, and resetting of the env var to the original value.

```suggestion
	t.Setenv(constants.EnvVarServiceAccountName, "test-sa")
```

operator/internal/webhook/register_test.go
```diff
@@ -0,0 +1,236 @@
...
+	defer func() {
+		if originalEnv != "" {
+			os.Setenv(constants.EnvVarServiceAccountName, originalEnv)
+		} else {
+			err := os.Unsetenv(constants.EnvVarServiceAccountName)
+			require.NoError(t, err)
+		}
+	}()
```
renormalize: https://github.com/ai-dynamo/grove/pull/230#discussion_r2499329029

operator/internal/webhook/admission/pcs/validation/util_test.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+// Helper functions
+func stringPtr(s string) *string {
+	return &s
+}
+
+func intPtr(i int) *int {
+	return &i
+}
```
renormalize: Use the generic function `ptr.To()` instead (`"k8s.io/utils/ptr"`). Cleaner, and we've already used that in quite a few places as well.

operator/internal/webhook/admission/pcs/validation/handler_test.go
```diff
@@ -0,0 +1,350 @@
...
+				WithPodCliqueTemplateSpec(
+					testutils.NewPodCliqueTemplateSpecBuilder("test").
+						WithReplicas(1).
+						WithRoleName("test-role").
+						WithMinAvailable(1).
+						Build()).
+				Build(),
+			expectError:   true,
+			errorContains: "field is immutable",
+		},
```
renormalize: Logic in `validatePodCliqueUpdate` and `validatePodCliqueScalingGroupConfigsUpdate` are not being tested by this unit test. Can we enhance this test?

renormalize (commented):
thanks for the PR @gflarity!

PR #231, opened 2025-10-23
test coverage for internal/utils

gflarity:
#### What type of PR is this?

Testing

Increased test coverage for internal/utils tree.

Prompt:

```release-note
None
```

### DIFF
--- a/operator/internal/utils/concurrent_test.go
+++ b/operator/internal/utils/concurrent_test.go
@@ -18,6 +18,7 @@ package utils
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"testing"


operator/internal/utils/concurrent_test.go
```diff
@@ -206,3 +206,48 @@ func createErringTaskFn(name string) func(ctx context.Context) error {
...
+	assert.NotNil(t, aggregatedErr)
+	assert.Contains(t, aggregatedErr.Error(), "task failed")
+
+	// Multiple errors returns joined error
+	err2 := fmt.Errorf("another failure")
+	result = RunResult{Errors: []error{err1, err2}}
+	aggregatedErr = result.GetAggregatedError()
+	assert.NotNil(t, aggregatedErr)
+	assert.Contains(t, aggregatedErr.Error(), "task failed")
+	assert.Contains(t, aggregatedErr.Error(), "another failure")
```
renormalize:
```suggestion
	assert.True(t, errors.Is(aggregatedErr, err1))
	assert.True(t, errors.Is(aggregatedErr, err2))
```

operator/internal/utils/concurrent_test.go
```diff
@@ -206,3 +206,48 @@ func createErringTaskFn(name string) func(ctx context.Context) error {
...
+		SuccessfulTasks: []string{"task1", "task2"},
+		FailedTasks:     []string{"task3"},
+		SkippedTasks:    []string{"task4", "task5"},
+	}
+	summary = result.GetSummary()
+	assert.Contains(t, summary, "task1")
+	assert.Contains(t, summary, "task2")
+	assert.Contains(t, summary, "task3")
+	assert.Contains(t, summary, "task4")
+	assert.Contains(t, summary, "task5")
```
renormalize: Establishes the task exists in the summary, along with the categorization; instead of just the former in the current test. Also is consistent with the empty result test above.

```suggestion
	assert.Contains(t, summary, "SuccessfulTasks: [task1 task2]")
	assert.Contains(t, summary, "FailedTasks: [task3]")
	assert.Contains(t, summary, "SkippedTasks: [task4 task5]")
```

operator/internal/utils/concurrent_test.go
```diff
@@ -206,3 +206,48 @@ func createErringTaskFn(name string) func(ctx context.Context) error {
...
+	// No errors returns nil
+	result := RunResult{Errors: []error{}}
+	assert.Nil(t, result.GetAggregatedError())
+
+	// Single error returns that error
+	err1 := fmt.Errorf("task failed")
+	result = RunResult{Errors: []error{err1}}
+	aggregatedErr := result.GetAggregatedError()
+	assert.NotNil(t, aggregatedErr)
+	assert.Contains(t, aggregatedErr.Error(), "task failed")
```
renormalize: Since the errors are `Join()`ed, I think this is a better way to verify things.

```suggestion
	assert.True(t, errors.Is(aggregatedErr, err1))
```

operator/internal/utils/concurrent_test.go
renormalize:
```suggestion
	assert.True(t, errors.Is(aggregatedErr, err1))
	assert.True(t, errors.Is(aggregatedErr, err2))
```

operator/internal/utils/kubernetes/pod_test.go
```diff
@@ -113,3 +115,552 @@ func TestIsPodScheduled(t *testing.T) {
...
+
+	// Check started but not ready pods
+	assert.Len(t, categories[PodStartedButNotReady], 1)
+	assert.Equal(t, "started-not-ready-pod", categories[PodStartedButNotReady][0].Name)
+}
+
+// Helper function to create a bool pointer
+func boolPtr(b bool) *bool {
+	return &b
+}
```
renormalize: As mentioned in https://github.com/ai-dynamo/grove/pull/230#discussion_r2499400091, let's use the generic function `ptr.To()`.

renormalize (changes requested):
thanks for the PR @gflarity!

PR #233, opened 2025-10-24
Manage external cert

kangclzjc:
#### What type of PR is this?
Currently we do the certificate generation ourselves. If they already have certificate in their system, we should manage external existing certificate by mounting them

We add support for allowing user to use external cert 

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/operator/charts/templates/authorizer-webhook-config.yaml
+++ b/operator/charts/templates/authorizer-webhook-config.yaml
@@ -4,7 +4,6 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: authorizer-webhook
-  namespace: {{ .Release.Namespace }}
   {{- if .Values.webhooks.authorizerWebhook.annotations }}
   annotations:
     {{- toYaml .Values.webhooks.authorizerWebhook.annotations | nindent 4 }}
--- a/operator/charts/templates/pcs-defaulting-webhook-config.yaml
+++ b/operator/charts/templates/pcs-defaulting-webhook-config.yaml
@@ -3,7 +3,6 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
   name: podcliqueset-defaulting-webhook
-  namespace: {{ .Release.Namespace }}
   {{- if .Values.webhooks.podCliqueSetDefaultingWebhook.annotations }}
   annotations:
     {{- toYaml .Values.webhooks.podCliqueSetDefaultingWebhook.annotations | nindent 4 }}
--- a/operator/charts/templates/pcs-validating-webhook-config.yaml
+++ b/operator/charts/templates/pcs-validating-webhook-config.yaml
@@ -3,7 +3,6 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: podcliqueset-validating-webhook
-  namespace: {{ .Release.Namespace }}
   {{- if .Values.webhooks.podCliqueSetValidationWebhook.annotations }}
   annotations:
     {{- toYaml .Values.webhooks.podCliqueSetValidationWebhook.annotations | nindent 4 }}
--- a/operator/charts/templates/webhook-server-cert-secret.yaml
+++ b/operator/charts/templates/webhook-server-cert-secret.yaml
@@ -1,3 +1,9 @@
+{{- /*
+  Create this Secret if autoProvision is either:
+  - Not specified (defaults to true behavior), OR
+  - Explicitly set to true
+  This allows the operator to auto-generate webhook TLS certs at startup.
+*/ -}}
 {{- if or (not (hasKey .Values.config.server.webhooks "autoProvision")) .Values.config.server.webhooks.autoProvision }}
 ---
 apiVersion: v1
--- a/operator/internal/controller/cert/cert.go
+++ b/operator/internal/controller/cert/cert.go
@@ -62,7 +62,7 @@ func ManageWebhookCerts(mgr ctrl.Manager, certDir string, secretName string, aut
 	rotator := &cert.CertRotator{
 		SecretKey: types.NamespacedName{
 			Namespace: namespace,
-			Name:      "grove-webhook-server-cert",
+			Name:      secretName,
 		},
 		CertDir:        certDir,
 		CAName:         certificateAuthorityName,
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -45,6 +45,8 @@ jobs:
             test_pattern: "^Test_RU"
           - test_name: startup_ordering
             test_pattern: "^Test_SO"
+          - test_name: cert_management
+            test_pattern: "^Test_CM"
     name: E2E - ${{ matrix.test_name }}
     steps:
       # print runner specs so we have a record incase of failures
--- a/operator/e2e/setup/grove.go
+++ b/operator/e2e/setup/grove.go
@@ -1,0 +1,120 @@
+// /*
+// Copyright 2025 The Grove Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// */
+
+package setup
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+
+	"github.com/ai-dynamo/grove/operator/e2e/utils"
+	"gopkg.in/yaml.v3"
+	"k8s.io/client-go/rest"
+)
+
+// UpgradeGrove upgrades the Grove operator with the specified Helm values.
+//
+// This uses Helm upgrade (rather than Skaffold) because:
+// 1. Grove is initially installed via Skaffold, which uses Helm under the hood
+// 2. For config-only changes (like switching cert modes), rebuilding images is unnecessary
+// 3. Helm upgrade with ReuseValues preserves the image configuration that Skaffold set
+//
+// This approach avoids wasteful rebuilds while staying compatible with the Skaffold installation.
+func UpgradeGrove(ctx context.Context, restConfig *rest.Config, values map[string]interface{}, logger *utils.Logger) error {
+	chartPath, err := getGroveChartPath()
+	if err != nil {
+		return fmt.Errorf("failed to get Grove chart path: %w", err)
+	}
+
+	chartVersion, err := getChartVersion(chartPath)
+	if err != nil {
+		return fmt.Errorf("failed to get chart version: %w", err)
+	}
+
+	// Configure Helm upgrade using shared constants to stay in sync with Skaffold installation.
+	// - ReleaseName: Uses OperatorDeploymentName from constants.go (matches skaffold.yaml's deploy.helm.releases[0].name)
+	// - ChartVersion: Read from Chart.yaml to avoid version string duplication
+	// - ReuseValues: Preserves image configuration that Skaffold set during initial install
+	helmConfig := &HelmInstallConfig{
+		RestConfig:     restConfig,
+		ReleaseName:    OperatorDeploymentName,
+		ChartRef:       chartPath,
+		ChartVersion:   chartVersion,
+		Namespace:      OperatorNamespace,
+		ReuseValues:    true,
+		Values:         values,
+		HelmLoggerFunc: logger.Debugf,
+		Logger:         logger,
+	}
+
+	logger.Debug("Upgrading Grove operator")
+
+	if _, err := UpgradeHelmChart(helmConfig); err != nil {
+		return fmt.Errorf("helm upgrade failed: %w", err)
+	}
+
+	// Wait for Grove operator pod to be ready after upgrade
+	if err := utils.WaitForPodsInNamespace(ctx, OperatorNamespace, restConfig, 1, defaultPollTimeout, defaultPollInterval, logger); err != nil {
+		return fmt.Errorf("grove operator pod not ready after upgrade: %w", err)
+	}
+
+	logger.Debug("Grove upgrade completed successfully")
+	return nil
+}
+
+// chartYAML represents the structure of Chart.yaml for version extraction
+type chartYAML struct {
+	Version string `yaml:"version"`
+}
+
+// getChartVersion reads the version from Chart.yaml in the given chart directory.
+// We read from Chart.yaml rather than hardcoding the version to maintain a single source
+// of truth. This ensures the upgrade uses the same version as the chart itself, avoiding
+// configuration drift between the chart definition and the e2e test code.
+func getChartVersion(chartPath string) (string, error) {
+	chartFile := filepath.Join(chartPath, "Chart.yaml")
+	data, err := os.ReadFile(chartFile)
+	if err != nil {
+		return "", fmt.Errorf("failed to read Chart.yaml: %w", err)
+	}
+
+	var chart chartYAML
+	if err := yaml.Unmarshal(data, &chart); err != nil {
+		return "", fmt.Errorf("failed to parse Chart.yaml: %w", err)
+	}
+
+	if chart.Version == "" {
+		return "", fmt.Errorf("version not found in Chart.yaml")
+	}
+
+	return chart.Version, nil
+}
+
+// getGroveChartPath returns the absolute path to the Grove Helm chart.
+// It uses runtime.Caller to find the path relative to this source file.
+func getGroveChartPath() (string, error) {
+	_, currentFile, _, ok := runtime.Caller(0)
+	if !ok {
+		return "", fmt.Errorf("failed to get current file path")
+	}
+	// This file is at operator/e2e/setup/grove.go
+	// Chart is at operator/charts
+	chartPath := filepath.Join(filepath.Dir(currentFile), "../../charts")
+	return filepath.Abs(chartPath)
+}
--- a/operator/e2e/tests/cert_management_test.go
+++ b/operator/e2e/tests/cert_management_test.go
@@ -1,0 +1,714 @@
+//go:build e2e
+
+// /*
+// Copyright 2025 The Grove Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// */
+
+package tests
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"path/filepath"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/ai-dynamo/grove/operator/e2e"
+	"github.com/ai-dynamo/grove/operator/e2e/setup"
+	"github.com/ai-dynamo/grove/operator/e2e/utils"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+)
+
+// certManagerGroveValues returns Grove Helm values for cert-manager mode.
+// This disables auto-provisioning and configures webhook annotations for cert-manager CA injection.
+func certManagerGroveValues() map[string]interface{} {
+	certManagerAnnotation := map[string]interface{}{
+		"cert-manager.io/inject-ca-from": "grove-system/grove-webhook-server-cert",
+	}
+	return map[string]interface{}{
+		"installCRDs": true,
+		"config": map[string]interface{}{
+			"server": map[string]interface{}{
+				"webhooks": map[string]interface{}{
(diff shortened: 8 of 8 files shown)


operator/api/config/v1alpha1/types.go
```diff
@@ -135,6 +135,27 @@ type WebhookServer struct {
...
+	CertManagement *WebhookCertManagement `json:"certManagement,omitempty"`
+}
+
+// WebhookCertManagement defines how webhook certificates are managed.
+type WebhookCertManagement struct {
+	// AutoProvision enables automatic certificate generation and rotation.
+	// If set to false, you must provide your own certificates via Secret.
+	// Default: true
+	// +optional
+	AutoProvision *bool `json:"autoProvision,omitempty"`
```
gflarity: Hmm... maybe just default to auto provision in the code if SecretName isn't set, then the user doesn't have to remember to disable and we don't need to track this variable.

kangclzjc: You're right. It seems that we can detect secretName or certFilesPath user want to mount to decide whether to auto provision

operator/charts/templates/authorizer-webhook-config.yaml
```diff
@@ -5,12 +5,20 @@ kind: ValidatingWebhookConfiguration
 metadata:
   name: authorizer-webhook
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.config.server.webhooks.certManagement.certManagerEnabled }}
+  annotations:
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ .Values.config.server.webhooks.certManagement.secretName }}
```
gflarity: Let's avoid cert manager requirements. The webhook just needs the CA bundle.

operator/charts/templates/pcs-defaulting-webhook-config.yaml
```diff
@@ -4,12 +4,20 @@ kind: MutatingWebhookConfiguration
 metadata:
   name: podcliqueset-defaulting-webhook
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.config.server.webhooks.certManagement.certManagerEnabled }}
+  annotations:
```
gflarity, line 7: Here as well. Let's not reference cert manager specific annotations. Webhook config's just need the CA Bundle.

operator/charts/templates/pcs-validating-webhook-config.yaml
```diff
@@ -4,12 +4,20 @@ kind: ValidatingWebhookConfiguration
 metadata:
   name: podcliqueset-validating-webhook
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.config.server.webhooks.certManagement.certManagerEnabled }}
```
gflarity: Here as well.

operator/charts/templates/webhook-server-cert-secret.yaml
```diff
@@ -1,11 +1,24 @@
...
-  name: grove-webhook-server-cert
+  name: {{ .Values.config.server.webhooks.certManagement.secretName }}
   namespace: {{ .Release.Namespace }}
   labels:
 {{- include "operator.server.secret.labels" . | nindent 4 }}
 type: kubernetes.io/tls
 data:
+  {{- if .Values.config.server.webhooks.certManagement.certFilesPath }}
+  # Certificate files provided: read from chart directory
+  tls.crt: {{ .Files.Get (printf "%s/server.crt" .Values.config.server.webhooks.certManagement.certFilesPath) | b64enc }}
```
gflarity: I don't understand why we're using .File.Get here?  

Cert Manager will bundle the following into a secret file:
```
tls.crt: The signed public certificate, often followed by the full intermediate certificate chain.

tls.key: The private key for the certificate.

ca.crt: The root CA certificate (this is common when using a private CA).

```

All that's needed is to mount this secret into the webhook deployment. Then the webhook service can use the certificate and private key. Note the ca.crt isn't used by the deployment itself. 

For those who aren't using Cert Manager, they can just create a similar secret however they want and everything will work.

operator/charts/templates/authorizer-webhook-config.yaml
kangclzjc: Yeah, it should be right we remove this cert manager requirements.

operator/charts/templates/webhook-server-cert-secret.yaml
kangclzjc: Here I want to implement this:
When customer give certFilesPath, we create the secret based on these files at the time we install grove operator. 

So do you want to leave creating secret tasks to users themselves, even when they know the certFilesPath, right ?

gflarity: Yes, cert manager will create the secrets for you. If you don't use cert manager, you can create them by hand.

kangclzjc: Now, I change the logic to this:
if user provide secretName, then we use external cert else we use autoProvision(cert controller) to generate our own cert.

operator/charts/templates/authorizer-webhook-config.yaml
```diff
@@ -11,6 +11,13 @@ webhooks:
   - admissionReviewVersions:
       - v1
     clientConfig:
+      {{- if .Values.config.server.webhooks.secretName }}
+      # CA Bundle from secret
+      {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.config.server.webhooks.secretName }}
```
gflarity: Don't try to read the CA from secret like this, it requires changing the secret name if you want to update the secret and ca bundle. Instead we can make it much more flexible and simpler. Consider the following two cases.

## 1. Using cert-manager
In this case the operator/admin is going to want to use the annotation feature of cert-manager
annotations:
```yaml
metadata:
  annotations:
    # This tells cert-manager to find the secret and inject the CA automatically
    cert-manager.io/inject-ca-from: "default/my-webhook-secret"
```

But to keep things flexible, just allow operator/admin to provide ANY annotations they want. 

```yaml
  {{- if .Values.mutatingWebhookConfiguration.annotations }}
  annotations:
    {{- toYaml .Values.mutatingWebhookConfiguration.annotations | nindent 4 }}
  {{- end }}
```

## 2. Using Anything Else (even hand creating them)
The operator/admin wants to specify everything by hand.  So give them values for the CA Bundle (which will change less frequently) and a secret with the cert + private key (which could change much more frequently).  This way they could change them independently and have maximum control/flexibility. 

```yaml
clientConfig:
    caBundle: {{ .Values.caBundle }}
```

## 3. Auto Provision (Default)
In this case, no annotation or CA is provided via values, so just leave them empty, "". 


```yaml
metadata:
  name: {{ .Values.name }} 
  annotations:   #template will render this empty, just like it now is
webhooks:
- admissionReviewVersions:
  - v1
  - v1beta1
  clientConfig:
    caBundle:   # template will render this empty, just like it now is
```
---

Please see the following for a working example of 1/2: https://github.com/CentML/ecr-anywhere/blob/main/charts/ecr-anywhere/templates/mutatingwebhookconfiguration.yaml 

For auto provision, only disable it if secretName isn't set. Use the default autoprovisioned values for the mounts etc.

kangclzjc: Addressed. Currently these three are all supported.

gflarity (changes requested):
Thanks @kangclzjc, gave it a quick review today. Looks good, I'll do a slower review tomorrow morning first thing. In the mean time:

One thing we'll need is a simple E2E test to verifying everything works using cert manager. Please take a look at the current E2E tests, all it really needs to do is install cert manager, then update the helm values to use it. If you start a simple grove deployment successfully, it means the webhooks are still working with this change. 

There  is code in the cluster setup utilities for installing a helm chart (cert manager). You should be able to reuse some of that logic to install cert manager for this specific test. Please let me know if you have any questions.

kangclzjc:
> Thanks @kangclzjc, gave it a quick review today. Looks good, I'll do a slower review tomorrow morning first thing. In the mean time:
> 
> One thing we'll need is a simple E2E test to verifying everything works using cert manager. Please take a look at the current E2E tests, all it really needs to do is install cert manager, then update the helm values to use it. If you stat a simple grove deployment successfully, it means the webhooks are still working with this change.
> 
> There code in the cluster setup utilities for installing a helm chart (cert manager). You should be able to reuse some of that logic to install cert manager for this specific test. Please let me know if you have any questions.

Thanks @gflarity , let me check the E2E test first. And then do a simple E2E test to verifying everything works using cert manager.

docs/api-reference/operator-api.md
```diff
@@ -898,5 +898,7 @@ _Appears in:_
 | `bindAddress` _string_ | BindAddress is the IP address on which to listen for the specified port. |  |  |
 | `port` _integer_ | Port is the port on which to serve requests. |  |  |
 | `serverCertDir` _string_ | ServerCertDir is the directory containing the server certificate and key. |  |  |
+| `secretName` _string_ | SecretName is the name of the Kubernetes Secret containing webhook certificates.<br />The Secret must contain tls.crt, tls.key, and ca.crt. |  |  |
+| `autoProvision` _boolean_ | AutoProvision enables automatic certificate generation and management via cert-controller.<br />When true: cert-controller automatically generates self-signed certificates and stores them in the Secret.<br />When false: certificates are expected to be provided externally (e.g., by cert-manager, cluster admin).<br />Default: true (auto-generate certificates for easy setup) |  |  |
```
gflarity: There's a default column, let's fill it in. 

```suggestion
| `secretName` _string_ | SecretName is the name of the Kubernetes Secret containing webhook certificates.<br />The Secret must contain tls.crt, tls.key, and ca.crt. | grove-webhook-server-cert |  |
| `autoProvision` _boolean_ | AutoProvision enables automatic certificate generation and management via cert-controller.<br />When true: cert-controller automatically generates self-signed certificates and stores them in the Secret.<br />When false: certificates are expected to be provided externally (e.g., by cert-manager, cluster admin).| true |  |
```

operator/charts/values.yaml
```diff
@@ -46,6 +46,25 @@ config:
   server:
     webhooks:
       port: 9443
+      # serverCertDir is the directory path where certificate files are located.
+      # The operator will look for tls.crt and tls.key in this directory.
+      # Default: /etc/grove-operator/webhook-certs
+      serverCertDir: /etc/grove-operator/webhook-certs
+      # secretName is the name of the Kubernetes Secret containing webhook certificates.
+      # The Secret must contain tls.crt, tls.key, and ca.crt.
+      # Default: ""
```
gflarity: Set this to "grove-webhook-server-cert" to match the default in the code? That way anyone reading the values knows where to look in Kubernetes.

operator/charts/values.yaml
```diff
@@ -46,6 +46,25 @@ config:
...
+      # autoProvision enables automatic certificate generation via cert-controller.
+      # When true (default): cert-controller automatically generates self-signed certificates.
+      #   - Best for development/testing
+      #   - No manual certificate management needed
+      #   - Certificates auto-rotate before expiry
+      # When false: certificates must be provided externally (e.g., cert-manager, manual).
+      #   - Best for production
+      #   - Use your own CA or Let's Encrypt
+      #   - Requires Secret to exist before operator starts
+      # Default: if secretName is not set, then autoProvision is true, otherwise it is false
```
gflarity: I don't think this is true? If secretName is not set it gets defaulted to "grove-webhook-server-cert" in defaults.go?

Its fine to require `autoProvision: false` in order to use external certificates. We can keep it this way, just fix the comment.

operator/charts/values.yaml
kangclzjc: Yes, no problem

docs/api-reference/operator-api.md
kangclzjc: OK~

operator/charts/values.yaml
kangclzjc: I check this only based on the secretName. If secretName is not set, I will check only depends on the secretName in _helpers.tpl. But let's use the simple way you mention like require autoProvision: false in order to use external certificates. Otherwise we will use autoProvision.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	cmConfig := &setup.HelmInstallConfig{
+		RestConfig:      restConfig,
+		ReleaseName:     deps.HelmCharts.CertManager.ReleaseName,
+		ChartRef:        deps.HelmCharts.CertManager.ChartRef,
+		ChartVersion:    deps.HelmCharts.CertManager.Version,
+		Namespace:       deps.HelmCharts.CertManager.Namespace,
+		CreateNamespace: true,
+		Wait:            true,
+		Timeout:         10 * time.Minute, // cert-manager needs time to download images and start
```
gflarity: Normally we don't wait, but install the helm chart and then poll for the pods to be ready with a timeout. Is there a ready why we're waiting here instead of doing something like that?

operator/e2e/dependencies.yaml
```diff
@@ -48,3 +48,11 @@ helmCharts:
     version: v25.3.4
     namespace: gpu-operator
     repoURL: https://helm.ngc.nvidia.com/nvidia
+
```
gflarity, line 53: Can you Include the cert manager images so that it's faster if you have them in docker locally?

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+		t.Fatalf("cert-manager pods not ready: %v", err)
+	}
+
+	// 3. Create Issuer and Certificate
+	logger.Info("Creating ClusterIssuer and Certificate...")
+	if err := applyYAMLContent(ctx, restConfig, certManagerIssuerYAML); err != nil {
+		t.Fatalf("Failed to apply ClusterIssuer: %v", err)
+	}
+
+	// Wait a bit for ClusterIssuer
```
gflarity: Rather than setting an arbitrary time out, poll to get the cluster issuer custom resource  directly and check the status. You can find an example of getting a custom resource in ScalePodCliqueScalingGroupWithClient (in setup.go).

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+		Logger:         logger,
+	}
+
+	logger.Info("Installing cert-manager...")
+	if _, err := setup.InstallHelmChart(cmConfig); err != nil {
+		t.Fatalf("Failed to install cert-manager: %v", err)
+	}
+
+	// Wait for cert-manager webhook to be ready (InstallHelmChart with Wait=true should handle it, but to be safe)
+	if err := utils.WaitForPodsInNamespace(ctx, cmConfig.Namespace, restConfig, 3, 2*time.Minute, 5*time.Second, logger); err != nil {
```
gflarity: Do we require 2 minutes? Can we try defaultPullTimeout instead?

Also just use defaultPullInterval here (it's also 5 seconds).

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+	// Wait a bit for ClusterIssuer
+	time.Sleep(5 * time.Second)
+
+	if err := applyYAMLContent(ctx, restConfig, certManagerCertificateYAML); err != nil {
+		t.Fatalf("Failed to apply Certificate: %v", err)
+	}
+
+	// 4. Wait for Secret to be created
+	logger.Info("Waiting for certificate secret to be created...")
+	err = utils.PollForCondition(ctx, 1*time.Minute, 2*time.Second, func() (bool, error) {
```
gflarity: Use defaultPollTimeout and defaultPollInterval instead.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("Certificate secret was not created: %v", err)
+	}
+
+	// 5. Update Grove to use the external secret via Helm upgrade
+	logger.Info("Upgrading Grove to use cert-manager secret...")
+
+	chartPath := "../../charts"
```
gflarity: Please follow the runtime.Caller(0) pattern used else where.  See CreateDefaultKaiQueues and waitForWebhookReady for examples.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+	logger.Infof("Using chart path: %s", chartPath)
+
+	// Delete the old auto-provisioned webhook secret if it exists
+	logger.Info("Deleting old webhook secret if exists...")
+	_ = clientset.CoreV1().Secrets("grove-system").Delete(ctx, "grove-webhook-server-cert", metav1.DeleteOptions{})
+	time.Sleep(30 * time.Second)
```
gflarity: What's this sleep for?  Please poll for the require state changes instead.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	groveUpgradeConfig := &setup.HelmInstallConfig{
+		RestConfig:   restConfig,
+		ReleaseName:  "grove-operator",
+		ChartRef:     chartPath,
+		ChartVersion: "0.1.0-dev",
+		Namespace:    "grove-system",
+		Wait:         true,
+		Timeout:      10 * time.Minute, // Allow time for pod restart
```
gflarity: Similar to above, why wait, and why 10 minutes? As opposed to polling with a timeout on the pods etc?

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+			},
+		},
+		HelmLoggerFunc: logger.Debugf,
+		Logger:         logger,
+	}
+
+	logger.Info("Starting Helm upgrade (this may take several minutes)...")
+	if _, err := setup.UpgradeHelmChart(groveUpgradeConfig); err != nil {
+		// Get more info about what went wrong
+		logger.Error("Helm upgrade failed, checking Grove pods status...")
```
gflarity: Duplicate. t.Fatalf will already provide this information.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+		pods, podErr := clientset.CoreV1().Pods("grove-system").List(ctx, metav1.ListOptions{})
+		if podErr == nil {
+			logger.Infof("Found %d pods in grove-system:", len(pods.Items))
+			for _, pod := range pods.Items {
+				logger.Infof("  - %s: %s (Ready: %v)", pod.Name, pod.Status.Phase, pod.Status.Conditions)
```
gflarity: This should probably be a Debugf instead.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+			}
+		}
+		t.Fatalf("Failed to upgrade Grove: %v", err)
+	}
+
+	time.Sleep(30 * time.Second)
```
gflarity: Another long sleep. Please poll.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	workloadPath := "../yaml/workload1.yaml"
+	tc := TestContext{
+		T:             t,
+		Ctx:           ctx,
+		Clientset:     clientset,
+		DynamicClient: dynamicClient,
+		RestConfig:    restConfig,
+		Namespace:     "default",
+		Timeout:       2 * time.Minute,
```
gflarity: Use defaultTimeout and defaultPoll unless there's a very specific reason. Let's discuss if there is.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	if _, err := deployAndVerifyWorkload(tc); err != nil {
+		t.Fatalf("Failed to verify workload: %v", err)
+	}
+
+	logger.Info("✅ Cert-manager integration test passed!")
+}
+
+// helper to apply YAML string
+func applyYAMLContent(ctx context.Context, restConfig *rest.Config, content string) error {
```
gflarity: utils/k8s_client.go has an unexported function applyYAMLData (line 132) that applies YAML content directly as bytes. Let's  export applyYAMLData as a public function so cert_manager_test.go can use it directly without the temp file workaround.

operator/e2e/tests/cert_manager_test.go
```diff
@@ -0,0 +1,247 @@
...
+    kind: ClusterIssuer
+  dnsNames:
+  - grove-operator
+  - grove-operator.grove-system
+  - grove-operator.grove-system.svc
+  - grove-operator.grove-system.svc.cluster.local
+`
+)
+
+func TestCertManagerIntegration(t *testing.T) {
```
gflarity: This tests going from autoProvision => cert manager. Let's create another test that starts with cert manager and goes to auto provsion, so we're testing both paths.  Were ever possible, don't forget to write helper functions in order to keep things DRY: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

operator/e2e/tests/cert_manager_test.go
kangclzjc: Yes, I see that I can use WaitForPodsInNamespace to poll for the pods

operator/e2e/dependencies.yaml
kangclzjc: @gflarity , definitely it will be faster, however it will make the repo larger. Would you still like me to include this images ?

kangclzjc: Yes, definitely

gflarity:
Just circling back on this @kangclz so we can get it merged.  What's the current status? 

Please resolve any comments you believe have now been addressed.  

Also, looks like the E2E tests are failing. Please try rebasing from main to see if that helps with the tests.

kangclzjc:
> Just circling back on this @kangclz so we can get it merged. What's the current status?
> 
> Please resolve any comments you believe have now been addressed.
> 
> Also, looks like the E2E tests are failing. Please try rebasing from main to see if that helps with the tests.


> Just circling back on this @kangclz so we can get it merged. What's the current status?
> 
> Please resolve any comments you believe have now been addressed.
> 
> Also, looks like the E2E tests are failing. Please try rebasing from main to see if that helps with the tests.

@gflarity Sorry for replying late. I resolve all the conversations before and format the test name and logs refer to other tests. PTAL.

operator/internal/controller/cert/cert.go
```diff
@@ -38,13 +38,27 @@ const (
...
+
+	if !autoProvision {
+		logger.Info("Using externally provided certificates",
+			"certDir", certDir, "secretName", secretName)
+		// Certificates are managed externally, signal ready immediately
+		close(certsReadyCh)
+		return nil
+	}
+
+	logger.Info("Auto-provisioning certificates using cert-controller",
```
gflarity, line 60: secretName isn't used below:

```
Name:      "grove-webhook-server-cert",
```

operator/charts/templates/authorizer-webhook-config.yaml
```diff
@@ -4,12 +4,20 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: authorizer-webhook
+  namespace: {{ .Release.Namespace }}
```
gflarity: ValidatingWebhookConfiguration is cluster scoped, namespace does nothing here and should be removed.

PR #238, opened 2025-10-31
added missed missed PR feedback

gflarity:
#### What type of PR is this?

Enhancement

This PR implements feedback made by @renormalize that was missed during the review process. 

Here's the specific conversations:
https://github.com/ai-dynamo/grove/pull/207#discussion_r2480411030
https://github.com/ai-dynamo/grove/pull/207#discussion_r2480537223
https://github.com/ai-dynamo/grove/pull/207#discussion_r2480667411
https://github.com/ai-dynamo/grove/pull/207#discussion_r2480595268

```release-note
NONE
```

```docs

```

### DIFF
--- a/operator/e2e/setup/shared_cluster.go
+++ b/operator/e2e/setup/shared_cluster.go
@@ -146,20 +146,18 @@ func (scm *SharedClusterManager) Setup(ctx context.Context, testImages []string)
 	scm.logger.Info("🚀 Setting up shared k3d cluster for all e2e tests...")
 
 	restConfig, cleanup, err := SetupCompleteK3DCluster(ctx, customCfg, relativeSkaffoldYAMLPath, scm.logger)
+	// Defer cleanup on error - only call if setup was not successful and we have a cleanup function
+	defer func() {
+		if !setupSuccessful && cleanup != nil {
+			cleanup()
+		}
+	}()
 	if err != nil {
 		return fmt.Errorf("failed to setup shared k3d cluster: %w", err)
 	}
 
 	scm.restConfig = restConfig
 	scm.cleanup = cleanup
-
-	// Defer cleanup on error - only call if setup was not successful and we have a cleanup function
-	defer func() {
-		if !setupSuccessful && cleanup != nil {
-			scm.logger.Error("⚠️ Setup failed, cleaning up cluster...")
-			cleanup()
-		}
-	}()
 
 	// Create clientset from restConfig
 	clientset, err := kubernetes.NewForConfig(restConfig)


operator/e2e/setup/shared_cluster.go
```diff
@@ -41,6 +41,22 @@ const (
...
+// resourceType represents a Kubernetes resource type for cleanup operations
+type resourceType struct {
+	group    string
+	version  string
+	resource string
+	name     string
+}
+
+// resourceTypes defines all Grove CRD types that need to be tracked for cleanup
+var resourceTypes = []resourceType{
```
gflarity: @renormalize I looked into adding the addition CRDs you mentioned, but there will be resources owned by resources other than Grove included and it'll stall the waiting. Is there a label or something I can use to know they were owned by Grove?

renormalize: All resources created by grove-operator will have the following label `app.kubernetes.io/managed-by: grove-operator` applied on them.
We can also filter based on the name of the PodCliqueSet that the resource was created for `app.kubernetes.io/part-of: <pcs-name>`.

This util is used to add the mentioned two labels on all resources created by grove-operator for a `PodCliqueSet`:

https://github.com/ai-dynamo/grove/blob/141155e714cb5df99f1a291a548aa9fe911062f5/operator/api/common/labels.go#L86-L92

operator/e2e/setup/shared_cluster.go
```diff
@@ -125,29 +141,25 @@ func (scm *SharedClusterManager) Setup(ctx context.Context, testImages []string)
 	// Create clientset from restConfig
 	clientset, err := kubernetes.NewForConfig(restConfig)
 	if err != nil {
-		cleanup()
 		return fmt.Errorf("failed to create clientset: %w", err)
```
renormalize, line 165: The defer-ed call to `cleanup()` as suggested in https://github.com/ai-dynamo/grove/pull/207#discussion_r2480411030 when an error occurs has not been added. Can it be added? Thanks.

operator/e2e/setup/shared_cluster.go
```diff
@@ -296,18 +303,7 @@ func (scm *SharedClusterManager) resetNodeStates(ctx context.Context) error {
...
-		version  string
-		resource string
-		name     string
-	}{
-		{"grove.io", "v1alpha1", "podcliquesets", "PodCliqueSets"},
-		{"grove.io", "v1alpha1", "podcliquescalinggroups", "PodCliqueScalingGroups"},
-		{"grove.io", "v1alpha1", "podgangsets", "PodGangSets"},
-		{"scheduler.grove.io", "v1alpha1", "podgangs", "PodGangs"},
-	}
+	resourceTypes := resourceTypes
```
renormalize: I might be missing something obvious, but I do not see a need for this declaration. We do not mutate `resourceTypes` anywhere either.

operator/e2e/setup/shared_cluster.go
gflarity: @renormalize It turns out the cleanup on err wasn't necessary as the Teardown() function calls this already, so these cleanups were actually unnecessary.

operator/e2e/setup/shared_cluster.go
gflarity: Thanks, I'll give that shot and update.

operator/e2e/setup/shared_cluster.go
gflarity: Hmm, I'll take a look.

operator/e2e/setup/shared_cluster.go
renormalize: @gflarity 

`Teardown()` isn't called when there are multiple test suites: `TestMain()` sets up the cluster and delegates running all individial test cases to `m.Run()`.

https://github.com/ai-dynamo/grove/blob/d76f83f3865c1f8f2bc347d139c76c92302dc32d/operator/e2e/tests/main_test.go#L46-L57

If `Setup()` fails, `Teardown()` is never called. Please correct me if I am missing something obvious.

The only other invocation of `Teardown()` is:

https://github.com/ai-dynamo/grove/blob/d76f83f3865c1f8f2bc347d139c76c92302dc32d/operator/e2e/tests/setup.go#L100-L103

which is only called when it's not the entire suite being run.

We just need to call `Teardown()` in `TestMain()` too.

gflarity: You're right, we need to make sure there's a cleanup if there's an error during Setup. However I think using 'err' could be fragile as it's easy to mask it using the short variable declaration (:=) etc. So i've used a flag that only lifts at the very end when everything is successful. 

Also, the isRunningFullSuite logic was left over from a earlier iteration and not necessary anymore. TestMain will always run whether it's one test with --run  or all of them. So I removed it and cleaned it up. 

PTAL

operator/e2e/setup/shared_cluster.go
gflarity: This is done.

renormalize (approved):
thanks for all the changes @gflarity!

renormalize:
Let's merge this first, and then #239.

PR #241, opened 2025-11-03
add support for prepulling images to speed up tests on slow networks

gflarity:
#### What type of PR is this?

Testing Feature (Improvement)

Pulling images on slow networks can easily double the shared cluster setup time. However these images will not change very often and can be cached locally. This PR does this transparently. In particular this mades debugging failing tests faster when you don't have a 500mbps + internet connection. 

```release-note
NONE
```

```docs

```

### DIFF
--- a/operator/e2e/dependencies.go
+++ b/operator/e2e/dependencies.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"sync"
 
-	"gopkg.in/yaml.v3"
+	"sigs.k8s.io/yaml"
 )
 
 //go:embed dependencies.yaml
--- a/operator/go.mod
+++ b/operator/go.mod
@@ -17,7 +17,6 @@ require (
 	github.com/spf13/pflag v1.0.9
 	github.com/stretchr/testify v1.11.1
 	go.uber.org/zap v1.27.0
-	gopkg.in/yaml.v3 v3.0.1
 	helm.sh/helm/v3 v3.19.0
 	k8s.io/api v0.34.1
 	k8s.io/apiextensions-apiserver v0.34.1
@@ -27,6 +26,7 @@ require (
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
 	sigs.k8s.io/controller-runtime v0.21.0
+	sigs.k8s.io/yaml v1.6.0
 )
 
 require (
@@ -196,6 +196,7 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/apiserver v0.34.1 // indirect
 	k8s.io/component-base v0.34.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20250814151709-d7b6acb124c3 // indirect
@@ -206,7 +207,6 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.20.1 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
-	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
 replace (


operator/e2e/setup/k8s_clusters.go
```diff
@@ -621,3 +684,238 @@ func restartNodeContainer(ctx context.Context, nodeName string, logger *utils.Lo
 	logger.Debugf("  ✅ Container restarted successfully: %s", targetContainer.ID[:12])
 	return nil
 }
+
+// prepullImages pre-pulls container images and pushes them to the local k3d registry
+// This significantly speeds up cluster startup by avoiding image pull delays during pod creation.
```
shayasoolin, line 769: This makes me think of another potential optimization, of keeping the registry in place (at least optionally), even after the tests are done (success/fail) in order to allow for faster startup time in the next test run.
I'm talking about the local case, where a developer runs the tests again and again with some minor changes in the code in between. WDYT?

operator/e2e/setup/k8s_clusters.go
```diff
@@ -621,3 +684,238 @@ func restartNodeContainer(ctx context.Context, nodeName string, logger *utils.Lo
...
+	logger.Info("✅ Container images pre-loaded into local registry")
+	return nil
+}
+
+// verifyRegistryImages verifies that images are accessible in the registry
+func verifyRegistryImages(images []string, registryPort string, logger *utils.Logger) error {
+	logger.Debug("🔍 Verifying images are in registry...")
+
+	for _, img := range images {
+		// Extract the image path WITHOUT the registry domain (same logic as prepullImages)
```
shayasoolin, line 872: Same code, consider separating to a function for reuse

shayasoolin (dismissed):

operator/e2e/setup/k8s_clusters.go
gflarity: Hmm... definitely worth trying.  I'll make a todo to circle back on it once I get through a few other prs.

shayasoolin (dismissed):

operator/e2e/setup/k8s_clusters.go
```diff
@@ -51,6 +53,22 @@ const (
...
+		"registry.k8s.io/nfd/node-feature-discovery:v0.17.3",
+		"nvcr.io/nvidia/gpu-operator:v25.3.4",
+		"ghcr.io/nvidia/kai-scheduler/admission:v0.9.3",
+		"ghcr.io/nvidia/kai-scheduler/binder:v0.9.3",
+		"ghcr.io/nvidia/kai-scheduler/operator:v0.9.3",
+		"ghcr.io/nvidia/kai-scheduler/podgroupcontroller:v0.9.3",
+		"ghcr.io/nvidia/kai-scheduler/podgrouper:v0.9.3",
+		"ghcr.io/nvidia/kai-scheduler/queuecontroller:v0.9.3",
+		"ghcr.io/nvidia/kai-scheduler/scheduler:v0.9.3",
+	}
```
sanjaychatterjee: Why are we hardcoding the image tags here? Do you plan to keep updating or do we plan to have configs to overrides?

gflarity: ### Why these images are hardcoded
The version of the helm charts we're using for KAI and GPU Operator are already hardcoded in the tests. So these need to match for there to be any speed up, but it's not the end of the world if they don't. That's why I put in a warning check.

### Why not try to pull the latest helm charts for GPU Operator/KAI? 
The tests will become non-deterministic between runs against the same Grove commit (deps have changed silently). For example, they might fail one day and then pass (or vice versa) the next because of a KAI/GPU Operator change only.  Which makes fixing potential grove issues harder. The goal is to test Grove with the E2E tests ultimately, not valid Grove against new releases of either. 

### Updating dependencies
That said, we still need to figure out a process for updating these dependencies.  For golang we have dependabot which is nice. Perhaps there's something similar for Helm chart dependencies? I'm not sure. I'd recommend creating a ticket for this. IMO, it doesn't seem high priority right now.

sanjaychatterjee: We need to find a way to make any hardcoded tags configurable by a release pipeline. Otherwise, we cannot automate the testing frameworks.
@renormalize can you please take a look?

gflarity: @sanjaychatterjee I'd abstracted out everything that was hardcoded and put a config file. So at least now it's all together and easy to update.  I looked into automatically pull in the helm chart images, but it was too tricky and was missing images. So they'll need to be updated as well.

shayasoolin (approved):

operator/go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/spf13/pflag v1.0.9
 	github.com/stretchr/testify v1.11.1
 	go.uber.org/zap v1.27.0
+	gopkg.in/yaml.v3 v3.0.1
```
renormalize: `operator/e2e/dependencies.go` brings in `gopkg.in/yaml.v3`, I'd prefer direct dependencies being SIG projects. I'm not 100% sure if the APIs are fully compatible but it should be easy enough to use the suggested dependency instead. 

Our dependencies really bloated up out of nowhere, and we should start pruning things out with time once things stabilize.

```suggestion
	sigs.k8s.io/yaml v1.6.0
```

operator/e2e/dependencies.go
```diff
@@ -0,0 +1,103 @@
...
+// */
+
+package e2e
+
+import (
+	_ "embed"
+	"fmt"
+	"sync"
+
+	"gopkg.in/yaml.v3"
```
renormalize: Can this too be removed? The direct dependency still exists.

operator/go.mod
renormalize: https://github.com/ai-dynamo/grove/pull/241#discussion_r2505007465

renormalize (approved):
Posted a minor nit. But everything else looks good to me @gflarity! Thanks!

sanjaychatterjee (commented):
LGTM. Once @renormalize reviews, I can approve.

sanjaychatterjee (approved):

PR #242, opened 2025-11-03
e2e tests gang scheduling

gflarity:
#### What type of PR is this?

Tests for gang scheduling as per the 'Grove Multi-Node QA Test Plan'.

Implements e2e tests for gang scheduling test scheduling. 

```release-note
NONE
```

```docs

```

### DIFF
--- a/operator/e2e/setup/k8s_clusters.go
+++ b/operator/e2e/setup/k8s_clusters.go
@@ -308,7 +308,7 @@ func SetupCompleteK3DCluster(ctx context.Context, cfg ClusterConfig, skaffoldYAM
 
 	// Nvidia Operator seems to take the longest to be ready, so we wait for it last
 	// to get the most done while waiting. (0 = skip count validation)
-	if err := utils.WaitForPodsInNamespace(ctx, nvidiaConfig.Namespace, nvidiaConfig.RestConfig, 0, defaultPollTimeout, defaultPollInterval, nvidiaConfig.Logger); err != nil {
+	if err := utils.WaitForPodsInNamespace(ctx, gpuOperatorConfig.Namespace, gpuOperatorConfig.RestConfig, 0, defaultPollTimeout, defaultPollInterval, gpuOperatorConfig.Logger); err != nil {
 		cleanup()
 		return nil, nil, fmt.Errorf("NVIDIA GPU Operator not ready: %w", err)
 	}


operator/e2e/tests/setup.go
```diff
@@ -233,3 +236,83 @@ func verifyPodsArePendingWithUnschedulableEvents(ctx context.Context, clientset
...
+	})
+	if err != nil {
+		t.Fatalf("Failed to marshal PCSG patch: %v", err)
+	}
+
+	if _, err := dynamicClient.Resource(pcsgGVR).Namespace(namespace).Patch(ctx, pcsgName, types.MergePatchType, patchBytes, metav1.PatchOptions{}); err != nil {
+		t.Fatalf("Failed to scale PodCliqueScalingGroup %s: %v", pcsgName, err)
+	}
+
+	totalPods, runningPods, pendingPods, err := waitForPodConditions(ctx, clientset, namespace, labelSelector, expectedTotalPods, expectedPending, 5*time.Minute, 5*time.Second)
```
shayasoolin: I'm afraid of long timeout as 5 minutes. IIUC a common scale operation should take seconds to complete. With a too long timeout, in case of some issue, tests will hang for too long before reporting a failure.
Perhaps 1 minute is good enough?

operator/e2e/tests/setup.go
```diff
@@ -233,3 +236,83 @@ func verifyPodsArePendingWithUnschedulableEvents(ctx context.Context, clientset
...
+	})
+	if err != nil {
+		t.Fatalf("Failed to marshal PCS patch: %v", err)
+	}
+
+	if _, err := dynamicClient.Resource(pcsGVR).Namespace(namespace).Patch(ctx, pcsName, types.MergePatchType, patchBytes, metav1.PatchOptions{}); err != nil {
+		t.Fatalf("Failed to scale PodCliqueSet %s: %v", pcsName, err)
+	}
+
+	totalPods, runningPods, pendingPods, err := waitForPodConditions(ctx, clientset, namespace, labelSelector, expectedTotalPods, expectedPending, 1*time.Minute, 1*time.Second)
```
shayasoolin: Here you're using a 1-minute timeout :) 
Consider having those as consts, for easier and global control.

operator/e2e/tests/gang_scheduling_test.go
```diff
@@ -114,3 +120,2846 @@ func Test_GS1_GangSchedulingWithFullReplicas(t *testing.T) {
...
+	}
+
+	logger.Info("4. Uncordon 1 node to allow scheduling and verify pods get scheduled")
+	firstNodeToUncordon := nodesToCordon[0]
+	if err := utils.CordonNode(ctx, clientset, firstNodeToUncordon, false); err != nil {
+		t.Fatalf("Failed to uncordon node %s: %v", firstNodeToUncordon, err)
+	}
+
+	logger.Info("5. Wait for pods to become ready")
+	if err := utils.WaitForPods(ctx, restConfig, []string{workloadNamespace}, workloadLabelSelector, defaultPollTimeout, defaultPollInterval, logger); err != nil {
```
shayasoolin: I'd expect to see the expectedPods variable as a function parameter here. What if some pods got deleted somehow before reaching this check? We cannot rely solely on the label selector.

operator/e2e/tests/gang_scheduling_test.go
```diff
@@ -114,3 +120,2846 @@ func Test_GS1_GangSchedulingWithFullReplicas(t *testing.T) {
...
+
+	_, err = utils.ApplyYAMLFile(ctx, workloadYAMLPath, workloadNamespace, restConfig, logger)
+	if err != nil {
+		t.Fatalf("Failed to apply workload YAML: %v", err)
+	}
+
+	expectedPods := 10
+
+	var pods *v1.PodList
+	err = utils.PollForCondition(ctx, defaultPollTimeout, defaultPollInterval, func() (bool, error) {
```
shayasoolin: This should be a utility function - waiting for number of pods.

operator/e2e/tests/gang_scheduling_test.go
shayasoolin: Same goes for verifyPodsArePendingWithUnschedulableEvents, should accept the expected number of nodes to be pending.

operator/e2e/tests/gang_scheduling_test.go
```diff
@@ -114,3 +120,2846 @@ func Test_GS1_GangSchedulingWithFullReplicas(t *testing.T) {
...
+	pods, err = clientset.CoreV1().Pods(workloadNamespace).List(ctx, metav1.ListOptions{
+		LabelSelector: workloadLabelSelector,
+	})
+	if err != nil {
+		t.Fatalf("Failed to list workload pods: %v", err)
+	}
+
+	runningPods := 0
+	for _, pod := range pods.Items {
+		if pod.Status.Phase == v1.PodRunning {
```
shayasoolin: Something's not clear to me - WaitForPods already verifies that all pods are Ready, and Ready > Running. Why do we need this extra check here?
Another note, regarding WaitForPods, specifically isPodReady - I think it should first check that Phase is PodRunning, and then that the Ready condition is true. Not that I know a case where a pod can be ready but not running, but still..

operator/e2e/tests/gang_scheduling_test.go
```diff
@@ -114,3 +120,2846 @@ func Test_GS1_GangSchedulingWithFullReplicas(t *testing.T) {
...
+			runningPods++
+		}
+	}
+
+	if runningPods != len(pods.Items) {
+		t.Fatalf("Expected all %d pods to be running, but only %d are running", len(pods.Items), runningPods)
+	}
+	assertPodsOnDistinctNodes(t, pods.Items)
+
+	logger.Info("6. Scale PCSG replicas to 3 and verify 4 new pending pods")
```
shayasoolin, line 142: This should be a utility function. Both in order to shorten the test code and make it more concise and clearer, and in order to reuse these functions in other tests.

shayasoolin (commented):
I reviewed the infra code (shared_cluster.go and setup.go) changes, and the first new test in gang_scheduling_test.go.
I stopped there since my main comment is that I strongly suggest to invest time on having as many utility test functions as possible, and then reuse them for the test cases themselves.
We'll not just gain reusability and make it easier to maintain, but the test cases themselves will get much clearer as they will show as high-level function calls that look ideally as the comment above the test case: a list of steps.
Since this requires a non-trivial refactoring, I prefer to review the test logic itself after it's done. Thanks!

operator/e2e/tests/setup.go
gflarity: Agreed.  I'm going to update the functions to get passed in the timeout/interval and then use the defaults instead. Just missed these.

operator/e2e/tests/setup.go
gflarity: Thanks, I'm on it.

gflarity:
Thanks @shayasoolin.  I abstracted as much as I could into utility functions. PTAL.

operator/e2e/tests/setup.go
```diff
@@ -221,3 +239,190 @@ func verifyPodsArePendingWithUnschedulableEvents(ctx context.Context, clientset
...
+
+	return nodesToCordon
+}
+
+// WorkloadConfig defines configuration for deploying and verifying a workload.
+type WorkloadConfig struct {
+	Name          string
+	YAMLPath      string
+	Namespace     string
+	LabelSelector string
```
shayasoolin: I think the LabelSelector field can be calculated from Name ("app.kubernetes.io/part-of=<Name>"), instead of repeating the label and format again and again in the tests themselves.

operator/e2e/tests/gang_scheduling_test.go
```diff
@@ -41,76 +39,1106 @@ func Test_GS1_GangSchedulingWithFullReplicas(t *testing.T) {
...
+		t.Fatalf("Failed to deploy workload: %v", err)
+	}
+
+	logger.Info("3. Verify all workload pods are pending due to insufficient resources")
+	if err := verifyPodsArePendingWithUnschedulableEvents(ctx, clientset, workloadNamespace, workloadConfig.LabelSelector, true, expectedPods, defaultPollTimeout, defaultPollInterval); err != nil {
+		t.Fatalf("Failed to verify all pods have Unschedulable events: %v", err)
+	}
+
+	logger.Info("4. Uncordon the node and verify all pods get scheduled")
+	uncordonNodesAndWaitForPods(t, ctx, clientset, restConfig, []string{workerNodeToCordon}, workloadNamespace, "", expectedPods, defaultPollTimeout, defaultPollInterval)
```
shayasoolin: Better pass nodesToCordon instead of []string{workerNodeToCordon}

operator/e2e/tests/setup.go
```diff
@@ -221,3 +239,190 @@ func verifyPodsArePendingWithUnschedulableEvents(ctx context.Context, clientset
...
+	// Need to use a sleep here unfortunately, see: https://github.com/NVIDIA/grove/issues/226
+	time.Sleep(30 * time.Second)
+	if err := verifyAllPodsArePending(ctx, clientset, namespace, labelSelector, timeout, interval); err != nil {
+		t.Fatalf("Failed to verify all pods are pending: %v", err)
+	}
+}
+
+// uncordonNodesAndWaitForPods uncordons the specified nodes and waits for pods to be ready.
+// This helper combines the common pattern of uncordoning nodes followed by waiting for pods.
+func uncordonNodesAndWaitForPods(t *testing.T, ctx context.Context, clientset kubernetes.Interface, restConfig *rest.Config, nodes []string, namespace, labelSelector string, expectedPods int, timeout, interval time.Duration) {
```
shayasoolin: One more general suggestion about the utility funcs. Many of them have many parameters in common, including:
t *testing.T, ctx context.Context, clientset kubernetes.Interface, restConfig *rest.Config, nodes []string, namespace, timeout, interval time.Duration
I mean, the actual values that are passed are common across different test cases (same client, same default timeouts, etc.)
It can be nice to have a struct that holds all of them, and is initialized once and used by all tests. Will make those functions' signatures shorter.

shayasoolin (approved):

operator/e2e/tests/setup.go
gflarity: Done

operator/e2e/tests/setup.go
gflarity: Done

operator/e2e/tests/setup.go
```diff
@@ -221,3 +295,204 @@ func verifyPodsArePendingWithUnschedulableEvents(ctx context.Context, clientset
...
+	totalPods, runningPods, pendingPods, err := waitForPodConditions(tc, expectedTotalPods, expectedPending)
+	if err != nil {
+		tc.T.Fatalf("Failed to wait for expected pod conditions after PCS scaling: %v. Final state: total=%d, running=%d, pending=%d (expected: total=%d, pending=%d)",
+			err, totalPods, runningPods, pendingPods, expectedTotalPods, expectedPending)
+	}
+}
+
+// cordonNodes cordons or uncordons multiple nodes.
+// This helper reduces repetition of the cordon/uncordon loop pattern found throughout tests.
+func cordonNodes(tc TestContext, nodes []string, cordon bool) {
```
julienmancuso: nit: consider a tiny wrapper uncordonNodes to help readability in the tests and to remove the check introduced to format the error string.

operator/e2e/tests/gang_scheduling_test.go
```diff
@@ -38,79 +36,1255 @@ func Test_GS1_GangSchedulingWithFullReplicas(t *testing.T) {
...
-	// Poll for pod creation and verify they are pending
-	expectedPods := 10 // pc-a: 2 replicas, pc-b: 1*2 (scaling group), pc-c: 3*2 (scaling group) = 2+2+6=10
+	if err := utils.VerifyPodPhases(pods, expectedPods, 4); err != nil {
+		t.Fatalf("Pod phase verification failed: %v", err)
+	}
+
+	logger.Info("7. Uncordon remaining nodes and verify all pods get scheduled")
+	uncordonNodesAndWaitForPods(tc, nodesToCordon[1:], expectedScaledPods)
+
+	// Verify that each pod is scheduled on a unique nodede
```
julienmancuso: nit: unique node

operator/e2e/tests/setup.go
gflarity: I agree, I was thinking the same but after a while it had too much momentum. I want to rename the functions and add cordon/uncordon wrappers.  I'll do that in a fast follow to avoid re-reviews being required.

operator/e2e/tests/gang_scheduling_test.go
gflarity: Fixed

PR #247, opened 2025-11-05
Feat/Topology Configuration Infrastructure

Ronkahn21:
# PR Description: Topology Configuration Infrastructure

## What type of PR is this?

/kind feature
/kind api

## What this PR does / why we need it

This PR implements the foundational infrastructure for topology-aware scheduling in Grove by adding operator-level configuration and validation for topology support.

**Key Changes:**

1. **Operator Configuration**: Adds `TopologyConfiguration` to operator config with:
   - `enabled` field to enable/disable topology features (defaults to `false`)
   - `name` field to specify ClusterTopology resource name (defaults to `"grove-topology"`)
   - Validation that topology name is required when enabled

2. **Startup Validation**: Operator validates topology configuration at startup:
   - When `topology.enabled=true`, verifies ClusterTopology resource exists
   - Exits with clear error if enabled but ClusterTopology not found
   - Prevents operator from running in invalid topology state

3. **RBAC Permissions**: Updates cluster role to grant Grove operator read access to ClusterTopology resources

4. **Documentation**: Adds TopologyConfiguration section to operator API reference

5. **Sample Configuration**: Provides example topology configuration YAML

**Why This Matters:**

Topology-aware scheduling is critical for Grove's multi-node inference workloads because:
- Network locality improves high-bandwidth communication between leaders and workers
- Coordinated placement of related components (model shards) improves performance
- Minimizing network hops between inference components reduces latency

This PR establishes the operator-side foundation that will enable workloads to specify topology constraints in future PRs.

## Special notes for your reviewer

**Implementation Scope:**

This is **Phase 1** of topology-aware scheduling implementation, focusing on operator configuration infrastructure. It includes:
- Operator config schema and defaults
- Validation logic for topology settings
- Startup-time ClusterTopology existence checks
- RBAC permissions

**What's NOT in this PR:**

Future PRs will add:
- ClusterTopology CRD definition and controller
- TopologyConstraint fields in PodCliqueSet/PodClique APIs
- Webhook validation for topology constraint hierarchy
- PodGang topology constraint translation

**Testing Notes:**

- Added unit tests for config defaults and validation
- Tests verify proper error messages and default values
- Startup validation tested via config validation tests

**Migration Path:**

- Default `topology.enabled=false` ensures backward compatibility
- Existing workloads unaffected until topology explicitly enabled
- No breaking changes to existing APIs

## Does this PR introduce an API change?

```release-note
Add topology configuration to Grove operator config, enabling foundational support for topology-aware scheduling. Topology support is disabled by default and requires explicit configuration.
```

## Additional documentation

```docs
See design document: docs/designs/topology.md

Operator configuration reference updated with TopologyConfiguration section showing:
- How to enable topology support
- How to configure ClusterTopology resource name
- Validation requirements and startup behav

### DIFF
--- a/operator/api/core/v1alpha1/crds/embed.go
+++ b/operator/api/core/v1alpha1/crds/embed.go
@@ -37,6 +37,7 @@ func PodCliqueSetCRD() string {
 	return podCliqueSetCRD
 }
 
+// ClusterTopologyCRD returns the ClusterTopology CRD
 func ClusterTopologyCRD() string {
 	return clusterTopologyCRD
 }

### CODE REVIEW

File: operator/charts/templates/clusterrole.yaml [hunk 1]
```diff
@@ -36,6 +36,16 @@ rules:
...
+- apiGroups:
+  - grove.io
+  resources:
+  - clustertopologies
+  - clustertopologies/finalizers
+  verbs:
+  - get
+  - list
+  - watch
+  - update
```
shayasoolin: What is update needed for?

shayasoolin (dismissed):

operator/charts/templates/clusterrole.yaml
Ronkahn21: I was trying to allow only update to finalizers but as I decided to aligned with the current RBAC where there is no separation

operator/api/config/v1alpha1/defaults.go
```diff
@@ -114,3 +127,10 @@ func SetDefaults_PodCliqueScalingGroupControllerConfiguration(obj *PodCliqueScal
 		obj.ConcurrentSyncs = ptr.To(1)
 	}
 }
+
+// SetDefaults_TopologyConfiguration sets defaults for the TopologyConfiguration.
+func SetDefaults_TopologyConfiguration(obj *TopologyConfiguration) {
```
sanjaychatterjee: nit: add a friendly var name like `topoConfig` instead of `obj`.

operator/api/config/v1alpha1/defaults.go
```diff
@@ -27,8 +27,21 @@ const (
 	defaultLeaderElectionResourceLock    = "leases"
 	defaultLeaderElectionResourceName    = "grove-operator-leader-election"
 	defaultWebhookServerTLSServerCertDir = "/etc/grove-operator/webhook-certs"
+	defaultTopologyName                  = "grove-topology"
 )
 
+// SetDefaults sets defaults for the OperatorConfiguration.
+func SetDefaults(config *OperatorConfiguration) {
```
sanjaychatterjee: Why do we need this separately?

operator/api/core/v1alpha1/crds/grove.io_topologydomains.yaml
```diff
@@ -1,95 +0,0 @@
----
```
sanjaychatterjee, line 1: Why are we removing this?

operator/api/config/v1alpha1/defaults.go
Ronkahn21: when loading the configuration its the current scheme default was not affected at all, so I am calling it explicit

operator/api/core/v1alpha1/crds/grove.io_topologydomains.yaml
Ronkahn21: its was the old CRD name, we missed removing it
now we use only the clusterTopology name

operator/api/config/validation/validation_test.go
```diff
@@ -0,0 +1,93 @@
...
+		{
+			name: "invalid: enabled with whitespace-only name",
+			config: configv1alpha1.TopologyConfiguration{
+				Enabled: true,
+				Name:    "   ",
+			},
+			expectError: true,
+			errorField:  "topology.name",
+		},
+	}
```
gflarity, line 77: What about cases with garbage characters? "%$@#"?

operator/cmd/main.go
```diff
@@ -107,3 +118,15 @@ func printFlags() {
...
+
+func validateTopology(ctx context.Context, reader client.Reader, topologyConfig configv1alpha1.TopologyConfiguration) error {
+	if !topologyConfig.Enabled {
+		return nil
+	}
+	var clusterTopology grovecorev1alpha1.ClusterTopology
+	if err := reader.Get(ctx, types.NamespacedName{Name: topologyConfig.Name}, &clusterTopology); err != nil {
+		return fmt.Errorf("failed to fetch ClusterTopology %s: %w", topologyConfig.Name, err)
+	}
+	logger.Info("topology validated successfully", "cluster topology", topologyConfig.Name)
```
gflarity: Is there any other validation that might be necessary besides just existing? Is it possible for a topology with 0 levels to exist?  Just checking.

operator/api/config/validation/validation.go
```diff
@@ -111,3 +116,11 @@ func mustBeGreaterThanZeroDuration(duration metav1.Duration, fldPath *field.Path
 	}
 	return allErrs
 }
+
+func validateTopologyConfiguration(topologyCfg configv1alpha1.TopologyConfiguration, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if topologyCfg.Enabled && len(strings.TrimSpace(topologyCfg.Name)) == 0 {
```
gflarity: Seems like this validation should probably check it's a valid kubernetes resource name? There's got to be a function for that that's easy to import and use? A quick serach points to "k8s.io/apimachinery/pkg/util/validation".

gflarity (changes requested):
Mostly just questions, but I think the validation one is work looking into.

operator/cmd/main.go
gflarity: I see we're covering the case of 0 levels in the webhook, seems like the right place. https://github.com/ai-dynamo/grove/pull/251/files

operator/api/config/validation/validation.go
gflarity: Oh, looks like you can accomplish this with: 

```
	allErrs = append(allErrs, apivalidation.ValidateObjectMeta(&v.ct.ObjectMeta, false,
		apivalidation.NameIsDNSSubdomain, field.NewPath("metadata"))...)
```
(from the webhook draft).

operator/api/config/validation/validation.go
```diff
@@ -111,3 +116,11 @@ func mustBeGreaterThanZeroDuration(duration metav1.Duration, fldPath *field.Path
 	}
 	return allErrs
 }
+
```
gflarity, line 115: nit, I'm a fan of documenting every function as a habit, even it's pretty clear form the name and package scoped.

operator/api/config/validation/validation.go
gflarity: I see that we check for existance in main, but might as well catch it sooner and more specifically than checking for a resource with a  bad name.

operator/api/config/v1alpha1/defaults.go
sanjaychatterjee: I see but trying to understand why `SetObjectDefaults_OperatorConfiguration` [here](https://github.com/ai-dynamo/grove/blob/main/operator/api/config/v1alpha1/zz_generated.defaults.go#L36) was not invoked?
@unmarshall can you please also review this function?

Ronkahn21: from what I see it, i think its assume the config is crd when you get it with the client, but we use config map with the config embedded with in

operator/api/config/v1alpha1/types.go
```diff
@@ -187,3 +188,13 @@ type AuthorizerConfig struct {
 	// +optional
 	ExemptServiceAccountUserNames []string `json:"exemptServiceAccountUserNames,omitempty"`
 }
+
+// TopologyConfiguration defines the configuration for topology-aware scheduling.
+type TopologyConfiguration struct {
```
unmarshall: Can we rename this to `ClusterTopologyConfiguration`? This will bring it in line with the CRD resource which is called `ClusterTopology`

operator/api/config/validation/validation.go
```diff
@@ -23,17 +23,23 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/utils/ptr"
 )
 
+const (
+	errTopologyNameRequired = "topology name is required"
```
unmarshall: it is only used in one place. Can you remove this constant as it adds no value?

operator/api/config/validation/validation.go
```diff
@@ -111,3 +117,21 @@ func mustBeGreaterThanZeroDuration(duration metav1.Duration, fldPath *field.Path
 	}
 	return allErrs
 }
+
+// validateTopologyConfiguration validates the topology configuration.
+// When topology is enabled, it ensures the topology name is provided and is a valid DNS subdomain name
+// that can be used as a Kubernetes resource name.
+func validateTopologyConfiguration(topologyCfg configv1alpha1.TopologyConfiguration, fldPath *field.Path) field.ErrorList {
```
unmarshall:
```suggestion
func validateClusterTopologyConfiguration(clusterTopologyCfg configv1alpha1.ClusterTopologyConfiguration, fldPath *field.Path) field.ErrorList {
```

operator/api/config/v1alpha1/types.go
```diff
@@ -62,6 +62,7 @@ type OperatorConfiguration struct {
 	LogLevel         LogLevel                `json:"logLevel"`
 	LogFormat        LogFormat               `json:"logFormat"`
 	Authorizer       AuthorizerConfig        `json:"authorizer"`
+	Topology         TopologyConfiguration   `json:"topology"`
```
unmarshall: Rename this to `ClusterTopology` and adjust the JSON tag as well.

operator/api/config/validation/validation.go
```diff
@@ -111,3 +117,21 @@ func mustBeGreaterThanZeroDuration(duration metav1.Duration, fldPath *field.Path
...
+	allErrs := field.ErrorList{}
```
unmarshall, line 119: There is no need to check if the clusterTopologyConfig.Enabled is true twice.
Instead do this:
```go
	allErrs := field.ErrorList{}
    if !clusterTopologyCfg.Enabled {
      return allErrs
    }
    // now do the rest of the validations

File: operator/api/config/validation/validation.go [hunk 14]
```diff
@@ -111,3 +117,21 @@ func mustBeGreaterThanZeroDuration(duration metav1.Duration, fldPath *field.Path
...
+	if topologyCfg.Enabled && len(strings.TrimSpace(topologyCfg.Name)) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("name"), errTopologyNameRequired))
+		return allErrs
+	}
+
+	if topologyCfg.Enabled {
+		if errs := validation.IsDNS1123Subdomain(topologyCfg.Name); len(errs) > 0 {
```
unmarshall: Do you really need this check?
When a cluster administrator creates a `ClusterTopology` CR then a `Name` must be specified as part of `metav1.ObjectMeta` and if you have a name which fails the `DNS1123Subdomain` check of then you will never be able to create the CR instance itself.

To prove that I changed the name in the sample for ClusterTopology to `Grove-topology` and applied the yaml. The creation failed with the following error:
```
The ClusterTopology "Grove-topology" is invalid: metadata.name: Invalid value: "Grove-topology": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

Therefore there is no chance that the reference name of ClusterTopology CR will not adhere to DNS1123Subdomain check.

operator/api/config/validation/validation_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"strings"
+	"testing"
+
+	configv1alpha1 "github.com/ai-dynamo/grove/operator/api/config/v1alpha1"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func TestValidateTopologyConfiguration(t *testing.T) {
```
unmarshall: You can remove all tests that validate `validation.IsDNS1123Subdomain` as these are not required.
All you need to check for is that if it is enabled then a non-empty `name` must be specified.

operator/charts/templates/clusterrole.yaml
```diff
@@ -27,6 +27,7 @@ rules:
   - podcliques/status
   - podcliquescalinggroups
   - podcliquescalinggroups/status
+  - clustertopologies
```
unmarshall: why should `Grove` operator have `create, delete, deletecollection, patch and update` roles for `clustertopologies`?

operator/charts/values.yaml
```diff
@@ -60,6 +60,9 @@ config:
       concurrentSyncs: 3
   logLevel: info
   logFormat: json
+  topology:
```
unmarshall: rename this to `clusterTopology`

operator/cmd/main.go
```diff
@@ -61,6 +66,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx := ctrl.SetupSignalHandler()
+
+	if err = validateTopology(ctx, mgr.GetAPIReader(), operatorCfg.Topology); err != nil {
```
unmarshall: As part of the design in `docs/designs/topology.md` the following is mentioned:
```
If ClusterTopology with that name doesn't exist → operator exits with ENOENT and crashloops
```
This is not correct, consider the following situation:

- There are a lot of workloads that do not use `pack constraints` feature. These workloads are deployed and are currently running and are being actively managed by the operator.
- Now the admin decides that ClusterTopology should be created and the new feature offered by Grove should be used. However the admin makes a mistake and does not create the ClusterTopology CR but enables the feature.

Consequence:
- Upon restart the, operator continuously crash loops. For this time all existing workloads that do not use `pack constraints` are no longer actively managed. This will have severe consequences if the admin does not notice the crash looping operator in time.

Ideally we should always start the operator but reject reconciliation of PCS that define pack constraints. For all other PCS which do not define any pack constraints it continues to reconcile those.

operator/cmd/opts/options.go
```diff
@@ -64,6 +64,8 @@ func (o *CLIOptions) Complete() error {
 	if err = runtime.DecodeInto(configDecoder, data, o.Config); err != nil {
 		return fmt.Errorf("error decoding config: %w", err)
 	}
+	// set defaults
+	configv1alpha1.SetDefaults(o.Config)
```
unmarshall: why is this required?
If you go to `operator/api/config/v1alpha1/register.go` you will see that the defaulting function has already been registered and should be invoked. There is no point in redefining `SetObjectDefaults_OperatorConfiguration` again as `SetDefaults`

operator/samples/clustertopology/cluster-topology-host-only.yaml
```diff
@@ -0,0 +1,20 @@
+# ClusterTopology with minimal single-level hierarchy
```
unmarshall, line 1: rename path to `operator/samples/clustertopology`
Can you have just one sample which already has all the supported topology levels defined?

operator/api/config/validation/validation.go
Ronkahn21: I agree, I will remove this

operator/cmd/main.go
Ronkahn21: We said it's a major error, I agree it as severe consequences that the operator is down but it should be for a short time, and the admin role is to fix this, rather of us handle this, you need to remember that we add automatic configuration for all workload when its enabled so now we need to advertise this is not happing and more.

for now I feel its best to keep the logic as it, and document it very well

operator/samples/clustertopology/cluster-topology-host-only.yaml
Ronkahn21: I will change the name, and we can add later a sample, but for now for testing I used this, for dev env this good as you might not have any other topology level

operator/cmd/main.go
shayasoolin: I must say I agree with @Ronkahn21 here. We're doing the simpler thing this way, both on the implementation and usage levels.

sanjaychatterjee: I too agree with @Ronkahn21 that burden of correct cluster topology config should be on the admin and not on the operator.

operator/cmd/opts/options.go
sanjaychatterjee: @unmarshall can we verify this issue raised by @Ronkahn21 ? If required, we need to make a subsequent PR to fix and get the workaround in for now.

operator/charts/templates/clusterrole.yaml
sanjaychatterjee: Agreed not required for now.

operator/cmd/main.go
unmarshall: I disagree, but since i am the minority view, i would therefore not push further.

operator/cmd/opts/options.go
unmarshall: I just tested it and the defaulting happens properly.

Ronkahn21: You were correct, I had issue from my end , I tested it again and saw it invoked, I deleted the function

unmarshall (approved):
looks good

PR #251, opened 2025-11-06
Add validation webhook for ClusterTopology resource 

shmuel-runai:
#### What type of PR is this?

/kind feature
/kind api

Implements validation webhook for ClusterTopology 

```release-note
NONE
```

```docs
Design ref: docs/designs/topology.md
```

### DIFF
--- a/operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
+++ b/operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/samber/lo"
 	admissionv1 "k8s.io/api/admission/v1"
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	metav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
--- a/operator/internal/webhook/admission/clustertopology/validation/clustertopology_crd_test.go
+++ b/operator/internal/webhook/admission/clustertopology/validation/clustertopology_crd_test.go
@@ -34,6 +34,185 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 )
 
+func TestClusterTopologyCRDValidation(t *testing.T) {
+	testEnv, k8sClient := setupEnvTest(t)
+	defer teardownEnvTest(t, testEnv)
+
+	tests := []struct {
+		name          string
+		ct            *grovecorev1alpha1.ClusterTopology
+		expectError   bool
+		errorContains string
+	}{
+		{
+			name: "valid cluster topology with single level",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-valid-single"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{
+							Domain: grovecorev1alpha1.TopologyDomainRegion,
+							Key:    "topology.kubernetes.io/region",
+						},
+					},
+				},
+			},
+			expectError: false,
+		},
+		{
+			name: "valid cluster topology with multiple levels",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-valid-multiple"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{
+							Domain: grovecorev1alpha1.TopologyDomainRegion,
+							Key:    "topology.kubernetes.io/region",
+						},
+						{
+							Domain: grovecorev1alpha1.TopologyDomainZone,
+							Key:    "topology.kubernetes.io/zone",
+						},
+						{
+							Domain: grovecorev1alpha1.TopologyDomainHost,
+							Key:    "kubernetes.io/hostname",
+						},
+					},
+				},
+			},
+			expectError: false,
+		},
+		{
+			name: "reject invalid domain",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-invalid-domain"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{
+							Domain: "invalid-domain", // Not in enum
+							Key:    "test.io/key",
+						},
+					},
+				},
+			},
+			expectError:   true,
+			errorContains: "supported values",
+		},
+		{
+			name: "reject empty key",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-empty-key"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{
+							Domain: grovecorev1alpha1.TopologyDomainRegion,
+							Key:    "", // Empty key
+						},
+					},
+				},
+			},
+			expectError:   true,
+			errorContains: "spec.levels[0].key",
+		},
+		{
+			name: "reject key too long",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-key-too-long"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{
+							Domain: grovecorev1alpha1.TopologyDomainRegion,
+							// Create a key longer than 63 characters (MaxLength validation)
+							Key: "this-is-a-very-long-key-that-exceeds-the-maximum-allowed-length-of-63-characters",
+						},
+					},
+				},
+			},
+			expectError:   true,
+			errorContains: "spec.levels[0].key",
+		},
+		{
+			name: "reject empty levels array",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-empty-levels"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{}, // Empty array
+				},
+			},
+			expectError:   true,
+			errorContains: "spec.levels",
+		},
+		{
+			name: "reject too many levels",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-too-many-levels"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{Domain: grovecorev1alpha1.TopologyDomainRegion, Key: "key1"},
+						{Domain: grovecorev1alpha1.TopologyDomainZone, Key: "key2"},
+						{Domain: grovecorev1alpha1.TopologyDomainDataCenter, Key: "key3"},
+						{Domain: grovecorev1alpha1.TopologyDomainBlock, Key: "key4"},
+						{Domain: grovecorev1alpha1.TopologyDomainRack, Key: "key5"},
+						{Domain: grovecorev1alpha1.TopologyDomainHost, Key: "key6"},
+						{Domain: grovecorev1alpha1.TopologyDomainNuma, Key: "key7"},
+						{Domain: "region", Key: "key8"}, // 8th level - exceeds max
+					},
+				},
+			},
+			expectError:   true,
+			errorContains: "spec.levels",
+		},
+		{
+			name: "reject missing domain",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-missing-domain"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{
+							Key: "test.io/key", // Missing domain
+						},
+					},
+				},
+			},
+			expectError: true,
+			// API server should reject due to Required validation
+		},
+		{
+			name: "reject missing key",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-missing-key"},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
+					Levels: []grovecorev1alpha1.TopologyLevel{
+						{
+							Domain: grovecorev1alpha1.TopologyDomainRegion,
+							// Missing key
+						},
+					},
+				},
+			},
+			expectError: true,
+			// API server should reject due to Required validation
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := k8sClient.Create(context.Background(), tt.ct)
+
+			if tt.expectError {
+				assert.Error(t, err)
+				if tt.errorContains != "" {
+					assert.Contains(t, err.Error(), tt.errorContains)
+				}
+			} else {
+				assert.NoError(t, err)
+				// Clean up successfully created resource
+				_ = k8sClient.Delete(context.Background(), tt.ct)
+			}
+		})
+	}
+}
+
 // setupEnvTest creates an envtest environment with ClusterTopology CRDs loaded
 func setupEnvTest(t *testing.T) (*envtest.Environment, client.Client) {
 	// Check if KUBEBUILDER_ASSETS is set
@@ -70,136 +249,4 @@ func setupEnvTest(t *testing.T) (*envtest.Environment, client.Client) {
 func teardownEnvTest(t *testing.T, testEnv *envtest.Environment) {
 	err := testEnv.Stop()
 	require.NoError(t, err)
-}
-
-func TestClusterTopologyCRDValidation(t *testing.T) {
-	testEnv, k8sClient := setupEnvTest(t)
-	defer teardownEnvTest(t, testEnv)
-
-	t.Run("reject invalid domain", func(t *testing.T) {
-		ct := &grovecorev1alpha1.ClusterTopology{
-			ObjectMeta: metav1.ObjectMeta{Name: "test-invalid-domain"},
-			Spec: grovecorev1alpha1.ClusterTopologySpec{
-				Levels: []grovecorev1alpha1.TopologyLevel{
-					{
-						Domain: "invalid-domain", // Not in enum
-						Key:    "test.io/key",
-					},
-				},
-			},
-		}
-
-		err := k8sClient.Create(context.Background(), ct)
-		assert.Error(t, err)
-		assert.Contains(t, err.Error(), "supported values")
-	})
-
-	t.Run("reject empty key", func(t *testing.T) {
-		ct := &grovecorev1alpha1.ClusterTopology{
-			ObjectMeta: metav1.ObjectMeta{Name: "test-empty-key"},
-			Spec: grovecorev1alpha1.ClusterTopologySpec{
-				Levels: []grovecorev1alpha1.TopologyLevel{
-					{
-						Domain: grovecorev1alpha1.TopologyDomainRegion,
-						Key:    "", // Empty key
-					},
-				},
-			},
-		}
-
-		err := k8sClient.Create(context.Background(), ct)
-		assert.Error(t, err)
-		assert.Contains(t, err.Error(), "spec.levels[0].key")
-	})
-
-	t.Run("reject key too long", func(t *testing.T) {
-		// Create a key longer than 63 characters (MaxLength validation)
-		longKey := "this-is-a-very-long-key-that-exceeds-the-maximum-allowed-length-of-63-characters"
-		ct := &grovecorev1alpha1.ClusterTopology{
-			ObjectMeta: metav1.ObjectMeta{Name: "test-key-too-long"},
-			Spec: grovecorev1alpha1.ClusterTopologySpec{
-				Levels: []grovecorev1alpha1.TopologyLevel{
-					{
-						Domain: grovecorev1alpha1.TopologyDomainRegion,
-						Key:    longKey,
-					},
-				},
-			},
-		}
-
-		err := k8sClient.Create(context.Background(), ct)
-		assert.Error(t, err)
-		assert.Contains(t, err.Error(), "spec.levels[0].key")
-	})
-
-	t.Run("reject empty levels array", func(t *testing.T) {
-		ct := &grovecorev1alpha1.ClusterTopology{
-			ObjectMeta: metav1.ObjectMeta{Name: "test-empty-levels"},
-			Spec: grovecorev1alpha1.ClusterTopologySpec{
-				Levels: []grovecorev1alpha1.TopologyLevel{}, // Empty array
-			},
-		}
-
-		err := k8sClient.Create(context.Background(), ct)
-		assert.Error(t, err)
-		assert.Contains(t, err.Error(), "spec.levels")
-	})
-
-	t.Run("reject too many levels", func(t *testing.T) {
-		// Create 8 levels (exceeds MaxItems=7)
-		ct := &grovecorev1alpha1.ClusterTopology{
-			ObjectMeta: metav1.ObjectMeta{Name: "test-too-many-levels"},
-			Spec: grovecorev1alpha1.ClusterTopologySpec{
-				Levels: []grovecorev1alpha1.TopologyLevel{
-					{Domain: grovecorev1alpha1.TopologyDomainRegion, Key: "key1"},
-					{Domain: grovecorev1alpha1.TopologyDomainZone, Key: "key2"},
-					{Domain: grovecorev1alpha1.TopologyDomainDataCenter, Key: "key3"},
-					{Domain: grovecorev1alpha1.TopologyDomainBlock, Key: "key4"},
-					{Domain: grovecorev1alpha1.TopologyDomainRack, Key: "key5"},
-					{Domain: grovecorev1alpha1.TopologyDomainHost, Key: "key6"},
-					{Domain: grovecorev1alpha1.TopologyDomainNuma, Key: "key7"},
-					{Domain: "region", Key: "key8"}, // 8th level - exceeds max
-				},
-			},
-		}
-
-		err := k8sClient.Create(context.Background(), ct)
-		assert.Error(t, err)
-		assert.Contains(t, err.Error(), "spec.levels")
-	})
-
-	t.Run("reject missing domain", func(t *testing.T) {
-		ct := &grovecorev1alpha1.ClusterTopology{
-			ObjectMeta: metav1.ObjectMeta{Name: "test-missing-domain"},
-			Spec: grovecorev1alpha1.ClusterTopologySpec{
-				Levels: []grovecorev1alpha1.TopologyLevel{
-					{
-						Key: "test.io/key", // Missing domain
-					},
-				},
-			},
-		}
-
-		err := k8sClient.Create(context.Background(), ct)
-		assert.Error(t, err)
-		// API server should reject due to Required validation
-	})
-
-	t.Run("reject missing key", func(t *testing.T) {
-		ct := &grovecorev1alpha1.ClusterTopology{
-			ObjectMeta: metav1.ObjectMeta{Name: "test-missing-key"},
-			Spec: grovecorev1alpha1.ClusterTopologySpec{
-				Levels: []grovecorev1alpha1.TopologyLevel{
-					{
-						Domain: grovecorev1alpha1.TopologyDomainRegion,
-						// Missing key
-					},
-				},
-			},
-		}
-
-		err := k8sClient.Create(context.Background(), ct)
-		assert.Error(t, err)
-		// API server should reject due to Required validation
-	})
 }
--- a/operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
+++ b/operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
@@ -321,7 +321,7 @@ func TestValidateUpdate(t *testing.T) {
 				},
 			},
 			expectedErr:    true,
-			expectedErrMsg: "domain is immutable",
+			expectedErrMsg: "field is immutable",
 		},
 		{
 			name: "invalid - level added",
@@ -485,7 +485,7 @@ func TestValidateUpdate(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			_, err := handler.ValidateUpdate(ctx, tt.newClusterTopology, tt.oldClusterTopology)
+			_, err := handler.ValidateUpdate(ctx, tt.oldClusterTopology, tt.newClusterTopology)
 
 			if tt.expectedErr {
 				assert.Error(t, err)
--- a/operator/internal/controller/cert/cert.go
+++ b/operator/internal/controller/cert/cert.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/ai-dynamo/grove/operator/internal/constants"
+	clustertopologyvalidation "github.com/ai-dynamo/grove/operator/internal/webhook/admission/clustertopology/validation"
 	authorizationwebhook "github.com/ai-dynamo/grove/operator/internal/webhook/admission/pcs/authorization"
 	defaultingwebhook "github.com/ai-dynamo/grove/operator/internal/webhook/admission/pcs/defaulting"
 	validatingwebhook "github.com/ai-dynamo/grove/operator/internal/webhook/admission/pcs/validation"
@@ -86,6 +87,10 @@ func getWebhooks(authorizerEnabled bool) []cert.WebhookInfo {
 		{
 			Type: cert.Validating,
 			Name: validatingwebhook.Name,
+		},
+		{
+			Type: cert.Validating,
+			Name: clustertopologyvalidation.Name,
 		},
 	}
 	if authorizerEnabled {


operator/charts/templates/clustertopology-validating-webhook-config.yaml
```diff
@@ -0,0 +1,35 @@
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: clustertopology-validating-webhook
+  namespace: {{ .Release.Namespace }}
```
gflarity: `ValidatingWebhookConfiguration` is a cluster-scoped resource and Kubernetes will reject the manifest if `metadata.namespace` is populated. Helm will render this with the release namespace, so `kubectl apply` will fail.

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,195 @@
...
+		allErrs = append(allErrs, errs...)
+	}
+
+	return warnings, allErrs.ToAggregate()
+}
+
+// validateClusterTopologySpec validates the specification of a ClusterTopology object.
+func (v *ctValidator) validateClusterTopologySpec(fldPath *field.Path) ([]string, field.ErrorList) {
+	allErrs := field.ErrorList{}
+	var warnings []string
```
gflarity: never used?

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,195 @@
...
+
+// validateTopologyLevel validates a single topology level in isolation.
+func (v *ctValidator) validateTopologyLevel(level grovecorev1alpha1.TopologyLevel, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate domain is one of the allowed values
+	domainPath := fldPath.Child("domain")
+	if _, exists := topologyDomainOrder[level.Domain]; !exists {
+		// Build list of allowed domains from the map keys
+		allowedDomains := make([]string, 0, len(topologyDomainOrder))
```
gflarity: Just move this into a const, or build exactly on in an init... No need to make it on the fly.

gflarity: Also could use .Keys() here if you really want it on the fly.

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,195 @@
...
+		for domain := range topologyDomainOrder {
+			allowedDomains = append(allowedDomains, string(domain))
+		}
+		allErrs = append(allErrs, field.NotSupported(domainPath, level.Domain, allowedDomains))
+	}
+
+	// Validate key is a valid Kubernetes label key
+	keyPath := fldPath.Child("key")
+	if level.Key == "" {
+		allErrs = append(allErrs, field.Required(keyPath, "key is required"))
```
gflarity: Note since we have +kubebuilder:validation:Required on the Key in the go struct, the API service reject this. We also have a min length check in there. 

```// TopologyLevel defines a single level in the topology hierarchy.
type TopologyLevel struct {
	// Domain is the predefined level identifier used in TopologyConstraint references.
	// Must be one of: region, zone, datacenter, block, rack, host, numa
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Enum=region;zone;datacenter;block;rack;host;numa
	Domain TopologyDomain `json:"domain"`

	// Key is the node label key that identifies this topology domain.
	// Must be a valid Kubernetes label key (qualified name).
	// Examples: "topology.kubernetes.io/zone", "kubernetes.io/hostname"
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=63
	Key string `json:"key"`
}
```
So this is isn't necessary.  Doesn't really hurt but just FYI.

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,195 @@
...
+
+		prevOrder := topologyDomainOrder[prevDomain]
+		currOrder := topologyDomainOrder[currDomain]
+
+		// Current level must have a higher order (narrower scope) than previous level
+		if currOrder <= prevOrder {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("domain"), currDomain,
+				fmt.Sprintf("topology levels must be in hierarchical order (broadest to narrowest). Domain '%s' at position %d cannot come after '%s' at position %d",
+					currDomain, i, prevDomain, i-1)))
+		}
```
gflarity: Tweak, you can distinguish between order and dupes this way. 
```suggestion
		// Current level must have a higher order (narrower scope) than previous level
		if currOrder < prevOrder {
			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("domain"), currDomain,
				fmt.Sprintf("topology levels must be in hierarchical order (broadest to narrowest). Domain '%s' at position %d cannot come after '%s' at position %d",
					currDomain, i, prevDomain, i-1)))
		} else if currOrder == prevOrder {
			allErrs = append(allErrs, field.Duplicate(fldPath.Index(i).Child("domain"), currDomain))
		}
```

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,195 @@
...
+		newLevel := newSpec.Levels[i]
+		oldLevel := oldSpec.Levels[i]
+
+		// Validate that domain is immutable
+		if newLevel.Domain != oldLevel.Domain {
+			allErrs = append(allErrs, field.Invalid(levelPath.Child("domain"), newLevel.Domain,
+				fmt.Sprintf("domain is immutable, cannot change from '%s' to '%s'", oldLevel.Domain, newLevel.Domain)))
+		}
+
+		// Note: Key is allowed to change (not in the requirements), but we validate it's still a valid label key
```
gflarity:
```suggestion
		// Note: Key is allowed to change (not in the requirements), but validation has already occurred in validate()
```

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
```diff
@@ -0,0 +1,541 @@
...
+			expectedErr:    true,
+			expectedErrMsg: "at least one topology level must be defined",
+		},
+		{
+			name: "invalid - duplicate domain (caught by order validation)",
+			ct: &grovecorev1alpha1.ClusterTopology{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-topology",
+				},
+				Spec: grovecorev1alpha1.ClusterTopologySpec{
```
gflarity, line 85: I'd add another test case that is dupes + plus before and after levels that are correct.

gflarity (changes requested):
Left a few comments.  Also, the design calls that both domains and keys must be unique but I didn't notice a check for the Keys being unique.

operator/charts/templates/clustertopology-validating-webhook-config.yaml
shmuel-runai: You are correct, this should also apply to other `ValidatingWebhookConfiguration`, in this PR won't have the `metadata.namespace`

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: Correct and removed, I saw that a minute after sending the draft PR

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: Thanks for the feedback. I've decided to keep the current approach for the following reasons:
Maintainability: Using a constant for the allowed domain list would make code maintenance difficult, as any updates to the list would also require updating the constant.
Memory Allocation: Creating the list during initialization would allocate memory unnecessarily, as it's only used in rare error events (less than 0.01% of the time).
Therefore, I prefer to generate the list on the fly only when an error occurs.

Regarding Keys(), using a simple range is considered the idiomatic way to code this.

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: cool, taking it

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: cool, taking it

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
shmuel-runai: done

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+		// Use Kubernetes label name validation
+		allErrs = append(allErrs, metav1validation.ValidateLabelName(level.Key, keyPath)...)
+	}
+
+	return allErrs
+}
+
+// validateTopologyLevelOrder validates that topology levels are in the correct hierarchical order
+// (broadest to narrowest: Region > Zone > DataCenter > Block > Rack > Host > Numa).
+// This function assumes all domains have already been validated to be valid topology domains.
```
danbar2, line 104: you also assume that the levels are ordered by domain, which is fine since its the mutation webhook job to do that - but please add it to the comments

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+
+// topologyDomainOrder defines the hierarchical order of topology domains from broadest to narrowest.
+// Lower index = broader scope (e.g., Region is broader than Zone).
+var topologyDomainOrder = map[grovecorev1alpha1.TopologyDomain]int{
+	grovecorev1alpha1.TopologyDomainRegion:     0,
+	grovecorev1alpha1.TopologyDomainZone:       1,
+	grovecorev1alpha1.TopologyDomainDataCenter: 2,
+	grovecorev1alpha1.TopologyDomainBlock:      3,
+	grovecorev1alpha1.TopologyDomainRack:       4,
+	grovecorev1alpha1.TopologyDomainHost:       5,
```
Ronkahn21: I would like a better to check this, a method for the stuct itself
level.Broader, .. etc

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// ctValidator validates ClusterTopology resources for create and update operations.
+type ctValidator struct {
+	operation admissionv1.Operation
+	ct        *grovecorev1alpha1.ClusterTopology
+}
+
+// newCTValidator creates a new ClusterTopology validator for the given operation.
+func newCTValidator(ct *grovecorev1alpha1.ClusterTopology, operation admissionv1.Operation) *ctValidator {
```
Ronkahn21: use full name, CT is not a good shortcut, ClusterTopology is better

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+	return &ctValidator{
+		operation: operation,
+		ct:        ct,
+	}
+}
+
+// ---------------------------- validate create of ClusterTopology -----------------------------------------------
+
+// validate validates the ClusterTopology object.
+func (v *ctValidator) validate() error {
```
Ronkahn21: should be exported

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+	allErrs := field.ErrorList{}
+
+	allErrs = append(allErrs, apivalidation.ValidateObjectMeta(&v.ct.ObjectMeta, false,
+		apivalidation.NameIsDNSSubdomain, field.NewPath("metadata"))...)
+	fldPath := field.NewPath("spec")
```
Ronkahn21: could be inline in line 65

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+		return allErrs
+	}
+
+	// Validate max items (should be enforced by kubebuilder validation, but check here too)
+	if len(v.ct.Spec.Levels) > len(topologyDomainOrder) {
+		allErrs = append(allErrs, field.TooMany(levelsPath, len(v.ct.Spec.Levels), len(topologyDomainOrder)))
+	}
+
+	// First, validate each level individually (domain validity, key format, etc.)
+	for i, level := range v.ct.Spec.Levels {
```
Ronkahn21: use lo package there ForEach

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+	return allErrs
+}
+
+// validateTopologyLevel validates a single topology level in isolation.
+func (v *ctValidator) validateTopologyLevel(level grovecorev1alpha1.TopologyLevel, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate domain is one of the allowed values
+	domainPath := fldPath.Child("domain")
+	if _, exists := topologyDomainOrder[level.Domain]; !exists {
```
Ronkahn21: the current  crd  definiation already have validation for name with enum  so does not required to cehck

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Validate that at least one level is defined
+	if len(v.ct.Spec.Levels) == 0 {
+		allErrs = append(allErrs, field.Required(levelsPath, "at least one topology level must be defined"))
+		return allErrs
+	}
+
+	// Validate max items (should be enforced by kubebuilder validation, but check here too)
+	if len(v.ct.Spec.Levels) > len(topologyDomainOrder) {
+		allErrs = append(allErrs, field.TooMany(levelsPath, len(v.ct.Spec.Levels), len(topologyDomainOrder)))
```
Ronkahn21: this validation is part of the CRD and validated by the api server

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+// validateTopologyLevel validates a single topology level in isolation.
+func (v *ctValidator) validateTopologyLevel(level grovecorev1alpha1.TopologyLevel, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate domain is one of the allowed values
+	domainPath := fldPath.Child("domain")
+	if _, exists := topologyDomainOrder[level.Domain]; !exists {
+		// Build list of allowed domains from the map keys
+		allowedDomains := make([]string, 0, len(topologyDomainOrder))
+		for domain := range topologyDomainOrder {
```
Ronkahn21: there is map to list funtionlity from lo packge, but also you dont need it if it handle by the api server validation

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+			allowedDomains = append(allowedDomains, string(domain))
+		}
+		allErrs = append(allErrs, field.NotSupported(domainPath, level.Domain, allowedDomains))
+	}
+
+	// Validate key is a valid Kubernetes label key
+	keyPath := fldPath.Child("key")
```
Ronkahn21, line 96: this is also validated by the api server using the crd validation markdown

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+		prevOrder := topologyDomainOrder[prevDomain]
+		currOrder := topologyDomainOrder[currDomain]
+
+		// Current level must have a higher order (narrower scope) than previous level
+		if currOrder < prevOrder {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("domain"), currDomain,
+				fmt.Sprintf("topology levels must be in hierarchical order (broadest to narrowest). Domain '%s' at position %d cannot come after '%s' at position %d",
+					currDomain, i, prevDomain, i-1)))
+		} else if currOrder == prevOrder {
+			allErrs = append(allErrs, field.Duplicate(fldPath.Index(i).Child("domain"), currDomain))
```
Ronkahn21: this could be simplify use lo package

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+		if firstIndex, exists := seenKeys[level.Key]; exists {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("key"), level.Key,
+				fmt.Sprintf("duplicate key: already used at levels[%d]", firstIndex)))
+		} else {
+			seenKeys[level.Key] = i
+		}
+	}
+
+	return allErrs
+}
```
Ronkahn21, line 141: also this could be simpilify just trasfer to map using lo packge and then check the size

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+	levelsPath := fldPath.Child("levels")
+
+	// Validate that the number of levels hasn't changed
+	if len(newSpec.Levels) != len(oldSpec.Levels) {
+		allErrs = append(allErrs, field.Forbidden(levelsPath, "not allowed to add or remove topology levels"))
+		return allErrs
+	}
+
+	// Validate that the order and domains of levels haven't changed
+	for i := range newSpec.Levels {
```
Ronkahn21: this need to be in new function

operator/internal/webhook/admission/clustertopology/validation/handler.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	oldCT, err := castToClusterTopology(oldObj)
+	if err != nil {
+		return nil, errors.WrapError(err, ErrValidateUpdateClusterTopology, string(admissionv1.Update), "failed to cast old object to ClusterTopology")
+	}
+	validator := newCTValidator(newCT, admissionv1.Update)
+	if err := validator.validate(); err != nil {
+		return nil, err
+	}
+	return nil, validator.validateUpdate(oldCT)
```
Ronkahn21: I dont see the point to sperate it, it should just calll validateUpdate or just validate and pass the operataion as arg

Ronkahn21 (changes requested):

Ronkahn21 (commented):
make sure to use lo packge, flows can be simplified but the overall idea looks good

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: mmm, I don't assume that, this function validates that the levels are in order and there isn't duplication. despite the fact that the will be a mutation webhook the ensure that.
I can add this to the comment (with the mutation webhook)

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: renmaed here and everywhere

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: why? Those methods are called internally by the cluster topology validation handler.
The handler has exported ValidateCreate, ValidateUpdate and ValidateDelete

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+	admissionv1 "k8s.io/api/admission/v1"
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	metav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// topologyDomainOrder defines the hierarchical order of topology domains from broadest to narrowest.
+// Lower index = broader scope (e.g., Region is broader than Zone).
+var topologyDomainOrder = map[grovecorev1alpha1.TopologyDomain]int{
+	grovecorev1alpha1.TopologyDomainRegion:     0,
```
shayasoolin: I think this list should be in a separate file, along with the level names (strings). A separate file that doesn't involve the webhook code, but maybe only some struct-related validators, if needed.

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+	errs := v.validateClusterTopologySpec(fldPath)
+	if len(errs) != 0 {
+		allErrs = append(allErrs, errs...)
+	}
+
+	return allErrs.ToAggregate()
+}
+
+// validateClusterTopologySpec validates the specification of a ClusterTopology object.
+func (v *ctValidator) validateClusterTopologySpec(fldPath *field.Path) field.ErrorList {
```
shayasoolin: Reading the other comments, which I also agree with, the API server is validating most of the things for us.
So what's left is only the uniqueness of the levels?
Regarding the uniqueness of the keys - I'm not sure but I think at some point @sanjaychatterjee or @unmarshall said it should be possible to use the same key for two (consecutive) levels. For example for clusters with less physical levels, but where you still want the user to be able to use all levels (and make the reduction on Grove/PodGang side).

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,215 @@
...
+		newLevel := newSpec.Levels[i]
+		oldLevel := oldSpec.Levels[i]
+
+		// Validate that domain is immutable
+		if newLevel.Domain != oldLevel.Domain {
+			allErrs = append(allErrs, field.Invalid(levelPath.Child("domain"), newLevel.Domain,
+				fmt.Sprintf("domain is immutable, cannot change from '%s' to '%s'", oldLevel.Domain, newLevel.Domain)))
+		}
+
+		// Note: Key is allowed to change (not in the requirements), but validation has already occurred in validate()
```
shayasoolin, line 178: This comment is not clear to me. What validation do you refer to?

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
```diff
@@ -0,0 +1,564 @@
...
+							Key:    "topology.kubernetes.io/rack",
+						},
+					},
+				},
+			},
+			expectedErr:    true,
+			expectedErrMsg: "topology levels must be in hierarchical order",
+		},
+		{
+			name: "invalid - levels out of order (rack before datacenter)",
```
shayasoolin: Why do we need 3 different tests for "levels out of order"? What does each of them add?

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
```diff
@@ -0,0 +1,564 @@
...
+							Domain: grovecorev1alpha1.TopologyDomainNuma,
+							Key:    "topology.kubernetes.io/numa",
+						},
+					},
+				},
+			},
+			expectedErr: false,
+		},
+		{
+			name: "valid - correct hierarchical order with gaps (region > rack > numa)",
```
shayasoolin: I think such case was already validated in one of the first positive tests in this file.

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
```diff
@@ -0,0 +1,564 @@
...
+func TestValidateUpdate(t *testing.T) {
+	tests := []struct {
+		name           string
+		oldCT          *grovecorev1alpha1.ClusterTopology
+		newCT          *grovecorev1alpha1.ClusterTopology
+		expectedErr    bool
+		expectedErrMsg string
+	}{
+		{
+			name: "valid update - key changed",
```
shayasoolin, line 266: Please add a negative test where a key is changed to an already existing key, s.t. the result key list is no longer unique.

operator/internal/webhook/admission/clustertopology/validation/handler.go
shayasoolin: But in create you have only one arg, and certain validations,
and in update you have two args (old and new) and different validations. ?

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: why? What is the benefit? using range is the idiomatic go way.

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: fix

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: done

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: moved

operator/internal/webhook/admission/clustertopology/validation/handler.go
shmuel-runai: This is the common pattern used in the PCS validation webhook as well, aligning the code with that

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: moved

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: The current design requires key uniqueness, will double check to ensure that

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
shmuel-runai: added

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
shmuel-runai: removed

operator/internal/webhook/admission/clustertopology/validation/clustertopology_test.go
shmuel-runai: no real need, removing the extra test cases

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
Ronkahn21: its the prefer stye in grove

operator/internal/webhook/admission/clustertopology/validation/handler.go
Ronkahn21: in this case you dont really need to run the first validation as we only allow to update the key, as it might be waste time to check evertything from the start just to denied it on changing somethings is not allowed in the firstplace

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: removed

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: removed

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: removed

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: Are you sure? The only validations is on the key are
1. required
2. len between 1 to 63, nothing on the content

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: The comment was removed due to other code changes

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: do you really think it is better 🤮

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
shmuel-runai: Isn't it what I'm already doing?

operator/charts/templates/clustertopology-validating-webhook-config.yaml
```diff
@@ -0,0 +1,34 @@
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: clustertopology-validating-webhook
+  labels:
+{{- include "operator.clustertopology.validating.webhook.labels" . | nindent 4 }}
```
renormalize, line 7: Shouldn't this webhook configuration also be namespaced, like the other webhooks?

```suggestion
metadata:
  name: clustertopology-validating-webhook
  namespace: {{ .Release.Namespace }}
  labels:
{{- include "operator.clustertopology.validating.webhook.labels" . | nindent 4 }}
```

operator/internal/webhook/admission/clustertopology/validation/handler.go
```diff
@@ -0,0 +1,102 @@
...
+	h.logValidatorFunctionInvocation(ctx)
+	clusterTopology, err := castToClusterTopology(obj)
+	if err != nil {
+		return nil, errors.WrapError(err, ErrValidateCreateClusterTopology, string(admissionv1.Create), "failed to cast object to ClusterTopology")
+	}
+	return nil, newClusterTopologyValidator(clusterTopology, admissionv1.Create).validate()
+}
+
+// ValidateUpdate validates a ClusterTopology update request.
+func (h *Handler) ValidateUpdate(ctx context.Context, newObj, oldObj runtime.Object) (admission.Warnings, error) {
```
renormalize: The fields are misordered. Checking `CustomValidator` documentation:

```
	// ValidateUpdate validates the object on update.
	// The optional warnings will be added to the response as warning messages.
	// Return an error if the object is invalid.
	ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (warnings Warnings, err error)
```

```suggestion
func (h *Handler) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
```


A similar error was made in the PCS validator previously, which got fixed in #256.

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,188 @@
...
+
+// validateUpdate validates the update to a ClusterTopology object.
+func (v *clusterTopologyValidator) validateUpdate(oldClusterTopology *grovecorev1alpha1.ClusterTopology) error {
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, v.validateClusterTopologySpecUpdate(&v.clusterTopology.Spec, &oldClusterTopology.Spec, field.NewPath("spec"))...)
+	return allErrs.ToAggregate()
+}
+
+// validateClusterTopologySpecUpdate validates updates to the ClusterTopology specification.
+func (v *clusterTopologyValidator) validateClusterTopologySpecUpdate(newSpec, oldSpec *grovecorev1alpha1.ClusterTopologySpec, fldPath *field.Path) field.ErrorList {
```
renormalize: Sticking to the convention.

```suggestion
func (v *clusterTopologyValidator) validateClusterTopologySpecUpdate(oldSpec, newSpec *grovecorev1alpha1.ClusterTopologySpec, fldPath *field.Path) field.ErrorList {
```

operator/internal/webhook/admission/clustertopology/validation/clustertopology.go
```diff
@@ -0,0 +1,188 @@
...
+		oldLevel := oldLevels[i]
+
+		// Validate that domain is immutable
+		if newLevel.Domain != oldLevel.Domain {
+			allErrs = append(allErrs, field.Invalid(levelPath.Child("domain"), newLevel.Domain,
+				fmt.Sprintf("domain is immutable, cannot change from '%s' to '%s'", oldLevel.Domain, newLevel.Domain)))
+		}
+
+		// Note: Key is allowed to change (not in the requirements), but validation has already occurred in validate()
+	}
```
renormalize, line 179: Simplify the code by relying on the APIs exposed by `apimachinery`. Import `apivalidation "k8s.io/apimachinery/pkg/api/validation"`.

```suggestion
	for i := range newLevels {
		levelPath := fldPath.Index(i)

		allErrs = append(allErrs, apivalidation.ValidateImmutableField(newLevels[i].Domain, oldLevels[i].Domain, levelPath.Child("domain"))...)
		// Note: Key is allowed to change (not in the requirements), but validation has already occurred in validate()
	}
```

operator/internal/webhook/admission/clustertopology/validation/clustertopology_crd_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+// setupEnvTest creates an envtest environment with ClusterTopology CRDs loaded
```
renormalize, line 216: Nit: can the setup functions be moved to the end of the file?

operator/internal/webhook/admission/clustertopology/validation/clustertopology_crd_test.go
```diff
@@ -0,0 +1,205 @@
...
+	return testEnv, k8sClient
+}
+
+// teardownEnvTest stops the envtest environment
+func teardownEnvTest(t *testing.T, testEnv *envtest.Environment) {
+	err := testEnv.Stop()
+	require.NoError(t, err)
+}
+
+func TestClusterTopologyCRDValidation(t *testing.T) {
```
renormalize: Can this test be modified to be table driven? Would make it consistent with the rest and also easier to read. 
Not too strong an opinion on this, however.

Also, can one or two positive tests be added as well?

renormalize (changes requested):
Thanks @shmuel-runai! Posted a few fits - please take a look.

operator/charts/templates/clustertopology-validating-webhook-config.yaml
shmuel-runai: no, it was wrong to place them at all, I open a different PR to fix them (see https://github.com/ai-dynamo/grove/pull/258)

operator/internal/webhook/admission/clustertopology/validation/clustertopology_crd_test.go
shmuel-runai: moved

operator/internal/webhook/admission/clustertopology/validation/clustertopology_crd_test.go
shmuel-runai: fix

sanjaychatterjee (approved):

operator/api/core/v1alpha1/clustertopology.go
```diff
@@ -77,6 +77,41 @@ const (
...
+}
+
+// Compare compares this domain with another domain.
+// Returns:
+//   - negative value if this domain is broader than other
+//   - zero if domains are equal
+//   - positive value if this domain is narrower than other
+//
+// This method assumes both domains are valid (enforced by kubebuilder validation).
+func (d TopologyDomain) Compare(other TopologyDomain) int {
```
unmarshall: This method is only used within this package. Any reason to expose this method?

operator/api/core/v1alpha1/clustertopology.go
```diff
@@ -77,6 +77,41 @@ const (
...
+
+// BroaderThan returns true if this domain represents a broader scope than the other domain.
+// For example, Region.BroaderThan(Zone) returns true.
+func (d TopologyDomain) BroaderThan(other TopologyDomain) bool {
+	return d.Compare(other) < 0
+}
+
+// NarrowerThan returns true if this domain represents a narrower scope than the other domain.
+// For example, Zone.NarrowerThan(Region) returns true.
+func (d TopologyDomain) NarrowerThan(other TopologyDomain) bool {
```
unmarshall: This method is not used anywhere. Where do you envision this will be used?

renormalize:
@Ronkahn21 can we get an approval from you as well, since you requested changes last and that is blocking the merge of this PR? Thanks.

@shmuel-runai while merging the PR please clean up unnecessary commit logs created during the review cycle in the commit message. Thanks.

PR #259, opened 2025-11-13
Cluster topology deletion prevention

shayasoolin:
#### What type of PR is this?

/kind feature

Implements the ClusterTopology controller with deletion prevention logic. The controller:
- Automatically manages the `grove.io/clustertopology` finalizer on all ClusterTopology resources
- Prevents deletion when the topology is in use by PodCliqueSet resources (via `grove.io/cluster-topology-name` label)
- Prevents deletion when the topology is actively configured in the operator settings
- Requeues deletion attempts every 30s until blockers are cleared

Follows existing Grove controller patterns and achieves 90.6% test coverage.

**Dependencies:**
- Requires the TopologyConfiguration API changes (separate PR)

**Deletion Prevention Logic:**
The controller uses a conservative approach - it blocks deletion if any PodCliqueSet references the topology OR if the topology is configured in operator settings. The default topology name is `grove-topology` if not specified.

**Test Coverage:**
25 test cases covering creation, deletion, edge cases, and error handling. Uses fake clients following existing Grove test patterns.

```docs
To reference a ClusterTopology from a PodCliqueSet, add the grove.io/topology-name label.

To delete a ClusterTopology, ensure no PodCliqueSet resources reference it and it's not configured in the operator settings. The controller will automatically retry deletion every 30 seconds.
```

### DIFF
--- a/docs/designs/topology.md
+++ b/docs/designs/topology.md
@@ -229,7 +229,7 @@ Deletion Workflow:
 2. Kubernetes blocks deletion (finalizer `grove.io/clustertopology` present)
 3. Controller reconciles:
    - Detects deletion request (deletion timestamp set)
-   - Checks if any PodCliqueSet (in any namespace) references this ClusterTopology (via `grove.io/topology-name` label)
+   - Checks if any PodCliqueSet (in any namespace) references this ClusterTopology (via `grove.io/cluster-topology-name` label)
    - Checks if topology is enabled in operator config (`clusterTopology.enabled: true`) and this specific topology is configured (considering both the default topology name and the `clusterTopology.name` setting)
    - If ANY PodCliqueSet references this topology OR topology is enabled and set to this topology: Keeps finalizer, deletion blocked
    - If NO PodCliqueSet references this topology AND topology is disabled or not set to this topology: Removes finalizer, deletion proceeds
@@ -238,7 +238,7 @@ Deletion Workflow:
 Key Points:
 
 - Admin must satisfy BOTH conditions before deletion:
-    - Delete all PodCliqueSet resources that reference this ClusterTopology (check `grove.io/topology-name` label)
+    - Delete all PodCliqueSet resources that reference this ClusterTopology (check `grove.io/cluster-topology-name` label)
     - Disable TAS in operator config (`clusterTopology.enabled: false`) and restart operator
 - Controller checks both conditions before allowing deletion
 - Controller continuously reconciles deletion requests


operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := r.client.List(ctx, pcsList, labelSelector); err != nil {
+		return ctrlcommon.ReconcileWithErrors("failed to list PodCliqueSet resources", err)
+	}
+	if len(pcsList.Items) > 0 {
+		logger.Info("Cannot delete ClusterTopology: referenced by PodCliqueSet resources",
+			"topologyName", ct.Name,
+			"podCliqueSetCount", len(pcsList.Items))
+		r.eventRecorder.Eventf(ct, corev1.EventTypeWarning, groveconstants.ReasonClusterTopologyDeleteBlocked,
+			"Cannot delete ClusterTopology %s: referenced by %d PodCliqueSet resource(s)", ct.Name, len(pcsList.Items))
+		return ctrlcommon.ReconcileAfter(30*time.Second, "topology referenced by PodCliqueSet resources")
```
danbar2: Do we want to reconcile in that case? It could lead to a infinite loop

shayasoolin: Good point. I changed it to be done using a watch on PCS, so we'll get an event when a PCS is deleted or changed. Thanks

operator/internal/controller/clustertopology/register.go
```diff
@@ -32,5 +41,69 @@ func (r *Reconciler) RegisterWithManager(mgr manager.Manager) error {
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Reconcile if topology label was removed or changed
+			// Empty values are treated as non-existent
+			oldLabel, oldExists := e.ObjectOld.GetLabels()[apicommon.LabelClusterTopologyName]
+			oldHasValue := oldExists && oldLabel != ""
+			newLabel, newExists := e.ObjectNew.GetLabels()[apicommon.LabelClusterTopologyName]
+			newHasValue := newExists && newLabel != ""
+
+			// Label removed (had value, now doesn't) or changed (both have values but different)
+			return (oldHasValue && !newHasValue) || (oldHasValue && newHasValue && oldLabel != newLabel)
```
danbar2: the event is on the new obj, so if newHasValue  is false => the map function will drop the event

shayasoolin: I didn't get you. There are two cases in which the event is NOT dropped:
1. Label removed (had value, now doesn't) --> (oldHasValue && !newHasValue)
2. Label changed (both have values but different) --> (oldHasValue && newHasValue && oldLabel != newLabel)
Which case are you referring to with the newHasValue==false?

danbar2: the first one -
you will get to the mapping function with the new obj => read the topology name => its empty => no reconcile

actually, the second one is also problematic -
you will get to the mapping function with the new obj => read the topology name => its the new one => if the old one is a topology that is being deleted and now no PCS using it we will not reconcile it

docs/designs/topology.md
```diff
@@ -228,11 +228,11 @@ Deletion Workflow:
 1. Admin runs `kubectl delete clustertopology <name>`
 2. Kubernetes blocks deletion (finalizer `grove.io/clustertopology` present)
 3. Controller reconciles:
-    - Detects deletion request (deletion timestamp set)
+   - Detects deletion request (deletion timestamp set)
    - Checks if any PodCliqueSet (in any namespace) references this ClusterTopology (via `grove.io/topology-name` label)
```
gflarity: I believe this grove.io/cluster-topology-name now?

operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,107 @@
...
+func (r *Reconciler) removeFinalizer(ctx context.Context, logger logr.Logger, ct *grovecorev1alpha1.ClusterTopology) ctrlcommon.ReconcileStepResult {
+	if !controllerutil.ContainsFinalizer(ct, constants.FinalizerClusterTopology) {
+		logger.Info("Finalizer not found", "ClusterTopology", ct.Name)
+		return ctrlcommon.ContinueReconcile()
+	}
+	logger.Info("Removing finalizer", "ClusterTopology", ct.Name, "finalizerName", constants.FinalizerClusterTopology)
+	if err := ctrlutils.RemoveAndPatchFinalizer(ctx, r.client, ct, constants.FinalizerClusterTopology); err != nil {
+		logger.Error(err, "failed to remove finalizer", "ClusterTopology", ct.Name, "finalizerName", constants.FinalizerClusterTopology)
+		r.eventRecorder.Eventf(ct, corev1.EventTypeWarning, groveconstants.ReasonClusterTopologyDeleteFailed,
+			"Cannot delete ClusterTopology %s: error removing finalizer: %s from ClusterTopology: %s: %w", constants.FinalizerClusterTopology, ct.Name, err)
```
gflarity: %w is only supported by fmt.Errorf unfortunately. Pretty sure the underlying code is using fmt.Sprintf so this won't work properly.  So you'll need to use a %v I think.

gflarity: Also check the arguments. ct.Name isn't first, but looks like that what you want for the %s.

operator/internal/controller/clustertopology/reconciledelete_test.go
```diff
@@ -0,0 +1,357 @@
...
+			},
+			wantRequeue: false,
+			wantError:   false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.topologyConfig.Name == "" {
+				tt.topologyConfig.Name = configv1alpha1.DefaultClusterTopologyName // TODO remove this after taking care of registering the defaults properly
```
gflarity, line 162: Nit, create a github issue or something else to track this outside of code?

operator/internal/controller/clustertopology/reconciledelete_test.go
```diff
@@ -0,0 +1,357 @@
...
+			wantRequeue:          false,
+			wantError:            false,
+		},
+		{
+			name: "blocked deletion flow - topology enabled",
+			ct: testutils.NewClusterTopologyBuilder(configv1alpha1.DefaultClusterTopologyName).
+				WithFinalizer(constants.FinalizerClusterTopology),
+			topologyConfig: configv1alpha1.ClusterTopologyConfiguration{
+				Enabled: true,
+				Name:    configv1alpha1.DefaultClusterTopologyName, // TODO remove when defaults are registered properly
```
gflarity, line 302: Same here...

operator/internal/controller/clustertopology/register.go
shayasoolin: Changed the implementation to use custom handler. This way, both the predicate and the mapping have both old and new objects to work with. So we return a list of both topologies - old and new, so both will be reconciled.

operator/internal/controller/clustertopology/reconciledelete_test.go
shayasoolin: I still hope to figure this out before merging this PR. Otherwise will do.

docs/designs/topology.md
```diff
@@ -228,17 +228,17 @@ Deletion Workflow:
...
+   - Checks if topology is enabled in operator config (`topology.enabled: true`) and this specific topology is configured (considering both the default topology name and the `topology.name` setting)
+   - If ANY PodCliqueSet references this topology OR topology is enabled and set to this topology: Keeps finalizer, deletion blocked
+   - If NO PodCliqueSet references this topology AND topology is disabled or not set to this topology: Removes finalizer, deletion proceeds
 4. Once finalizer removed, Kubernetes deletes ClusterTopology
 
 Key Points:
 
 - Admin must satisfy BOTH conditions before deletion:
-    - Delete all PodCliqueSet resources that reference this ClusterTopology (check `grove.io/topology-name` label)
+    - Delete all PodCliqueSet resources that reference this ClusterTopology (check `grove.io/cluster-topology-name` label)
```
Ronkahn21, line 241: you can revert this changes as I created a PR with this changes

docs/designs/topology.md
```diff
@@ -280,11 +280,11 @@ topology:
 3. Admin restarts operator (see Startup Behavior section for details)
 4. Operator validates ClusterTopology CR exists
 5. For existing workloads:
-    - Workloads with topology constraints: operator uses topology from PodCliqueSet label (`grove.io/topology-name`)
+    - Workloads with topology constraints: operator uses topology from PodCliqueSet label (`grove.io/cluster-topology-name`)
```
Ronkahn21, line 283: same

docs/designs/topology.md
```diff
@@ -280,11 +280,11 @@ topology:
...
     - Workloads without topology constraints: no impact (checked via label presence)
 6. For new workloads:
     - Topology constraints validated against ClusterTopology CR
-   - Mutation webhook adds `grove.io/topology-name` label to PodCliqueSet (see Mutation Webhook section)
+   - Mutation webhook adds `grove.io/cluster-topology-name` label to PodCliqueSet (see Mutation Webhook section)
```
Ronkahn21, line 287: same

operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,108 @@
...
+
+// triggerDeletionFlow handles the deletion of a ClusterTopology with deletion prevention checks.
+func (r *Reconciler) triggerDeletionFlow(ctx context.Context, logger logr.Logger, ct *grovecorev1alpha1.ClusterTopology) ctrlcommon.ReconcileStepResult {
+	logger.Info("Triggering deletion flow", "ClusterTopology", ct.Name)
+	deleteStepFns := []ctrlcommon.ReconcileStepFn[grovecorev1alpha1.ClusterTopology]{
+		r.checkDeletionConditions,
+		r.removeFinalizer,
+	}
+
+	for _, fn := range deleteStepFns {
```
Ronkahn21, line 44:
```suggestion
	for _, deleteStepFn := range deleteStepFns {
```

operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	return ctrlcommon.DoNotRequeue()
+}
+
+// checkDeletionConditions verifies that ClusterTopology can be safely deleted.
+// Deletion is blocked if:
+// 1. Any PodCliqueSet references this topology (via grove.io/cluster-topology-name label)
+// 2. Topology is enabled AND this specific topology is configured
+func (r *Reconciler) checkDeletionConditions(ctx context.Context, logger logr.Logger, ct *grovecorev1alpha1.ClusterTopology) ctrlcommon.ReconcileStepResult {
```
Ronkahn21, line 57:
```suggestion
func (r *Reconciler) checkDeletionConditions(ctx context.Context, logger logr.Logger, cTopology *grovecorev1alpha1.ClusterTopology) ctrlcommon.ReconcileStepResult {
```

operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,108 @@
...
+	logger.Info("Checking deletion conditions", "ClusterTopology", ct.Name)
+	// Condition 1: Check for PodCliqueSet references
+	pcsList := &grovecorev1alpha1.PodCliqueSetList{}
+	labelSelector := client.MatchingLabels{
+		apicommon.LabelClusterTopologyName: ct.Name,
+	}
+	if err := r.client.List(ctx, pcsList, labelSelector); err != nil {
```
Ronkahn21, line 75: as we only search for existing resource we can limit the result to 1 with listOptions, so if there is at least one dont allow to delete

operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,108 @@
...
+		return ctrlcommon.ReconcileWithErrors("failed to list PodCliqueSet resources", err)
+	}
+	if len(pcsList.Items) > 0 {
+		logger.Info("Cannot delete ClusterTopology: referenced by PodCliqueSet resources",
+			"topologyName", ct.Name,
+			"podCliqueSetCount", len(pcsList.Items))
```
Ronkahn21, line 81: when the cluster is big enough listing all the PCS is just waste, so we can limit to only one result and dont show the  podCliqueSetCount

operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,108 @@
...
+		r.eventRecorder.Eventf(ct, corev1.EventTypeWarning, groveconstants.ReasonClusterTopologyDeleteBlocked,
+			"Cannot delete ClusterTopology %s: referenced by %d PodCliqueSet resource(s)", ct.Name, len(pcsList.Items))
+		// Don't requeue - the watch on PodCliqueSet will trigger reconciliation when PCS is deleted
```
Ronkahn21, line 84: this is dangerous, as you might watch for delete, but it might take some time it acutely getting deleted, and if the the last one (pcs) you might want get a reconcile for it anymore and therefore it hangs

operator/internal/controller/clustertopology/reconciledelete.go
```diff
@@ -0,0 +1,108 @@
...
+		return ctrlcommon.DoNotRequeue()
+	}
+
+	// Condition 2: Check if topology is enabled and configured to use this ClusterTopology
+	if r.config.ClusterTopology.Enabled && r.config.ClusterTopology.Name == ct.Name {
```
Ronkahn21: I will first check this and then go for checking if there is pcs ref it, as we prefer non api calls first

operator/internal/constants/constants.go
```diff
@@ -96,3 +96,17 @@ const (
...
+// constants for ClusterTopology lifecycle events
+const (
+	// ReasonClusterTopologyCreateOrUpdateSuccessful is an event reason which represents a successful creation or update of a ClusterTopology.
+	ReasonClusterTopologyCreateOrUpdateSuccessful = "ClusterTopologyCreateOrUpdateSuccessful"
+	// ReasonClusterTopologyCreateOrUpdateFailed is an event reason which represents that the creation or update of a ClusterTopology failed.
+	ReasonClusterTopologyCreateOrUpdateFailed = "ClusterTopologyCreateOrUpdateFailed"
+	// ReasonClusterTopologyDeleteBlocked is an event reason which represents that the deletion of a ClusterTopology is blocked.
+	ReasonClusterTopologyDeleteBlocked = "ClusterTopologyDeleteBlocked"
+	// ReasonClusterTopologyDeleteSuccessful is an event reason which represents a successful deletion of a ClusterTopology.
+	ReasonClusterTopologyDeleteSuccessful = "ClusterTopologyDeleteSuccessful"
```
Ronkahn21, line 109: the controller only look on it self, there might another finlizer created by someone else, so it better to stay a way deleted successful, just finlizer removed successfully

operator/internal/controller/clustertopology/register.go
```diff
@@ -0,0 +1,137 @@
...
+type podCliqueSetEventHandler struct{}
+
+// Create handles PodCliqueSet creation events.
+// We don't reconcile on create since topology isn't blocked when PCS is created.
+func (h *podCliqueSetEventHandler) Create(_ context.Context, _ event.TypedCreateEvent[client.Object], _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	// No-op: don't reconcile on create
+}
+
+// Update handles PodCliqueSet update events.
+// Reconciles both old and new topologies if the topology label was changed or removed.
```
Ronkahn21, line 63: we dont allow to change this label, so it might redundant to look for this,  so no reconcile for this either

operator/internal/controller/clustertopology/register.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Generic handles generic events for PodCliqueSet.
+// We don't reconcile on generic events.
+func (h *podCliqueSetEventHandler) Generic(_ context.Context, _ event.TypedGenericEvent[client.Object], _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	// No-op: don't reconcile on generic events
+}
+
+// getTopologyName extracts the topology name from a PodCliqueSet's labels.
+// Returns empty string if the label doesn't exist or is empty.
+func getTopologyName(pcs *grovecorev1alpha1.PodCliqueSet) string {
```
Ronkahn21: I would move this to util as it usefull to the PCS controller as well

operator/internal/controller/register.go
```diff
@@ -26,7 +27,7 @@ import (
 )
 
 // RegisterControllers registers all controllers with the manager.
-func RegisterControllers(mgr ctrl.Manager, controllerConfig configv1alpha1.ControllerConfiguration) error {
+func RegisterControllers(mgr ctrl.Manager, controllerConfig configv1alpha1.ControllerConfiguration, topologyConfig configv1alpha1.OperatorConfiguration) error {
```
Ronkahn21: its not only the topology config, so you need to change the name topologyConfig

sanjaychatterjee:
@shayasoolin can you please rebase your changes and resolve conflicts? Also, please address the comments where @Ronkahn21 mentions that those changes are done in a separate PR.

operator/internal/controller/clustertopology/reconciledelete.go
shayasoolin: I chose to be aligned with all other usages of this pattern

operator/internal/constants/constants.go
shayasoolin: Again I'm aligned with other similar consts. When we add another finalizer we can change that, for now we're the only controller looking at our own CRD

operator/internal/controller/clustertopology/register.go
shayasoolin: I prefer to be resilient and independent of such design decisions. The price is low, so I think it's better to support this case as well. We may decide to allow changing topology for a PCS (it's actually not so well defined yet, as I see it.)

PR #269, opened 2025-11-21
E2E tests for startup ordering

gflarity:
#### What type of PR is this?

E2E Tests

This PR implements the startup ordering scenarios defined in the Grove Multi-Node QA Test Plan. It confirms that everything is working as expected as PR the test plan. 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/operator/e2e/tests/startup_ordering_test.go
+++ b/operator/e2e/tests/startup_ordering_test.go
@@ -93,8 +93,8 @@ func Test_SO1_InorderStartupOrderWithFullReplicas(t *testing.T) {
 	logger.Info("   pcs-0-pc-a")
 	logger.Info("   pcs-0-sg-x-0-pc-b, pcs-0-sg-x-1-pc-b")
 	logger.Info("   pcs-0-sg-x-0-pc-c, pcs-0-sg-x-1-pc-c")
-	verifyCliqueGroupStartupOrder(t, tc, "pc-a", 2, "pc-b", 2)
-	verifyCliqueGroupStartupOrder(t, tc, "pc-b", 2, "pc-c", 6)
+	verifyPodCliqueStartupOrder(t, tc, "pc-a", 2, "pc-b", 2)
+	verifyPodCliqueStartupOrder(t, tc, "pc-b", 2, "pc-c", 6)
 
 	logger.Info("🎉 Inorder startup order with full replicas test completed successfully!")
 }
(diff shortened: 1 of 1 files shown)


operator/e2e/tests/startup_ordering_test.go
```diff
@@ -0,0 +1,601 @@
...
+		}
+		if readyTime.After(latest) {
+			latest = readyTime
+		}
+	}
+	return latest
+}
+
+// CliqueGroupSpec defines a group of pods by clique name and expected count
+type CliqueGroupSpec struct {
```
shayasoolin: IIUC this better be called PodCliqueSpec.
We have PCLQ and PCSG, so I'd go with PodCliqueSpec and ScalingGroupOrderSpec, respectively.
If you agree, please also rename verifyCliqueGroupStartupOrder to be verifyPodCliqueStartupOrder.

operator/e2e/tests/startup_ordering_test.go
```diff
@@ -0,0 +1,601 @@
...
+
+	// Get running pods
+	var runningPodsList []v1.Pod
+	for _, pod := range pods.Items {
+		if pod.Status.Phase == v1.PodRunning {
+			runningPodsList = append(runningPodsList, pod)
+		}
+	}
+
+	// Build a map of clique name -> pods for all groups
```
shayasoolin: What is a group here? A single PCLQ?

operator/e2e/tests/startup_ordering_test.go
```diff
@@ -0,0 +1,601 @@
...
+	cliquePods := make(map[string][]v1.Pod)
+
+	// Collect prefix group pods
+	for _, group := range spec.PrefixGroups {
+		cliquePods[group.Name] = getPodsByCliquePattern(runningPodsList, group.Name)
+	}
+
+	// Collect scaling group pods for each clique in the within-replica order
+	for _, group := range spec.WithinReplicaOrder {
```
shayasoolin, line 408: Consider joining the two lists (PrefixGroups and WithinReplicaOrder) and collect their pods in the same loop. I think it will be clearer.

operator/e2e/tests/startup_ordering_test.go
```diff
@@ -0,0 +1,601 @@
...
+	// Verify counts for prefix groups
+	for _, group := range spec.PrefixGroups {
+		pods := cliquePods[group.Name]
+		if len(pods) != group.ExpectedCount {
+			t.Fatalf("Expected %d running %s pods, got %d", group.ExpectedCount, group.Name, len(pods))
+		}
+	}
+
+	// Verify counts for within-replica order groups
+	for _, group := range spec.WithinReplicaOrder {
```
shayasoolin: If you join the lists above, you can also merge these two for loops here.

shayasoolin (commented):
Great job!

operator/e2e/tests/startup_ordering_test.go
gflarity: Thanks, this was poorly written as "group" here refers to a PodCliqueSpec. I've fixed the inline docs here.

operator/e2e/tests/startup_ordering_test.go
shayasoolin: Don't you want to use allGroups below also for these two loops, joined into one? If you do, just make sure to handle repeats such that you don't handle the same group name twice.

shayasoolin (approved):

Issue #270, opened 2025-11-21
Add automatic support for MNNVL


julienmancuso:
### What you would like to be added?

Automatic MNNVL support (MultiNode NVLink) needs to be added to grove.

**What is MNNVL?**
MNNVL (Multi-Node NVLink) is NVIDIA's technology that enables GPUs in different servers to communicate at full NVLink bandwidth through NVIDIA NVLink Switches, transforming an entire rack into a single, unified GPU fabric. This is particularly important for systems like the NVIDIA GB200 NVL72

**What are ComputeDomains?**
A ComputeDomain is an abstraction for robust and secure Multi-Node NVLink that guarantees MNNVL-reachability between pods that are in the ComputeDomain, and secure isolation from other pods that are not in the ComputeDomain. When a workload requests a ComputeDomain, NVIDIA's DRA Driver for GPUs performs all the heavy lifting required for sharing GPU memory securely via NVLink among all pods that comprise the workload installation

### Why is this needed?

Need to support systems like the NVIDIA GB200 NVL72

PR #273, opened 2025-11-21
feat: add grep-270 for automatic support for MNNVL

julienmancuso:
#### What type of PR is this?

/kind documentation

GREP for #270

### DIFF
--- a/docs/greps/270-MNNVL/README.md
+++ b/docs/greps/270-MNNVL/README.md
@@ -17,12 +17,24 @@ We need a way for a `PodCliqueSet` (PCS) workload to:
   - Injecting **pod-level resource claims** and **container-level resource claim usage**.
   - Cleaning up these resources when the workload terminates.
 
+### How ComputeDomains work
+
+A **ComputeDomain** is a NVIDIA abstraction that enables workloads to use MNNVL (Multi-Node NVLink) across one or more physical racks. Key aspects:
+
+- **One ComputeDomain per PCS replica**, regardless of how many MNNVL racks the pods are distributed across.
+- The ComputeDomain controller **automatically orchestrates IMEX domains** (one per physical rack).
+  - Pods on the same rack communicate via fast MNNVL/NVLink.
+  - Pods on different racks fall back to the fastest available inter-rack fabric (slower than NVLink).
+- **Auto-scaling support**: New pods joining a workload can automatically join existing IMEX domains on their respective racks (requires DRA driver 25.8.0+ and GPU driver 570.150.x+).
+
+**Bottom line**: One ComputeDomain per PCS replica, automatic multi-rack orchestration, with fast MNNVL communication within racks but not across them.
+
 ### Goals and non-goals
 
 - **Goals**
   - Allow a PCS to declare MNNVL usage declaratively in its spec.
   - Support **per-replica ComputeDomains** so each PCS replica has its own isolated GPU fabric.
-  - Automatically manage the lifecycle of ComputeDomains and RCTs for managed mode.
+  - Automatically manage the lifecycle of ComputeDomains and RCTs.
   - Ensure pods are correctly wired via Kubernetes `resourceClaims` and container `resources.claims`.
 
 ### User-facing API changes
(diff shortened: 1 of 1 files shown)


docs/greps/270-MNNVL/README.md
```diff
@@ -0,0 +1,189 @@
...
+**Status**: Draft  
+**Tracking issue**: [Add automatic support for MNNVL](https://github.com/ai-dynamo/grove/issues/270)
+
+### Problem statement
+
+Grove today does not provide first-class support for NVIDIA Multi-Node NVLink (MNNVL). As a result, workloads running on systems like the NVIDIA GB200 NVL72 cannot automatically form secure, high-bandwidth GPU fabrics across nodes using NVIDIA ComputeDomains and Dynamic Resource Allocation (DRA).
+
+We need a way for a `PodCliqueSet` (PCS) workload to:
+
+- Express that it **requires a ComputeDomain** (i.e. wants to run with MNNVL).
```
shayasoolin, line 13: Ideally, Grove defaults to creating a ComputeDomain resource, even if not requested.
The user should be able to opt-out, but by default, MNNVL should be enabled.
In Run:ai we implemented those 3 request levels for MNNVL: no, yes and auto (aka none, required, preferred).
Auto means - best effort: only if all preconditions are met, a CD will be created and used.
Required means - if any precondition is not met, the workload will fail.
None means - opt-out, never create a CD.

There are some product decisions that need to be made around that. I think we should agree on that before finalizing the design.

docs/greps/270-MNNVL/README.md
```diff
@@ -0,0 +1,189 @@
...
+  - Pods on different racks fall back to the fastest available inter-rack fabric (slower than NVLink).
+- **Auto-scaling support**: New pods joining a workload can automatically join existing IMEX domains on their respective racks (requires DRA driver 25.8.0+ and GPU driver 570.150.x+).
+
+**Bottom line**: One ComputeDomain per PCS replica, automatic multi-rack orchestration, with fast MNNVL communication within racks but not across them.
+
+### Goals and non-goals
+
+- **Goals**
+  - Allow a PCS to declare MNNVL usage declaratively in its spec.
+  - Support **per-replica ComputeDomains** so each PCS replica has its own isolated GPU fabric.
```
shayasoolin, line 36: Do we mandate all the PCS components (PCLQs) to use the ComputeDomain?
We should consider allowing specific PCLQ/PCSGs to opt-out of it, e.g. when there's a frontend component alongside prefill and decode ones, one may don't want the frontend to specify the CD RCT.
Another approach could be not to add the CD RCT for components without a GPU requirement. It's weaker than the control level I suggested here. But it may be necessary even regardless the control level discussion - even if the CD is always applied to all components, it should skip non-GPU ones, for the spec to be valid. I might be wrong here though - not really sure what is the CD behavior in these cases.

shayasoolin, line 36: Do we want to allow having multiple CDs per PCS replica?
I'm not sure if we have a valid use case today, e.g. one PCS running two separate models? But if it makes sense, let's at least discuss how such a feature can be added in the future without breaking the proposed API.
I don't think we should support it today anyway, but if it makes sense as a future extension, let's think about it in advance.

docs/greps/270-MNNVL/README.md
```diff
@@ -0,0 +1,189 @@
...
+   - Register these in the PCS controller's operator registry.
+
+5. **RBAC updates** (`operator/charts/templates/clusterrole.yaml`):
+   - Add permissions for `resource.nvidia.com/v1alpha1/computedomains` (CRUD).
+   - Add permissions for `resource.k8s.io/v1beta1/resourceclaimtemplates` (CRUD).
+
+### Open questions for discussion
+
+- What is the desired failure behavior if the NVIDIA stack (ComputeDomain CRD/DRA) is not installed or misconfigured?
+- Should we add a status field to PCS to surface per-replica ComputeDomain readiness, or rely on users inspecting ComputeDomain objects directly?
```
shayasoolin, line 187: +1 for these two open questions :)
Status reporting on the PCS level (per replica) is super important IMHO.

PR #280, opened 2025-11-26
E2e tests rolling updates

gflarity:
PR for Rolling Updates tests. I was able to get most of them working

1. Please take a look at the fix in rollingupdate.go. Seems like a simple enough typo/cut and pass issue?
2. These fails are failing still:
    * 10 (always fails) and
    * 18 fails intermittently

### DIFF
--- a/operator/e2e/tests/debug_utils.go
+++ b/operator/e2e/tests/debug_utils.go
@@ -27,6 +27,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+const logBufferSize = 64 * 1024 // 64KB
+
 // captureOperatorLogs fetches and logs the operator logs from the specified namespace and deployment.
 // Uses tc.Ctx and tc.Clientset but takes namespace and deploymentPrefix as parameters since they
 // typically differ from the test namespace (e.g., "grove-system" vs "default").
@@ -59,8 +61,7 @@ func captureOperatorLogs(tc TestContext, namespace, deploymentPrefix string, fil
 					continue
 				}
 
-				// Read logs
-				buf := make([]byte, 64*1024) // 64KB buffer
+				buf := make([]byte, logBufferSize)
 				for {
 					n, err := logStream.Read(buf)
 					if n > 0 {
@@ -87,14 +88,17 @@ func captureOperatorLogs(tc TestContext, namespace, deploymentPrefix string, fil
 
 // containsRollingUpdateTag checks if a line contains rolling update related tags
 func containsRollingUpdateTag(line string) bool {
-	tags := []string{"[ROLLING_UPDATE]", "rolling update", "processPendingUpdates", "deleteOldPending", "nextPodToUpdate"}
+	tags := []string{
+		"[ROLLING_UPDATE]",
+		"rolling update",
+		"processPendingUpdates",
+		"deleteOldPending",
+		"nextPodToUpdate",
+	}
+
 	for _, tag := range tags {
-		if len(line) >= len(tag) {
-			for i := 0; i <= len(line)-len(tag); i++ {
-				if line[i:i+len(tag)] == tag {
-					return true
-				}
-			}
+		if strings.Contains(line, tag) {
+			return true
 		}
 	}
 	return false


operator/e2e/tests/setup.go
```diff
@@ -324,12 +340,16 @@ func waitForPodConditions(tc TestContext, expectedTotalPods, expectedPending int
 	return lastTotal, lastRunning, lastPending, err
 }
 
-// scalePCSGAndWait scales a PCSG and waits for the expected pod conditions to be reached.
-func scalePCSGAndWait(tc TestContext, pcsgName string, replicas int32, expectedTotalPods, expectedPending int) {
+// scalePCSGAndWait attempts to scale a PCSG via the PCS template and waits for the expected pod conditions.
+// WARNING: This only affects NEW PCS replicas created after the change. Existing PCSG instances are NOT updated.
+// DEPRECATED: Use scalePCSGAcrossAllReplicasAndWait instead to scale existing PCSG instances directly.
```
shayasoolin: Why keep this function here if it's deprecated and unused?

operator/e2e/utils/k8s_client.go
```diff
@@ -485,6 +485,75 @@ func ScalePodCliqueScalingGroupWithClient(ctx context.Context, dynamicClient dyn
 	return scaleCRD(ctx, dynamicClient, pcsgGVR, namespace, name, replicas)
 }
 
+// ScalePCSGInPCSWithClient modifies the PCS template's podCliqueScalingGroups[].replicas field.
+// WARNING: This only affects NEW PCS replicas created after the change. Existing PCSG instances
```
shayasoolin: Is this function in use currently? If not, can we remove it?

operator/e2e/tests/rolling_updates_test.go
```diff
@@ -0,0 +1,2670 @@
...
+// getPodIdentifier returns a stable identifier for a pod based on its logical position in the workload.
+// This identifier remains the same when a pod is replaced during rolling updates.
+//
+// Stability is guaranteed because the hostname is set by the operator based on the pod's logical
+// position in the PodClique (pclqName-podIndex), not the pod's generated name. During a rolling
+// update:
+//  1. The old pod is deleted (e.g., hostname: "workload1-pc-a-0-0")
+//  2. The PodClique remains unchanged (same name and replica count)
+//  3. The new pod is created to fill the same logical position
+//  4. The new pod receives the same hostname (e.g., "workload1-pc-a-0-0")
```
shayasoolin: Why is hostname promised to be stable? A pod can potentially be scheduled on a different host, post deletion, isn't it?

operator/e2e/tests/rolling_updates_test.go
```diff
@@ -0,0 +1,2670 @@
...
+
+	// After scale-in from 3 to 2 pc-a: 2 PCS replicas × (2 pc-a + 8 sg-x) = 2 × 10 = 20 pods
+	if len(pods.Items) != 20 {
+		t.Fatalf("Expected 20 pods after PodClique scale-in, got %d", len(pods.Items))
+	}
+
+	logger.Info("🎉 Rolling Update with PodClique scale-in before update test (RU-21) completed successfully!")
+}
+
+// podEvent represents a pod lifecycle event during rolling update
```
shayasoolin: Test-infra code from here till the end of this file better be split to a utils files.

operator/e2e/tests/rolling_updates_test.go
```diff
@@ -0,0 +1,2670 @@
...
+		tc.T.Fatalf("Pod %s does not have hostname set, cannot determine stable identifier", pod.Name)
+	}
+
+	return pod.Spec.Hostname
+}
+
+// verifyOnePodDeletedAtATime verifies only one pod globally is being deleted at a time
+// during rolling updates.
+//
+// This function handles the fact that Kubernetes watch events can arrive out of order.
```
shayasoolin: I'm curious if KAI-scheduler has similar tests where it's required to check the order of events, and how these validations are implemented there. @romanbaron

operator/e2e/tests/rolling_updates_test.go
```diff
@@ -0,0 +1,2670 @@
...
+func Test_RU7_RollingUpdatePCSPodClique(t *testing.T) {
+	ctx := context.Background()
+
+	logger.Info("1. Initialize a 10-node Grove cluster")
+	clientset, restConfig, dynamicClient, cleanup := prepareTestCluster(ctx, t, 10)
+	defer cleanup()
+
+	logger.Info("2. Deploy workload WL1, and verify 10 newly created pods")
+	expectedPods := 10
+	tc := TestContext{
```
shayasoolin: I think it worth the time to try and minimize repeated code over the test cases here. Including the initialization of the text context (most of the parameters are the same for all tests), as well as the deploy, wait, tracker code that comes after it.
It will help us maintain the code moving forward.

operator/e2e/tests/setup.go
gflarity: You're right, it should be removed. Removing it now.

operator/e2e/utils/k8s_client.go
gflarity: Yes, thanks. It didn't show up in linter because it's capitalized.

operator/e2e/tests/rolling_updates_test.go
gflarity: AH, this isn't the Node host name. This pod.Spec.Hostname, the internal hostname of the pod itself. it's what the hostname command would return inside the container. See /internal/controller/podclique/components/pod/pod.go:

```go
// configurePodHostname sets the pod hostname and subdomain for service discovery
func configurePodHostname(pcsName string, pcsReplicaIndex int, pclqName string, pod *corev1.Pod, podIndex int) {
	// Set hostname for service discovery (e.g., "my-pclq-0")
	pod.Spec.Hostname = fmt.Sprintf("%s-%d", pclqName, podIndex)

	// Set subdomain to headless service name (reusing existing logic)
	pod.Spec.Subdomain = apicommon.GenerateHeadlessServiceName(
		apicommon.ResourceNameReplica{Name: pcsName, Replica: pcsReplicaIndex})
}
```
But I definitely agree this is confusing. I'll update the comment to give a better explanation.

operator/e2e/tests/setup.go
gflarity: Thanks, linter missed this because it's referenced by another function that's not used anymore. Removing.

gflarity:
Thanks @shayasoolin, please take a look: 

* removed the unused functions I missed
* moved the helper functions into separate files 
* reduced repetition in the tests themselves

T

operator/e2e/tests/rolling_updates_test.go
romanbaron: Actually I cannot recall any use case in KAI where multiple events are recorded with a specific order between them.

operator/e2e/tests/debug_utils.go
```diff
@@ -0,0 +1,170 @@
...
+					}
+				}
+				logStream.Close()
+			}
+		}
+	}
+}
+
+// splitLines splits a string into lines
+func splitLines(s string) []string {
```
shmuel-runai: what is wrong with `strings.Split(s, "\n")`

operator/e2e/tests/debug_utils.go
```diff
@@ -0,0 +1,170 @@
...
+				buf := make([]byte, 64*1024) // 64KB buffer
+				for {
+					n, err := logStream.Read(buf)
+					if n > 0 {
+						// Print logs line by line, filtering for [ROLLING_UPDATE] tags
+						lines := string(buf[:n])
+						for _, line := range splitLines(lines) {
+							if len(line) > 0 {
+								// Print all lines, but highlight rolling update related ones
+								if containsRollingUpdateTag(line) {
```
shmuel-runai: a bit YANGI but,`containsRollingUpdateTag` should be a function param to this function (filterFN), so that this captureOperatorLogs could be reused with other testing than rolling update

operator/e2e/tests/rolling_update_tracker.go
```diff
@@ -0,0 +1,128 @@
...
+// newRollingUpdateTracker creates a new rolling update tracker
+func newRollingUpdateTracker() *rollingUpdateTracker {
+	return &rollingUpdateTracker{
+		ready: make(chan struct{}),
+	}
+}
+
+// Start begins watching pod events.
+// Uses tc.Ctx, tc.Clientset, tc.Namespace, and tc.getLabelSelector() for watch configuration.
+func (t *rollingUpdateTracker) Start(tc TestContext) error {
```
shmuel-runai, line 54: nit, you can make the ready channel internal to this function, and avoid "waiting for ready" function, and it usage at the caller end

operator/e2e/tests/rolling_update_utils.go
```diff
@@ -0,0 +1,1149 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/util/retry"
+)
+
+// triggerRollingUpdate triggers a rolling update on the specified cliques and returns an errgroup.Group
+// that completes when the rolling update finishes.
+// Uses tc.Workload.Name as the PCS name and tc.Timeout for the wait timeout.
+func triggerRollingUpdate(tc TestContext, expectedReplicas int32, cliqueNames ...string) *errgroup.Group {
+	g := new(errgroup.Group)
```
shmuel-runai: I'm not sure I understand the benefit of using the errgroup here?
maybe if it was
```
	for _, cliqueName := range cliqueNames {

		g.Go(func() error {
			startTime := time.Now()
			// Trigger synchronously first
			if err := triggerPodCliqueRollingUpdate(tc, cliqueName); err != nil {
				return fmt.Errorf("failed to update PodClique %s spec: %w", cliqueName, err)
  ...
```

operator/e2e/tests/rolling_update_utils.go
```diff
@@ -0,0 +1,1149 @@
...
+		}
+
+		return nil
+	})
+}
+
+// patchPCSWithSIGTERMIgnoringCommand patches all containers in the PCS to use a command that ignores SIGTERM
+// and sets the termination grace period to 5 seconds. This makes pods ignore graceful shutdown but still
+// allows rolling updates to progress in a reasonable time for testing.
+// Uses tc.Workload.Name as the PCS name.
```
shmuel-runai: nit: last line of the comment is redundant

operator/internal/controller/podclique/components/pod/rollingupdate.go
```diff
@@ -60,7 +60,7 @@ func (w *updateWork) getPodNamesPendingUpdate(deletionExpectedPodUIDs []types.UI
 // getNextPodToUpdate selects the next ready pod with old template hash to update, prioritizing oldest pods first
 func (w *updateWork) getNextPodToUpdate() *corev1.Pod {
 	if len(w.oldTemplateHashReadyPods) > 0 {
-		slices.SortFunc(w.oldTemplateHashPendingPods, func(a, b *corev1.Pod) int {
+		slices.SortFunc(w.oldTemplateHashReadyPods, func(a, b *corev1.Pod) int {
```
shmuel-runai, line 63: great catch

operator/e2e/tests/rolling_update_utils.go
gflarity: This needs to run asynchronously, see the call sites. This just seemed a like a clean way to do that without re-inventing the wheel.  However after a bit more research is seems like <-chan error is a bit more idiomatic. Either works for me.

operator/e2e/tests/debug_utils.go
```diff
@@ -0,0 +1,156 @@
...
+				})
+
+				logStream, err := req.Stream(tc.Ctx)
+				if err != nil {
+					logger.Errorf("Failed to get logs for container %s: %v", container.Name, err)
+					continue
+				}
+
+				// Read logs
+				buf := make([]byte, 64*1024) // 64KB buffer
```
shmuel-runai: nit: should be a const, not a comment

operator/e2e/tests/debug_utils.go
```diff
@@ -0,0 +1,156 @@
...
+					}
+				}
+				logStream.Close()
+			}
+		}
+	}
+}
+
+// containsRollingUpdateTag checks if a line contains rolling update related tags
+func containsRollingUpdateTag(line string) bool {
```
shmuel-runai, line 90:
```
func containsRollingUpdateTag(line string) bool {
	tags := []string{
		"[ROLLING_UPDATE]",
		"rolling update",
		"processPendingUpdates",
		"deleteOldPending",
		"nextPodToUpdate",
	}

	for _, tag := range tags {
		if strings.Contains(line, tag) {
			return true
		}
	}
	return false
}
```

shmuel-runai (approved):
Nice work

PR #282, opened 2025-11-30
ci: add e2e tests to GitHub Actions

shmuel-runai:
#### What type of PR is this?

/kind e2e

Add e2e test workflow that runs on PRs (non-draft or with 'run-e2e' label).

- New workflow: .github/workflows/e2e-test.yaml
- Add make test-e2e targets with automatic CRD preparation
- Update documentation with CI integration and troubleshooting

Tests run in ~10-15 minutes on GitHub-hosted runners with automatic cleanup and log artifact uploads on failure.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -26,7 +26,7 @@ jobs:
     # Run on non-draft PRs or draft PRs with 'run-e2e' label
     if: github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'run-e2e')
     runs-on: ubuntu-latest
-    timeout-minutes: 45
+    timeout-minutes: 60
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -47,16 +47,12 @@ jobs:
           sudo install skaffold /usr/local/bin/
           skaffold version
 
-      - name: Log in to GitHub Container Registry
-        uses: docker/login-action@v3
-        with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Run e2e tests
         run: make test-e2e
 
+      # The test code handles cleanup via Teardown(), but this step provides
+      # extra safety in case of timeout or panic. Also good practice to ensure
+      # clean state even though GitHub Actions runners are ephemeral.
       - name: Cleanup k3d cluster
         if: always()
         run: |
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -98,7 +98,7 @@ test-e2e:
 	@echo "> Preparing charts (copying CRDs)..."
 	@$(MODULE_HACK_DIR)/prepare-charts.sh
 	@echo "> Running e2e tests..."
-	@cd e2e && go test -tags=e2e ./tests/... -v -timeout 30m
+	@cd e2e && go test -tags=e2e ./tests/... -v -timeout 45m
 
 # Make targets for local development and testing
 # -------------------------------------------------------------
--- a/operator/e2e/README.md
+++ b/operator/e2e/README.md
@@ -8,15 +8,22 @@ This directory contains the E2E test infrastructure for the Grove Operator, incl
 
 The following tools must be installed:
 - **Docker** - For running containers and k3d
-- **k3d** (v5.x) - For creating local Kubernetes clusters
 - **skaffold** (v2.x) - For deploying Grove operator
 - **helm** - For deploying Helm charts
 - **Go** (1.24.5+) - For running the tests
 
+The following tools are nice to have
+- **k3d** (v5.x) - For creating local Kubernetes clusters
+
 On macOS with Homebrew:
 ```bash
-brew install docker k3d skaffold helm go
+brew install docker skaffold helm go
 ```
+To install k3d
+```bash
+brew install k3d
+```
+
 
 ### Running Locally
 
@@ -35,7 +42,7 @@ The test suite will:
 1. Copy CRDs to Helm charts (via `prepare-charts.sh`)
 2. Create a k3d cluster with 28 worker nodes
 3. Install Grove, Kai Scheduler, and GPU Operator
-4. Run all gang scheduling tests (~10-15 minutes)
+4. Run all e2e testing suites
 5. Clean up the cluster
 
 ### Running in CI/CD
@@ -91,21 +98,12 @@ images:
 
 ## Troubleshooting
 
-### Docker Authentication Errors
-
-If you see errors like "Pull failed due to an unauthenticated request":
-```bash
-docker login
-```
-
-The e2e tests check for cached images before pulling, but some images may need to be pulled fresh.
-
 ### "No matches for kind PodCliqueSet" Error
 
 This means CRDs weren't copied to the Helm charts. The `make test-e2e` target automatically runs `prepare-charts.sh`, but if you're running tests manually:
 ```bash
 ./hack/prepare-charts.sh
-cd e2e && go test -tags=e2e ./tests/... -v
+cd e2e && go clean -testcache && go test -tags=e2e ./tests/... -v ---timeout 45m
 ```
 
 ### Stale k3d Cluster


operator/Makefile
```diff
@@ -92,6 +92,14 @@ test-envtest: $(SETUP_ENVTEST)
 	@echo "Running envtest with CRD validation..."
 	@KUBEBUILDER_ASSETS=$$($(SETUP_ENVTEST) use -p path) go test ./internal/webhook/admission/clustertopology/validation -v -run TestClusterTopologyCRDValidation
 
+# Run e2e tests
+.PHONY: test-e2e
+test-e2e:
+	@echo "> Preparing charts (copying CRDs)..."
+	@$(MODULE_HACK_DIR)/prepare-charts.sh
+	@echo "> Running e2e tests..."
+	@cd e2e && go test -tags=e2e ./tests/... -v -timeout 30m
```
danbar2: tests currently take ±26min, I suggest to increase the timeout (both here and on the whole github action)

operator/e2e/README.md
```diff
@@ -1,7 +1,55 @@
...
+```bash
+cd operator
+make test-e2e
+```
+
+The test suite will:
+1. Copy CRDs to Helm charts (via `prepare-charts.sh`)
+2. Create a k3d cluster with 28 worker nodes
+3. Install Grove, Kai Scheduler, and GPU Operator
+4. Run all gang scheduling tests (~10-15 minutes)
```
shayasoolin: Not only gang-sched tests, there are more suites - some of them already merged, the others are on the way.

operator/Makefile
shmuel-runai: fix

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,73 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: "1.24.5"
+
+      - name: Install k3d
```
gflarity, line 39: Is this truly necessary? I'm surprised because I run everything through the go modules directly. So all the dependencies should be installed by `go test`.... Same with skaffold below.  Please try running the tests without these.

operator/e2e/README.md
```diff
@@ -1,7 +1,55 @@
...
 
 This directory contains the E2E test infrastructure for the Grove Operator, including dependency management for external components.
 
+## Running E2E Tests
+
+### Prerequisites
+
+The following tools must be installed:
+- **Docker** - For running containers and k3d
+- **k3d** (v5.x) - For creating local Kubernetes clusters
```
gflarity: Again, `go test` should install the modules k3d and scaffold, and even helm. So these shouldn't be required installs unless I'm missing something.

operator/e2e/README.md
```diff
@@ -40,3 +88,36 @@ images:
   - name: docker.io/myorg/myimage
     version: v1.2.3
 ```
+
+## Troubleshooting
+
+### Docker Authentication Errors
+
+If you see errors like "Pull failed due to an unauthenticated request":
```
gflarity: ? Not sure what would require docker login.  In the past these errors were related to the registry not being running, things like that.

operator/e2e/README.md
gflarity: I guess  for development/debugging etc these might come in handy though...

operator/e2e/README.md
gflarity: Note that the E2E tests build the grove images locally and upload them to the local registry running in docker. Everything else should be public already.

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,73 @@
...
+          curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
+          k3d version
+
+      - name: Install skaffold
+        run: |
+          curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-amd64
+          sudo install skaffold /usr/local/bin/
+          skaffold version
+
+      - name: Log in to GitHub Container Registry
```
gflarity: This shouldn't be necessary. The E2E tests build the Grove images locally and upload them to the local registry running in docker.  They also modify the helm values to use these build-for-test images. Everything else should be public already.

.github/workflows/e2e-test.yaml
```diff
@@ -0,0 +1,73 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run e2e tests
+        run: make test-e2e
+
+      - name: Cleanup k3d cluster
```
gflarity, line 56: Not needed, the tests themselves do this. Also, the github runner will be deleted after the run if there anything still lying around due to a panic that isn't caught.

operator/e2e/README.md
```diff
@@ -1,7 +1,55 @@
...
+### Prerequisites
+
+The following tools must be installed:
+- **Docker** - For running containers and k3d
+- **k3d** (v5.x) - For creating local Kubernetes clusters
+- **skaffold** (v2.x) - For deploying Grove operator
+- **helm** - For deploying Helm charts
+- **Go** (1.24.5+) - For running the tests
+
+On macOS with Homebrew:
```
gflarity, line 18: For development/debugging only. The tests use go modules directly 🎉

operator/e2e/README.md
```diff
@@ -40,3 +88,36 @@ images:
...
+```
+
+The e2e tests check for cached images before pulling, but some images may need to be pulled fresh.
+
+### "No matches for kind PodCliqueSet" Error
+
+This means CRDs weren't copied to the Helm charts. The `make test-e2e` target automatically runs `prepare-charts.sh`, but if you're running tests manually:
+```bash
+./hack/prepare-charts.sh
+cd e2e && go test -tags=e2e ./tests/... -v
```
gflarity: I'd add -timeout 30m because I think we're getting over 20m now to run. Default timeout is 15m I think.

gflarity: I also add -count 1 to run the tests without caching. But this isn't as important as the timeout.

operator/e2e/README.md
shmuel-runai: fix

.github/workflows/e2e-test.yaml
shmuel-runai: Good catch! You're right that the k3d Go library is included in our `go.mod`, so the tests themselves don't need the k3d binary installed.
However, I'd like to keep the k3d installation for the following reasons:

1. Cleanup Safety Net
While our test code handles cleanup, this won't execute if the tests timeout or panic unexpectedly. The if: always() cleanup step ensures we delete the k3d cluster even in failure scenarios. 
Although GitHub Actions runners are ephemeral and will be destroyed anyway, explicitly cleaning up is good hygiene and helps with feature changes and unknown.


Regarding skaffold,I think that it is required
Unlike k3d, skaffold cannot be handled by Go modules because we invoke it as a binary via `exec.CommandContext(ctx, "skaffold", args...)`

Let me know if you'd prefer a different approach!

.github/workflows/e2e-test.yaml
shmuel-runai: removed

.github/workflows/e2e-test.yaml
shmuel-runai: See comment above, if we still thinks it's better to remove, I will

operator/e2e/README.md
shmuel-runai: removed

operator/e2e/README.md
shmuel-runai: added (to 45min we don't know how not powerfull the host) + go clean -testcache

operator/e2e/README.md
shmuel-runai: fix

.github/workflows/e2e-test.yaml
gflarity: I think it's fine to include them as dependencies in the README for debugging/manual clean up etc.  I'd like the test to fail if the automatic clean up fails (I think this is the case, but double check). If you want to leave manual clean up that's fine, but let's add comment in the workflow yaml explaining it's only for back up.

shmuel-runai: fix

gflarity (approved):

PR #287, opened 2025-12-09
Stablize E2E Tests

gflarity:
#### What type of PR is this?

BUG

Currently our E2E tests fail intermittently at different spots.  I found ~~two~~ three issues: 

1. The standard/free github runners only have 4cpu and 16G RAM. This sometimes causes non-deterministic timeouts.
2. There's no check that the validating webhook is running before we start the tests. This only seems to be an issue when there's not enough CPU and things are slower. 
3. 30 seconds just isn't a long enough for the default time out, bumped it to 60 to avoid needing to re-run tests due to random timeouts.

This PR fixes both of these issues. 

```release-note
NONE

```

```docs

```

### DIFF
--- a/operator/e2e/setup/k8s_clusters.go
+++ b/operator/e2e/setup/k8s_clusters.go
@@ -60,6 +60,25 @@ const (
 	defaultPollInterval = 5 * time.Second
 )
 
+// transientErrors contains error substrings that indicate a webhook is not ready yet
+var transientErrors = []string{
+	"no endpoints available",
+	"connection refused",
+	"i/o timeout",
+	"Bad Gateway",
+	"Service Unavailable",
+}
+
+// isTransient checks if an error string contains any transient error patterns
+func isTransient(errStr string) bool {
+	for _, s := range transientErrors {
+		if strings.Contains(errStr, s) {
+			return true
+		}
+	}
+	return false
+}
+
 // NodeTaint represents a Kubernetes node taint
 type NodeTaint struct {
 	Key    string
@@ -1072,11 +1091,7 @@ func waitForWebhookReady(ctx context.Context, restConfig *rest.Config, logger *u
 		if err != nil {
 			errStr := err.Error()
 			// These errors indicate the webhook is not ready yet
-			if strings.Contains(errStr, "no endpoints available") ||
-				strings.Contains(errStr, "connection refused") ||
-				strings.Contains(errStr, "i/o timeout") ||
-				strings.Contains(errStr, "Bad Gateway") ||
-				strings.Contains(errStr, "Service Unavailable") {
+			if isTransient(errStr) {
 				logger.Debugf("  Webhook not ready yet: %v", err)
 				return false, nil
 			}


operator/e2e/setup/k8s_clusters.go
```diff
@@ -1004,3 +1014,92 @@ func buildLDFlagsForE2E() string {
...
+	}
+
+	// Define the GVR for PodCliqueSet
+	pcsGVR := schema.GroupVersionResource{
+		Group:    "grove.io",
+		Version:  "v1alpha1",
+		Resource: "podcliquesets",
+	}
+
+	// Create a minimal PodCliqueSet for testing the webhook
```
shayasoolin: Can we have it as a separate yaml file? Or alternatively use one of the existing test yamls, if it makes sense?
This code with the deep nesting might be hard to read and maintain.

.github/workflows/e2e-test.yaml
```diff
@@ -25,12 +25,25 @@ jobs:
   e2e:
     # Run on non-draft PRs or draft PRs with 'run-e2e' label
     if: github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'run-e2e')
-    runs-on: ubuntu-latest
+    # use Nvidia hosted runner with 8 vCPUs and 30GB RAM
```
shayasoolin: Nvidia hosted?

.github/workflows/e2e-test.yaml
```diff
@@ -25,12 +25,25 @@ jobs:
...
+    runs-on: cpu-amd-m5-2xlarge
```
shmuel-runai, line 30: The comment should inform why the workload uses this spec (and not the default), not what the code does.
also +1 on Shay's comment on nvidia host

.github/workflows/e2e-test.yaml
```diff
@@ -25,12 +25,25 @@ jobs:
...
+      # print runner specs so we have a record incase of failures
+      - name: Print runner specs
+        run: |
+          echo "CPUs: $(nproc)"
+          echo "RAM: $(free -h | awk '/^Mem:/ {print $2}')"
+
       - name: Checkout code
         uses: actions/checkout@v4
 
+      # Nvidia hosted runners don't have make installed by default
```
shmuel-runai: same (nvidia)

operator/e2e/setup/k8s_clusters.go
```diff
@@ -1004,3 +1014,92 @@ func buildLDFlagsForE2E() string {
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	return utils.PollForCondition(ctx, defaultPollTimeout, defaultPollInterval, func() (bool, error) {
```
shmuel-runai, line 1080: For setup time, we would like a different (longer) poll timeout. (unlike testing, where everything should be somewhat ready and faster)

operator/e2e/setup/k8s_clusters.go
```diff
@@ -1004,3 +1014,92 @@ func buildLDFlagsForE2E() string {
...
+			testPCS,
+			metav1.CreateOptions{
+				DryRun: []string{metav1.DryRunAll},
+			},
+		)
+
+		if err != nil {
+			errStr := err.Error()
+			// These errors indicate the webhook is not ready yet
+			if strings.Contains(errStr, "no endpoints available") ||
```
shmuel-runai: nit
```
var transientErrors = []string{
    "no endpoints available",
    "connection refused",
    "i/o timeout",
    "Bad Gateway",
    "Service Unavailable",
}

func isTransient(errStr string) bool {
    for _, s := range transientErrors {
        if strings.Contains(errStr, s) {
            return true
        }
    }
    return false
}
```

.github/workflows/e2e-test.yaml
gflarity: Self hosted runners available to nvidia projects on github.  Technically this one is in AWS (all of them are I think).

.github/workflows/e2e-test.yaml
gflarity: I can make the comment more informative, don't quite follow the issue with saying Nvidia. I can just call them self hosted runners I guess.

.github/workflows/e2e-test.yaml
gflarity: oh, captialization? Sure. Never got the memo about that one though heh.

operator/e2e/setup/k8s_clusters.go
gflarity: Ok

.github/workflows/e2e-test.yaml
shayasoolin: Oh I see, sorry for the confusion. Yes I think self-hosted will be less ambiguous :)

.github/workflows/e2e-test.yaml
shayasoolin: Correct capitalization is NVIDIA but I don't think that's what he meant :)

PR #288, opened 2025-12-09
New TAS Design

Ronkahn21:
#### What type of PR is this?

  /kind documentation
  /kind feature

  This PR adds a comprehensive design document for topology-aware scheduling in the Grove operator. The design introduces a flexible topology system that enables optimal placement of multinode inference
  workloads based on cluster network topology.

  Fixes #

 
  suggest change

### DIFF
--- a/docs/designs/topology.md
+++ b/docs/designs/topology.md
@@ -394,7 +394,7 @@ For new workloads:
 
 For Existing Workloads:
 - If constraint references non-existent level:
-  - Remove required constraint for invalid level
+  - Remove required constraint for invalid level from PodGang
   - Update preferred constraint to new strictest level in updated ClusterTopology
   - Update PodCliqueSet status with constraint removal reason
 - If constraint still valid:
@@ -651,7 +651,7 @@ The operator translates user's level names to keys and builds required/preferred
 **Preferred Constraints (Auto-Generated):**
 
 - Operator ALWAYS generates preferred constraint at all three levels
-- Uses key of strictest (narrowest) level configured in ClusterTopology
+- Uses key of strictest (narrowest) level configured in ClusterTopology (for example if the ClusterTopology is host, rack, zone use "host" level, `"kubernetes.io/hostname"`)
 - Example: If levels include "host", preferred = `"kubernetes.io/hostname"`
 - Writes to PodGang: `TopologyConstraint.PackConstraint.Preferred = "kubernetes.io/hostname"`
 - Enables out-of-box optimization even without user configuration


docs/designs/topology.md
```diff
@@ -17,8 +17,8 @@ applications require:
 
 - Provide flexible, cluster-agnostic topology hierarchy definition via ClusterTopology CRD
 - Enable packing constraints for network locality across all Grove scalable resources
-- Immutable topology configuration ensuring scheduling consistency
-- Hierarchical constraint validation (child stricter than parent)
+- Mutable topology configuration allowing runtime updates
```
shayasoolin: This sounds contradicting to the below non-goal of "Dynamic topology reconfiguration after creation", please fix/clarify.

docs/designs/topology.md
```diff
@@ -17,8 +17,8 @@ applications require:
...
+- Flexible topology level ordering without enforced hierarchy
 
 ## Non-Goals
```
shayasoolin, line 25: See below - "Grove implements topology-aware scheduling through a ClusterTopology CRD" - needs to be updated, it's no longer exactly the case as this CRD is internal while the user is working with the operator configuration not the CRD itself.

docs/designs/topology.md
```diff
@@ -17,8 +17,8 @@ applications require:
 
 - Provide flexible, cluster-agnostic topology hierarchy definition via ClusterTopology CRD
 - Enable packing constraints for network locality across all Grove scalable resources
-- Immutable topology configuration ensuring scheduling consistency
-- Hierarchical constraint validation (child stricter than parent)
+- Mutable topology configuration allowing runtime updates
+- Flexible topology level ordering without enforced hierarchy
```
shayasoolin, line 23: Another goal is the creation of the downstream topology CR that is required by the scheduler.

docs/designs/topology.md
```diff
@@ -44,17 +44,25 @@ while allowing users to specify required constraints for strict placement (upper
...
+│  │ ClusterTopology      │          │ KAI Topology         │             │
+│  │ "grove-topology"     │─────────▶│ "grove-topology"     │             │
+│  │ (operator-managed)   │  Manage  │ (operator-managed)   │             │
+│  └──────────┬───────────┘          └────────────┬─────────┘             │
 │             │                                   │                       │
-│             │                                   │                       │
-│  Operator Config: clusterTopology.enabled=true  │                       |                       |
-│             │                                   │                       │
-│             │ (validates against)               │ (referenced by)       │
+│             │ (validates against)               │ (used by)             │
```
shayasoolin: KAI topology is not used by grove operator but by KAI scheduler (should fix the diagram).
It's created by grove-operator.

docs/designs/topology.md
```diff
@@ -122,52 +130,95 @@ const (
...
+    ConditionReasonKAITopologyCreationFailed = "KAITopologyCreationOrUpdateFailed"
+)
 ```
 
+**Ready Condition Semantics:**
+
+- **Condition Type**: `Ready`
+- **Status Values**:
+  - `True` = ClusterTopology is ready and KAI Topology CR successfully created
+  - `False` = ClusterTopology has errors or KAI Topology creation failed
```
shayasoolin: "ClusterTopology has errors" - can you give an example? If indeed there's such case, let's also add a dedicated Reason for that.

docs/designs/topology.md
```diff
@@ -186,128 +237,169 @@ spec:
...
+**Configuring ClusterTopology:**
+
+The Manged ClusterTopology CR is generated and managed sole by the Grove operator. To configure it:
+
+1. Define topology levels in OperatorConfiguration under `clusterTopology.levels` (see OperatorConfiguration section below)
+2. Set `clusterTopology.enabled: true` in OperatorConfiguration
+3. Restart the Grove operator
+4. Operator creates ClusterTopology CR named "grove-topology"
+5. Operator creates and continuously reconciles KAI Topology CR
+6. If configuration is invalid or CR creation fails → operator exits with error
```
shayasoolin: Be more specific - it's only the ClusterTopology CR creation that causes the operator to fail. The KAI topology CR creation doesn't cause it to fail, but just to report error status.

shayasoolin: I now see you elaborate about it later below. Still please just clarify which CR you refer to in this bullet item.

docs/designs/topology.md
```diff
@@ -186,128 +237,169 @@ spec:
...
-    - `name` not specified → defaults to "grove-topology"
-  - Operator looks for ClusterTopology with configured name (defaults to "grove-topology")
-  - If ClusterTopology with that name doesn't exist → operator exits with ENOENT and crashloops
+  - Operator generates ClusterTopology CR named "grove-topology"
+  - Operator validates topology config at startup
+  - If config invalid (duplicate domains, invalid keys, etc.) → operator exits with error
+  - If ClusterTopology CR creation fails → operator exits with error
+  - Operator generates KAI Topology CR
+  - If KAI Topology creation fails → reflected in ClusterTopology Ready condition and LastErrors (not operator failure)
 - If `clusterTopology.enabled: false`: topology features disabled
```
shayasoolin: Should specify here that the operator deletes the ClusterToplogy CR

docs/designs/topology.md
```diff
@@ -186,128 +237,169 @@ spec:
...
-3. Admin restarts operator (see Startup Behavior section for details)
-4. Operator validates ClusterTopology CR exists
-5. For existing workloads:
-    - Workloads with topology constraints: operator uses topology from PodCliqueSet label (`grove.io/cluster-topology-name`)
-    - Workloads without topology constraints: no impact (checked via label presence)
-6. For new workloads:
-    - Topology constraints validated against ClusterTopology CR
-   - Mutation webhook adds `grove.io/cluster-topology-name` label to PodCliqueSet (see Mutation Webhook section)
+***For existing workloads:***
+* operator validates constraints, removes invalid ones, updates status
```
shayasoolin: Please clarify what's "invalid ones" and a bit about this "status update"

docs/designs/topology.md
```diff
@@ -186,128 +237,169 @@ spec:
...
-*note: in the future, we may support dynamic updates to ClusterTopology without disabling topology first.*
+For New Workloads:
+- Validation webhook rejects workloads with invalid constraints
+- Error message indicates which constraint is invalid
+- Users must update workload spec to match available topology levels
+
+Preferred Constraint Updates:
+- When strictest (narrowest) topology level changes (e.g., host → numa)
+- Operator updates preferred constraint to new strictest level
+- Applies to all three levels (PodGang, TopologyConstraintGroup, PodGroup)
```
shayasoolin: What are these 3 levels? This is the first time I read about them, looks either wrong or misplaced.

docs/designs/topology.md
```diff
@@ -528,26 +613,18 @@ Fields Added:
...
 
-**Topology Annotation:**
+**Scheduler Topology Discovery:**
 
-- Operator adds annotation `grove.io/topology-name: "<topology-name>"` to PodGang metadata
-- Annotation value matches the ClusterTopology name from operator configuration
-- KAI scheduler uses this annotation to locate the corresponding Kueue Topology CRD
-- Annotation approach provides API flexibility for future changes without breaking spec
+- KAI will use topology annotation on the podGroup to allow KAI discover which KAI topology is used.
+- The annotation key will be `grove.io/topology-name` and the value will be the name of the ClusterTopology resource created by Grove operator, which is `grove-topology`.
```
shayasoolin: I'm missing something here - I thought that once Grove translates the levels to topology keys in PodGang, the grove topology is not longer relevant. KAI only needs to know those keys plus the KAI topology name.
Why do we store the "name of the ClusterTopology resource created by Grove operator" as part of the pod gang/group?

shayasoolin (commented):
Looks good!

docs/designs/topology.md
```diff
@@ -17,8 +17,8 @@ applications require:
 
 - Provide flexible, cluster-agnostic topology hierarchy definition via ClusterTopology CRD
 - Enable packing constraints for network locality across all Grove scalable resources
-- Immutable topology configuration ensuring scheduling consistency
-- Hierarchical constraint validation (child stricter than parent)
+- Mutable topology configuration allowing runtime updates
+- Flexible topology level ordering without enforced hierarchy
```
sanjaychatterjee: This need not be a goal.
I would rather add the following as goals: 
1. Grove will define a standard topology model in the Grove API that is uniform across all cloud and on-prem clusters.
2. The mapping from the Grove standard topology model to cluster specific node labels will be provided by the cluster admin.
3. User workloads will be able to define topology constraints on a Grove workload at PCS, PCSG and PCLQ levels.
4. Grove operator will translate the user defined topology constraints into cluster specific scheduling constraints in the scheduler api.
5. Grove will be responsible for generating the appropriate topology CR for any downstream scheduler.
6. Initial implementation will focus on packing constraints.

docs/designs/topology.md
```diff
@@ -83,18 +89,20 @@ while allowing users to specify required constraints for strict placement (upper
...
 
-- **Cluster-scoped resource**: Can have multiple ClusterTopology resources defined
-- **Default name**: In operator configuration, clusterTopologyName defaults to "grove-topology" when not specified
-- **Partially immutable**: After creation, only `key` field values can be updated; `domain` fields and level ordering
-  are immutable
-- **List-ordered hierarchy**: Index 0 represents the broadest category (e.g., region), and the final index represents the narrowest (e.g., host).
-- **Supported topology levels**: Region > Zone > DataCenter > Block > Rack > Host > Numa (broadest to narrowest)
-- **Webhook-validated**: Webhook validates constraint hierarchy and immutability
+- **Cluster-scoped resource**: Only one ClusterTopology resource managed by operator: "grove-topology"
+- **Operator-managed resource**: Created and managed by Grove operator based on OperatorConfiguration
```
sanjaychatterjee, line 113: We can merge the first two points as:
- **Grove-managed ClusterTopology CR**: Grove will create and manage the lifecycle of one ClusterTopology CR with a reserved name called "grove-topology". This contents of this managed CR will be generated from Grove OperatorConfig values.

docs/designs/topology.md
```diff
@@ -186,128 +237,169 @@ spec:
...
+- Note: This validates OperatorConfiguration topology config, not ClusterTopology CR
+- Note: ClusterTopology CR has separate validation layers (API server + webhook - see Validation section above)
 
 **Admin Responsibilities:**
 
-- Manually create Kueue Topology with same name as Grove ClusterTopology for KAI scheduler
-- Ensure topology levels align between Grove ClusterTopology and Kueue Topology
+- Define topology levels in OperatorConfiguration
+- Restart operator when changing topology configuration
+- Grove operator automatically creates and manages both ClusterTopology and KAI Topology CRs
```
sanjaychatterjee, line 366: nit: Last point is not admin responsibility, so can be omitted.

docs/designs/topology.md
```diff
@@ -186,128 +237,169 @@ spec:
...
+- Keys must be valid Kubernetes label keys
+- If validation fails → operator exits with descriptive error
+- Note: This validates OperatorConfiguration topology config, not ClusterTopology CR
+- Note: ClusterTopology CR has separate validation layers (API server + webhook - see Validation section above)
 
 **Admin Responsibilities:**
 
-- Manually create Kueue Topology with same name as Grove ClusterTopology for KAI scheduler
-- Ensure topology levels align between Grove ClusterTopology and Kueue Topology
+- Define topology levels in OperatorConfiguration
```
sanjaychatterjee, line 364: Also add one more point in admin responsibility to:
- Enable/disable topology-aware scheduling through `clusterTopology.enabled` flag.

docs/designs/topology.md
sanjaychatterjee: @shayasoolin the scheduler API has 3 types of topology constraints, one at the PodGang level, one for a group of PodGroups that represent a PCSG and one for a PodGroup.
@Ronkahn21 maybe we can rephrase this as:
```
- Applies uniformly to all topology constraints defined in the scheduler API
```

docs/designs/topology.md
shmuel-runai: But the first goal is (line 18)

> Provide flexible, cluster-agnostic topology hierarchy definition via ClusterTopology CRD

docs/designs/topology.md
Ronkahn21: KAI required the topology name we are refers to as INPUT, as KAI required the name, Its better to have annotation with the name rather then KAI assume using some name, in the future we will have more than one topology or allow diffrent name. I dont want to be really on scheduler side logic, we should tell the down stream scheduler what is the name

docs/designs/topology.md
```diff
@@ -15,25 +15,28 @@ applications require:
...
 
 ## Non-Goals
 
-- Spread constraints across topology domains (ReplicaSpreadDomain)
-- Root domain constraints for entire resource (RootDomain)
-- Ratio-based affinity groups between scaling groups (AffinityGroups with PackRatio)
-- Dynamic topology reconfiguration after creation
-- Automatic suggest topology according to workload characteristics
+- Spread constraints across topology domains (anti-packing)
+- Root domain constraints for entire workloads
```
gflarity, line 27: What is meant by root domain constraints for entire workload?  Doesn't packDomain on a PCS do precisely this? Example,  schedule a PCS replica onto a single rack?

docs/designs/topology.md
```diff
@@ -83,18 +92,20 @@ while allowing users to specify required constraints for strict placement (upper
...
-- **Cluster-scoped resource**: Can have multiple ClusterTopology resources defined
-- **Default name**: In operator configuration, clusterTopologyName defaults to "grove-topology" when not specified
-- **Partially immutable**: After creation, only `key` field values can be updated; `domain` fields and level ordering
-  are immutable
-- **List-ordered hierarchy**: Index 0 represents the broadest category (e.g., region), and the final index represents the narrowest (e.g., host).
-- **Supported topology levels**: Region > Zone > DataCenter > Block > Rack > Host > Numa (broadest to narrowest)
-- **Webhook-validated**: Webhook validates constraint hierarchy and immutability
+- **Cluster-scoped resource**: Only one ClusterTopology resource managed by operator: "grove-topology"
+- **Operator-managed resource**: Created and managed by Grove operator based on OperatorConfiguration
+- **User-created CRs allowed**: Users can create additional ClusterTopology CRs but Grove does not use them
```
gflarity, line 114: Ok, but is there a reason why we allow it? Seems like just preventing it might be safer?

docs/designs/topology.md
```diff
@@ -83,18 +92,20 @@ while allowing users to specify required constraints for strict placement (upper
...
+- **Fixed name for managed CR**: Operator-managed CR always named "grove-topology"
+- **Mutability**:
+  - **Operator-managed CR** ("grove-topology"): Only modified by operator at startup (not runtime)
+  - **User-created CRs**: Fully mutable - can be modified anytime
+- **Flexible ordering**: Levels can be specified in any order - no hierarchical ordering enforced in OperatorConfiguration
```
gflarity: I think you just mean the order in which labels are assigned to topology levels within the config file does not matter. The hierarchical ordering is already defined:

Region, Zone, DataCenter, Block, Rack, Host, Numa

If so consider wording similar to this instead:

```suggestion
- **Order-Independent Configuration**: The mapping of topology domains to kubernetes labels, in order to define levels, can be specified in any order - no hierarchical ordering enforced in OperatorConfiguration
```

docs/designs/topology.md
```diff
@@ -83,18 +92,20 @@ while allowing users to specify required constraints for strict placement (upper
...
+- **Supported topology levels**: Region, Zone, DataCenter, Block, Rack, Host, Numa
+- **Webhook-validated**: Webhook validates domain/key uniqueness, key format, and authorization (only operator can modify "grove-topology")
 
 **TopologyDomain Definitions:**
```
gflarity: Nit: Consider moving this before the characteristics for someone who reviews this after the fact. I think the reader should understand this before characteristics are defined.

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
-  topology is enabled)
+**A. Webhook Validation**
 
-**Mutation Webhook:**
+The validation webhook enforces business logic constraints:
 
-The mutation webhook automatically modifies ClusterTopology resources on CREATE operations:
+- **On CREATE**: Validates domain uniqueness, key uniqueness, and key format
+- **On UPDATE**: Runs same CREATE validations (uniqueness, format)
+- **Authorization**: Only operator service account can CREATE/UPDATE/DELETE the Manged ClusterTopology
```
gflarity:
```suggestion
- **Authorization**: Only operator service account can CREATE/UPDATE/DELETE the Managed ClusterTopology with the resource Name grove-topology
```

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
+- **Important**: No hierarchical order enforcement - levels can be specified in any order
```
gflarity:
```suggestion
- **Important**: No hierarchical order enforcement - levels can be specified in any order. The order is already defined (Region, Zone, DataCenter, Block, Rack, Host, Numa)
```

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
+- **Important**: Resource is fully mutable - all fields can be updated after creation
```
gflarity: Can you talk a bit more about this? What the expected mutation situations?

docs/designs/topology.md
gflarity: Or actually, looks like we describe this below. Maybe this is just duplicate and we should remove it favour of the section below?

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
-- **Subset Support**: Admins can define any subset of levels; webhook reorders only the levels provided
-- **Consistency**: Ensures all ClusterTopology resources follow the same ordering convention across the cluster
-- **Example**: If admin defines levels as `[host, rack, region]`, webhook reorders to `[region, rack, host]`
+
+**Authorization Validation**
+
+Validation webhook ensures only operator service account can modify operator-managed ClusterTopology ("grove-topology"):
+- Operator service account: Can CREATE/UPDATE/DELETE "grove-topology"
+- User service accounts: Cannot modify "grove-topology" (rejected by webhook)
+- User service accounts: CAN create/modify/delete their own ClusterTopology CRs (Grove ignores them)
```
gflarity, line 274: Again, I'm confused why we might allow this? Assuming there is a specific reasons, let's call it out explicitly.  Otherwise just keep these locked down?

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
+
+*Note: Webhooks unavailable during operator downtime, but not a concern for operator-managed workflow.*
 
 #### ClusterTopology Controller
 
-The ClusterTopology controller manages the ClusterTopology resource lifecycle:
+The ClusterTopology controller manages the ClusterTopology resource lifecycle and KAI Topology synchronization.
```
gflarity, line 280: I'm curious how the KAI Topology CR differs?  What do we gain with a separate Grove CR currently (ie with no other schedulers currently viable?)

Does it make sense to just create the KAI CR for now, and then when more schedulers are viable make it possible to specify which Topology CR we produce via a config setting?

Just curious, seems like it might be less work to just do that.

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
+**KAI Topology CR Generation and Reconciliation**
 
-**Deletion Protection**
+The controller continuously reconciles KAI Topology CR to keep it synchronized with ClusterTopology:
 
-Prevents ClusterTopology deletion while in use, requiring both conditions for deletion using Kubernetes finalizer.
+- On ClusterTopology creation → Create corresponding KAI Topology CR
+- On ClusterTopology update → Update KAI Topology CR to match
+- On reconciliation → Verify KAI Topology exists and matches ClusterTopology spec
+- Name: Always matches ClusterTopology name (e.g., "grove-topology" for managed CR)
```
gflarity, line 289: Is there any constraints on who can edit the KAI Topology CR?

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
-3. Admin restarts operator (see Startup Behavior section for details)
-4. Operator validates ClusterTopology CR exists
-5. For existing workloads:
-    - Workloads with topology constraints: operator uses topology from PodCliqueSet label (`grove.io/cluster-topology-name`)
-    - Workloads without topology constraints: no impact (checked via label presence)
-6. For new workloads:
-    - Topology constraints validated against ClusterTopology CR
-   - Mutation webhook adds `grove.io/cluster-topology-name` label to PodCliqueSet (see Mutation Webhook section)
+***For existing workloads:***
+* operator validates constraints, removes constraints referencing non-existent topology levels from PodGang, updates PodCliqueSet status to indicate constraints are ignored
```
gflarity, line 373: As per our discussion around this during the requirements phase, can we more explicitly mention putting warnings in the logs as well.  Some warning events on the the effected resources seem like a good idea also given that describing resources and reading through events is a common workflow for SREs.

docs/designs/topology.md
```diff
@@ -379,38 +460,84 @@ type PodCliqueTemplateSpec struct {
...
+```yaml
+status:
+  conditions:
+  - type: TopologyConstraintSatisfied
+    status: "False"
+    observedGeneration: 5
+    lastTransitionTime: "2025-12-08T10:10:00Z"
+    reason: TopologyLevelNotFound
+    message: "Topology level 'block' not found in ClusterTopology 'grove-topology'. Remove packDomain or update ClusterTopology."
+```
```
gflarity: I believe this status is just meant to inform that the constraint is being ignored due to the non-existent Topology level. If so let's mention that, something like:

```suggestion

```yaml
status:
  conditions:
  - type: TopologyConstraintSatisfied
    status: "False"
    observedGeneration: 5
    lastTransitionTime: "2025-12-08T10:10:00Z"
    reason: TopologyLevelNotFound
    message: "Topology level 'block' not found in ClusterTopology 'grove-topology'.  Topology constraint will be ignored. Please remove packDomain or update ClusterTopology."
```
```

File: docs/designs/topology.md [hunk 27]
```diff
@@ -186,128 +240,170 @@ spec:
...
 
-1. Admin disables topology per above workflow
-2. Restart operator
-3. Admin edits ClusterTopology CR (only `key` field values can be changed; `domain` fields immutable)
-4. Admin enables topology per above workflow
-5. Restart operator
-6. Operator reconciles existing workloads with updated topology configuration
+For Existing Workloads:
+- If constraint references non-existent level:
+  - Remove required constraint for invalid level
```
gflarity: Remove constraint? From the PodClique etc, or the PodGang?  See note below, I thought were just ignoring these  at the PCS,PCSG, PCLQ levels?

docs/designs/topology.md
gflarity: I see down below we mention this: 

Validation uses semantic order, NOT OperatorConfiguration list index. We should maybe just throw this information near the very top. Makes sense.

docs/designs/topology.md
```diff
@@ -563,11 +645,21 @@ The operator translates user's level names to keys and builds required/preferred
 **Preferred Constraints (Auto-Generated):**
 
 - Operator ALWAYS generates preferred constraint at all three levels
-- Uses key of strictest level (e.g., `"kubernetes.io/hostname"` for "host" level)
+- Uses key of strictest (narrowest) level configured in ClusterTopology
```
gflarity: Can we include a concrete example for prefered constraints. I think I get it but I think it would be helpful as a future reference for others.

docs/designs/topology.md
```diff
@@ -563,11 +645,21 @@ The operator translates user's level names to keys and builds required/preferred
...
+- Example: If levels include "host", preferred = `"kubernetes.io/hostname"`
 - Writes to PodGang: `TopologyConstraint.PackConstraint.Preferred = "kubernetes.io/hostname"`
 - Enables out-of-box optimization even without user configuration
 - Scheduler can fallback to less strict levels if preferred cannot be satisfied
 
+**Edge Cases:**
+
+1 **Strictest Level Changes During Topology Update**:
+   - Initial topology: rack, host (strictest = host)
+   - Updated topology: rack, host, numa (strictest = numa)
```
gflarity, line 664: Hmm, just realized I don't completely understand the numa domain. Maybe we should just define all the domains and their hierarchy/order near the top for reference?

docs/designs/topology.md
```diff
@@ -563,11 +645,21 @@ The operator translates user's level names to keys and builds required/preferred
...
+   - Operator updates preferred constraints to "topology.kubernetes.io/numa" for all affected PodGangs
```
gflarity, line 665: A full concrete example would be great.

gflarity (changes requested):
Nothing major, just a bunch of clarifications requested along with some suggestions. 

I still need to review from the Security and RBAC section and onward, but I've run out of time today.  So I'll submit this review now and finish it off tomorrow.

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
-  - If ClusterTopology with that name doesn't exist → operator exits with ENOENT and crashloops
-- If `clusterTopology.enabled: false`: topology features disabled
-- Admin must create ClusterTopology with matching name OR disable topology in operator config
+  - Operator generates ClusterTopology CR named "grove-topology"
+  - Operator validates topology config at startup
+  - If config invalid (duplicate domains, invalid keys, etc.) → operator exits with error
+  - If ClusterTopology CR creation fails → operator exits with error
+  - Operator generates KAI Topology CR
+  - If KAI Topology creation fails → reflected in ClusterTopology Ready condition and LastErrors (not operator failure)
+- If `clusterTopology.enabled: false`: topology features disabled, ClusterTopology CR remains (operator does not modify or delete it)
```
sanjaychatterjee: Can you please clarify why ClusterTopology CR should remain if `clusterTopology.enabled` is set to `false`? Would it be cleaner to delete it than keep it lying around?

docs/designs/topology.md
shayasoolin, line 280: Either here or earlier in this document, I think you should mention that as of today, "KAI topology CRD" means "Kueue topology". I understand and agree that you leave it as an implementation detail, but at least once in the document it should be specified in order to clarify what it is about exactly.
Otherwise it might look as if we're creating some unique KAI CRD and then KAI generates the Kueue CRD.

docs/designs/topology.md
```diff
@@ -186,128 +240,170 @@ spec:
...
+**OperatorConfiguration API Structure:**
+
+```go
+type OperatorConfiguration struct {
+    // Topology configuration for cluster topology hierarchy
+    // +optional
+    ClusterTopology *TopologyConfiguration `json:"clusterTopology,omitempty"`
+}
+
+type TopologyConfiguration struct {
```
shayasoolin:
```suggestion
type ClusterTopologyConfiguration struct {
```

docs/designs/topology.md
shayasoolin: I also think it should be deleted. It's an operator-managed CR. If the operator doesn't need it, it should delete it (as it created it).

docs/designs/topology.md
```diff
@@ -528,26 +617,19 @@ Fields Added:
...
 #### Translation Logic
 
-The operator translates Grove operator API to Grove Scheduler API with three-level topology constraint hierarchy:
+The operator translates Grove operator API to Grove Scheduler API with three-level topology constraint hierarchy.
+
+**Scheduler Topology Discovery:**
 
-**Topology Annotation:**
+- KAI will use topology annotation on the podGroup to allow KAI discover which KAI topology is used.
+- The annotation key will be `grove.io/topology-name` and the value will be the name of the KAI Topology CR created by Grove operator, which is `grove-topology`.
```
shayasoolin, line 634: I'm not sure this is the right decision here. It feels like a too tight coupling between Grove and KAI. 
An alternative approach could be that KAI is just configured with the topology name (e.g. "grove-topology") by the admins, who know that this is the name of the toplogy that Grove creates.
The advantage here is that if the admins want to start KAI with a different topology, they can, and the PodGang CRs will still be effective - no annotation changes required - of course as long as the domain levels are aligned, i.e. included in the new topology CR.
Let's discuss if we really wanna have this annotation or can go without it.

shayasoolin (commented):
Looks good, added a few minor comment and one item for discussion - on the necessity of the topology name annotation for KAI.

docs/designs/topology.md
Ronkahn21: Kai required the topology you all want to use as an input.  You have to give it as an input for KAI.
So you have two options here. Either put it as a spec or an annotation. I prefer to go with annotation first because if we want to change it in the future it will be easy because it's not an API.
but either options the information must pass as some arg to KAI pod grouper or in the future any downstream integraion

docs/designs/topology.md
Ronkahn21: We didn't want on adding more validation logic, Basically because the Grove does not look at them, it does not affect anything. So we could do this validation, but it was part of the requirement that user can create whatever they want. And we don't really care.

docs/designs/topology.md
Ronkahn21: From PogGang it gets removed and from PCS its get a status info, I will try to make more clear

docs/designs/topology.md
Ronkahn21: KAI topology CR is basically the same as kueu but they decided they want to have their own CR for now, either way. We don't want to align with them as our purpose are different from KAI. We wanted to have a nice names map to Labels. This allows us to create more functionality in the future. Also we don't want to be relying on KAI if they are changing. We already know that supporting other schedulers will be in the near future. It makes perfect sense that we have our own topology CR that is perfect for our situation. We create any downstream topology CR once we have more integration, but for now it is KAI.

docs/designs/topology.md
Ronkahn21: As this operator, we enforcing the changes upon the CR,
BC its not our CRD, its not a good practice to have a webhook for them

Ronkahn21: so even if some one changes the CR we will change it back

docs/designs/topology.md
Ronkahn21: Logs are meant for admin and for debug when somethings wrongs but this by design, we will add logs. but I am not expecting the users to log the operator, its not a good practice, if the users really on logs its means we done something wrong. also users might not have RBAC to log the operator at all.

Events are good and we will add them, but the best way is to update the status

docs/designs/topology.md
Ronkahn21: Ok, I will change this, but failing to delete the CT CR should not consider as fatal error, just log and event

docs/designs/topology.md
Ronkahn21: I made it clearer

docs/designs/topology.md
Ronkahn21: what is missing, I added this example and it make sense to me.
what should the full example should look like ?

gflarity: A grove-topology YAML along with a PodCliqueSet YAML with constraints referencing it. Then finally  PodGang YAML the required and preferred values set.  Just makes it very easy to understand.  It doesn't have to go in this GREP necessarily, but we should have and end to end example in the documentation somewhere.

shayasoolin (approved):

PR #296, opened 2025-12-23
add mnnvl requirements GREP file

danbar2:
#### What type of PR is this?

/kind documentation

#### Does this PR introduce a API change?

```release-note
NONE
```

#### Additional documentation e.g., enhancement proposals, usage docs, etc.:

```docs
GREP-270: mnnvl requirements
```

PR #297, opened 2025-12-23
MNNVL support design doc

shmuel-runai:
MNNVL support documentation

#### What type of PR is this?

/kind documentation
/kind feature


docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+config:
+	mnnvl:
+		enabled: false
+```
+
+### Feature validity 
+
+When the Grove operator starts, it will check for MNNVL support (only if the feature is enabled). Support is confirmed by the presence of the ComputeDomain CRD in the cluster.
+
+If the cluster lacks MNNVL support, the Grover operator will terminate and log an appropriate error.
```
julienmancuso: Grover -> Grove

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Creation
+
+The PodCliqueSet (PCS) controller determines if a ComputeDomain is necessary during reconciliation, based on these criteria:
+
+1. The feature is activated within Grove’s `OperatorConfiguration`.  
+2. The PCS has not explicitly opted out of the feature.  
+3. At least one contained PodClique (PC) requires a GPU.
+
+The controller will create a ComputeDomain resource for each PCS replica if one does not already exist. This resource will be named `{pcs-name}-cd-{replica-index}` and will have the PCS set as its owner reference.
```
julienmancuso: To remain consistent with existing named resources it should only be `{pcs-name}-{replica-index}` : currently other resources (like podcliques , pcsg, ..) doesn't include the resource type in their name

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+* Ties the feature to Grove.
+
+### Watching the compute domain
+
+To ensure the PCS controller reconciles when a `ComputeDomain` is deleted or its status changes, a watch must be added in the PCS controller's RegisterWithManager function. 
+
+When the controller creates a `ComputeDomain` resource, it should attach the label `grove.io/podcliqueset` with the value `<pcs-name>` to identify which PCS owns it. 
+
+As part of the registration process, a new watch will be used to garud the `ComputeDomain` created by the controller (using the `grove.io/podcliqueset`label as a filter).
```
julienmancuso: garud -> guard

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+metadata:
+ labels:
+   grove.io/podcliqueset: <pcs-name>
+```
+
+### Scale-Out and Scale-In
+
+When scaling out (replicas increased), the subsequent reconciliation process will identify the ComputeDomains missing for the new replica indices and create them using the identical logic as the initial creation.
+
+When scaling in( replicas decreased), the PCS controller determines which `ComputeDomains` to delete by comparing the existing ones against the new replica count. Any `ComputeDomain` with a replica index equal to or greater than the new count is removed. This process mirrors the "desired versus existing" logic used for managing `PodClique` and `PCSG`.
```
julienmancuso: 'in( ' -> 'in ('

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+      reason: Ready               # or "Timeout" / "AllocationFailed"
+      message: "ComputeDomain my-pcs-replica-0 is ready"
+      lastTransitionTime: "2024-01-15T10:00:00Z"
+```
+
+When the `PodClique` controller reconciles pods, the Pod component checks its parent PodClique's ComputeDomainReady condition. If the condition is True, the Pod is created with a resourceClaims entry that references the ComputeDomain's ResourceClaimTemplate (RCT). 
+
+If the condition is False, the Pod is created without the RCT reference, allowing the workload to run on any available GPUs (without the NVLink optimization).
+
+The RCT name follows a deterministic naming convention based on the PCS name and replica index: {pcs-name}-replica-{index}-rct. This allows the Pod component to derive the correct RCT name without querying the ComputeDomain directly.
```
julienmancuso: same thing here for the name, we don't need to add `rct`

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Overview
+
+This design document details the plan to enable the Grove operator to automatically leverage MNNVL for appropriate workloads.
+
+## Motivation
+
+The MNNVL support feature in Grove is guided by three core design principles:
+
+**Simplicity:** Deploying inference workloads with MNNVL in Kubernetes currently demands familiarity with complex concepts like ComputeDomain CRDs, ResourceClaimTemplates, and Dynamic Resource Allocation (DRA) APIs. Grove, however, must completely abstract this complexity, minimizing required configuration and hiding all underlying implementation details from the user.
```
unmarshall: >ResourceClaimTemplates, and Dynamic Resource Allocation (DRA) APIs

`ResourceClaimTemplate` is part of DRA API so why mention that explicitly?

>Grove, however, must completely abstract this complexity, minimizing required configuration and hiding all underlying implementation details from the user.

We should phrase it differently. Please be aware that DRA APIs which are going to be used by the workload operator are `ResourceClaim` & `ResourceClaimTemplate`. Now these are k8s native APIs and if people are wanting to optimize scheduling of their workloads then they need to have working knowledge of these APIs. Therefore there are 2 sets of users:
* Power users: These are the users that are familiar with k8s in general and DRA in particular and would like to more control over over how DRA has been configured.
* Normal users: These are users who are experts in designing AI/ML/HPC workloads but find DRA + device specific optimizations (e.g. ComputeDomain) difficult to configure.

What this proposal in its first does is to make life simpler for `Normal` users only.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+**Portability:** The Grove Custom Resource Definitions (CRDs)—specifically PodCliqueSet, PodCliqueScalingGroup, and PodClique—are designed to be generic, excluding any direct references to MNNVL, ComputeDomain, or other NVIDIA-specific components.
+
+To maintain portability, all MNNVL-related configuration is kept external, residing in the operator configuration and runtime annotations. This approach allows the same PodCliqueSet manifest to be used without alteration, regardless of whether the target cluster supports MNNVL.
```
unmarshall, line 27: >To maintain portability, all MNNVL-related configuration is kept external, residing in the operator configuration and runtime annotations.

`OperatorConfiguration` for Grove operator is also API which is going to be used by a cluster admin who is setting up Grove. For now it is fine to keep it under operator configuration, but in the end we should make device specific configurations/optimizations pluggable and keep Grove as a whole independent of them.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+**Failure Resilience:** MNNVL is treated as an optimization, not a hard requirement. If encountering a failure creating the MNNVL resources, Grove does not block the workload. Instead, pods are created without the RCT reference and are scheduled using standard GPU allocation mechanisms.
```
unmarshall: Creating PCS pods while ignoring the errors during creation of MNNVL resources is ideally a degradation of a functionality when this has explicitly been enabled. I think this decision has also been taken because we enable this on Grove operator level and do not provide this as part of an individual PCS since we are only supporting homogeneous clusters in this phase.

So mentioning the above is not really `failure resilience`.

What i do not see in this document as a section on `Limitations` or `Next Steps` detailing out the choices/constraints with which the initial version of this feature is going to be developed and the consequences of such choices and in brief mentioning the next steps.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Using MNNVL in a K8S cluster
+
+To use **MNNVL (Multi-Node NVLink)** in Kubernetes with NVIDIA’s DRA driver, you start by creating a **ComputeDomain (CD)**. The ComputeDomain represents a logical GPU fabric spanning multiple nodes and defines how inter-node NVLink/IMEX channels are allocated. In its `spec.channel`, the ComputeDomain references a **ResourceClaimTemplate**; this template describes the DRA resource class (MNNVL) that will be used to provision the interconnect. When the ComputeDomain is created, the NVIDIA DRA controller prepares the underlying GPU fabric and ensures that nodes participating in the domain can securely communicate over MNNVL.
```
unmarshall: >In its `spec.channel`, the ComputeDomain references 

Just reword this to:
[ComputeDomainSpec.Channel](https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/f4960132a50b58079e913aa746991fdbf8e097f3/api/nvidia.com/resource/v1beta1/computedomain.go#L88) references a `ResourceClaimTemplate`.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+Next, pods that want to use MNNVL simply **reference the ResourceClaimTemplate** in their pod spec. Each pod declares a `resourceClaims` entry pointing to the template name, and the container lists that claim under `resources.claims`. Kubernetes then automatically creates a **ResourceClaim per pod** from the template. These per-pod claims are handed to the NVIDIA DRA driver, which allocates the necessary MNNVL/IMEX channels for each pod within the ComputeDomain.
+
+As a result, multiple pods—possibly scheduled on different nodes—are joined into the same ComputeDomain and can communicate using **multi-node NVLink semantics** without manually wiring GPUs or fabric resources. The ComputeDomain handles reachability and isolation, the ResourceClaimTemplate enables automatic per-pod allocation, and the pod spec remains simple: create a CD once, then reference its template in every pod that needs MNNVL.
```
unmarshall: >create a CD once

create a ComputeDomain once

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Goals
+
+The following key goals, derived from the requirements document, guide the MNNVL design:
```
unmarshall, line 61: >derived from the requirements document

If you are mentioning this then you should give a reference to the requirements document, or else you should put the requirements publicly as part of this doc or you should simply remove any mention of some requirements document.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+* **Cluster Support:** The feature will only support homogeneous clusters.
```
unmarshall: Rename `Cluster Support` to `Homogeneous Cluster Support` then write about what is homogeneous which is completely missing.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+* **Global Configuration:** A global configuration for the feature is required. If the cluster does not support MNNVL, the operator must exit.
```
unmarshall: > A global configuration for the feature is required. If the cluster does not support MNNVL, the operator must exit.

In k8s operator is a very overloaded word. Therefore please be precise and reword it to say `Grove` Operator will exit with a non-zero exit code.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+* **Opt-Out and Workload Impact:**
```
unmarshall: These are two different things and I would recommended to cover these in 2 different points.
`Opt-Out` - If MNNVL feature is enabled at the operator level then workload operators should be given a provision to opt out of automatic creation of DRA + ComputeDomain resources for specific workloads.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+  * An immutable, granular-level opt-out option must be available in the PCS.
+  * Failure to create MNNVL resources must not prevent the workload from running.
+  * Changes to the feature configuration should not impact a currently running workload.
```
unmarshall: Not clear, which configuration?

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+Next, pods that want to use MNNVL simply **reference the ResourceClaimTemplate** in their pod spec. Each pod declares a `resourceClaims` entry pointing to the template name, and the container lists that claim under `resources.claims`. Kubernetes then automatically creates a **ResourceClaim per pod** from the template. These per-pod claims are handed to the NVIDIA DRA driver, which allocates the necessary MNNVL/IMEX channels for each pod within the ComputeDomain.
+
+As a result, multiple pods—possibly scheduled on different nodes—are joined into the same ComputeDomain and can communicate using **multi-node NVLink semantics** without manually wiring GPUs or fabric resources. The ComputeDomain handles reachability and isolation, the ResourceClaimTemplate enables automatic per-pod allocation, and the pod spec remains simple: create a CD once, then reference its template in every pod that needs MNNVL.
+
+# Goals
```
unmarshall, line 59: Shouldn't you mention that these goals are for the initial version of `MNNVL` support and will be revised in future iterations?

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Enabling the feature
+
+Enabling and disabling the feature will be done by the cluster admin by setting a flag in the Grove OperatorConfiguration.
+
+```go
+// MNNVLConfiguration defines the configuration for MNNVL (Multi-Node NVLink) support.
+type MNNVLConfiguration struct {
+   // Enabled indicates whether MNNVL support is enabled.
+   // When true, the operator validates that the ComputeDomain CRD is installed at startup.
```
unmarshall, line 100:
```suggestion
   // When true, the operator validates that the ComputeDomain CRD is installed at startup.
  // When MNNVL support is enabled, cluster admin should ensure that the ComputeDomain CRD has been installed.
// If this prerequisite fails then Grove will exit with a non-zero exit code.
```

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+The value could be set from a Helm chart under the config attribute
+
+```yaml
+config:
+	mnnvl:
+		enabled: false
+```
+
+### Feature validity
```
unmarshall, line 120: you should explicitly mention that in the initial version of MNNVL support it is assumed that it is a homogeneous cluster. It is not the responsibility of Grove to ascertain the cluster homogeneity. It is upon the cluster admin to ensure that this feature is only enabled for homogeneous clusters.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+		enabled: false
+```
+
+### Feature validity 
+
+When the Grove operator starts, it will check for MNNVL support (only if the feature is enabled). Support is confirmed by the presence of the ComputeDomain CRD in the cluster.
+
+If the cluster lacks MNNVL support, the Grover operator will terminate and log an appropriate error.
+
+## ComputeDomain life-cycle management
```
unmarshall: life-cycle - just use a single word lifecycle (its valid)

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+  conditions:
+    - type: ComputeDomainsReady
+      status: "True"   # All replicas succeeded
+      reason: "AllCreated"
+      message: "ComputeDomains created for all 3 replicas"
+```
+
+```yaml
+status:
+  conditions:
```
unmarshall: It is fine to have this. It will only become a problem when there are hundreds of PCS replicas and ComputeDomain could not be created for most of them. However if there are hundreds of replicas then this is still better rather having hundreds of conditions

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+Pro:
+
+* Per-replica granularity  
+* Easy look up
+
+Cons:
+
+* Dynamic type names (less common but valid)  
+* Need to clean up conditions on scale-down
```
unmarshall: Well even when you have a single condition you also need to clean up the message in that condition if the replicas change. So lets assume you go from 1 -> 2 PCS replicas. For the 2nd replica ComputeDomain could not be created. Later on you scale-in to 1 replica, then you will have to still clean up the message, right? So this con if listed here should also be listed in the option where you just have a single condition.

Also the correct term is `scale-in` and not `scale-down` (which is when you are vertically scaling)

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+```
+
+Pro:
+
+* A well-defined structure makes everything clearer.
+
+Cons:
+
+* Ties the feature to Grove.
```
unmarshall:
```suggestion
* Ties the feature to Grove API, thus violating its goal of device independent configuration.
```

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+Pro:
+
+* Per-replica granularity  
+* Easy look up
+
+Cons:
+
+* Dynamic type names (less common but valid)
```
unmarshall: The biggest con is potentially bloating the status when there are many replicas.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+* Per-replica granularity  
+* Easy look up
+
+Cons:
+
+* Dynamic type names (less common but valid)  
+* Need to clean up conditions on scale-down
+
+#### Alt option 2: Use a desiccated status structure (not recommended)
```
unmarshall: is `desiccated` the right word?

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Watching the compute domain
+
+To ensure the PCS controller reconciles when a `ComputeDomain` is deleted or its status changes, a watch must be added in the PCS controller's RegisterWithManager function. 
+
+When the controller creates a `ComputeDomain` resource, it should attach the label `grove.io/podcliqueset` with the value `<pcs-name>` to identify which PCS owns it. 
+
+As part of the registration process, a new watch will be used to garud the `ComputeDomain` created by the controller (using the `grove.io/podcliqueset`label as a filter).
+
+```yaml
```
unmarshall, line 260: can you move this yaml snippet just below the para where you mention the label? Also is there a need to provide a yaml when you have explicitly mentioned the label?

docs/designs/mnnvl-design.md
unmarshall:
```suggestion
As part of the registration process, a new watch will be added to enqueue events for `ComputeDomain` created by the controller (using the `grove.io/podcliqueset`label as a filter).
```

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+* A well-defined structure makes everything clearer.
+
+Cons:
+
+* Ties the feature to Grove.
+
+### Watching the compute domain
+
+To ensure the PCS controller reconciles when a `ComputeDomain` is deleted or its status changes, a watch must be added in the PCS controller's RegisterWithManager function.
```
unmarshall: Can you elaborate which changes in the `ComputeDomainStatus` will you react to and why?
For your quick reference the [status struct](https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/f4960132a50b58079e913aa746991fdbf8e097f3/api/nvidia.com/resource/v1beta1/computedomain.go#L108) has 2 fields: `Status` and `Nodes`.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Scale-Out and Scale-In
+
+When scaling out (replicas increased), the subsequent reconciliation process will identify the ComputeDomains missing for the new replica indices and create them using the identical logic as the initial creation.
+
+When scaling in( replicas decreased), the PCS controller determines which `ComputeDomains` to delete by comparing the existing ones against the new replica count. Any `ComputeDomain` with a replica index equal to or greater than the new count is removed. This process mirrors the "desired versus existing" logic used for managing `PodClique` and `PCSG`.
+
+The controller lists existing `ComputeDomains` by label selector, computes expected resources from the current spec, and deletes the excess. The `ComputeDomainsReady` condition is updated to reflect the new state after scale operations.
+
+PCS Deletion
```
unmarshall:
```suggestion
### PCS Deletion  
```

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+The controller lists existing `ComputeDomains` by label selector, computes expected resources from the current spec, and deletes the excess. The `ComputeDomainsReady` condition is updated to reflect the new state after scale operations.
+
+PCS Deletion  
+When a PodCliqueSet is deleted, all ComputeDomain resources are automatically garbage-collected by Kubernetes through the owner reference mechanism. 
+
+Since each ComputeDomain has controllerReference set to the PCS, Kubernetes will cascade-delete them when the PCS is removed. No explicit cleanup logic is required in the controller for this case.
+
+## PCS opt-out
+
+Users can opt-out of MNNVL support for a specific `PodCliqueSet` using the `grove.io/mnnvl-enabled` annotation.
```
unmarshall: Since this annotation is for opting out, shouldn't it be `grove.io/disable-mnnvl`? Then you only need to add this annotation when you wish to disable mnnvl support when it has been enabled at the Grove operator level.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+
+When the global MNNVL feature is enabled in the `OperatorConfiguration`, individual workloads inherit this setting by default. To disable MNNVL for a specific PCS, set the annotation to "`false`".  
+
+When opting out, the operator will not create a `ComputeDomain` for that PCS.
+
+This annotation is validated as immutable by the PCS validation webhook, meaning that once a PCS is created with or without the annotation, it cannot be changed or removed. This ensures consistent behavior throughout the lifecycle of the workload.
+
+## Pod reconciliation
+
+As described above, before the PCS controller creates a `PodClique` for a replica, it first creates a `ComputeDomain` for the replica. It waits for it to become ready (with a 3-minute timeout).
```
unmarshall: what is the rational for a 3 min timeout? Is this configurable? I hope the reconciler worker is not blocked waiting for 3 mins.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+When the global MNNVL feature is enabled in the `OperatorConfiguration`, individual workloads inherit this setting by default. To disable MNNVL for a specific PCS, set the annotation to "`false`".  
+
+When opting out, the operator will not create a `ComputeDomain` for that PCS.
+
+This annotation is validated as immutable by the PCS validation webhook, meaning that once a PCS is created with or without the annotation, it cannot be changed or removed. This ensures consistent behavior throughout the lifecycle of the workload.
+
+## Pod reconciliation
+
+As described above, before the PCS controller creates a `PodClique` for a replica, it first creates a `ComputeDomain` for the replica. It waits for it to become ready (with a 3-minute timeout).   
+The final `ComputeDomain` status is captured in the `PodClique`'s status conditions and is **never** overwritten afterward, ensuring all pods within the PodClique have a consistent configuration.
```
unmarshall: I am not sure i understand this. If there are transient problems in creating ComputeDomain then you will set this condition once after 3 mins worth of retries and thereafter this condition is fixed in stone till the PCS replica lives?

Also the title of this section is `Pod reconciliation` and the YAML is PCLQ

docs/designs/mnnvl-design.md
unmarshall: If you are using abbreviations then its advisable to have a section where you can list the abbreviations. Then you can easily use `CD`, `RC`, `RCT` etc. Right now this document has mixed conventions.

You chose to name RCT as `{pcs-name}-replica-{index}-rct` but you have a different naming convention for CD `{pcs-name}-cd-{replica-index}`. Should you not align here and also have the same convention for RCT, namely `{pcs-name}-rct-{replica-index}`

unmarshall (changes requested):

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: I wanted to reference the req doc, but it is not public yet, there is a PR for that (https://github.com/ai-dynamo/grove/pull/296). it should be merged before this PR. and I will reference it here

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: split, not sure why it was together at the first place

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: taken

docs/designs/mnnvl-design.md
shmuel-runai: added

docs/designs/mnnvl-design.md
shmuel-runai: added

docs/designs/mnnvl-design.md
shayasoolin: Where is it stated/stored for a PCS that it has been started with MNNVL enabled? Consider the following scenario:
* Grove is started with MNNVL enabled
* PCS is created (no opt-out label)
* Grove is restarted with MNNVL disabled
Now the PCS should still be reconciled with its CD & RCT (even for new replicas, right?); How does Grove know that this PCS should be handled as an MNNVL-enabled one, even that it is currently running without MNNVL?

One way is to have this mnnvl-enabled label in both cases - false and true - and decide on its (immutable) value in a mutation webhook.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,245 @@
...
+status:
+  conditions:
+    - type: ComputeDomainReady
+      status: "True"              # or "False" if CD failed/timed out
+      reason: Ready               # or "Timeout" / "AllocationFailed"
+      message: "ComputeDomain my-pcs-replica-0 is ready"
+      lastTransitionTime: "2024-01-15T10:00:00Z"
+```
+
+When the `PodClique` controller reconciles pods, the Pod component checks its parent PodClique's ComputeDomainReady condition. If the condition is True, the Pod is created with a resourceClaims entry that references the ComputeDomain's ResourceClaimTemplate (RCT).
```
shayasoolin: "If the condition is True..."
Assuming you go with a single condition type for all replicas: what if the condition is False due to other replicas having issues? The pod should still work properly if its own specific replica has no issues.

docs/designs/mnnvl-design.md
shmuel-runai: You are right, I renamed it "Graceful degradation."

docs/designs/mnnvl-design.md
shmuel-runai: typo

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: aligned

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: correct added

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: I removed the "or its status" which was added by mistake

docs/designs/mnnvl-design.md
shmuel-runai: From my experience, using a negative is counterintuitive and may cause confusion
Is disable-mnnvl: true means it on or off.
I know it's funny to write "enabled false" to turn it off, but from my experience it better

docs/designs/mnnvl-design.md
shmuel-runai: removed. CD is considered "ok" if it was able be created

docs/designs/mnnvl-design.md
athreesh: +1 to @unmarshall point about Power vs Normal users. Can we add a note that Phase 1 targets 'Normal AI/ML?HPC users' and outline rougly what Power user support might look like in future phases of MNNVL support?

It's ok for us to target "Normal Users" right now since the need for this MNNVL abstraction is coming from the inference deployment users.

docs/designs/mnnvl-design.md
athreesh: Agreed

docs/designs/mnnvl-design.md
athreesh: Curious, is the 3-minute timeout configurable? For larger deployments like DSR1, this probably would impact deployment latency quite a bit. Is there any  expected p99 for ComputeDomain readiness?

docs/designs/mnnvl-design.md
athreesh: Lack of next steps is a great callout, want to make sure we don't scope creep during implementation.

Based on #296, some suggestions to add in:
  - Phase 1 only supports homogeneous clusters
  - DRA-only GPU requests not supported (Phase 2, potentially needs more color around how we support "power users"
  - No per-PCLQ granularity for MNNVL toggle

docs/designs/mnnvl-design.md
nvrohanv: While I agree that DRA is more general and power users might have other uses for it, I disagree that a power user would want to deal with MNNVL and ComputeDomains. What we are trying to do with MNNVL is just bring things back up to where they were with regular NVLink where users didn't have any control over it at the k8s level it was just enabled by default. 

With MNNVL there actually isn't any case where you'd want to do any custom special stuff since the requirement can be summarized as simply as all pods in a workload should be able to go over MNNVL when possible, regardless of autoscaling. In fact it is my hope in the long term that we actually move away from DRA for MNNVL entirely since ideally the user is not involved at all and the concept of MNNVL enablement being put on the user in a PVC style is fundamentally flawed in my opinion.

docs/designs/mnnvl-design.md
shmuel-runai: To avoid a large number of conditions, we update the status only in case of failure, which should be a rare case.
(failure only of creating the CD resource)

docs/designs/mnnvl-design.md
shmuel-runai: update the section

docs/designs/mnnvl-design.md
shmuel-runai: I don't think those points should be part of the design doc; maybe they should be in the requirements document.

docs/designs/mnnvl-design.md
nvrohanv: Here my hope is eventually Grove is not involved at all in multinode nvlink, so in my head if phase 2 is support heterogenous cluster, phase 3 is evolve dra driver to a point where all grove does is have some workload id that maps to a pcs replica and then lower layers take care of the rest. For this PR I think this description is adequate, what do you think @unmarshall

docs/designs/mnnvl-design.md
gflarity: Not sure why there's a debate here around this paragraph in particular?  This proposal objectively reduces complexity for those who want that reduction. Personas would factor in more for the requirements I think, there's a separate PR (and previously a google doc) for that.

sanjaychatterjee: Maybe rephrase this as:

Currently, Kubernetes workloads can take advantage of NVIDIA MNNVL accelerated hardware by explicitly adding NVIDIA ComputeDomain APIs to their workloads. With this feature, when applicable, Grove will abstract the MNNVL hardware from users by automatically creating ComputeDomain custom resources, enabling ComputeDomains support to workloads, and managing the lifecycle of the CRs.

docs/designs/mnnvl-design.md
shmuel-runai: Added a comment about it in the "Enabling the feature" section

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+When the global MNNVL feature is enabled in the `OperatorConfiguration`, individual workloads inherit this setting by default. To disable MNNVL for a specific PCS, set the annotation to "`false`".  
+
+When opting out, the operator will not create a `ComputeDomain` for that PCS.
+
+This annotation is validated as immutable by the PCS validation webhook, meaning that once a PCS is created with or without the annotation, it cannot be changed or removed. This ensures consistent behavior throughout the lifecycle of the workload.
+
+## PC and PCSG Creation
+
+When the PCS controller creates a PC or PCSG for a replica, it checks whether a CD exists for that replica. If a CD exists, the controller adds an annotation to the PC/PCSG with the RCT name:
```
gflarity: Related to my suggestion above, this change closes the loop by explicitly explaining how the correct compute domain is found/determined:

```suggestion
When the PCS controller creates a PC or PCSG for a replica, it checks whether a CD exists for that replica using the labels mentioned above. If a CD exists, the controller adds an annotation to the PC/PCSG with the RCT name:
```

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+The PCS controller has a reconciliation flow for managing resources in a specific order. The `ComputeDomain` component is synced **before** creating PCs and PCSGs, ensuring the CD exists before pods that reference it are created.
+
+Before creating the `CD`, the controller checks the following criteria:
+
+1. The MNNVL feature is enabled in Grove's `OperatorConfiguration`.
+2. The PCS has not opted out via the `grove.io/mnnvl-enabled: "false"` annotation.
+3. At least one PC in the PCS requires a GPU.
+
+If all criteria are met, the controller creates a `ComputeDomain` resource for each PCS replica. The CD is named `{pcs-name}-cd-{replica-index}` and references an RCT named `{pcs-name}-rct-{replica-index}`. The PCS is set as the owner reference to enable automatic garbage collection.
```
gflarity: I'd suggest mentioning the labels purpose explicitly as they're quite important. 

```suggestion
If all criteria are met, the controller creates a `ComputeDomain` resource for each PCS replica. The CD is named `{pcs-name}-cd-{replica-index}` and references an RCT named `{pcs-name}-rct-{replica-index}`. The PCS is set as the owner reference to enable automatic garbage collection. The  'grove.io/podcliqueset' and 'grove.io/replica-index' labels will be used to determine which Compute Domain is associated with a specific podcliqueset replica. 
```

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+  * Each PCS replica must have its own dedicated ComputeDomain.
+  * The lifecycle of the ComputeDomain is tied directly to the lifecycle of the PCS replica.
+
+## Scope and Limitations
+
+This document covers **Phase 1** of MNNVL support in Grove. See GREP-270 for the full requirements.
+
+**Limitations:**
+
+- **Homogeneous Clusters Only:** This phase supports only homogeneous clusters where all nodes have identical GPU types and NVLink topology. Grove does not validate or enforce cluster homogeneity—it is the cluster administrator's responsibility to enable this feature only on clusters that meet this requirement. Enabling MNNVL on heterogeneous clusters may result in undefined scheduling behavior.
```
kangclzjc, line 80: What if customer have both GB200 and other GPUs in the same cluster ? Do we plan to support heterogeneous cluster ? Because it maybe common in real customer environment with multiple types GPUs.

nvrohanv: in phase 2 we plan to support it (potentially even just have it taken care of at level lower than grove) but for now just homogeneous.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+This design document details the plan to enable the Grove operator to automatically leverage MNNVL for appropriate workloads.
+
+## Abbreviations
+
+| Abbreviation | Full Name | Description |
+|--------------|-----------|-------------|
+| CD | ComputeDomain | NVIDIA CRD representing a logical GPU fabric spanning multiple nodes |
+| RCT | ResourceClaimTemplate | Kubernetes resource template for dynamic resource allocation |
+| PCS | PodCliqueSet | Grove CRD that manages a set of PodCliques and PodCliqueScalingGroups |
+| PC | PodClique | Grove CRD representing a group of related pods |
```
shayasoolin: The operator uses the PCLQ abbreviation, not PC.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+| PCSG | PodCliqueScalingGroup | Grove CRD that manages scaling of PodCliques |
+
+## Motivation
+
+The MNNVL support feature in Grove is guided by three core design principles:
+
+**Simplicity:** Currently, Kubernetes workloads can take advantage of NVIDIA MNNVL accelerated hardware by explicitly adding NVIDIA ComputeDomain APIs to their workloads. With this feature, when applicable, Grove will abstract the MNNVL hardware from users by automatically creating ComputeDomain custom resources, enabling ComputeDomains support to workloads, and managing the lifecycle of the CRs.
+
+Power users who require custom ComputeDomain or RCT configurations can opt-out of automatic MNNVL management and specify their own `resourceClaims` directly in the pod spec template.
```
shayasoolin: AFAIU we decided for opt-in, not opt-out. This should be updated across the whole document, in multiple places.

shayasoolin: What if the user wants to have his own RCTs for non-CD stuff - can we allow that? I.e. handle CD and CD-RCT for the user, while allowing having more RCTs in the PCS template.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+**Portability:** The Grove Custom Resource Definitions (CRDs)—specifically PodCliqueSet, PodCliqueScalingGroup, and PodClique—are designed to be generic, excluding any direct references to MNNVL, ComputeDomain, or other NVIDIA-specific components.
+
+To maintain portability, all MNNVL-related configuration is kept external, residing in the operator configuration and runtime annotations. This approach allows the same PodCliqueSet manifest to be used without alteration, regardless of whether the target cluster supports MNNVL.
```
shayasoolin, line 27: I'm not sure the suffix of this sentence still holds, considering the switch to opt-in approach. The user won't be able to use the same manifest for both MNNVL and non-MNNVL clusters.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+**Standard Kubernetes behavior:** ComputeDomain lifecycle follows the same reconciliation pattern as other Grove-managed resources. If CD creation fails (due to transient cluster issues), the sync stops and requeues for retry. Persistent failures are surfaced via Kubernetes Events, allowing cluster administrators to investigate.
```
shayasoolin, line 29: I suggest to surface failures also as part of the PCS status (e.g. using conditions), to make them easily available also for non-admin users.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Scope and Limitations
+
+This document covers **Phase 1** of MNNVL support in Grove. See GREP-270 for the full requirements.
+
+**Limitations:**
+
+- **Homogeneous Clusters Only:** This phase supports only homogeneous clusters where all nodes have identical GPU types and NVLink topology. Grove does not validate or enforce cluster homogeneity—it is the cluster administrator's responsibility to enable this feature only on clusters that meet this requirement. Enabling MNNVL on heterogeneous clusters may result in undefined scheduling behavior.
+
+- **PCS-Level Granularity:** The MNNVL feature is applied at the PCS level—it cannot be targeted to individual PCs or PCSGs within a PCS. Either all pods in a replica receive the RCT reference, or none do.
```
shayasoolin: Fair enough, but please mention that non-GPU PCLQs are excluded. Otherwise the user would expect to be able to opt-out specific PCLQ/PCSG.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **No ComputeDomain Customization:** The ComputeDomain and ResourceClaimTemplate configurations are automatically generated by Grove and cannot be customized. Power users who require custom configurations can opt-out of automatic MNNVL management and specify their own `resourceClaims` directly in the pod spec template.
+
+# Design Details
+
+## Enabling the feature
+
+Enabling and disabling the feature will be done by the cluster admin by setting a flag in the Grove OperatorConfiguration.
```
shayasoolin, line 94: Worth mentioning explicitly that the feature is disabled by default.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+```go
+// MNNVLConfiguration defines the configuration for MNNVL (Multi-Node NVLink) support.
+type MNNVLConfiguration struct {
+   // Enabled indicates whether MNNVL support is enabled.
+   // When true, the operator validates that the ComputeDomain CRD is installed at startup.
```
shayasoolin, line 100: What about the DRA driver, shouldn't we verify that it's installed and running? Or is CRD availability sufficient?
What happens if I try to install the CD CRDs on a non-MNNVL cluster? I assume it goes without issues, as opposed to the driver itself which I expect to fail.

docs/designs/mnnvl-design.md
shayasoolin: Naming should be consistent with other grove resources, AFAIK they do not have the kind abbreviation (e.g. "cd" or "rct") as part of their names.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+When opting out, the operator will not create a `ComputeDomain` for that PCS.
+
+This annotation is validated as immutable by the PCS validation webhook, meaning that once a PCS is created with or without the annotation, it cannot be changed or removed. This ensures consistent behavior throughout the lifecycle of the workload.
+
+## PC and PCSG Creation
+
+When the PCS controller creates a PC or PCSG for a replica, it checks whether a CD exists for that replica. If a CD exists, the controller adds an annotation to the PC/PCSG with the RCT name:
+
+**Annotation:**
```
shayasoolin: The current approach in grove operator is that PCLQ/PCSG sync extracts info on the PCS by referencing it (see `componentutils.GetPodCliqueSet`). Perhaps it's good enough that the MNNVL state is immutable (as already mentioned in the doc), and that RCT names are known, given the consistent naming pattern? Then the annotation on the PC is redundant.

docs/designs/mnnvl-design.md
Ronkahn21: I agree with @shayasoolin about this point.

I think it will be worth to use explicit labels for each case, where defaulting adding this labels as enabled by default.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+### PCS Deletion  
+When a PodCliqueSet is deleted, all ComputeDomain resources are automatically garbage-collected by Kubernetes through the owner reference mechanism. 
+
+Since each ComputeDomain has controllerReference set to the PCS, Kubernetes will cascade-delete them when the PCS is removed. No explicit cleanup logic is required in the controller for this case.
+
+## PCS opt-out
+
+Users can opt-out of MNNVL support for a specific `PodCliqueSet` using the `grove.io/mnnvl-enabled` annotation. 
+
+When the global MNNVL feature is enabled in the `OperatorConfiguration`, individual workloads inherit this setting by default. To disable MNNVL for a specific PCS, set the annotation to "`false`".
```
Ronkahn21: what would be the behavior for existing workload, I assume it should consider opt out

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+
+This annotation is validated as immutable by the PCS validation webhook, meaning that once a PCS is created with or without the annotation, it cannot be changed or removed. This ensures consistent behavior throughout the lifecycle of the workload.
+
+## PC and PCSG Creation
+
+When the PCS controller creates a PC or PCSG for a replica, it checks whether a CD exists for that replica. If a CD exists, the controller adds an annotation to the PC/PCSG with the RCT name:
+
+**Annotation:**
+- **Key:** `grove.io/compute-domain-rct`
+- **Value:** The RCT name for the replica (e.g., `my-pcs-rct-0`)
```
Ronkahn21: Could you clarify the need for this annotation? I expected the RCT would be defined directly within the Pod spec as part of PCLQ

Ronkahn21: I’d prefer to keep PCLQ and PCSG agnostic of this feature. Ideally, the implementation should be transparent to them, treating the RCT simply as a standard field within the Pod spec.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,285 @@
...
+  labels:
+    grove.io/podcliqueset: my-pcs
+    grove.io/replica-index: "0"
+```
+
+When a PCSG creates PCs, it propagates the `grove.io/compute-domain-rct` annotation to its child PCs.
+
+### Why Use an Annotation?
+
+A simpler approach would be to have the Pod controller check if the MNNVL support is required at pod creation time and decide whether to add the RCT reference dynamically. However, this approach has a critical flaw:
```
Ronkahn21: I don't fully agree with the annotation approach as a blanket solution. For PCLQ, since the CRD already includes the Pod spec, the RCT should be part of the CR from the start—this keeps the controller feature-agnostic while ensuring pod consistency. I recognize that PCSG currently lacks a Pod template in its CRD, so we might need to rely on the annotation there for now. However, for PCLQ, the template should be defined upfront so the controller just reconciles it without needing to 'know' about the feature.

docs/designs/mnnvl-design.md
kangclzjc: @nvrohanv  When would be phase 2, since customer is already using in heterogeneous environment. If we can't support this in phase 1 then we may give some workaround to avoid this issue.

nvrohanv: I see, in that case the need has arisen sooner than I expected. I'll ping you to get more details on the exact setup.

nvrohanv: Actually I believe in the current design they would be able to workaround by just opting out. I believe the current opt-out plan was to have one flag at the PCS level but we could potentially make it so that you can opt out at more granular levels as well (if this is not already the case)

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: fix

docs/designs/mnnvl-design.md
shmuel-runai: This is aligned with other resources the controllers manage, error are reported at the event

docs/designs/mnnvl-design.md
shmuel-runai: added

docs/designs/mnnvl-design.md
shmuel-runai: align with the requirement

docs/designs/mnnvl-design.md
shmuel-runai: added

docs/designs/mnnvl-design.md
shmuel-runai: removed

docs/designs/mnnvl-design.md
shmuel-runai: yes, update the doc

docs/designs/mnnvl-design.md
shmuel-runai: adjust

docs/designs/mnnvl-design.md
shmuel-runai: adjust

docs/designs/mnnvl-design.md
shmuel-runai: outdated

docs/designs/mnnvl-design.md
shmuel-runai: added

docs/designs/mnnvl-design.md
kangclzjc: Could you give more detail about the workaround? You mentioned PCS level, do you mean add a flag when customer apply PCS yaml ?

docs/designs/mnnvl-design.md
```diff
@@ -105,36 +111,108 @@ When the Grove operator starts, it will check for MNNVL support (only if the fea
...
+When a PodCliqueSet is created, a **mutating admission webhook** automatically determines and records the MNNVL enablement status by setting the `grove.io/mnnvl-enabled` annotation.
+
+
+The webhook behavior:
+
+1. **If annotation already exists** → Do not override (respects user opt-out)
+2. **If MNNVL feature is disabled** in `OperatorConfiguration` → Do not add annotation
+3. **If MNNVL feature is enabled**:
+   - PCS has GPU containers (`nvidia.com/gpu`) → Set `grove.io/mnnvl-enabled: "true"`
+   - PCS has no GPU containers → Set `grove.io/mnnvl-enabled: "false"`
```
shayasoolin: I prefer that we will never add a false annotation. We discussed it and understood we don't need it. The only case where we should see a PCS with a false annotation is when the user added it in order to flag us to opt-out.

docs/designs/mnnvl-design.md
```diff
@@ -105,36 +111,108 @@ When the Grove operator starts, it will check for MNNVL support (only if the fea
...
+
+**Opt-out Behavior:**
+
+If the annotation already exists on the PCS (user explicitly set it), the webhook does **not** override it. Users can opt-out of MNNVL support for a specific `PodCliqueSet` by explicitly setting `grove.io/mnnvl-enabled: "false"` in the PCS manifest **before creation**.
+When a PCS is submitted with this annotation already set, the mutating webhook will **not override** the user's choice. This allows workloads to explicitly disable MNNVL even when the feature is globally enabled and the workload requires GPUs.
```
shayasoolin: There's one more case to talk about - either here or later in this section. If the user sets the annotation to true, but mnnvl is disabled by the operator. It should fail the submission. For existing workloads it's okay, but not for new ones.

docs/designs/mnnvl-design.md
```diff
@@ -145,141 +223,133 @@ ComputeDomain creation follows the same observability pattern as other Grove-man
...
+        resources:
+          limits:
+            nvidia.com/gpu: "8"
+          claims:
+            - name: mnnvl-claim
 ```
 
+### PCS Creating PCSG
+
+When the PCS controller creates a PCSG, it propagates the `grove.io/mnnvl-enabled` annotation to the PCSG:
```
shayasoolin: Please consider not adding this annotation on the PCSG but read it from the PCS (when syncing the PCSG) - as many other attributes are already read from the PCS at exactly the same stage in the controller's reconcile loop.

docs/designs/mnnvl-design.md
```diff
@@ -105,36 +111,108 @@ When the Grove operator starts, it will check for MNNVL support (only if the fea
 
 If the cluster lacks MNNVL support, the Grove operator will terminate and log an appropriate error.
 
+### PCS MNNVL Eligibility Determination
+
+When a PodCliqueSet is created, a **mutating admission webhook** automatically determines and records the MNNVL enablement status by setting the `grove.io/mnnvl-enabled` annotation.
```
shayasoolin: Naming: let's consider `grove.io/mnnvl` with multiple values such as `disabled` and `required` with room for future extension for more modes of operation.

docs/designs/mnnvl-design.md
shmuel-runai: We talked about it, and agreed that it is better to "cut" the PCS lookup.

docs/designs/mnnvl-design.md
shmuel-runai: I want to go with `grove.io/auto-mnnvl`

shayasoolin (dismissed):

danbar2 (dismissed):

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,355 @@
...
+  * Enabling/Disabling of MNNVL feature should not impact a currently running workload.
+* **ComputeDomain Management:**
+  * Each PCS replica must have its own dedicated ComputeDomain.
+  * The lifecycle of the ComputeDomain is tied directly to the lifecycle of the PCS replica.
+
+## Scope and Limitations
+
+This document covers **Phase 1** of MNNVL support in Grove. See GREP-270 for the full requirements.
+
+**Limitations:**
```
Ronkahn21, line 78: Mention the one IMEX channel per node , And what it means from utilization.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,355 @@
...
+
+### PCS MNNVL Eligibility Determination
+
+When a PodCliqueSet is created, a **mutating admission webhook** automatically determines and records the MNNVL enablement status by setting the `grove.io/mnnvl-enabled` annotation.
+
+
+The webhook behavior:
+
+1. **If annotation already exists** → Do not override (respects user opt-out)
+2. **If MNNVL feature is disabled** in `OperatorConfiguration` → Do not add annotation
```
Ronkahn21: You should denied creation of this annotation when the feature is disabled.

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,355 @@
...
+
+## ComputeDomain lifecycle management
+
+### Creation
+
+The PCS controller has a reconciliation flow for managing resources in a specific order. The `ComputeDomain` component is synced **before** creating PCLQs and PCSGs, ensuring the CD exists before pods that reference it are created.
+
+Before creating the `CD`, the controller checks the `grove.io/mnnvl-enabled` annotation on the PCS:
+
+- If `grove.io/mnnvl-enabled: "true"` → Create ComputeDomains for each replica
```
Ronkahn21: We decided to change the name, mnnvl-enabled is more for feature flag in the configuration

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,355 @@
...
+  #   Warning  ComputeDomainFailed    Failed to create ComputeDomain for replica 2: <error>
+  ```
+
+- **Logs:** Detailed error information is logged by the operator.
+
+- **Requeue on failure:** If CD creation fails, the sync stops and requeues for retry. PCLQ and PCSG creation only proceeds after all CDs for the replicas have been successfully created.
+
+### Protecting the ComputeDomain
+
+Deleting a ComputeDomain while pods are actively using its RCT causes significant workload disruption. To prevent accidental deletion, Grove adds a **finalizer** to each ComputeDomain it creates.
```
Ronkahn21, line 254: Did you checked this? It might not required  if CT does it by it own

docs/designs/mnnvl-design.md
shmuel-runai: updated

docs/designs/mnnvl-design.md
shmuel-runai: updated

docs/designs/mnnvl-design.md
shmuel-runai: added

docs/designs/mnnvl-design.md
shmuel-runai: updated

docs/designs/mnnvl-design.md
shmuel-runai: check

docs/designs/mnnvl-design.md
```diff
@@ -0,0 +1,355 @@
...
+
+- If `grove.io/mnnvl-enabled: "true"` → Create ComputeDomains for each replica
+- If `grove.io/mnnvl-enabled: "false"` or annotation is absent → Skip ComputeDomain creation
+
+Since the annotation is set by the mutating webhook at PCS creation time (based on feature enablement and GPU requirements), the controller logic is simplified to a single annotation check.
+
+**Backward Compatibility:** Existing PCS resources created before the MNNVL feature was deployed will not have the `grove.io/mnnvl-enabled` annotation. These workloads will continue to operate without ComputeDomains, even after the feature is enabled globally. To enable MNNVL for an existing workload, the PCS must be deleted and recreated.
+
+If MNNVL is enabled, the controller creates a `ComputeDomain` resource for each PCS replica. The CD is named `{pcs-name}-{replica-index}` and references an RCT with the same name. The PCS is set as the owner reference to enable automatic garbage collection. The `app.kubernetes.io/part-of` and `grove.io/podcliqueset-replica-index` labels will be used to determine which ComputeDomain is associated with a specific PCS replica.
```
kangclzjc, line 214: Does it mean that if I have a PCS with 5 replicas, it will create 5 ComputeDomain. Then can pods communicate witch each other cross ComputeDomain. I'm afraid not, right ? But can we have a mode that these replicas have the same ComputeDomain.

docs/designs/mnnvl-design.md
shmuel-runai: added

docs/designs/mnnvl-design.md
shmuel-runai: AFAIK, it isn't common for pods across replicas to communicate with each other.
Maybe I'm wrong, but isn't the idea that each replica is its own isolated "realm" (or should I say "domain").
Maybe the pod should be in the same PCSG?

regardless 
1. Pods can communicate, but not using the same CD.
2. As stated, power users can provide their own CD and RCT ref in the pod spec template.
3. We can discuss use cases for sharing a CD between replicas for the next phase.

shayasoolin: @shmuel-runai you're not wrong at all. Replicas are considered fully independent of each other and pods from different replicas are not supposed to communicate with each other.

docs/designs/mnnvl-design.md
shmuel-runai: Resolving this conversation, as the request will be handled as part of Phase 2.

PR #300, opened 2026-01-04
Move header copyright from 2025 to 2026

kangclzjc:
#### What type of PR is this?
CI/CD will check the header and compare the year. 

Fixes #

```release-note

```

```docs

```


renormalize (commented):
I don't think we should have to update the year in the license for the generated files every year. The rest of the ecosystem does not do this either.

There seems to be a gap in the `add-license-headers` target which is causing this diff in the generated files.

PR #312, opened 2026-01-09
E2E stability fixes

gflarity:
#### What type of PR is this?

Testing Improvment/Fixes

This PR:

* Splits the E2E tests into the separate feature suites, this allows them to run in parallel and also you can now re-run specific suites 
* Disables a number of Rolling Update tests that are known to fail intermittently
* Adds retry logic to node cordoning (a randomly re-occurring failure) and increasing debug logging

```release-note
NONE
```


sanjaychatterjee (approved):
Approving it for now so we can let TAS through. We need to follow up on the rolling updates failures.

PR #337, opened 2026-01-16
skip patch ObservedGeneration if no change

xulinfei1996:
#### What type of PR is this?

/kind enhancement

Optimizes number of `Patch` calls made to the kube-apiserver by explicitly inspecting if there is a change done to `ObservedGeneration` w.r.t `Generation`. If an update is required, only then make the patch call else skips making the call.

Fixes #

```release-note
NONE
```

```docs

```


unmarshall:
@xulinfei1996 Thanks for the PR. It would be great if you could follow some expectations that are required to raise a PR:

1. Create an Issue
2. Ensure that it is labeled with a `kind`, `area`
3. In the Issue have sufficient description of what the issue is as this establishes a context for reviewers.
4. In the PR mention in brief what you propose and which issue it fixes.

For now i have labeled this issue and added some description to the PR.

PR #344, opened 2026-01-17
Support external certificate management for webhooks

gflarity:
#### What type of PR is this?

/kind feature

Note this PR was originally #233. I have taken it over it order to implement PR feedback and get things moving faster. 

This PR adds support for external certificate management for Grove's webhook server. Users can now choose between:

1. **Auto-provisioned certificates (default)**: Grove's built-in cert-controller automatically generates and manages self-signed certificates - best for development/testing
2. **External certificates**: Users can provide their own certificates (e.g., via cert-manager or manually) - best for production environments

Key changes:
- Added `autoProvision` configuration option (`config.server.webhooks.autoProvision`) to toggle between auto-provisioned and external certificates
- Added `secretName` configuration to specify the Kubernetes Secret containing webhook certificates
- Added `webhooks.caBundle` Helm value to inject a custom CA bundle into webhook configurations
- Added comprehensive e2e tests covering both certificate management modes and transitions between them

Fixes #

- When `autoProvision: true` (default), behavior is unchanged - Grove generates and manages certificates automatically
- When `autoProvision: false`, the operator expects certificates to exist in the specified Secret before starting
- The e2e tests verify transitioning from cert-manager managed certificates to autoprovision mode and back
- Webhook configurations now support an optional `caBundle` value for external CA certificates

```release-note
Added support for external certificate management. Users can now disable auto-provisioned certificates by setting `config.server.webhooks.autoProvision: false` and provide their own certificates (e.g., via cert-manager). A new `webhooks.caBundle` Helm value allows injecting custom CA bundles into webhook configurations.
```

```docs
Added docs/user-guide/certificate-management.md
```

### DIFF
--- a/operator/e2e/utils/k8s_client.go
+++ b/operator/e2e/utils/k8s_client.go
@@ -66,7 +66,11 @@ func ApplyYAMLFile(ctx context.Context, yamlFilePath string, namespace string, r
 		return nil, fmt.Errorf("failed to read YAML file %s: %w", yamlFilePath, err)
 	}
 
-	return ApplyYAMLData(ctx, yamlData, namespace, restConfig, logger)
+	dynamicClient, restMapper, err := CreateKubernetesClients(restConfig)
+	if err != nil {
+		return nil, err
+	}
+	return ApplyYAMLData(ctx, yamlData, namespace, dynamicClient, restMapper, logger)
 }
 
 // WaitForPods waits for pods to be ready in the specified namespaces
@@ -131,18 +135,8 @@ func WaitForPods(ctx context.Context, restConfig *rest.Config, namespaces []stri
 	})
 }
 
-// ApplyYAMLData applies YAML data to Kubernetes, creating clients from restConfig.
-// Note this is a convenience function that creates clients on every call and is deprecated (issue 378)
-func ApplyYAMLData(ctx context.Context, yamlData []byte, namespace string, restConfig *rest.Config, logger *Logger) ([]AppliedResource, error) {
-	dynamicClient, restMapper, err := CreateKubernetesClients(restConfig)
-	if err != nil {
-		return nil, err
-	}
-	return ApplyYAMLDataWithClients(ctx, yamlData, namespace, dynamicClient, restMapper, logger)
-}
-
-// ApplyYAMLDataWithClients applies YAML data to Kubernetes.
-func ApplyYAMLDataWithClients(ctx context.Context, yamlData []byte, namespace string, dynamicClient dynamic.Interface, restMapper meta.RESTMapper, logger *Logger) ([]AppliedResource, error) {
+// ApplyYAMLData applies YAML data to Kubernetes.
+func ApplyYAMLData(ctx context.Context, yamlData []byte, namespace string, dynamicClient dynamic.Interface, restMapper meta.RESTMapper, logger *Logger) ([]AppliedResource, error) {
 	decoder := yamlutil.NewYAMLOrJSONDecoder(strings.NewReader(string(yamlData)), 4096)
 	var appliedResources []AppliedResource
 
--- a/operator/e2e/tests/cert_management_test.go
+++ b/operator/e2e/tests/cert_management_test.go
@@ -161,12 +161,12 @@ func Test_CM1_CertManagementRoundTrip(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to create Kubernetes clients: %v", err)
 	}
-	if _, err := utils.ApplyYAMLDataWithClients(ctx, []byte(certManagerIssuerYAML), "", dynamicClient, restMapper, logger); err != nil {
+	if _, err := utils.ApplyYAMLData(ctx, []byte(certManagerIssuerYAML), "", dynamicClient, restMapper, logger); err != nil {
 		t.Fatalf("Failed to apply ClusterIssuer: %v", err)
 	}
 	waitForClusterIssuer(t, ctx, dynamicClient, "selfsigned-issuer")
 
-	if _, err := utils.ApplyYAMLDataWithClients(ctx, []byte(certManagerCertificateYAML), "", dynamicClient, restMapper, logger); err != nil {
+	if _, err := utils.ApplyYAMLData(ctx, []byte(certManagerCertificateYAML), "", dynamicClient, restMapper, logger); err != nil {
 		t.Fatalf("Failed to apply Certificate: %v", err)
 	}
 


operator/api/config/v1alpha1/types.go
```diff
@@ -138,6 +138,18 @@ type WebhookServer struct {
...
+	// +optional
+	// +kubebuilder:default="grove-webhook-server-cert"
+	SecretName string `json:"secretName,omitempty"`
+	// AutoProvision enables automatic certificate generation and management via cert-controller.
+	// When true: cert-controller automatically generates self-signed certificates and stores them in the Secret.
+	// When false: certificates are expected to be provided externally (e.g., by cert-manager, cluster admin).
+	// Default: true (auto-generate certificates for easy setup)
+	// +optional
+	// +kubebuilder:default=true
+	AutoProvision *bool `json:"autoProvision,omitempty"`
```
Ronkahn21: I would suggest rename the field 

```suggestion
	ProvisionMode CertProvisionMode`json:"certProvisionMode,omitempty"`
```

```go
type CertProvisionMode string

cosnt (
Auto CertProvisionMode = "auto"
Manual CertProvisionMode = "manual" 
)```

this would allow the API to be extended if needed

File: operator/charts/values.yaml [hunk 2]
```diff
@@ -143,21 +162,27 @@ leaseRoleBinding:
...
       app.kubernetes.io/component: operator-pcs-defaulting-webhook
       app.kubernetes.io/name: pcs-defaulting-webhook
       app.kubernetes.io/part-of: grove
+    annotations: {}
   authorizerWebhook:
     labels:
       app.kubernetes.io/component: operator-authorizer-webhook
       app.kubernetes.io/name: authorizer-webhook
       app.kubernetes.io/part-of: grove
+    annotations: {}
```
Ronkahn21, line 191: Is empty annotation needs to be present? If not lets remove this for cleaner code

operator/e2e/setup/grove.go
```diff
@@ -0,0 +1,120 @@
...
+
+// UpgradeGrove upgrades the Grove operator with the specified Helm values.
+//
+// This uses Helm upgrade (rather than Skaffold) because:
+// 1. Grove is initially installed via Skaffold, which uses Helm under the hood
+// 2. For config-only changes (like switching cert modes), rebuilding images is unnecessary
+// 3. Helm upgrade with ReuseValues preserves the image configuration that Skaffold set
+//
+// This approach avoids wasteful rebuilds while staying compatible with the Skaffold installation.
+func UpgradeGrove(ctx context.Context, restConfig *rest.Config, values map[string]interface{}, logger *utils.Logger) error {
```
Ronkahn21: I like the logic here! To make this more user-friendly, could we pass a dedicated config struct instead of a raw map[string]interface{}? The function could then handle the translation to JSON internally. Also, would it be possible to accept a controller-runtime client (or a specific typed client) instead of rest.Config to stay consistent with our other utilities?"

operator/e2e/setup/grove.go
```diff
@@ -0,0 +1,120 @@
...
+// chartYAML represents the structure of Chart.yaml for version extraction
+type chartYAML struct {
+	Version string `yaml:"version"`
+}
+
+// getChartVersion reads the version from Chart.yaml in the given chart directory.
+// We read from Chart.yaml rather than hardcoding the version to maintain a single source
+// of truth. This ensures the upgrade uses the same version as the chart itself, avoiding
+// configuration drift between the chart definition and the e2e test code.
+func getChartVersion(chartPath string) (string, error) {
```
Ronkahn21: Its not clear from the doc the file name must be Chart.yaml, chartPath imply for path and not file dir path.
 I would prefer just passing the full path to chart without relay on certain name,

operator/e2e/setup/helm.go
```diff
@@ -133,6 +138,81 @@ func InstallHelmChart(config *HelmInstallConfig) (*release.Release, error) {
 	return rel, nil
 }
 
+// UpgradeHelmChart upgrades a Helm chart with the given configuration.
+func UpgradeHelmChart(config *HelmInstallConfig) (*release.Release, error) {
```
Ronkahn21, line 142:
```suggestion
func UpdateGroveOperatorConfiguration(config *HelmInstallConfig) (*release.Release, error) {
```
The main goal is support changing configuration, Helm is just the way to do it

gflarity: Actually this is generic, it will work with any helm chart.  I'd rather keep it generic incase we want to change settings in KAI or another helm chart we add later.

operator/e2e/setup/grove.go
gflarity: I've created a GroveConfig struct with WebhooksConfig that handles the translation to Helm values internally via a toHelmValues() method. PTAL. 

Regarding rest.Config vs controller-runtime client: I kept rest.Config because the other utilities in operator/e2e/utils/ (like ApplyYAMLFile, WaitForPods, WaitForPodsInNamespace) all use rest.Config. So it's consistent in that way and used quite a bit. What other utilities did you mean?

operator/e2e/setup/grove.go
gflarity: So I use the the Helm go packages under the hood to actually install and upgrade Helm charts. These require the chart path anyways. A valid helm chart must have a Chart.yaml, so it's not arbitrary at all. So it's just easier to only require chartPath at the highest level and then use it everywhere.

operator/charts/values.yaml
gflarity: It was, but I think there's a better way to do this that won't require this.

operator/api/config/v1alpha1/types.go
gflarity: Done.

operator/charts/values.yaml
gflarity: Updated.

operator/e2e/setup/grove.go
Ronkahn21: As this method is exported, its consider a public API, if not meant to access directy lets change the functon to be internal only, and then the values arg is ok to be in this format

operator/e2e/setup/grove.go
Ronkahn21: can we add your explanation as comment, as some people (me included) are not familar with helm charts

operator/e2e/setup/grove.go
gflarity: setup is a different package  than the tests that use it though... It's meant to be accessed by E2E tests if needed... I'll take a look and see what we can do.

gflarity: Adding some documentation to the top of file explaining what it's intended for.

operator/e2e/setup/grove.go
```diff
@@ -0,0 +1,178 @@
...
+		annotations[k] = v
+	}
+
+	webhookAnnotations := map[string]interface{}{"annotations": annotations}
+
+	return map[string]interface{}{
+		"installCRDs": c.InstallCRDs,
+		"config": map[string]interface{}{
+			"server": map[string]interface{}{
+				"webhooks": map[string]interface{}{
```
Ronkahn21: Manual map nesting like this is pretty brittle—if we rename a field in GroveConfig but forget to update this string key, the Helm chart will break silently.

Can we refactor this to use json.Marshal/Unmarshal or a library like mapstructure?

operator/e2e/setup/grove.go
```diff
@@ -0,0 +1,178 @@
...
+// UpdateGroveConfiguration updates the Grove operator configuration.
+//
+// This uses Helm upgrade (rather than Skaffold) because:
+// 1. Grove is initially installed via Skaffold, which uses Helm under the hood
+// 2. For config-only changes (like switching cert modes), rebuilding images is unnecessary
+// 3. Helm upgrade with ReuseValues preserves the image configuration that Skaffold set
+//
+// This approach avoids wasteful rebuilds while staying compatible with the Skaffold installation.
+func UpdateGroveConfiguration(ctx context.Context, restConfig *rest.Config, config *GroveConfig, logger *utils.Logger) error {
+	chartDir, err := getGroveChartDir()
```
Ronkahn21: From a SOLID perspective the path should be passed to the function

operator/e2e/utils/k8s_client.go
```diff
@@ -129,8 +129,8 @@ func WaitForPods(ctx context.Context, restConfig *rest.Config, namespaces []stri
 	})
 }
 
-// applyYAMLData is the common function that applies YAML data to Kubernetes
-func applyYAMLData(ctx context.Context, yamlData []byte, namespace string, restConfig *rest.Config, logger *Logger) ([]AppliedResource, error) {
+// ApplyYAMLData is the common function that applies YAML data to Kubernetes
+func ApplyYAMLData(ctx context.Context, yamlData []byte, namespace string, restConfig *rest.Config, logger *Logger) ([]AppliedResource, error) {
 	dynamicClient, restMapper, err := createKubernetesClients(restConfig)
```
Ronkahn21: I would prefer to have a way we not to create client obj in every function that interact with the cluster, in long session it could take up to a minute

operator/internal/controller/cert/cert.go
```diff
@@ -38,17 +39,31 @@ const (
 	certificateAuthorityOrganization = "Grove"
 )
 
-// ManageWebhookCerts registers the cert-controller with the manager which will be used to manage
-// webhook certificates.
-func ManageWebhookCerts(mgr ctrl.Manager, certDir string, authorizerEnabled bool, certsReadyCh chan struct{}) error {
+// ManageWebhookCerts manages webhook certificates based on the CertProvisionMode configuration.
+// When mode=auto: uses cert-controller for automatic certificate generation and management.
+// When mode=manual: waits for externally provided certificates (e.g., from cert-manager, cluster admin).
+func ManageWebhookCerts(mgr ctrl.Manager, certDir string, secretName string, authorizerEnabled bool, certProvisionMode configv1alpha1.CertProvisionMode, certsReadyCh chan struct{}) error {
```
Ronkahn21, line 46: can you please add unit testing for the new functionality. It will  make sure it does not break when we might have more the current two modes

docs/user-guide/certificate-management.md
```diff
@@ -0,0 +1,298 @@
...
+To use external certificates, configure the following Helm values:
+
+```yaml
+config:
+  server:
+    webhooks:
+      # Use manual certificate provisioning (external)
+      certProvisionMode: manual
+      # Name of the Secret containing your certificates
+      secretName: "grove-webhook-server-cert"
```
renormalize: Nit:

```suggestion
      secretName: "grove-webhook-server-cert" # or <custom-secret-name>
```

operator/api/config/v1alpha1/defaults.go
```diff
@@ -79,6 +79,14 @@ func SetDefaults_ServerConfiguration(serverConfig *ServerConfiguration) {
 		serverConfig.Webhooks.ServerCertDir = defaultWebhookServerTLSServerCertDir
 	}
 
+	if serverConfig.Webhooks.SecretName == "" {
+		serverConfig.Webhooks.SecretName = "grove-webhook-server-cert"
```
renormalize: Nit: The string can be defined as a const in `config/types`, similar to how `CertProvisionMode*` were introduced as consts.

operator/e2e/utils/k8s_client.go
renormalize: We could perhaps just create the client at the top level, and pass it down instead of the restConfig?

renormalize (dismissed):
thanks! cert-manager and manual cert generation docs are both perfect.

operator/e2e/setup/grove.go
gflarity: Done

operator/e2e/utils/k8s_client.go
gflarity: Done.

docs/user-guide/certificate-management.md
gflarity: Done.

operator/api/config/v1alpha1/defaults.go
gflarity: Done.

operator/e2e/setup/grove.go
gflarity: Done.

operator/e2e/utils/k8s_client.go
gflarity: See comment about using this everywhere.

operator/e2e/utils/k8s_client.go
```diff
@@ -129,13 +129,18 @@ func WaitForPods(ctx context.Context, restConfig *rest.Config, namespaces []stri
 	})
 }
 
-// applyYAMLData is the common function that applies YAML data to Kubernetes
-func applyYAMLData(ctx context.Context, yamlData []byte, namespace string, restConfig *rest.Config, logger *Logger) ([]AppliedResource, error) {
-	dynamicClient, restMapper, err := createKubernetesClients(restConfig)
+// ApplyYAMLData applies YAML data to Kubernetes, creating clients from restConfig.
+// Note this is a convenience function that creates clients on every call and is deprecated (issue 378)
+func ApplyYAMLData(ctx context.Context, yamlData []byte, namespace string, restConfig *rest.Config, logger *Logger) ([]AppliedResource, error) {
```
Ronkahn21: I would preferred to remove this function and use only the ApplyYAMLDataWithClients  , I see the appeal but we want user to use the sec onc

Ronkahn21 (commented):
One last comment, but LGTM

operator/e2e/utils/k8s_client.go
gflarity: A long time ago this function was used in many places. I see it's just the CM tests now. My bad, this is done.

PR #346, opened 2026-01-18
Auto-MNNVL: add MNNVL configuration and startup validation

shmuel-runai:
/kind feature

Ref: GREP-270

Add MNNVL configuration support to the Grove operator. This includes configuration, startup validation, and Helm chart integration.

Changes:
- Add MNNVLConfiguration struct with Enabled field (default: false) and to the OperatorConfiguration
- Add validateMNNVLPrerequisites() to check for ComputeDomain CRD at startup
- Update Helm chart values.yaml and _helpers.tpl with mnnvl.enabled
- Add unit tests for config parsing and CRD detection logic

When MNNVL is enabled, but the ComputeDomain CRD is not installed, the operator exits with a non-zero exit code.

Tested
- UT

### DIFF
--- a/operator/api/config/v1alpha1/types.go
+++ b/operator/api/config/v1alpha1/types.go
@@ -207,7 +207,7 @@ type TopologyAwareSchedulingConfiguration struct {
 type NetworkAcceleration struct {
 	// AutoMNNVLEnabled indicates whether automatic MNNVL (Multi-Node NVLink) support is enabled.
 	// When enabled, the operator will automatically create and manage ComputeDomain resources
-	// for GPU workloads. If the cluster doesn't have the NVIDIA OpenDRA driver installed,
+	// for GPU workloads. If the cluster doesn't have the NVIDIA DRA driver installed,
 	// the operator will exit with a non-zero exit code.
 	// Default: false
 	AutoMNNVLEnabled bool `json:"autoMNNVLEnabled"`
--- a/docs/api-reference/operator-api.md
+++ b/docs/api-reference/operator-api.md
@@ -794,7 +794,7 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `autoMNNVLEnabled` _boolean_ | AutoMNNVLEnabled indicates whether automatic MNNVL (Multi-Node NVLink) support is enabled.<br />When enabled, the operator will automatically create and manage ComputeDomain resources<br />for GPU workloads. If the cluster doesn't have the NVIDIA OpenDRA driver installed,<br />the operator will exit with a non-zero exit code.<br />Default: false |  |  |
+| `autoMNNVLEnabled` _boolean_ | AutoMNNVLEnabled indicates whether automatic MNNVL (Multi-Node NVLink) support is enabled.<br />When enabled, the operator will automatically create and manage ComputeDomain resources<br />for GPU workloads. If the cluster doesn't have the NVIDIA DRA driver installed,<br />the operator will exit with a non-zero exit code.<br />Default: false |  |  |
 
 
 
--- a/operator/internal/mnnvl/preflight_test.go
+++ b/operator/internal/mnnvl/preflight_test.go
@@ -130,7 +130,7 @@ func Test_validateComputeDomainCRD(t *testing.T) {
 		{
 			name:      "discovery error - returns error",
 			resources: []*metav1.APIResourceList{},
-			reactor: func(action k8stesting.Action) (bool, runtime.Object, error) {
+			reactor: func(_ k8stesting.Action) (bool, runtime.Object, error) {
 				return true, nil, errors.New("connection refused")
 			},
 			expectError: true,


operator/cmd/main.go
```diff
@@ -121,6 +129,58 @@ func printFlags() {
 	logger.Info("Running with flags", flagKVs...)
 }
 
+// validateMNNVLPrerequisites checks if MNNVL prerequisites are met when the feature is enabled.
+// If MNNVL is enabled, it verifies that the ComputeDomain CRD is installed in the cluster.
+func validateMNNVLPrerequisites(operatorCfg *configv1alpha1.OperatorConfiguration) error {
```
Ronkahn21: Please move this function to different package so main file logic could remain clean.
by exporting all this logic to separate function in different package it will cleaner,

Ronkahn21: also the exported function should just get the MNNVL feature enablement bool, or checking it in main and then calling this function.

operator/api/common/constants/constants.go
```diff
@@ -132,3 +132,17 @@ const (
...
+	// ComputeDomainGroup is the API group for NVIDIA ComputeDomain resources.
+	ComputeDomainGroup = "resource.nvidia.com"
+	// ComputeDomainVersion is the API version for NVIDIA ComputeDomain resources.
+	ComputeDomainVersion = "v1beta1"
+	// ComputeDomainKind is the Kind for NVIDIA ComputeDomain resources.
+	ComputeDomainKind = "ComputeDomain"
+	// ComputeDomainResource is the plural resource name for ComputeDomain.
+	ComputeDomainResource = "computedomains"
+	// ComputeDomainCRDName is the full CRD name for ComputeDomain.
+	ComputeDomainCRDName = ComputeDomainResource + "." + ComputeDomainGroup
```
Ronkahn21: **Nit**: Avoid hardcoding these. Please import them from the compute domain repo if possible; otherwise, move them to the mnnvl package since they are only used for internal CRD check logic

operator/api/config/v1alpha1/types.go
```diff
@@ -200,3 +202,11 @@ type TopologyAwareSchedulingConfiguration struct {
 	// +optional
 	Levels []corev1alpha1.TopologyLevel `json:"levels,omitempty"`
 }
+
+// MNNVLConfiguration defines the configuration for MNNVL (Multi-Node NVLink) support.
+type MNNVLConfiguration struct {
+	// Enabled indicates whether MNNVL support is enabled.
+	// If the cluster doesn't have nvidia OpenDRA driver, the operator will exit with a non-zero exit code.
```
gflarity:
```suggestion
	// If the cluster doesn't have NVIDIA DRA driver, the operator will exit with a non-zero exit code.
```

operator/cmd/cli/cli_test.go
```diff
@@ -130,6 +130,15 @@ func TestLoadAndValidateOperatorConfig(t *testing.T) {
 				assert.Equal(t, 9445, config.Server.Metrics.Port)
 				assert.NotNil(t, config.Controllers.PodCliqueSet.ConcurrentSyncs)
 				assert.Equal(t, 3, *config.Controllers.PodCliqueSet.ConcurrentSyncs)
+				assert.False(t, config.MNNVL.Enabled, "MNNVL should be disabled by default")
+			},
+		},
```
gflarity, line 135: Nit: Maybe a test case for defaulting false.

sanjaychatterjee:
@unmarshall can you please review this as well.

operator/api/config/v1alpha1/types.go
shmuel-runai: fix

operator/cmd/main.go
shmuel-runai: moved

operator/api/common/constants/constants.go
shmuel-runai: moved

operator/charts/templates/_helpers.tpl
```diff
@@ -55,6 +55,10 @@ config.yaml: |
       []
     {{- end }}
   {{- end }}
+  {{- if .Values.config.mnnvl }}
```
gflarity: The Helm template generates `mnnvl.enabled` but the Go struct expects `network.autoMNNVLEnabled`.  Mismatch?

Shouldn't this be:
```yaml
network:
  autoMNNVLEnabled: {{ .Values.config.mnnvl.enabled | default false }}
```

operator/cmd/cli/testdata/valid-config.yaml
```diff
@@ -35,4 +35,6 @@ authorizer:
   enabled: false
 clusterTopology:
   enabled: false
+mnnvl:
```
gflarity: Same here, change `network.autoMNNVLEnabled: false` to match the Go struct, not `mnnvl.enabled`.

gflarity (commented):
PTAL at the mismatch, otherwise LGTM.

danbar2 (dismissed):
LGTM, please fix the missmatches @gflarity pointed out

operator/internal/mnnvl/preflight.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+var logger = ctrl.Log.WithName("mnnvl")
+
+// Preflight checks if MNNVL prerequisites are met when the feature is enabled.
+// If MNNVL is enabled, it verifies that the ComputeDomain CRD is installed in the cluster.
+func Preflight(operatorCfg *configv1alpha1.OperatorConfiguration) error {
+	if !operatorCfg.Network.AutoMNNVLEnabled {
+		return nil
+	}
```
Ronkahn21, line 36: After thinking about it, this more about business logic other functionality rather than just logic,
I would get this out to main

operator/internal/mnnvl/preflight.go
```diff
@@ -0,0 +1,79 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/discovery"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var logger = ctrl.Log.WithName("mnnvl")
+
+// Preflight checks if MNNVL prerequisites are met when the feature is enabled.
+// If MNNVL is enabled, it verifies that the ComputeDomain CRD is installed in the cluster.
+func Preflight(operatorCfg *configv1alpha1.OperatorConfiguration) error {
```
Ronkahn21, line 33: It will cleaner and easy to test if you will pass the Dynamic client interface as arg, instead of passing the entire configuration 

than this section can be in main 
	cfg, err := ctrl.GetConfig()
	if err != nil {
		return fmt.Errorf("failed to get cluster config for MNNVL validation: %w", err)
	}

	discoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg)
	if err != nil {
		return fmt.Errorf("failed to create discovery client for MNNVL validation: %w", err)
	}

operator/internal/mnnvl/preflight.go
```diff
@@ -0,0 +1,79 @@
...
+
+	cfg, err := ctrl.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to get cluster config for MNNVL validation: %w", err)
+	}
+
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create discovery client for MNNVL validation: %w", err)
+	}
```
Ronkahn21, line 59: Move this init client to main, it will make this function easier to test as it get only interface you can mock

operator/charts/templates/_helpers.tpl
shmuel-runai: cool catch, 10x

operator/cmd/cli/testdata/valid-config.yaml
shmuel-runai: same

operator/internal/mnnvl/preflight.go
shmuel-runai: fix

Ronkahn21 (approved):

operator/api/config/v1alpha1/types.go
```diff
@@ -200,3 +202,13 @@ type TopologyAwareSchedulingConfiguration struct {
 	// +optional
 	Levels []corev1alpha1.TopologyLevel `json:"levels,omitempty"`
 }
+
+// NetworkAcceleration defines the configuration for network acceleration features.
+type NetworkAcceleration struct {
+	// AutoMNNVLEnabled indicates whether automatic MNNVL (Multi-Node NVLink) support is enabled.
+	// When enabled, the operator will automatically create and manage ComputeDomain resources
+	// for GPU workloads. If the cluster doesn't have the NVIDIA OpenDRA driver installed,
```
shayasoolin: It's called DRA, not OpenDRA driver. Please fix in two places.

PR #353, opened 2026-01-19
Gang termination fix

### LINKED ISSUE #277: Gang Termination Doesn't Work
### What happened?

The newly created E2E tests are failing for gang termination.  The root cause is that the E2E tests use cordoning nodes and killing pods afterwards to creates breaches. Currently when podclique status is reconciled,  we avoid breaching if scheduledReplicas < minAvailable.  So the test cases don't actually trigger a breach as intended:

```
// If the number of scheduled pods is less than the minimum available, then minAvailable is not considered as breached.
	// Consider a case where none of the PodCliques have been scheduled yet, then it should not cause the PodGang to be recreated all the time.
	if scheduledReplicas < minAvailable {
		return metav1.Condition{
			Type:               constants.ConditionTypeMinAvailableBreached,
			Status:             metav1.ConditionFalse,
			Reason:             constants.ConditionReasonInsufficientScheduledPods,
			Message:            fmt.Sprintf("Insufficient scheduled pods. expected at least: %d, found: %d", minAvailable, scheduledReplicas),
			LastTransitionTime: now,
		}
	}
```

This also happens inside the PCSG status reconciliation as well:

```
	if scheduledReplicas < minAvailable {
		return metav1.Condition{
			Type:    constants.ConditionTypeMinAvailableBreached,
			Status:  metav1.ConditionFalse,
			Reason:  constants.ConditionReasonInsufficientScheduledPCSGReplicas,
		}
	}
```

I tried to a couple of "easy" fixes just to see. They didn't work for various reasons but maybe someone else has the needed context. 

Here's a quick summary of things tried:

## One-Way Ratchet
Once a PCLQ is healthy, track that it achieved this in it's conditions.  This didn't work because podcliques were getting re-created in certain situations, this status gets lost as well. 

## Grace Period
I tried implementing a grace period, during which pods are not marked as in breach, but afterwards they are. There ended up being a number of edge cases and the tests wouldn't pass. 

_No response_

### Environment

- Kubernetes version
- Grove version
- Scheduler details

gflarity:
#### What type of PR is this?

/kind feature
/kind bug
/kind api

This PR fixes and improves gang termination behavior in Grove with several key changes:

1. **Gang termination is now opt-in** - `terminationDelay` no longer defaults to 4 hours. When nil, gang termination is disabled entirely for the PodCliqueSet.

2. **Only ready pods count for breach detection** - Previously breach calculation considered scheduled pods; now it only counts ready pods, which more accurately reflects actual workload availability.

3. **PCSG-level terminationDelay override** - Individual PodCliqueScalingGroups can now override the PCS-level termination delay, allowing finer-grained control over gang termination timing.

4. **Bug fix: nil pointer when base gang doesn't exist** - Fixed a crash when the base gang no longer exists after gang termination.

Fixes #277

- **Start with the design doc** at `docs/designs/gang-termination.md` - it provides a high-level overview of the gang termination feature after these changes. Happy to update the code alongside any design doc feedback.
- Comprehensive E2E tests have been added covering multiple gang termination scenarios (GT1-GT5)
- Validation ensures PCSG-level `terminationDelay` can only be set when PCS-level `terminationDelay` is set

```release-note
Gang termination is now opt-in. The `terminationDelay` field no longer defaults to 4 hours - when not set, gang termination is disabled. To enable gang termination, explicitly set `spec.template.terminationDelay` on PodCliqueSet.  Administrators who were expecting Gang Termination at the PodCliqueSet or PodCliqueScalingGroup levels after 4 hours should update their PodCliqueSets according. 

Added `WasAvailable` status field to PodClique to track whether a workload has ever reached its MinAvailable threshold. Gang termination only triggers for workloads that were previously available, preventing termination during initial creation and startup.

Added `terminationDelay` field to PodCliqueScalingGroupConfig allowing per-PCSG override of the PCS-level termination delay.
```

```docs
docs/designs/gang-termination.md - Comprehensive design document explaining gang termination behavior, configuration, and debugging
```


docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+- **Multi-Level Granularity:** Support gang termination at PCLQ, PCSG, and PCS levels
+- **Opt-In Behavior:** Gang termination is disabled by default and must be explicitly enabled
+
+## Scope and Limitations
+
+**Limitations:**
+
+- **Ready Pods Only (PCLQ level):** Only ready pods count toward availability—starting pods are not considered
+- **Sticky WasAvailable:** Once a PodClique has reached MinAvailable, the WasAvailable flag never reverts, even if the workload is later degraded and recovers
+- **No Partial Recovery:** When gang termination triggers, the entire PCS replica is deleted—there is no attempt to recover individual failed pods
```
shayasoolin: Not 100% clear to me - do you mean that when gang termination triggers due to a breach in a certain PCLQ (for example), then the whole PCS replica is deleted? If so - can you please clarify the description? Also, if this is the case, the suffix of "individual failed pods" is also a bit unclear, as it's more about individual pod cliques, right? Individual pods are not expected to be recovered - this is the whole point behind gang termination, we terminate the gang after giving up of the individual pods that can't recover.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -230,6 +230,16 @@ func (v *pcsValidator) validatePodCliqueScalingGroupConfigs(fldPath *field.Path)
...
 		}
+
+		// validate TerminationDelay field
+		if scalingGroupConfig.TerminationDelay != nil {
+			// PCSG terminationDelay can only be set if PCS-level terminationDelay is set (gang termination is enabled)
+			if v.pcs.Spec.Template.TerminationDelay == nil {
+				allErrs = append(allErrs, field.Forbidden(fldPath.Index(i).Child("terminationDelay"), "terminationDelay can only be set on PodCliqueScalingGroupConfig when PodCliqueSetTemplateSpec.terminationDelay is set (gang termination is enabled)"))
+			} else if scalingGroupConfig.TerminationDelay.Duration <= 0 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Index(i).Child("terminationDelay"), scalingGroupConfig.TerminationDelay, "terminationDelay must be greater than 0"))
+			}
```
shayasoolin, line 241: and another else-if case, to validate that the PCSG-specific termination delay is not greater than the PCS one.

shayasoolin (commented):
Great job!

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
gflarity: See https://docs.google.com/document/d/11RvFMS1l5RH_FY54G6wd1Q0wSJ2RCobiPRAs2vAhzgM/edit?disco=AAAByqL0jIQ. I discussed with @nvrohanv and @athreesh and while I don't see why someone would do this, we're not going to babysit this.

docs/designs/gang-termination.md
gflarity: I updated the wording, does it make more sense now?

athreesh (dismissed):

docs/designs/gang-termination.md
shayasoolin: Yes, thanks, just fix a typo "PodCliqueswithin" -> "PodCliques within"

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -225,6 +226,12 @@ type PodCliqueScalingGroupConfig struct {
 	// +optional
 	// +kubebuilder:default=1
 	MinAvailable *int32 `json:"minAvailable,omitempty"`
+	// TerminationDelay overrides the PodCliqueSet-level terminationDelay for this scaling group.
+	// Can only be set if PodCliqueSetTemplateSpec.TerminationDelay is set (gang termination is enabled).
+	// When set, this value is used instead of the PodCliqueSet-level terminationDelay for gang termination
+	// decisions affecting this scaling group's replicas.
```
sanjaychatterjee, line 232: Do we need validation that this value should be less than or equal to the PCS one?

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
sanjaychatterjee: I agree with @shayasoolin. PCSG termination delay should not be more than PCS. We need to provide API where results are deterministic.

operator/internal/controller/podcliqueset/components/podcliquesetreplica/gangterminate.go
```diff
@@ -163,9 +175,11 @@ func (r _resource) getExistingPCLQsByNames(ctx context.Context, namespace string
 	return pclqs, notFoundPCLQFQNs, nil
 }
 
-// getMinAvailableBreachedPCSGInfo filters PodCliqueScalingGroups that have grovecorev1alpha1.ConditionTypeMinAvailableBreached set to true.
+// getMinAvailableBreachedPCSGInfoWithEffectiveDelay filters PodCliqueScalingGroups that have
+// grovecorev1alpha1.ConditionTypeMinAvailableBreached set to true, using the effective terminationDelay
+// for each PCSG (PCSG override if set, otherwise PCS default).
 // It returns the names of all such PodCliqueScalingGroups and minimum of all the waitDurations.
-func getMinAvailableBreachedPCSGInfo(pcsgs []grovecorev1alpha1.PodCliqueScalingGroup, terminationDelay time.Duration, since time.Time) ([]string, time.Duration) {
+func getMinAvailableBreachedPCSGInfoWithEffectiveDelay(pcsgs []grovecorev1alpha1.PodCliqueScalingGroup, pcs *grovecorev1alpha1.PodCliqueSet, since time.Time) ([]string, time.Duration) {
```
sanjaychatterjee, line 182: Can we add a unit test?

operator/internal/controller/podcliquescalinggroup/components/podclique/sync.go
```diff
@@ -76,17 +80,41 @@ func (r _resource) prepareSyncContext(ctx context.Context, logger logr.Logger, p
...
 
 	// pre-compute expected PodTemplateHash for each PCLQ
 	syncCtx.expectedPCLQPodTemplateHashMap = getExpectedPCLQPodTemplateHashMap(syncCtx.pcs, pcsg)
 
 	return syncCtx, nil
 }
 
+// findMatchingPCSGConfig finds the PodCliqueScalingGroupConfig that matches the given PCSG.
+// It returns nil if no matching config is found.
+func findMatchingPCSGConfig(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev1alpha1.PodCliqueScalingGroup) *grovecorev1alpha1.PodCliqueScalingGroupConfig {
```
sanjaychatterjee, line 103: Can we add a unit test?

operator/api/core/v1alpha1/podclique.go
```diff
@@ -121,6 +121,12 @@ type PodCliqueStatus struct {
 	// ScheduledReplicas is the number of Pods that have been scheduled by the kube-scheduler.
 	// +kubebuilder:default=0
 	ScheduledReplicas int32 `json:"scheduledReplicas"`
+	// WasAvailable indicates whether the PodClique has ever reached its MinAvailable threshold.
+	// Once set to true, it remains true for the lifetime of the PodClique.
+	// This field is used for gang termination: a PodClique can only be considered in breach of
+	// MinAvailable if it was previously available (WasAvailable=true).
+	// +kubebuilder:default=false
+	WasAvailable bool `json:"wasAvailable"`
```
sanjaychatterjee: Do we need this in the API? Wouldn't an update event tell us if the old version of the object was available but the new version is not?

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
gflarity: Let's chat with @nvrohanv and @athreesh. I'm ambivalent tbh, but we did a final pass on the requirements and this was the call.

operator/api/core/v1alpha1/podclique.go
gflarity: I mentioned this in the "GREP":

> **Motivation for WasAvailable:**
> 
> 1. **Ignore Initial Startup:** New PodCliques should not be terminated simply because they haven't yet reached their MinAvailable threshold. The WasAvailable gate ensures gang termination only triggers for workloads that were previously healthy and then became degraded—not for workloads that are still starting up.
> 
> 2. **Operator Crash Resilience:** The WasAvailable flag is persisted in the PodClique's status (not held in memory). This ensures that if the Grove operator crashes and restarts during the transition to availability, the flag is not lost. Because it's a sticky bit that never reverts to `false`, the operator can safely restart at any time without missing the transition or incorrectly treating a previously-available PodClique as never-available.

I'll go into a bit more detail in the doc, but  we do need to persist this information for a few reasons:
  • If the operator crashes and restarts, we can't reliably infer whether a PodClique was ever available just from the conditions. The UpdateInProgress reason doesn't tell us either way, so we'd be guessing.
  • There's also a race condition where pods could briefly reach MinAvailable then die before we reconcile. Without the sticky bit, we'd lose that info and wrongly treat it as "never available."

I also considered just adding another condition, but decided against it as it follows the Kubernetes pattern/convention of using status fields for internal bookkeeping (like observedGeneration) rather than conditions. Conditions are more for user-facing state they'd alert on it. So ultimately I went with the separate status boolean.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
nvrohanv: Yes the reason why I don't think we should babysit this is we assume if you choose to do gang termination you know what you are doing. In that scenario I can understand someone being more tolerant if out of 3 pcsg one is down they give it a longer amount of time to recover than if the pcs minAvailable is breached then they have less tolerance because the system is not functional.  Since theres a valid use case for setting it to be more and we assume this is a power-user feature I think we should provide full flexibility.

nvrohanv (dismissed):

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
+# Gang Termination Design Doc
```
unmarshall: `designs` directory is going to removed. Please rework on this document and use the GREP template instead. Put the document under `proposals`. Before writing a GREP please also read `docs/proposals/README.md`

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+
+# Overview
+
+Gang termination is a mechanism in Grove that ensures when a component becomes unhealthy (falls below minimum availability threshold), the entire group (gang) is terminated rather than leaving it in a degraded state. This is particularly important for distributed AI inference workloads where partial availability often means the workload cannot function properly.
```
unmarshall: Can you define a `component`?
Ideally one should introduce what a `Gang` is, then talk about gang termination. 

>Gang termination is a mechanism in Grove

Need to change this as well as it is not a new concept introduced in Grove but a commonly known concept in gang scheduled workloads.

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+
+## Abbreviations
```
unmarshall: If we are going to have this section, then we will have to repeat it for every proposal that is written, which i think is quite unnecessary. Instead create a separate document for commonly used abbreviations and all documents can refer to that.

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+
+| Abbreviation | Full Name | Description |
+|--------------|-----------|-------------|
+| PCS | PodCliqueSet | Grove CRD that manages a set of PodCliques and PodCliqueScalingGroups |
```
unmarshall: PCS does not manage PCLQ and PCSG, it is the controllers within Grove that manage these resources.

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+| PCLQ | PodClique | Grove CRD representing a group of related pods |
```
unmarshall: PCLQ is a group of pods that share the same PodSpecTemplate and serve a single purpose.

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+| PCSG | PodCliqueScalingGroup | Grove CRD that manages scaling of PodCliques |
```
unmarshall: Does not manage PodCliques.

unmarshall: All these are clearly defined https://github.com/ai-dynamo/grove?tab=readme-ov-file#how-it-works

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+- Block other workloads from being scheduled
+- Leave the system in an undefined state
+
+Gang termination addresses this by providing an automated mechanism to clean up degraded workloads, freeing resources for healthy workloads to be scheduled.
+
+## Background
+
+### Gang Scheduling in Grove
+
+Grove implements gang scheduling through the `MinAvailable` field, which ensures that a minimum number of units can be scheduled together before any are created. At the PodClique level, this means a minimum number of pods; at the PodCliqueScalingGroup level, this means a minimum number of PCSG replicas. This prevents partial scheduling where some units run while others are stuck pending.
```
unmarshall: You start the sentence with `Grove implements gang scheduling through the MinAvailable field`. This is a very simplistic view and needs to be enhanced.

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+
+### The Problem with Partial Availability
+
+After a workload is running, individual pods may fail due to:
+- Node failures
+- OOM kills
+- Application crashes
+- Network partitions
+
+Without gang termination, a workload that loses critical pods remains in a degraded state indefinitely. The `MinAvailable` threshold used for gang scheduling is reused to define "acceptable availability" for gang termination.
```
unmarshall: > The `MinAvailable` threshold used for gang scheduling is reused to define "acceptable availability" for gang termination.

Maybe you can use the API docs at `PodCliqueSpec.MinAvailable` and `PodCliqueScalingGroupConfig.MinAvailable`.
`acceptable availability` for gang termination needs to be rephrased as that is not clear.

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+
+# Goals
+
+- **Automatic Cleanup:** Automatically terminate workloads that fall below minimum availability threshold
+- **Configurable Grace Period:** Allow time for Kubernetes scheduler to recover before terminating
+- **Startup Protection:** Avoid terminating workloads that haven't yet reached full availability
+- **Rolling Update Safety:** Prevent false terminations during expected pod churn
+- **Multi-Level Granularity:** Support gang termination at PCLQ, PCSG, and PCS levels
```
unmarshall: What is gang termination at PCS level?

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+- **Opt-In Behavior:** Gang termination is disabled by default and must be explicitly enabled
+
+## Scope and Limitations
+
+**Limitations:**
+
+- **Ready Pods Only (PCLQ level):** Only ready pods count toward availability—starting pods are not considered
```
unmarshall: how is this a limitation?

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+- **Sticky WasAvailable:** Once a PodClique has reached MinAvailable, the WasAvailable flag never reverts, even if the workload is later degraded and recovers
+- **No Partial Recovery:** When gang termination triggers, the affected replica (PCS or PCSG) is deleted in its entirety, healthy PodCliques within that replica are not preserved.
```
unmarshall: how is this a limitation?

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+| Level | Resource | MinAvailable Meaning |
+|-------|----------|---------------------|
+| PCLQ | PodClique | Minimum number of **pods** that must be available |
+| PCSG | PodCliqueScalingGroup / PodCliqueScalingGroupConfig | Minimum number of **PCSG replicas** that must be available |
+
+**PCSG replica availability:** A PCSG replica is considered unavailable (breached) if **any** of its constituent PCLQs has `MinAvailableBreached=True`. The PCSG aggregates the breach status from all PCLQs within each replica to determine overall PCSG replica availability.
+
+### MinAvailableBreached Condition
+
+A status condition (`ConditionTypeMinAvailableBreached`) set on PodClique and PodCliqueScalingGroup resources when available replicas fall below `MinAvailable`. The condition includes:
```
unmarshall:
```suggestion
A status condition (`MinAvailableBreached`) set on PodClique and PodCliqueScalingGroup resources when available replicas fall below `MinAvailable`. The condition includes:
```

docs/designs/gang-termination.md
```diff
@@ -0,0 +1,375 @@
...
+A configurable duration specified at `PodCliqueSet.Spec.Template.TerminationDelay`. After `MinAvailable` is breached, this delay must pass before gang termination is triggered. This gives the Kubernetes scheduler time to potentially recover by rescheduling pods.
+
+**Important:** Gang termination is **disabled by default**. When `terminationDelay` is `nil` (not set), no gang termination will occur. To enable gang termination, you must explicitly set `terminationDelay` to a non-zero duration.
+
+### PCSG TerminationDelay Override
+
+Individual `PodCliqueScalingGroupConfig` entries can override the PCS-level `terminationDelay`:
+
+- If PCS-level `terminationDelay` is nil, gang termination is disabled for the entire PCS
+- If PCS-level `terminationDelay` is set, each PCSG can optionally override it with its own delay
```
unmarshall: Why did we choose to have this behavior? If we have `TerminationDelay` at PCS and PCSG level then it should be respected when defined and termination delay is enabled. The issue is that `nil` value of PCS termination delay has been taken as an indication that this feature has been enabled or disabled. This IMHO is not so nice and also not very intuitive.

From API design perspective when we define a delay at multiple levels, thus allowing overriding, then if PCS does not define TerminationDelay but this is defined at the PCSG level then it should be honored as long as this feature is enabled.

unmarshall (commented):
1/n reviews

docs/designs/gang-termination.md
nvrohanv: gang termination at PCS is whole PCS replica getting kicked back to scheduling.

docs/designs/gang-termination.md
nvrohanv: Agreed this seems like a feature. Also its only once-ready pods right.

docs/designs/gang-termination.md
nvrohanv: Also agreed, I believe this would only happen if minAvailable is breached at which point we want the whole thing triggered.

docs/designs/gang-termination.md
nvrohanv: hmm it seems a bit odd to do it that way as well because since minAvailable defaults to 1 (if I remember correctly) then it seems like you should be getting gang semantics throughout. If anything I would lean towards validating that you cant set it on just a pcsg. Something we should discuss more.

docs/designs/gang-termination.md
unmarshall: Maybe you can just move this document out of this PR. Lets create a single document for gang scheduling and termination. This doc as it stands today needs a bit of work else it will hold your fix for no good reason.

docs/designs/gang-termination.md
unmarshall: For PCLQ if `minAvailable` is not set then it is defaulted to pclq template replicas. For PCSG it defaults to 1. Gang semantics (scheduling and termination) applies to identified pod-gangs. Currently there are only 2 types of PodGangs:

**Base PodGang**
Comprises of:
* `minAvailable` replicas of stand-alone PCLQs
* `minAvailable` replicas of PCSG. Within PCSG if constituent PCLQ defines `minAvailable` then only those many pods.
This means that only these many pods are required for a functional application. If the number goes below that and stays like that for terminationDelay seconds then its time to terminate the gang.

**Scaled PodGang**
Similar behavior applies to scaled PodGang where only `minAvailable` of constituent PCLQs are considered to determine if `minAvailableBreached` condition is true.

Now what happens when you do not define `TerminationDelay` - i believe we then need to have a default (very large) termination delay.

Consider the following PCS composition:
* Standalone PCLQ - router
* PCSG - comprises of decode & prefill PCLQs

Case #1
You define a terminationDelay of 1hr on PCS and 2hr on PCSG. 
**Now what does it mean for the base PodGang?**
If router has minAvailableBreached set to true and it has already exceeded 1hr then it will gang terminate the base PodGang. A higher termination delay on PCSG would not come into play here.

**What is the behavior for the scaled PodGang?**
This is relatively simple. PCS termination delay will not come into effect. For all Scaled podgangs termination delay defined at the PCSG will only be considered.

Case #2 
You define a terminationDelay of 2hr on PCS and 1hr on PCSG. 
**Now what does it mean for the base PodGang?**
For base pod gang only PCS termination delay will apply. So if the PCSG PCLQs have their minAvailableBreached condition set for more than 1hr but less than 2hr, there will not be any gang termination. From the API perspective this behavior is utterly confusing.

So as soon as we introduce 2 level termination delay we should consider the behavior carefully.

unmarshall (commented):
2/n review comments

operator/api/core/v1alpha1/podclique.go
gflarity: **Update** - After going through the edge case below, I do see that if we respect the previous MinAvailableBreached=True on operator restart (and perhaps all the time) and not require a "FLIP" it would solve this edge case. I'll revive the implementation and see if there are other addition edge cases that pop up. That said, WasAvailable definitely has advantages in terms of simplicity and being explicitly observable though. 

---

For more background, here's the initial approach I tried and abandoned:

The PodClique controller watches "owned" Pods using a predicate. This triggers PodClique controller reconciliation when pods are deleted or have status changes (specifically when the Ready condition transitions). During this reconciliation, the "old" (current but to be reconciled) PodClique object is retrieved from the Kubernetes API. We then "Reconcile" and determine what the new values should be for status fields like PodClique.Status.ReadyReplicas. At this point we essentially have an "old" and "new" PodClique object, allowing us to notice changes in PodClique.Status.ReadyReplicas. Using this approach, we could avoid setting MinAvailableBreached to true unless oldPodClique.Status.ReadyReplicas >= minAvailable in the "old" object. This is insufficient given our requirements however. Consider the following edge case:

#### Common Starting Point

| Time | Event | MinAvailableBreached Status | Reason | WasAvailable |
|------|-------|------------------|--------|--------------|
| T0 | PCLQ created, 0 pods | False | InsufficientScheduledPods | false |
| T1 | 3 pods scheduled, starting | False | InsufficientScheduledPods | false |

At T1, both scenarios are identical: 3 pods scheduled and starting.

#### Scenario A: Workload Becomes Available Then Degrades

| Time | Event | MinAvailableBreached Status | Reason | WasAvailable |
|------|-------|------------------|--------|--------------|
| T2 | All 3 pods become ready | False | SufficientReadyPods | **true** |
| T3 | 1 pod dies (2/3 ready) | True | InsufficientReadyPods | true |
| T4 | Operator crashes | - | - | - |
| T5 | Operator restarts | True | InsufficientReadyPods | ? |

**Should gang terminate? YES** - workload was healthy, now degraded.

#### Scenario B: Workload Never Becomes Available

| Time | Event | MinAvailableBreached Status | Reason | WasAvailable |
|------|-------|------------------|--------|--------------|
| T2 | 1 pod crashes during init | True | InsufficientReadyPods | **false** |
| T3 | Operator crashes | - | - | - |
| T4 | Operator restarts | True | InsufficientReadyPods | ? |

**Should gang terminate? NO** - workload never achieved availability.

#### State After Restart Comparison

| Field | Scenario A (T5) | Scenario B (T4) |
|-------|-----------------|-----------------|
| `Status` | True | True |
| `Reason` | InsufficientReadyPods | InsufficientReadyPods |
| `ScheduledReplicas` | 3 | 3 |
| `ReadyReplicas` | 2 | 2 |
| **Should gang terminate?** | **YES** | **NO** |

The condition and status fields are **identical**. Without `WasAvailable`, the operator cannot distinguish between these scenarios.

gflarity: Removed in favour of respecting previously set status condition.

PR #356, opened 2026-01-20
Danbar/e2e rolling update 10

danbar2:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/operator/e2e/tests/rolling_updates_test.go
+++ b/operator/e2e/tests/rolling_updates_test.go
@@ -49,10 +49,8 @@ func Test_RU7_RollingUpdatePCSPodClique(t *testing.T) {
 		t.Fatalf("Failed to update PodClique spec: %v", err)
 	}
 
-	// With MaxSurge rolling update strategy, each pod requires: create surge pod -> wait Ready -> delete old pod
-	// This takes longer than delete-first approach, so we use a longer timeout
 	tcLongTimeout := tc
-	tcLongTimeout.Timeout = 5 * time.Minute
+	tcLongTimeout.Timeout = 1 * time.Minute
 	if err := waitForRollingUpdateComplete(tcLongTimeout, 1); err != nil {
 		// Diagnostics will be collected automatically by cleanup on test failure
 		t.Fatalf("Failed to wait for rolling update to complete: %v", err)
--- a/operator/internal/controller/podclique/reconcilestatus_test.go
+++ b/operator/internal/controller/podclique/reconcilestatus_test.go
@@ -1,0 +1,195 @@
+// /*
+// Copyright 2025 The Grove Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// */
+
+package podclique
+
+import (
+	"testing"
+	"time"
+
+	apicommon "github.com/ai-dynamo/grove/operator/api/common"
+	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
+
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+// TestMutateUpdatedReplica tests the mutateUpdatedReplica function across different PodClique states
+func TestMutateUpdatedReplica(t *testing.T) {
+	tests := []struct {
+		name                    string
+		pclq                    *grovecorev1alpha1.PodClique
+		existingPods            []*corev1.Pod
+		expectedUpdatedReplicas int32
+	}{
+		{
+			name: "rolling update in progress - count pods with new hash",
+			pclq: &grovecorev1alpha1.PodClique{
+				Status: grovecorev1alpha1.PodCliqueStatus{
+					RollingUpdateProgress: &grovecorev1alpha1.PodCliqueRollingUpdateProgress{
+						PodTemplateHash: "new-hash-v2",
+						UpdateStartedAt: metav1.Time{Time: time.Now()},
+						UpdateEndedAt:   nil, // Still in progress
+					},
+					CurrentPodTemplateHash: ptr.To("old-hash-v1"),
+				},
+			},
+			existingPods: []*corev1.Pod{
+				// 3 pods updated to new version
+				createPodWithHash("pod-1", "new-hash-v2"),
+				createPodWithHash("pod-2", "new-hash-v2"),
+				createPodWithHash("pod-3", "new-hash-v2"),
+				// 7 pods still on old version
+				createPodWithHash("pod-4", "old-hash-v1"),
+				createPodWithHash("pod-5", "old-hash-v1"),
+				createPodWithHash("pod-6", "old-hash-v1"),
+				createPodWithHash("pod-7", "old-hash-v1"),
+				createPodWithHash("pod-8", "old-hash-v1"),
+				createPodWithHash("pod-9", "old-hash-v1"),
+				createPodWithHash("pod-10", "old-hash-v1"),
+			},
+			expectedUpdatedReplicas: 3, // Only the 3 pods with new hash
+		},
+		{
+			name: "update just completed - use RollingUpdateProgress hash (edge case)",
+			pclq: &grovecorev1alpha1.PodClique{
+				Status: grovecorev1alpha1.PodCliqueStatus{
+					RollingUpdateProgress: &grovecorev1alpha1.PodCliqueRollingUpdateProgress{
+						PodTemplateHash: "new-hash-v2",
+						UpdateStartedAt: metav1.Time{Time: time.Now().Add(-5 * time.Minute)},
+						UpdateEndedAt:   &metav1.Time{Time: time.Now()}, // Just completed
+					},
+					// CurrentPodTemplateHash not updated yet - still has old hash
+					CurrentPodTemplateHash: ptr.To("old-hash-v1"),
+				},
+			},
+			existingPods: []*corev1.Pod{
+				// All pods updated to new version
+				createPodWithHash("pod-1", "new-hash-v2"),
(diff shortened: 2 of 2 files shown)


operator/e2e/setup/shared_cluster.go
```diff
@@ -586,9 +609,11 @@ func setupRegistryTestImages(registryPort string, images []string) error {
 		if err != nil {
 			// Image doesn't exist locally, need to pull it
 			pullOpts := image.PullOptions{}
```
shayasoolin: This section is duplicate with operator/e2e/setup/k8s_clusters.go:prepullImages, please consider extracting to a commonly used func.

shayasoolin (dismissed):

operator/e2e/setup/shared_cluster.go
```diff
@@ -495,6 +507,117 @@ func (scm *SharedClusterManager) Teardown() {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal auth config: %w", err)
+	}
+
+	return base64.StdEncoding.EncodeToString(encodedJSON), nil
+}
+
+// getDockerAuthForImage returns the base64-encoded auth string for pulling an image
+// by reading credentials from the Docker config file.
+func getDockerAuthForImage(imageName string, registryURL string, configPath string) (string, error) {
```
gflarity: Hmm... What image is requiring Docker Auth? I recently replaced the registry image we were using. Ideally we wouldn't get any images from docker hub (the paid requirement to avoid rate limits is annoying) at all. 

If we can just replace the image instead of worry about auth, I'd prefer it. It's not clear the auth would work in CI either?

operator/e2e/setup/k8s_clusters.go
```diff
@@ -777,22 +777,13 @@ func prepullImages(ctx context.Context, images []string, registryPort string, lo
...
-				_, err = io.Copy(logger.WriterLevel(utils.DebugLevel), pullReader)
-			} else {
-				_, err = io.Copy(io.Discard, pullReader)
+				outputWriter = logger.WriterLevel(utils.DebugLevel)
 			}
-			pullReader.Close()
-			if err != nil {
-				errChan <- fmt.Errorf("failed to read pull output for %s: %w", img, err)
+			if err := pullImageIfNotExists(ctx, dockerClient, img, outputWriter); err != nil {
+				errChan <- err
```
gflarity: We should log if pullImageIfNotExists fails... 


```suggestion
			if err := pullImageIfNotExists(ctx, dockerClient, img, outputWriter); err != nil {
				logger.Debugf("Pull failed for %s: %v", img, err)
				errChan <- err
```

operator/e2e/setup/shared_cluster.go
```diff
@@ -495,6 +507,117 @@ func (scm *SharedClusterManager) Teardown() {
 	}
 }
 
+// parseRegistryFromImage extracts the registry host from an image name.
+// Returns the provided defaultRegistry if the image doesn't specify a custom registry.
+func parseRegistryFromImage(imageName string, defaultRegistry string) string {
```
gflarity: We can replace this manual string parsing with the distribution/reference package, which is already an indirect dependency in our go.mod. It handles OCI edge cases and implicit Docker Hub domains (like index.docker.io) much more reliably than strings.SplitN.

```go
named, _ := reference.ParseNormalizedNamed(imageName)
registryHost := reference.Domain(named)
```

operator/e2e/setup/shared_cluster.go
gflarity: Also, if we really do need auth, the current manual JSON parsing of .docker/config.json will fail for developers using credential helpers (e.g., osxkeychain, pass, or ecr-login), as the password won't be in the JSON file. Since we already have github.com/docker/cli in our tree, we should use config.LoadDefaultConfigFile(os.Stderr) to handle auth natively.

```go
// Suggestion: Leverage Docker's own config loader to support credential helpers
cfg := config.LoadDefaultConfigFile(os.Stderr)
authConfig, err := cfg.GetAuthConfig(registryHost)
```

operator/e2e/setup/shared_cluster.go
```diff
@@ -495,6 +507,117 @@ func (scm *SharedClusterManager) Teardown() {
...
+		ioutil.CloseQuietly(pullReader)
+		if pullErr != nil {
+			return fmt.Errorf("failed to read pull output for %s: %w", imageName, pullErr)
+		}
+	}
+	return nil
+}
+
 // setupRegistryTestImages sets up test images in the registry
 func setupRegistryTestImages(registryPort string, images []string) error {
```
gflarity: Looks like I should have just used google/go-containerregistry as well here to significantly simplify this entire flow. Looks like there's a package called 'crane'  that would allow us to replace the manual Pull → Tag → Push logic (and all the associated auth boilerplate) with a single crane.Copy() call.

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -124,10 +124,15 @@ func mutateReplicas(pclq *grovecorev1alpha1.PodClique, podCategories map[corev1.
...
 	// else take it from the PodClique.Status.CurrentPodTemplateHash field
 	if componentutils.IsPCLQUpdateInProgress(pclq) {
 		expectedPodTemplateHash = pclq.Status.RollingUpdateProgress.PodTemplateHash
+	} else if componentutils.IsLastPCLQUpdateCompleted(pclq) {
+		// Rolling update just completed (UpdateEndedAt is set) - use the new hash from RollingUpdateProgress.
+		// This handles the window between markRollingUpdateEnd() and mutateCurrentHashes() where
+		// CurrentPodTemplateHash hasn't been updated yet but the rolling update has finished.
+		expectedPodTemplateHash = pclq.Status.RollingUpdateProgress.PodTemplateHash
 	} else if pclq.Status.CurrentPodTemplateHash != nil {
 		// CurrentPodTemplateHash will be set if RollingUpdateProgress is nil or if the last update has completed.
```
gflarity: This comment is now stale with change above:

```suggestion
		// Steady state: no rolling update has ever been tracked for this PodClique.
		// This handles fresh PodCliques that have been reconciled but never updated.
```

operator/e2e/setup/shared_cluster.go
danbar2: decided to revert the e2e setup changes, @Ronkahn21 raised concerns about it too and it does not related to the bug fix. Will do in future PR

operator/e2e/setup/shared_cluster.go
danbar2: decided to revert the e2e setup changes, @Ronkahn21 raised concerns about it too and it does not related to the bug fix. Will do in future PR

operator/e2e/setup/k8s_clusters.go
danbar2: decided to revert the e2e setup changes, @Ronkahn21 raised concerns about it too and it does not related to the bug fix. Will do in future PR

operator/e2e/setup/shared_cluster.go
danbar2: decided to revert the e2e setup changes, @Ronkahn21 raised concerns about it too and it does not related to the bug fix. Will do in future PR

operator/e2e/setup/shared_cluster.go
```diff
@@ -495,6 +507,117 @@ func (scm *SharedClusterManager) Teardown() {
...
+	// Check if image already exists locally
+	_, _, err := cli.ImageInspectWithRaw(ctx, imageName)
+	if err != nil {
+		// Image doesn't exist locally, need to pull it
+		pullOpts := image.PullOptions{}
+		authStr, authErr := getDockerAuthForImage(imageName, defaultDockerRegistry, defaultDockerConfigPath)
+		if authErr == nil {
+			pullOpts.RegistryAuth = authStr
+		}
+		// Ignore auth errors and attempt pull without auth - may work for public images
```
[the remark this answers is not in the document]
danbar2: decided to revert the e2e setup changes

operator/e2e/tests/rolling_updates_test.go
```diff
@@ -44,8 +49,10 @@ func Test_RU7_RollingUpdatePCSPodClique(t *testing.T) {
 		t.Fatalf("Failed to update PodClique spec: %v", err)
 	}
 
+	// With MaxSurge rolling update strategy, each pod requires: create surge pod -> wait Ready -> delete old pod
+	// This takes longer than delete-first approach, so we use a longer timeout
 	tcLongTimeout := tc
-	tcLongTimeout.Timeout = 1 * time.Minute
+	tcLongTimeout.Timeout = 5 * time.Minute
```
gflarity: Are we sure 5 minutes is required?  Seems very long.

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -124,12 +124,13 @@ func mutateReplicas(pclq *grovecorev1alpha1.PodClique, podCategories map[corev1.
 // mutateUpdatedReplica calculates and sets the number of pods with the expected template hash
 func mutateUpdatedReplica(pclq *grovecorev1alpha1.PodClique, existingPods []*corev1.Pod) {
 	var expectedPodTemplateHash string
-	// If the PCLQ update is in progress then take the expected PodTemplateHash from the PodClique.Status.RollingUpdateProgress.PodCliqueSetGenerationHash field
-	// else take it from the PodClique.Status.CurrentPodTemplateHash field
-	if componentutils.IsPCLQUpdateInProgress(pclq) {
+	// If RollingUpdateProgress exists (update in progress or recently completed), use the target hash from it.
+	// This covers both the active update phase and the window after completion before CurrentPodTemplateHash is synced.
+	if pclq.Status.RollingUpdateProgress != nil {
```
gflarity, line 129: Consider adding a unit test for `mutateUpdatedReplica` that specifically covers this "just-completed" state (where `RollingUpdateProgress != nil` and `UpdateEndedAt != nil`). The existing e2e tests validate the rolling update completes, but a focused unit test would help prevent regressions and document this edge case more explicitly.

Something like:
```go
{
    name: "update just completed - use RollingUpdateProgress hash",
    pclq: &PodClique{
        Status: PodCliqueStatus{
            RollingUpdateProgress: &RollingUpdateProgress{
                PodTemplateHash: "new-hash",
                UpdateEndedAt:   &metav1.Time{Time: time.Now()},
            },
            CurrentPodTemplateHash: ptr.To("old-hash"),
        },
    },
    // ... pods with "new-hash" should be counted as updated
}
```

gflarity (commented):
Looking good, one question and one comment.

operator/e2e/tests/rolling_updates_test.go
danbar2: it actually ends in 30s, so we can keep it with 1m if we hit timeout I suggest to increase that

PR #358, opened 2026-01-20
Gather all operator logs on e2e test failure

gflarity:
#### What type of PR is this?

/kind cleanup

This PR changes the e2e test diagnostic utility to gather all operator logs after a test failure, instead of just the last 2000 lines. This ensures complete diagnostic information is available when debugging test failures, rather than potentially missing earlier relevant logs that occurred before the failure.

NONE

Small change to `operator/e2e/tests/debug_utils.go`:
- Removed the `operatorLogLines` constant (previously set to 2000)
- Removed the `TailLines` option from the pod log request, which causes it to fetch all logs

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/e2e/tests/debug_utils.go
+++ b/operator/e2e/tests/debug_utils.go
@@ -120,20 +120,15 @@ func (m *multiWriteCloser) Close() error {
 	return errors.Join(errs...)
 }
 
-// createDiagnosticsOutput creates writer(s) for diagnostics output based on GROVE_E2E_DIAG_MODE.
+// createDiagnosticsOutput creates writer(s) for diagnostics output based on the provided mode.
 // Mode values:
 //   - "stdout": write to stdout only
 //   - "file": write to timestamped file only (default)
 //   - "both": write to both stdout and file
 //
-// When writing to file, GROVE_E2E_DIAG_DIR specifies the directory (default: current dir, fallback to /tmp).
+// The diagDir parameter specifies the directory for file output (empty string uses current dir, fallback to /tmp).
 // Returns the writer, filename (empty if stdout-only), and any error.
-func createDiagnosticsOutput(testName string) (io.WriteCloser, string, error) {
-	mode := os.Getenv(DiagnosticsModeEnvVar)
-	if mode == "" {
-		mode = DiagnosticsModeFile // default
-	}
-
+func createDiagnosticsOutput(testName, mode, diagDir string) (io.WriteCloser, string, error) {
 	var writers []io.WriteCloser
 	var filename string
 
@@ -144,7 +139,7 @@ func createDiagnosticsOutput(testName string) (io.WriteCloser, string, error) {
 
 	// Add file if mode is file or both
 	if mode == DiagnosticsModeFile || mode == DiagnosticsModeBoth {
-		file, name, err := createDiagnosticsFile(testName)
+		file, name, err := createDiagnosticsFile(testName, diagDir)
 		if err != nil {
 			// If we can't create a file but stdout is available, continue with stdout only
 			if mode == DiagnosticsModeBoth {
@@ -165,8 +160,9 @@ func createDiagnosticsOutput(testName string) (io.WriteCloser, string, error) {
 	return newMultiWriteCloser(writers...), filename, nil
 }
 
-// createDiagnosticsFile creates a timestamped diagnostics file
-func createDiagnosticsFile(testName string) (*os.File, string, error) {
+// createDiagnosticsFile creates a timestamped diagnostics file.
+// The diagDir parameter specifies the directory for the file (empty string uses current dir, fallback to /tmp).
+func createDiagnosticsFile(testName, diagDir string) (*os.File, string, error) {
 	// Sanitize test name for use in filename (replace / with _)
 	sanitizedName := strings.ReplaceAll(testName, "/", "_")
 
@@ -174,8 +170,7 @@ func createDiagnosticsFile(testName string) (*os.File, string, error) {
 	timestamp := time.Now().Format("2006-01-02_15-04-05")
 	baseFilename := fmt.Sprintf("e2e-diag-%s_%s.log", sanitizedName, timestamp)
 
-	// Determine the directory for the diagnostics file
-	diagDir := os.Getenv(DiagnosticsDirEnvVar)
+	// Use provided directory if specified
 	if diagDir != "" {
 		filename := filepath.Join(diagDir, baseFilename)
 		file, err := os.Create(filename)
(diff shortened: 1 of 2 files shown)


danbar2 (approved):

.github/workflows/e2e-test.yaml
```diff
@@ -81,6 +81,8 @@ jobs:
           skaffold version
 
       - name: Run e2e tests - ${{ matrix.test_name }}
+        env:
+          GROVE_E2E_DIAG_TO_STDOUT: "1"
         run: |
           cd operator
           echo "> Preparing charts (copying CRDs)..."
```
Ronkahn21: **Nit**: While I know this isn't strictly in scope for this PR, it would be great to aim for parity with our Makefile workflow. Keeping the environment variables and diagnostic behavior consistent across both the Makefile and the test runner makes the DX much smoother and prevents 'configuration whiplash' for the team.

operator/e2e/tests/debug_utils.go
```diff
@@ -35,12 +39,18 @@ const (
...
+	// DiagnosticsToStdoutEnvVar is the environment variable that controls diagnostics output.
+	// If set to any non-empty value, diagnostics are printed to stdout.
+	// If not set or empty, diagnostics are written to a timestamped file.
+	DiagnosticsToStdoutEnvVar = "GROVE_E2E_DIAG_TO_STDOUT"
+
+	// DiagnosticsDirEnvVar is the environment variable that specifies the directory
+	// where diagnostics files should be written. If not set, files are written to
+	// the current directory. This is ignored if DiagnosticsToStdoutEnvVar is set.
+	DiagnosticsDirEnvVar = "GROVE_E2E_DIAG_DIR"
 )
```
Ronkahn21: ###  Suggestion: Support simultaneous outputs (Both)

I'd suggest evolving the "either/or" logic into a more flexible **Mode** pattern. In E2E testing, we often need **both**: `stdout` for real-time visibility in the CI console and `file` for a persistent artifact we can download and analyze later.

Instead of having `DIAG_TO_STDOUT` override the file output, we could use a single "Mode" variable:

* **`GROVE_E2E_DIAG_MODE`**: Set to `stdout`, `file`, or `both` (defaulting to `both` or `file` for safety).
* **`GROVE_E2E_DIAG_DIR`**: Defines the destination for the file output (used when mode is `file` or `both`).

**Why this helps:**
1. **CI Visibility:** We can see what's happening live in the GitHub Actions runner logs.
2. **Persistence:** If a test fails 40 minutes in, we don't have to scroll through thousands of lines of terminal output; we can just grab the timestamped log file from the artifacts.

gflarity: This makes sense , I just didn't realize GitHub *easily* supported uploading artifacts from CI runs. But looks like there's an action we can so it's pretty straight forward 👍

.github/workflows/e2e-test.yaml
gflarity: Done.

operator/e2e/tests/debug_utils.go
```diff
@@ -78,34 +81,91 @@ var groveResourceTypes = []groveResourceType{
...
+}
+
+// createDiagnosticsOutput creates writer(s) for diagnostics output based on GROVE_E2E_DIAG_MODE.
+// Mode values:
+//   - "stdout": write to stdout only
+//   - "file": write to timestamped file only (default)
+//   - "both": write to both stdout and file
+//
+// When writing to file, GROVE_E2E_DIAG_DIR specifies the directory (default: current dir, fallback to /tmp).
+func createDiagnosticsOutput(testName string) (*diagnosticsOutput, error) {
```
Ronkahn21: This function is currently handling several distinct responsibilities: parsing configuration, managing filesystem paths, and orchestrating log writers.

To make this more maintainable, we might want to separate these concerns. If we pass in a Config struct and the stdout writer as arguments, it becomes a "pure" factory that's much easier to test in isolation without needing a real environment or disk.

operator/e2e/tests/debug_utils.go
```diff
@@ -78,34 +81,91 @@ var groveResourceTypes = []groveResourceType{
...
+	mode := os.Getenv(DiagnosticsModeEnvVar)
```
Ronkahn21: By calling os.Getenv directly inside this logic, it becomes tightly coupled to the global system environment.

We could improve the testability here by moving the environment variable parsing into a dedicated LoadConfig() helper. That way, we can pass specific configurations into this function during unit tests without affecting the rest of the process.

operator/e2e/tests/debug_utils.go
```diff
@@ -78,34 +81,91 @@ var groveResourceTypes = []groveResourceType{
...
+	if mode == "" {
+		mode = DiagnosticsModeFile // default
+	}
+
+	output := &diagnosticsOutput{}
+	var writers []io.Writer
+
+	// Add stdout if mode is stdout or both
+	if mode == DiagnosticsModeStdout || mode == DiagnosticsModeBoth {
+		writers = append(writers, os.Stdout)
```
Ronkahn21: Depending on a global like os.Stdout makes it difficult to "capture" logs if we ever want to run assertions on our diagnostic output in tests.

If we pass stdout as an io.Writer dependency, it gives us the flexibility to swap it for a buffer or a mock writer whenever we need to.

operator/e2e/tests/debug_utils.go
```diff
@@ -78,34 +81,91 @@ var groveResourceTypes = []groveResourceType{
...
+//   - "both": write to both stdout and file
+//
+// When writing to file, GROVE_E2E_DIAG_DIR specifies the directory (default: current dir, fallback to /tmp).
+func createDiagnosticsOutput(testName string) (*diagnosticsOutput, error) {
+	mode := os.Getenv(DiagnosticsModeEnvVar)
+	if mode == "" {
+		mode = DiagnosticsModeFile // default
+	}
+
+	output := &diagnosticsOutput{}
```
Ronkahn21: diagnosticsOutput is acting as a container to help track the file handle for later cleanup.

We might be able to simplify this using a MultiWriteCloser pattern (essentially a decorator for multiple io.WriteCloser objects). This would allow us to return a single interface, so the rest of the application doesn't need to worry about whether it's talking to one file, two files, or just the terminal.

operator/e2e/tests/debug_utils.go
```diff
@@ -78,34 +81,91 @@ var groveResourceTypes = []groveResourceType{
...
+		return nil, fmt.Errorf("no valid diagnostics output configured (mode=%s)", mode)
 	}
 
+	// Use MultiWriter if we have multiple outputs
+	if len(writers) == 1 {
+		output.writer = writers[0]
+	} else {
+		output.writer = io.MultiWriter(writers...)
+	}
+
```
Ronkahn21: MultiWriteCloser, we can treat 1 writer or 5 writers exactly the same way,  so the if is redundant 
```go
output.writer = io.MultiWriter(writers...)
```

shayasoolin (approved):

PR #364, opened 2026-01-22
Re-enable the authorizer webhook by default.

renormalize:
#### What type of PR is this?

/kind cleanup

The authorizer webhook was disabled by default due to a potential regression observed around the second alpha release in #223.

However, I have been unable to replicate this issue since, and would like to re-enable it by default. It being enabled by default will also help us in catching any potential bugs before the next release as all new PRs will indirectly check the behavior of the authorizer webhook, along with the E2E tests.

This PR also fixes E2E tests and introduces a new dynamic client which creates a bearer token via `TokenRequest` API using the service account with which Grove operator is started. The authorizer webhook allows calls to change managed resources when it comes from grove service account. 

Previous to this PR, user `system:admin` was making calls to scale-out/in PCLQ/PCSG and with authorizer webhook enabled this will be disallowed as `system:admin` is not an exempted account. Also we would not like to change the `values.yaml` to include test users that k3d brings in.

```release-note
N/A
```

```docs
N/A
```


Ronkahn21 (dismissed):
LGTM

operator/e2e/setup/k8s_clusters.go
```diff
@@ -1105,7 +1107,70 @@ func buildLDFlagsForE2E() string {
...
+	// Create Operator RestConfig by making a copy of the admin RestConfig and overriding the bearer token.
+	operatorRestConfig = rest.CopyConfig(adminRestConfig)
+	operatorRestConfig.BearerToken = tokenReq.Status.Token
+	// Clear cert data since we're using token auth
+	operatorRestConfig.CertData = nil
+	operatorRestConfig.KeyData = nil
+	operatorRestConfig.CertFile = ""
+	operatorRestConfig.KeyFile = ""
+
+	logger.Debug("✅ Created REST config using operator service account token (valid for 1 hour)")
```
gflarity, line 1176: Above this says: 
```go
expirationSeconds := int64(10800) // 3 hours - more than sufficient for e2e tests
```
Rather than hardcode it in the string just use Debugf and so if the value changes it's fine.

operator/e2e/setup/k8s_clusters.go
```diff
@@ -256,21 +257,21 @@ func ensureRegistryDoesNotExist(ctx context.Context, registryPort string, logger
...
+// CreateK3DClusterWithComponents creates a complete k3d cluster with Grove, Kai Scheduler components.
+func CreateK3DClusterWithComponents(ctx context.Context, cfg ClusterConfig, skaffoldYAMLPath string, logger *utils.Logger) (*rest.Config, func(), error) {
+	restConfig, cleanup, err := CreateK3DCluster(ctx, cfg, logger)
 	if err != nil {
 		return nil, cleanup, err
 	}
 
-	// Create clientset for node monitoring
-	clientset, err := kubernetes.NewForConfig(restConfig)
+	// Create clientSet for node monitoring
```
gflarity: This came up during the intial review this code, for wahtever reason if you look at the k8s code of NewForConfig it's called a Clientset... there's no camelcase.  Not a huge deal but we should match their casing.

operator/e2e/tests/setup.go
```diff
@@ -146,39 +147,49 @@ func waitForPodCountAndPhases(tc TestContext, expectedTotal, expectedRunning, ex
...
 
 	// Prepare cluster with required worker nodes
-	if err := sharedCluster.PrepareForTest(ctx, requiredWorkerNodes); err != nil {
+	if err = sharedCluster.PrepareForTest(ctx, requiredWorkerNodes); err != nil {
 		t.Fatalf("Failed to prepare shared cluster: %v", err)
 	}
 
 	// Get clients from shared cluster
-	clientset, restConfig, dynamicClient := sharedCluster.GetClients()
+	adminClientSet, adminRESTConfig, adminDynamicClient = sharedCluster.GetClients()
```
gflarity: I know it's odd, but they call them a Clientset /shrug.

operator/e2e/tests/setup.go
```diff
@@ -146,39 +147,49 @@ func waitForPodCountAndPhases(tc TestContext, expectedTotal, expectedRunning, ex
...
+			Namespace:             "default",
 		}
 
 		// Collect diagnostics BEFORE cleaning up if test failed
 		if t.Failed() {
 			CollectAllDiagnostics(diagnosticsTc)
 		}
 
-		if err := sharedCluster.CleanupWorkloads(ctx); err != nil {
+		if err = sharedCluster.CleanupWorkloads(ctx); err != nil {
```
gflarity, line 192: No need to change these all now,   but  `err = ` vs `err :=` is a style thing.  Changing this everywhere added a lot of extra diff lines to check for style preference IMO. If we really want to make this our style going forward we should just decide to do so and do them all everywhere in one big separate PR.  Out of habit, I always do `if err :=`, but otherwise I don't have a strong preference.

operator/internal/webhook/admission/pcs/authorization/handler.go
```diff
@@ -112,13 +116,32 @@ func (h *Handler) handleCreateOrUpdate(req admission.Request, resourceObjectKey
 		return admission.Allowed(fmt.Sprintf("admission allowed, creation/updation of resource: %v is initiated by exempt serviceaccount: %v", resourceObjectKey, req.UserInfo.Username))
 	}
 
+	h.printDiff(req, logger)
+	logger.Info("admission denied for create/update operation", "objectKey", resourceObjectKey)
 	return admission.Denied(fmt.Sprintf("admission denied, creation/updation of resource: %v is not allowed", resourceObjectKey))
 }
 
+func (h *Handler) printDiff(req admission.Request, logger logr.Logger) {
```
gflarity: Nit, this probably better called logDiff.

gflarity (commented):
Thanks for the PR!

A few small things in my comments.  I'd like to see the Clientset and hardcoded `(valid for 1 hour)` fixed if you don't disagree. The `printDiff` is just a nit.  I won't request changes so this doesn't get blocked on me. Ping me for a fast approval after, I'll also chat with @sanjaychatterjee and ask him to approve it I'm not immediately available.

operator/e2e/setup/k8s_clusters.go
unmarshall: Ok, lets change it back, unfortunately the type is `kubernetes.Clientset` - k8s codebase is full of contradictions and does not follow golang best practices.

operator/e2e/setup/k8s_clusters.go
unmarshall: Are you talking about `logger.Debugf` or something else? Also currently there are a lot of hard coded values in E2E tests overall. Do you have a proper configuration for e2e where this can be put?

operator/internal/webhook/admission/pcs/authorization/handler.go
unmarshall: This was only added for debugging, forgot to remove it.

operator/e2e/tests/setup.go
unmarshall: Repeated short assignments within a function scope is not really idiomatic go. So it is not merely style/bias. Short assign `:=` is convenience short hand for a combined declaration + initialization. Since `err` has already been defined within the functional scope, therefore by convention it should not be re-declared but just reassigned another value.

operator/e2e/tests/setup.go
unmarshall: will rename it back to `clientset`

operator/e2e/setup/k8s_clusters.go
unmarshall: For now i have created a constant `defaultOperatorSATokenExpirationSeconds` and put it together with other constants `defaultPollTimeout` and `defaultPollInterval`

operator/e2e/tests/setup.go
gflarity: Yes, it absolutely is. The variables are scoped to the control statement, in this case an if block otherwise the compiler would complain.  This pattern is used quite frequently within the go standard library itself: 


https://github.com/golang/go/blob/11722941452d04aa0364a5c6b60acffaa2776b1c/src/encoding/xml/marshal.go#L82
https://github.com/golang/go/blob/11722941452d04aa0364a5c6b60acffaa2776b1c/src/net/smtp/smtp.go#L246
etc... 

Here's an official go blog post on error handling, see the finally code at the end:

https://go.dev/blog/error-handling-and-go

unmarshall: >https://go.dev/blog/error-handling-and-go

This example that is given is fine because `if` block creates a scope. In the function all `err` variables are defined well within the `if` scope. There is not a single `err` variable defined outside that scope and in this case you will have to use short assign as there is no other way (well technically you could still define a var at the top but this pattern is fine). However if you have already defined/initialized an `err` variable in a function (outside the scope of an `if` block) and in the rest of the function you are continuing to use the same variable then there is no need to use short assign.

All your examples point to the same pattern.

In the changed code the following was introduced:
```go
operatorDynamicClient, err = setup.GetOperatorDynamicClient(ctx, adminRESTConfig, logger)
```
This is not included in the `if` block and therefore now you break that pattern and you no longer have all `err` defined within an `if` scope. Therefore it is now wise to not use `short assign` and instead just define `var err error` once in the function scope.

renormalize:
Close temporarily until we have clarity + bandwidth on what to do with this.

cc @danbar2

PR #371, opened 2026-01-27
Split e2e cluster creation

danbar2:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/operator/hack/e2e-cluster/create-e2e-cluster.py
+++ b/operator/hack/e2e-cluster/create-e2e-cluster.py
@@ -505,7 +505,7 @@ def main(
 
     config = ClusterConfig()
     script_dir = Path(__file__).resolve().parent
-    operator_dir = script_dir.parent
+    operator_dir = script_dir.parent.parent  # Go up from hack/e2e-cluster/ to operator/
 
     # With is_flag=True, Typer passes boolean flags correctly
     # No need for to_bool conversion, but keeping it for safety with environment variables


operator/hack/create-e2e-cluster.sh
```diff
@@ -0,0 +1,471 @@
...
+
+# Build configuration
+SKAFFOLD_PROFILE="${E2E_SKAFFOLD_PROFILE:-topology-test}"
+
+# Retry configuration
+MAX_RETRIES="${E2E_MAX_RETRIES:-3}"
+CLUSTER_TIMEOUT="120s"
+
+# Topology configuration (from topology.go)
+NODES_PER_ZONE=28
```
shayasoolin: Are those used by the TAS test code? I don't see the change there.

operator/hack/create-e2e-cluster.sh
```diff
@@ -0,0 +1,471 @@
+#!/bin/bash
```
shayasoolin: I think this better be written as a Python script, will be much more readable and extensible. We can discuss it separately in order not to delay this PR.

.github/workflows/build-check-test.yaml
```diff
@@ -138,7 +138,22 @@ jobs:
...
+        id: create-cluster
+        run: |
+          # Use the centralized k3d cluster creation script
+          # The script handles: cluster creation, Kai Scheduler, Grove, topology labels
+          ./operator/hack/create-e2e-cluster.sh
+
       - name: Run e2e tests - ${{ matrix.test_name }}
+        env:
+          # E2E_REGISTRY_PORT is the local registry port for test images (k3d default)
+          E2E_REGISTRY_PORT: "5001"
```
shayasoolin: It's 5001 here and in some other scripts, but then all the expected yaml files are changed from 5001 to 5000. Why is that?

shayasoolin (commented):
Looks great

operator/hack/create-e2e-cluster.sh
danbar2: it was set by the TAS tests, to label the nodes - you can see the code I removed was handling that on the cluster setup

shayasoolin: yes I saw that, I'm referring to hard-coded consts inside the TAS tests. for example:
```	logger.Info("1. Initialize a 28-node Grove cluster for topology testing")
	clientset, restConfig, dynamicClient, cleanup := prepareTestCluster(ctx, t, 28)
	defer cleanup()

	expectedPods := 7 // worker-rack: 3 pods, worker-block: 4 pods
```
do we want to make it more dynamic by using the value from the env vars?

danbar2: I think it is the next step, we need to make the whole connection between the cluster and test more configurable. See item #2 https://docs.google.com/document/d/1-rVHXspsvndXczcgRFPe1XRthNH6-HzN0f8-wvmix9g/edit?tab=t.0

.github/workflows/build-check-test.yaml
danbar2: I had an error there, fixed

operator/hack/create-e2e-cluster.sh
gflarity: One thing that was lost in the translation was the "image pre-pulling" functionality.   Some of the deleted Go code copied docker images that already exist locally (already docker pulled) into the k3d  kubernetes registry and sped the cluster setup quite a bit. IIRC cluster startup time improved a lot when we got rid of the Nvidia GPU Operator, but there's still value in optimizing this (poor internet connections, mobile tethered development etc) so that startup time is fast and uses as little data as possible. 

Also, is there a specific reason bash is preferred given all the tooling (helm, skaffold, k3d etc)  is written in go anyways? I suppose it was easy enough to port it over using modern tooling, but is there any actual advantage?  It's completely personal preference thing, but having written a lot more bash than I would have liked in my career, I now view it being less and less necessary to get things done. As @shayasoolin mentioned, python would be better. But even then, why not Go, it's great :)

gflarity: Also, looks like we orphaned some stuff and it should be cleaned now if we're moving to bash/python: 

  Orphaned in `dependencies.go` / `dependencies.yaml`

  | Item | Status |
  |------|--------|
  | GetImagesToPrePull() | Orphaned - only called from unit tests, no production use |
  | FullImageName() | Only used by GetImagesToPrePull() |
  | Images list (lines 10-43 in yaml) | Orphaned - never pre-pulled anymore |
  | HelmCharts.KaiScheduler | Orphaned - only validated in unit tests, bash script installs Kai directly |
  | HelmCharts.GPUOperator | Orphaned - defined in struct but not in YAML, never used |

danbar2: I consulted with Shay offline and tried to refactor to python, it didnt improve the overall script by much (line reduced to ±400) and added more dependencies. so we agreed to keep it with bash, Im a bit concerned about doing it with go - I want a clear separation from the code base, E2E and cluster creation.

I'll take care of the rest of the items you raised

danbar2: actually, digging into it, for image prepulling python has an advantage of existing docker SDK and parralism. so we got 2 birds with one stone  :)

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+def require_command(cmd: str):
+    """Check if a command exists."""
+    try:
+        sh.which(cmd)
+    except sh.ErrorReturnCode:
+        console.print(f"[red]❌ Required command '{cmd}' not found. Please install it first.[/red]")
+        raise typer.Exit(1)
+
+
+def run_cmd(cmd, *args, **kwargs) -> Tuple[int, any]:
```
gflarity: This needs to be updated to Any (capitalized) from the typing module.

In Python, lowercase any is a built-in function, whereas Any is the correct type hint. Using the lowercase version here will cause static analysis tools to fail and probably is highlighted in IDEs.

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+
+def deploy_grove_operator(config: ClusterConfig, operator_dir: Path):
+    """Deploy Grove operator using Skaffold."""
+    console.print(Panel.fit("Deploying Grove operator", style="bold blue"))
+
+    # Delete existing installation (ignore errors)
+    run_cmd(sh.helm, "uninstall", "grove-operator", "-n", "grove-system", _ok_code=[0, 1])
+
+    # Set environment for skaffold build
+    build_date = datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%SZ")
```
gflarity: `datetime.utcnow()` is deprecated in Python 3.12+ and 3.14.x is currently the most recent release. Let's use `datetime.now(timezone.utc)` instead (requires importing `timezone` from `datetime`).

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+
+    console.print("[green]✅ Grove operator deployed[/green]")
+
+    # Wait for Grove pods
+    console.print("[yellow]ℹ️  Waiting for Grove pods to be ready...[/yellow]")
+    sh.kubectl("wait", "--for=condition=Ready", "pods", "--all", "-n", "grove-system", "--timeout=5m")
+
+    # Wait for webhook
+    console.print("[yellow]ℹ️  Waiting for Grove webhook to be ready...[/yellow]")
+    for i in range(1, 61):
```
gflarity: Consider extracting these magic numbers into named constants for clarity:
```python
WEBHOOK_READY_MAX_RETRIES = 60
WEBHOOK_READY_POLL_INTERVAL = 5  # seconds
```
Same applies to line 487 with the Kai queue retries.

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+            _ok_code=[0, 1]
+        )
+
+        # Get output text - handle both string and ErrorReturnCode
+        if isinstance(result, str):
+            output = result.lower()
+        else:
+            output = (str(result.stdout) + str(result.stderr)).lower()
+
+        if any(kw in output for kw in ["validated", "denied", "error", "invalid", "created", "podcliqueset"]):
```
gflarity: This condition is a bit confusing - checking for "error" as a positive signal for webhook readiness seems counterintuitive. Maybe rename or restructure to make it clearer that we're checking "did the webhook respond at all" vs "did it succeed"?

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+class ClusterConfig(BaseModel):
+    """Configuration loaded from environment variables with defaults."""
+
+    cluster_name: str = Field(default_factory=lambda: os.getenv("E2E_CLUSTER_NAME", "shared-e2e-test-cluster"))
+    registry_port: str = Field(default_factory=lambda: os.getenv("E2E_REGISTRY_PORT", "5001"))
+    api_port: str = Field(default_factory=lambda: os.getenv("E2E_API_PORT", "6560"))
+    lb_port: str = Field(default_factory=lambda: os.getenv("E2E_LB_PORT", "8090:80"))
+    worker_nodes: int = Field(default_factory=lambda: int(os.getenv("E2E_WORKER_NODES", "30")))
```
gflarity: Let's just use pydantic to handle validation and defaulting of these environment variables. Its pretty slick: 

```

from pydantic import Field
from pydantic_settings import BaseSettings, SettingsConfigDict


class ClusterConfig(BaseSettings):
    """Config auto-loaded from E2E_* environment variables."""

    model_config = SettingsConfigDict(env_prefix="E2E_", extra="ignore")

    cluster_name: str = "shared-e2e-test-cluster"
    registry_port: int = Field(default=5001, ge=1, le=65535)
    api_port: int = Field(default=6560, ge=1, le=65535)
    lb_port: str = "8090:80"
    worker_nodes: int = Field(default=30, ge=1, le=100)
    worker_memory: str = Field(default="150m", pattern=r"^\d+[mMgG]?$")
    k3s_image: str = "rancher/k3s:v1.33.5-k3s1"
    kai_version: str = Field(default="v0.13.0-rc1", pattern=r"^v[\d.]+(-[\w.]+)?$")
    skaffold_profile: str = "topology-test"
    max_retries: int = Field(default=3, ge=1, le=10)

    # Constants (not from env), makes it easy to keep this all in one spot
    cluster_timeout: str = "120s"
    nodes_per_zone: int = 28
    nodes_per_block: int = 14
    nodes_per_rack: int = 7
  
 
```

Then you can just go `config = ClusterConfig()` and it's all done for you. Ie `config.registry_port` etc.

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+    ci_kubeconfig_path = operator_dir / "hack" / "kubeconfig"
+    default_kubeconfig_dir = Path.home() / ".kube"
+    default_kubeconfig_dir.mkdir(parents=True, exist_ok=True)
+    default_kubeconfig_path = default_kubeconfig_dir / "config"
+
+    # Write kubeconfig to both locations for maximum compatibility
+    console.print(f"[yellow]Exporting kubeconfig...[/yellow]")
+    # Use k3d kubeconfig merge to write directly to the file
+    # The --overwrite flag ensures we write a clean kubeconfig (important for CI)
+    sh.k3d("kubeconfig", "merge", config.cluster_name, "-o", str(default_kubeconfig_path), "--overwrite")
```
gflarity: I pretty sure that using --overwrite here means this isn't a merge at all but will overwrite the current kubeconfig. This might surprise some people.

I believe if you drop the --overwrite flag, k3d performs a 'smart merge' instead—it will update the specific cluster context if it already exists or append it if it's new, without touching any of the user's other existing configurations."

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+
+    Environment variables can override defaults (see E2E_* variables).
+    """
+
+    config = ClusterConfig()
+    script_dir = Path(__file__).resolve().parent
+    operator_dir = script_dir.parent
+
+    # Debug: Print all parameters
+    console.print(f"[yellow]DEBUG: Parameters received:[/yellow]")
```
gflarity: Looks like some debugging left in by mistake? Alternatively we can some sort of verbosity mechanism.

operator/e2e/tests/cert_management_test.go
```diff
@@ -102,6 +101,13 @@ func updateGroveToAutoProvision(t *testing.T, ctx context.Context, restConfig *r
 }
 
 const (
+	// Cert-manager Helm chart configuration
```
gflarity, line 104: The images that get prepulled should be synced with this... If we keep the dependencies.yaml we can just colocate this all in one spot centrally.

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,541 @@
...
+app = typer.Typer(help="k3d cluster setup for local E2E testing")
+console = Console(stderr=True)
+
+
+# ============================================================================
+# Configuration
+# ============================================================================
+
+# Kai Scheduler images that need to be pre-pulled for faster cluster creation
+KAI_SCHEDULER_IMAGES = [
```
gflarity: Also need the versions and this needs to be sync with the helm chart and centralized. 

Also we should pre pull cert manager images too. Also 

I'd suggest we just keep depdencies.yaml in some form and just load them up from that. The cert manager tests need the helm chart version and it needs to match the images, so keeping it one place is desirable.

gflarity (commented):
Thanks for switching to python.  I found a couple issues but otherwise it's looking good.

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,619 @@
...
+
+def run_cmd(cmd, *args, **kwargs) -> Tuple[int, Any]:
+    """Run a command, handling errors gracefully. Returns (exit_code, output)."""
+    try:
+        output = cmd(*args, **kwargs)
+        return 0, output
+    except sh.ErrorReturnCode as e:
+        if not kwargs.get("_ok_code"):
+            raise
+        return e.exit_code, e
```
gflarity: Nit: The try/except here is effectively dead code. The `_ok_code` kwarg is passed through to `sh`, which means `sh` itself won't raise `ErrorReturnCode` for those exit codes. The except block only fires for codes *not* in `_ok_code`, where it silently swallows the error instead of propagating -- which is likely unintended. Since `sh` already handles the suppression, this can be simplified to:

```suggestion
def run_cmd(cmd, *args, **kwargs) -> Tuple[int, Any]:
    """Run a command, handling errors gracefully. Returns (exit_code, output).

    Pass _ok_code=[0, 1, ...] to suppress exceptions for expected exit codes.
    The sh library handles this natively - it won't raise for codes in _ok_code.
    """
    output = cmd(*args, **kwargs)
    return output.exit_code, output
```

gflarity (approved):
Left one nit. Looks like there's a conflicted rebase required, so I'd suggest fixing the nit as I believe a re-approval will be required anyways. But if you can merge it go for it, it's not a big deal.

gflarity (dismissed):

shayasoolin (approved):

operator/hack/create-e2e-cluster.py
```diff
@@ -0,0 +1,645 @@
+#!/usr/bin/env python3
+# /*
+# Copyright 2025 The Grove Authors.
```
shmuel-runai: 2026 :)

operator/hack/e2e-cluster/create-e2e-cluster.py
```diff
@@ -0,0 +1,645 @@
+#!/usr/bin/env python3
```
shmuel-runai, line 1: Maybe it is better to place this script in its own folder (with the requirements.txt), other py scripts will have different requirements

operator/e2e/setup/k8s_clusters.go
```diff
@@ -14,703 +14,73 @@
...
-		} else {
-			logger.Debug("✅ Grove installation completed successfully")
+// GetRestConfig returns a REST config for connecting to a Kubernetes cluster.
+// It tries the following methods in order:
+//  1. KUBECONFIG environment variable (if set)
+//  2. Default kubeconfig at ~/.kube/config
+//
+// For local development with k3d, run './operator/hack/create-e2e-cluster.py' first
+// to create a cluster and configure kubectl.
+func GetRestConfig() (*rest.Config, error) {
```
shmuel-runai: nit: getRestConfig

operator/e2e/setup/k8s_clusters.go
```diff
@@ -14,703 +14,73 @@
...
-			logger.Debugf("🔄 Retrying %s installation (attempt %d/%d)...", componentName, attempt, maxRetries)
-			time.Sleep(retryDelay)
-		}
-
-		err := installFunc()
-		if err == nil {
-			if attempt > 1 {
-				logger.Debugf("✅ %s installation succeeded on attempt %d/%d", componentName, attempt, maxRetries)
+	// Try to load from kubeconfig file
+	if kubeconfigPath != "" {
```
shmuel-runai: nit: rev the if statement 
if kubeConfigPath == "" {
  return error failed to ...
}
load from configPath

operator/e2e/setup/shared_cluster.go
```diff
@@ -75,7 +83,9 @@ var groveManagedResourceTypes = []resourceType{
 	{"autoscaling", "v2", "horizontalpodautoscalers", "HorizontalPodAutoscalers"},
 }
 
-// SharedClusterManager manages a shared (singleton) k3d cluster for E2E tests
+// SharedClusterManager manages a shared (singleton) Kubernetes cluster for E2E tests.
```
shmuel-runai: nit: emit (singleton) from the comment

operator/Makefile
```diff
@@ -86,23 +86,20 @@ cover-html: test-cover
...
-# Variables:
-#   TEST_PATTERN - Test name pattern to match (optional)
-#   DIAG_DIR     - Directory for diagnostic files (default: operator/)
-#   DIAG_MODE    - Diagnostic output mode: stdout, file, both (default: file)
-DIAG_DIR ?= $(MODULE_ROOT)
-DIAG_MODE ?= file
+# Environment variables:
+#   GROVE_E2E_DIAG_MODE - Diagnostic output mode: stdout, file, both (default: file)
+#   GROVE_E2E_DIAG_DIR  - Directory for diagnostic files (default: operator/)
 .PHONY: test-e2e
```
shmuel-runai: nit: rename this to run-e2e

shmuel-runai (approved):
Great update, very needed
LGTM

PR #380, opened 2026-01-29
Fix tas label

### LINKED ISSUE #379: TAS e2e test fai randomly
### What happened?

When do e2e test, sometimes node can be unready, then we will restart a new node. From log, I can see that it won't apply labels to the new node.

```2026-01-29T04:54:43.059Z	INFO	setup/k8s_clusters.go:776	WARN: 🚨 Found not ready node: k3d-shared-e2e-test-cluster-agent-1
2026-01-29T04:54:44.023Z	INFO	setup/k8s_clusters.go:783	WARN: ✅ Successfully replaced not ready node: k3d-shared-e2e-test-cluster-agent-1
2026-01-29T04:54:55.353Z	INFO	tests/topology_test.go:47	Get all pods once for verification
2026-01-29T04:54:55.359Z	INFO	tests/topology_test.go:390	3. Verify all 4 pods in same zone (PCS zone constraint)
topology_test.go:392: Failed to verify pods in same zone: node k3d-shared-e2e-test-cluster-agent-1 does not have topology label kubernetes.io/zone
```

Replaced node should also have the label with the previous one

### Environment

- Kubernetes version
- Grove version
- Scheduler details

kangclzjc:
#### What type of PR is this?
/kind bug

Fix TAS randomly failure

Fixes #379 

```release-note

```

```docs

```

### DIFF
--- a/operator/e2e/setup/k8s_clusters.go
+++ b/operator/e2e/setup/k8s_clusters.go
@@ -41,6 +41,7 @@ import (
 	k3dlogger "github.com/k3d-io/k3d/v5/pkg/logger"
 	"github.com/k3d-io/k3d/v5/pkg/runtimes"
 	k3d "github.com/k3d-io/k3d/v5/pkg/types"
+	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -1182,12 +1183,7 @@ func waitForWebhookReady(ctx context.Context, restConfig *rest.Config, logger *u
 // reapplyNodeLabels reapplies the original labels to a replaced node
 func reapplyNodeLabels(ctx context.Context, clientset *kubernetes.Clientset, node *v1.Node, labels map[string]string, logger *utils.Logger) error {
 	// Merge original labels with current labels (preserving any new system labels)
-	if node.Labels == nil {
-		node.Labels = make(map[string]string)
-	}
-	for k, v := range labels {
-		node.Labels[k] = v
-	}
+	node.Labels = lo.Assign(node.Labels, labels)
 
 	// Update the node with the merged labels
 	_, err := clientset.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
--- a/operator/e2e/utils/node.go
+++ b/operator/e2e/utils/node.go
@@ -41,10 +41,9 @@ func IsNodeReady(node *v1.Node) bool {
 }
 
 // WaitAndGetReadyNode waits for a specific node to become ready after container restart and returns the ready node
-func WaitAndGetReadyNode(ctx context.Context, clientset *kubernetes.Clientset, nodeName string, timeout time.Duration, logger *Logger) (*v1.Node, error) {
-	var readyNode *v1.Node
-	err := PollForCondition(ctx, timeout, defaultPollInterval, func() (bool, error) {
-		node, err := clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+func WaitAndGetReadyNode(ctx context.Context, clientset *kubernetes.Clientset, nodeName string, timeout time.Duration, logger *Logger) (node *v1.Node, err error) {
+	err = PollForCondition(ctx, timeout, defaultPollInterval, func() (bool, error) {
+		node, err = clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
 		if err != nil {
 			// If node is not found, it's expected (k3d agent still starting up), continue waiting
 			if errors.IsNotFound(err) {
@@ -53,19 +52,12 @@ func WaitAndGetReadyNode(ctx context.Context, clientset *kubernetes.Clientset, n
 			}
 			return false, err
 		}
-
 		if IsNodeReady(node) {
 			logger.Debugf("✅ Node %s has rejoined and is ready", nodeName)
-			readyNode = node
 			return true, nil
 		}
-
-		logger.Debugf("  Node %s found but not ready yet, waiting...", nodeName)
+		logger.Debugf("⏳ Node %s found but not ready yet, waiting...", nodeName)
 		return false, nil
 	})
-
-	if err != nil {
-		return nil, err
-	}
-	return readyNode, nil
+	return
 }


operator/e2e/setup/k8s_clusters.go
```diff
@@ -813,6 +818,30 @@ func replaceNotReadyNode(ctx context.Context, node *v1.Node, clientset *kubernet
...
+	}
+
+	if hasTopologyLabels {
+		logger.Debugf("🏷️  Reapplying topology labels to replaced node: %s", nodeName)
+		if err := reapplyTopologyLabelsToNode(ctx, clientset, nodeName, logger); err != nil {
+			// Log warning but don't fail - node is functional, just missing topology labels
+			logger.Warnf("⚠️  Failed to reapply topology labels to node %s: %v", nodeName, err)
+		}
+	}
+
```
Ronkahn21: Can you export this logic to sperate function

operator/e2e/setup/k8s_clusters.go
```diff
@@ -813,6 +818,30 @@ func replaceNotReadyNode(ctx context.Context, node *v1.Node, clientset *kubernet
...
+		return fmt.Errorf("node %s did not rejoin cluster: %w", nodeName, err)
+	}
+
+	// Step 4: Reapply topology labels if they existed on the original node
+	// Check if this node had topology labels before replacement
+	hasTopologyLabels := false
+	for key := range originalLabels {
+		if key == TopologyLabelZone || key == TopologyLabelBlock || key == TopologyLabelRack {
+			hasTopologyLabels = true
+			break
```
Ronkahn21: We are adding topology labels to all of the nodes by default.As this does not effect to other test,.
so I think it safe to assume the topology labels are applied to any worker  (no need to check)

operator/e2e/setup/k8s_clusters.go
```diff
@@ -1174,3 +1203,88 @@ func waitForWebhookReady(ctx context.Context, restConfig *rest.Config, logger *u
 		return true, nil
 	})
 }
+
+// waitForSingleNodeReady waits for a specific node to become ready after container restart
+func waitForSingleNodeReady(ctx context.Context, clientset *kubernetes.Clientset, nodeName string, timeout time.Duration, logger *utils.Logger) error {
```
Ronkahn21: Can you move this function under utils (e2e)

operator/e2e/setup/k8s_clusters.go
```diff
@@ -1174,3 +1203,88 @@ func waitForWebhookReady(ctx context.Context, restConfig *rest.Config, logger *u
...
+				return nil
+			}
+
+			logger.Debugf("  Node %s found but not ready yet, waiting...", nodeName)
+		}
+	}
+}
+
+// reapplyTopologyLabelsToNode reapplies topology labels to a single node after replacement
+func reapplyTopologyLabelsToNode(ctx context.Context, clientset *kubernetes.Clientset, nodeName string, logger *utils.Logger) error {
```
Ronkahn21: As you save the original labels of the given node, can you simply reapply the topology once back to the new  Node

Ronkahn21 (changes requested):
Good catch finding the edge case bug, I added a few comments

operator/e2e/setup/k8s_clusters.go
kangclzjc: After remove hasTopolopy, this shouldn't be a problem

operator/e2e/setup/k8s_clusters.go
kangclzjc: Yes, we could do that, let me do it.

operator/e2e/utils/k8s_client.go
```diff
@@ -531,3 +531,45 @@ func scaleCRD(ctx context.Context, dynamicClient dynamic.Interface, gvr schema.G
 
 	return nil
 }
+
+// IsNodeReady checks if a node is in Ready state
+func IsNodeReady(node *v1.Node) bool {
```
unmarshall: It ideally does not belong here. The file is named `k8s_client.go` but in this method you are not really making any calls to the KAPI via the client. I would suggest that you create a new file under `utils` and call it `node.go` and move this there.

operator/e2e/utils/k8s_client.go
```diff
@@ -531,3 +531,45 @@ func scaleCRD(ctx context.Context, dynamicClient dynamic.Interface, gvr schema.G
...
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == v1.NodeReady {
+			return condition.Status == v1.ConditionTrue
+		}
+	}
+	return false // If no Ready condition is found, consider the node not ready
+}
+
+// WaitForSingleNodeReady waits for a specific node to become ready after container restart
+func WaitForSingleNodeReady(ctx context.Context, clientset *kubernetes.Clientset, nodeName string, timeout time.Duration, logger *Logger) error {
```
unmarshall: Also move this to `node.go`. Having one file `k8s_client.go` is a dumping place for all things to do with any k8s resource. Lets create more specific files. This has also been done already in `internal/utils/kubernetes`

operator/e2e/utils/k8s_client.go
```diff
@@ -531,3 +531,45 @@ func scaleCRD(ctx context.Context, dynamicClient dynamic.Interface, gvr schema.G
...
+	const defaultPollInterval = 2 * time.Second
```
unmarshall: There is another function `isPodReady` - can you please remove that function and instead use `IsPodReady` as defined in `internal/utils/kubernetes/pod.go` to endure duplicate code? This is not really related to this PR but i am seeing this code for the first time. Did not get a chance to review it before. This is a small change.

operator/e2e/setup/k8s_clusters.go
```diff
@@ -813,6 +804,18 @@ func replaceNotReadyNode(ctx context.Context, node *v1.Node, clientset *kubernet
 		return fmt.Errorf("failed to restart container for node %s: %w", nodeName, err)
 	}
 
+	// Step 3: Wait for the node to rejoin and become ready
+	logger.Debugf("⏳ Waiting for node to rejoin cluster: %s", nodeName)
```
unmarshall: should we wait for the node to be ready in order to re-apply the labels? The original code did not wait for node to be ready.

operator/e2e/utils/k8s_client.go
```diff
@@ -531,3 +531,45 @@ func scaleCRD(ctx context.Context, dynamicClient dynamic.Interface, gvr schema.G
...
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
```
unmarshall: you can re-write this function and use an existing function `PollForCondition` instead.

```go
func WaitForNodeReady(ctx context.Context, logger *Logger, clientset *kubernetes.Clientset, nodeName string, timeout time.Duration) {
  return PollForCondition(ctx, timeout, defaultPollInterval, func() (bool, error) {
    node, err := clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
    if err != nil {
			// Node might not have rejoined yet (k3d agent still starting up)
			logger.Debugf("  Node %s not found yet (k3d agent still connecting), waiting...", nodeName)
			return false, nil
	}
    if IsNodeReady(node) {
			logger.Debugf("✅ Node %s has rejoined and is ready", nodeName)
			return true, nil
	}
    logger.Debugf("  Node %s found but not ready yet, waiting...", nodeName)
	return false, nil
  }
}
```

Alternatively you can rename this function to `WaitAndGetReadyNode` - since you are already making GET calls here there is no reason to make another GET call in `reapplyNodeLabels` which can simply be passed a ready `Node` object.

operator/e2e/setup/k8s_clusters.go
kangclzjc: That's correct, actually we don't need wait for node to be ready and then apply the label. But since we replace a node, we should guarantee that new node should be Ready, otherwise, in next round check, this node may still be replaced ?

operator/e2e/utils/k8s_client.go
kangclzjc: Yes, it will save one GET.

operator/e2e/utils/k8s_client.go
kangclzjc: Done

operator/e2e/utils/k8s_client.go
kangclzjc: That's a great idea. Learned a lot

unmarshall (approved):

Issue #377, opened 2026-01-28
Prevent triggering of e2e tests unless "Build, test, check" jobs succeed


unmarshall:
/area dev-productivity
/kind enhancement

### What you would like to be added?

Prevent E2E tests to be triggered unless and until `build`, `test` and `check` jobs have completed successfully.

### Why is this needed?

E2E tests require additional resources (bigger runners) and also take a long time to finish. When commits are made to a PR, then there are often cases where `make check` or `make build` or `make test` fails but it still triggers e2e tests which is completely unnecessary.

PR #381, opened 2026-01-29
run e2e only if code changed and build pass

danbar2:
#### What type of PR is this?

/kind cleanup

This PR optimizes E2E test execution by:
1. Consolidating the E2E workflow into the "Build, Check and Test" workflow
2. E2E tests now only run when there are changes to `operator/**` or `.github/**` paths
3. E2E tests are gated behind successful completion of `test`, `build`, and `check` jobs

This reduces unnecessary CI resource consumption by skipping expensive E2E tests when:
- Changes don't affect operator or workflow code
- Build/test/check jobs have already failed

Fixes https://github.com/ai-dynamo/grove/issues/377

- Deleted `.github/workflows/e2e-test.yaml` and merged its contents into `build-check-test.yaml`
- Uses `dorny/paths-filter@v3` action to detect path changes
- E2E jobs have `needs: [test, build, check, changes]` to enforce ordering

### DIFF
--- a/.github/workflows/build-check-test.yaml
+++ b/.github/workflows/build-check-test.yaml
@@ -139,12 +139,8 @@ jobs:
           skaffold version
 
       - name: Run e2e tests - ${{ matrix.test_name }}
-        run: |
-          cd operator
-          echo "> Preparing charts (copying CRDs)..."
-          ./hack/prepare-charts.sh
-          echo "> Running e2e tests for ${{ matrix.test_name }}..."
-          cd e2e && go test -tags=e2e ./tests/... -v -timeout 45m -run '${{ matrix.test_pattern }}'
+        run: make test-e2e TEST_PATTERN='${{ matrix.test_pattern }}'
+        working-directory: operator
 
       # The test code handles cleanup via Teardown(), but this step provides
       # extra safety in case of timeout or panic. Also good practice to ensure


.github/workflows/build-check-test.yaml
```diff
@@ -55,3 +78,87 @@ jobs:
           go-version: "1.24.5"
       - name: check
         run: make check
+
+  # E2E tests - only run after build, check, and test jobs succeed
+  # Only triggered by changes to operator or .github folders
+  e2e:
```
unmarshall, line 84: Does it make sense to allow people to also trigger e2e tests for DRAFT PRs to validate if all tests are running fine before the PR is opened up for review?
If you agree then we can add the following if condition:
```yaml
e2e:
  if: |
    (github.event_name == 'pull_request' && (github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'run-e2e')))
```

danbar2: it does make sense and already implemented, look at line 88

.github/workflows/build-check-test.yaml
```diff
@@ -55,3 +78,87 @@ jobs:
...
+      needs.changes.outputs.e2e-relevant == 'true' &&
+      (github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'run-e2e'))
+    # use NVIDIA self-hosted runner with 8 vCPUs and 30GB of ram (on AWS)
+    # the tests are unstable using the default runner with 4 vCPUs and 16GB of ram
+    runs-on: cpu-amd-m5-2xlarge
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
```
shayasoolin, line 99: Do we have to include specific tests? Won't it be better to run all of them, or just exclude some if needed? I'm thinking about newly added tests (new prefix) that will be skipped to do this policy.

danbar2: this pr is just moving the current E2E step to be in the same workflow as the build and unit tests.
The matrix was created for paralism - I agree we need to define a better solution for that, but it is our of scope for this pr

shayasoolin (approved):

.github/workflows/build-check-test.yaml
```diff
@@ -55,3 +78,87 @@ jobs:
...
+          sudo install skaffold /usr/local/bin/
+          skaffold version
+
+      - name: Run e2e tests - ${{ matrix.test_name }}
+        run: |
+          cd operator
+          echo "> Preparing charts (copying CRDs)..."
+          ./hack/prepare-charts.sh
+          echo "> Running e2e tests for ${{ matrix.test_name }}..."
+          cd e2e && go test -tags=e2e ./tests/... -v -timeout 45m -run '${{ matrix.test_pattern }}'
```
Ronkahn21: can you use the makefile e2e command instead directly running the command

Ronkahn21 (commented):
Just one comment, removing this duplication from the makefile

Ronkahn21 (approved):

PR #385, opened 2026-02-02
Auto-MNNVL: add PodSpec injection for MNNVL resourceClaims

shmuel-runai:
/kind feature

ref: GREP-270

Implement InjectAutoMNNVLIntoPodSpec to inject MNNVL resourceClaims into PodSpecs for GPU workloads. The function:
- Adds a pod-level PodResourceClaim referencing the RCT
- Adds claim references to each container that requests GPU
- Is idempotent (safe to call multiple times)
- Logs injection actions at debug level

Integration with PCS + PCSG controller:
- Inject MNNVL claims when PCS controller creates PodClique resources
- Propagate grove.io/auto-mnnvl annotation from PCS to PCSG
- Inject MNNVL claims into PodCliques created via PCSG path

Refactor test organization:
- Move helper function tests to helpers_test.go

### DIFF
--- a/operator/internal/controller/podcliqueset/components/podcliquescalinggroup/podcliquescalinggroup.go
+++ b/operator/internal/controller/podcliqueset/components/podcliquescalinggroup/podcliquescalinggroup.go
@@ -217,7 +217,9 @@ func (r _resource) buildResource(pcsg *grovecorev1alpha1.PodCliqueScalingGroup, 
 	pcsg.Spec.CliqueNames = pcsgConfig.CliqueNames
 	pcsg.Labels = getLabels(pcs, pcsReplica, client.ObjectKeyFromObject(pcsg))
 
-	// Propagate MNNVL annotation from PCS to PCSG
+	// Propagate MNNVL annotation from PCS to PCSG.
+	// Only propagate when enabled - if not enabled, PCSG controller has nothing to inject,
+	// so the annotation is not needed.
 	if mnnvl.IsAutoMNNVLEnabled(pcs.Annotations) {
 		if pcsg.Annotations == nil {
 			pcsg.Annotations = make(map[string]string)
--- a/operator/internal/mnnvl/injection.go
+++ b/operator/internal/mnnvl/injection.go
@@ -23,11 +23,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// InjectAutoMNNVLIntoPodSpec injects MNNVL resourceClaims into the PodSpec if it has GPU requirements.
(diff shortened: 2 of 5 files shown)


danbar2 (approved):

operator/internal/mnnvl/injection.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// InjectAutoMNNVLIntoPodSpec injects MNNVL resourceClaims into the PodSpec if it has GPU requirements.
+// It adds a PodResourceClaim referencing the generated RCT name, and adds claim references to each
+// container that requests GPU resources. The operation is idempotent.
+// Should only be called when MNNVL is enabled (check IsAutoMNNVLEnabled first).
+func InjectAutoMNNVLIntoPodSpec(logger logr.Logger, podSpec *corev1.PodSpec, pcsNameReplica apicommon.ResourceNameReplica) {
```
shayasoolin: I suggest renaming to InjectMNNVLIntoPodSpec. The 'Auto' part is not required here, it was already decided, and this function just adds it.
Another option is InjectMNNVLResourceClaimIntoPodSpec

operator/internal/controller/podcliqueset/components/podcliquescalinggroup/podcliquescalinggroup.go
```diff
@@ -215,6 +216,15 @@ func (r _resource) buildResource(pcsg *grovecorev1alpha1.PodCliqueScalingGroup,
 	pcsg.Spec.MinAvailable = pcsgConfig.MinAvailable
 	pcsg.Spec.CliqueNames = pcsgConfig.CliqueNames
 	pcsg.Labels = getLabels(pcs, pcsReplica, client.ObjectKeyFromObject(pcsg))
+
+	// Propagate MNNVL annotation from PCS to PCSG
+	if mnnvl.IsAutoMNNVLEnabled(pcs.Annotations) {
```
shayasoolin, line 223: I think that's fine but writing to you just to make sure: if MNNVL is explicitly disabled for the PCS, don't we need to propagate it to the PCSG? (I think we don't, as we control the PCSG, the user can't change it, so we're safe. Right?)

operator/internal/mnnvl/injection.go
shmuel-runai: renamed

operator/internal/controller/podcliqueset/components/podcliquescalinggroup/podcliquescalinggroup.go
shmuel-runai: added a comment

PR #403, opened 2026-02-06
GREP-291: `OnDelete` update strategy for `PodCliqueSet`

### LINKED ISSUE #291: Enhance the rolling update approach for PodClique

Rolling update dynamically update a PodClique's podTemplate `nodeAffinity` rules, without triggering pod deletions/restarts (non-disruptive update). 

### Why is this needed?

Training workloads may encounter machine failures. The mitigation strategy includes:
1. Automated/Manual Action: Adjust Job affinity to prevent scheduling new pods on failed nodes.
2. Pod Recovery: Recreate affected pods with strict `nodeAffinity` rules against faulty nodes.
3. Preservation: Unaffected pods continue running to minimize disruption.

Currently, in step1, Grove will delete all pods during rolling update, including unaffected ones.

renormalize:

/kind feature
/kind api

Introduces GREP-291.

#291 mentions certain realities of performing day 2 operations on a PodCliqueSet. The author mentions changing `nodeAffinity` in the pod template of the PodCliqueSet to ensure that pods are not scheduled onto failed nodes.

However, changing `nodeAffinity` will cause all pods of the PodClique to roll, which is undesirable since there might only be a small subset of pods that were on failed nodes. Therefore, to avoid the rolling of all pods in the PodClique, taking inspiration from Kubernetes StatefulSet, I would like to introduce an `OnDelete` update strategy for PodCliqueSet, where users can then update the `nodeAffinity`, and avoid rolling of all the pods in a PodClique.

Fixes #291 

N/A

```release-note
Introduce the new `spec.updateStrategy` field to specify `OnDelete` update strategy for `PodCliqueSet`
```

```docs
Introduce GREP-291 for the `OnDelete` update strategy for `PodCliqueSet`
```


renormalize:
In a discussion with the maintainers, they had expressed the following opinions:

- Change the scope of the GREP to define `OnDelete` for both PodCliques and ScalingGroups. Implementation can be split into different stages.
- Rename `RollingUpdate` to `RollingRecreate`, and define what `RollingRecreate` is. The update strategy currently implemented in grove is not strictly `RollingUpdate` as referred to by the community, and therefore it would be incorrect to call it that.
- Keep update strategy only at the PodCliqueSet level. In the future, update strategy can be defined at each hierarchy to specify different behavior for different PodCliques/ScalingGroups. This would align with the convention used for topology configuration in PCS.
- ScalingGroups and (standalone) PodCliques both will follow the same strategy. This can be configured hierarchically at all levels later on if needed, as mentioned above.

@unmarshall @sanjaychatterjee @Ronkahn21 @gflarity had suggested the above.

Please refrain from reviews before I adapt the GREP with the above suggestions, thanks.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,461 @@
...
+- Implementing partition-based rolling updates (as in StatefulSet's `RollingUpdate` with `partition` field).
+- Providing a `maxUnavailable` or `maxSurge` configuration for rolling updates (can be addressed in a future GREP).
+- Automatic detection of failed nodes and selective pod recreation. This should be the responsibility of the scheduler.
+- Defining update strategy at each hierarchy level (PodCliqueSet, PodCliqueScalingGroup, PodClique) individually (following the design used for topology configuration in a PodCliqueSet). This can be addressed in a future GREP to allow different strategies for different components.
+
+## Proposal
+
+Introduce a new `spec.updateStrategy` field in the `PodCliqueSet` specification. This field will accept an `UpdateStrategy` object that defines how replicas should be updated when templates change.
+
+Two update strategy types will be supported:
```
sanjaychatterjee, line 66: This is fine for now since the default `RollingRecreate` strategy is the currently implemented strategy. At some point, we also need to introduce a rolling recreate strategy that rolls entire PCS replicas. How would you name that, `PCSRecreate`? Just trying to ensure we name these strategies correctly. cc @nvrohanv

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,461 @@
...
+    // PodCliqueScalingGroups.
+    // +optional
+    UpdateStrategy *PodCliqueSetUpdateStrategy `json:"updateStrategy,omitempty"`
+    
+    // Template describes the template spec for PodGangs that will be 
+    // created in the PodCliqueSet.
+    Template PodCliqueSetTemplateSpec `json:"template"`
+}
+```
+#### PodCliqueSpec Changes
```
sanjaychatterjee: I am not sure if we want to support different update strategies at different levels. E.g. if you have `RollingRecreate` at higher level and `OnDelete` at lower level, the update will get stuck. On the other hand, `OnDelete` at higher level and `RollingReceate` at lower level does not make sense because the user may delete pods at random and may not be following any order. I think we need to think hard whether we should support simultaneous multiple update strategies at all. cc @unmarshall @nvrohanv

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,461 @@
...
+  	// If MinAvailable is not set, then it will default to the template Replicas.
+  	// +optional
+  	MinAvailable *int32 `json:"minAvailable,omitempty"`
+...
+}
+```
+
+#### PodCliqueScalingGroupSpec Changes
+
+The `PodCliqueScalingGroupSpec` will be extended to include the `updateStrategy` field:
```
sanjaychatterjee: Same comment as above regarding multiple update strategies per PCS.

docs/proposals/291-ondelete-update/README.md
renormalize: Thanks for the review @sanjaychatterjee!

The changes to `PodCliqueSpec` do not suggest support for different update strategies at different levels. In this proposal, my intent is to *pass the update strategy specified at the PodCliqueSet level to all descendant resources*. This field is only used by the PodClique controller to *enact* the behavior that is specified at the  PodCliqueSet level.

If we do intend to support different update strategies at different levels, this is not the field that the user would be configuring. They would be configuring a field in `PodCliqueSetTemplateSpec.PodCliqueTemplateSpec`, which is the template that is used by the PodCliqueSet controller to create PodClique resources.

PodCliques are resources not meant for direct modification by users too, so this is not an API field that is exposed to them, therefore I don't think my proposed changes would imply multiple strategy support.

docs/proposals/291-ondelete-update/README.md
renormalize: Same argument as https://github.com/ai-dynamo/grove/pull/403#discussion_r2791325086 but for PodCliqueScalingGroup.

docs/proposals/291-ondelete-update/README.md
sanjaychatterjee: Sounds good. Thanks for the clarification.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,461 @@
...
+- `Status.UpdatedReplicas`: Number of replicas running with the current template specification.
+- `Status.Replicas`: Total number of replicas.
+- `Status.CurrentGenerationHash`: Hash of the current desired specification.
+- `PodCliqueStatus.CurrentPodTemplateHash`: Hash identifying the template version for each PodClique.
+
+When `UpdatedReplicas < Replicas` with `OnDelete` strategy, it indicates that some replicas are running with an outdated specification.
+
+**Proposed Condition:**
+
+A new condition `UpdatePending` can be added to `PodCliqueSetStatus.Conditions`:
```
sanjaychatterjee: Not sure what benefit this condition would bring. If using the current status, then `Status.UpdatedReplicas` less than `Status.Replicas` would mean the same as this condition being `True`. And like you mentioned above, a comparison of hash would be able to identify the replicas that are not yet updated.

sanjaychatterjee (commented):
Mostly looks good to me. Not convinced about the need for the extra condition.

docs/proposals/291-ondelete-update/README.md
renormalize: I do agree that the condition does not add too much value to consumers the way the `MinAvailableBreached` condition does, for example.

I don't have a strong opinion on introducing this condition; I wanted to get some feedback on whether this condition will be a better user experience for consumers. I will go ahead and remove this condition from the proposal.

renormalize: Removed in [0b06d30](https://github.com/ai-dynamo/grove/pull/403/commits/0b06d3028b3d2003df45da8e6c6e41f00d5571cf).

sanjaychatterjee (dismissed):
Looks good to me. Thanks!

renormalize:
After a proof of concept implementation, I had second thoughts about certain fields. I've reworked the GREP based on these opinions.
- I've removed the `.spec` changes to `PodClique` and `PodCliqueScalingGroup`. I feel we should not include the update strategy as a part of the spec for these resources now, and only include them when absolutely necessary. The update strategy can be fetched from the PodCliqueSet resource. I think this aligns with the original comments from the codeowners as well.
- I've renamed the `rollingUpdateProgress` status fields to `updateProgress`, to more generically track information about updates no matter what the update strategy is.

Ronkahn21 (commented):
Looks good overall. Thoughts on the API break—do we introduce it now or wait for v1? @sanjaychatterjee @unmarshall

sanjaychatterjee:
> > Looks good overall. Thoughts on the API break—do we introduce it now or wait for v1? @sanjaychatterjee @unmarshall
> 
> ~@Ronkahn21 thanks for your review! Can you please elaborate on what you mean by the API break in this case? My proposed changes are backwards compatible.~
> 
> We are indeed changing the status API that we expose to consumers by renaming this field.
> 
> @julienmancuso Could you let us know if renaming this field will cause issues for dynamo/operator?

Julien already confirmed on slack that Dynamo is not using the field but still we need to follow deprecation process as part of changelog notifications for two consecutive releases. cc @danbar2

sanjaychatterjee (dismissed):

Ronkahn21 (dismissed):

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+    - [Phase 1: Standalone PodCliques, and PodCliqueScalingGroups implementation](#phase-1-standalone-podcliques-and-podcliquescalinggroups-implementation)
+    - [Phase 2: Comprehensive tests](#phase-2-comprehensive-tests)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This GREP proposes adding an `updateStrategy` field to the `PodCliqueSet` spec, similar to the Kubernetes `StatefulSet` update strategy. The primary goal is to introduce an `OnDelete` update strategy that allows users to update PodClique and PodCliqueScalingGroup specifications (such as `nodeAffinity` rules) without automatically triggering replica deletions/restarts. This enables non-disruptive updates where only manually deleted replicas are recreated with the new specification.
```
unmarshall: The primary goal comes first - which is to introduce `OnDelete` update strategy. In order for allow users to choose this strategy over the existing rolling-recreate strategy it is important that a new `updateStrategy` field in PCS be introduced. So that is secondary.

One possible rephrase of this section:

```
## Summary

This GREP introduces a new `OnDelete` update strategy for `PodCliqueSet` that allows users to update `PodClique` and `PodCliqueScalingGroup` specifications (such as `nodeAffinity` rules) without automatically triggering replica deletions/restarts. With `OnDelete` strategy, only manually deleted replicas are recreated with the new specification, enabling non-disruptive updates.

Currently, Grove implements a `rolling recreate` strategy that automatically deletes and recreates replicas one at a time when templates are updated. To support the new `OnDelete` strategy alongside the existing `rolling recreate` behavior, this GREP introduces a new `updateStrategy` field to the `PodCliqueSet` spec, enabling users to explicitly choose their preferred update strategy.

The update strategy is defined at the `PodCliqueSet` level and applies uniformly to both standalone `PodCliques` and `PodCliqueScalingGroups`. In the future, update strategy could be defined at each hierarchy level to specify different behavior for different `PodCliques/PodCliqueScalingGroups`, following the convention used for [topology configuration](/docs/proposals/244-topology-aware-scheduling/README.md) in `PodCliqueSet`.
```

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+The update strategy is defined at the PodCliqueSet level and applies uniformly to both standalone PodCliques and PodCliqueScalingGroups. In the future, update strategy could be defined at each hierarchy level to specify different behavior for different PodCliques/PodCliqueScalingGroups, following the convention used for [topology configuration](/docs/proposals/244-topology-aware-scheduling/README.md) in PodCliqueSet.
+
+Implementation will be split into phases, with Phase 1 covering standalone PodCliques and Phase 2 extending support to PodCliqueScalingGroups.
```
unmarshall: This does not belong in `Summary` section and should be removed.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Motivation
+
+Training workloads and long-running inference workloads may encounter machine failures during execution. The typical mitigation strategy involves:
```
unmarshall: I am not sure if it is a typical migration strategy. This is a very specific strategy deployed/used by a specific customer.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+1. **Automated/Manual Action**: Adjust workload affinity to prevent scheduling new pods on failed nodes.
+2. **Pod Recovery**: Recreate only the affected pods with strict `nodeAffinity` rules against faulty nodes.
```
unmarshall: I am not sure if `NodeAffinity` should be mentioned here.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+3. **Preservation**: Unaffected pods continue running to minimize disruption.
+
+Currently, when a user updates a standalone PodClique's `podTemplate` (e.g., changing `nodeAffinity` rules), Grove triggers a rolling recreate that deletes and recreates **all** pods one at a time, including unaffected ones. This behavior is disruptive and wasteful for workloads that only need to recreate and reschedule a subset of pods. Similar behavior is displayed in PodCliqueScalingGroups where a rolling recreate deletes and recreates **all** replicas one at a time.
```
unmarshall: This is not entirely true - not `all` pods are affected. If there are PCLQs that do not have any updates, their Pods will not be recreated.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+By introducing an `OnDelete` update strategy (inspired by Kubernetes StatefulSet), users gain fine-grained control over when replicas are updated. The new specification is applied only when a replica is manually deleted, allowing unaffected replicas to continue running undisturbed.
```
unmarshall: This section should focus on what is the current problems that users have with the rolling recreate strategy and how does OnDelete addresses this.  You can then state an example and make that case stronger.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+### Goals
+
+- Introduce a new `spec.updateStrategy.type` field to `PodCliqueSet` that allows users to choose between `RollingRecreate` (default) and `OnDelete` update strategies.
+- The update strategy applies uniformly to both standalone PodCliques and PodCliqueScalingGroups within a PodCliqueSet.
+- When `OnDelete` strategy is configured:
+  - Changes to the template do not automatically trigger replica deletions.
+  - New replicas (created due to scale-out or manual deletion) use the updated template.
+  - Replica deletions prefer replicas with the older template.
+  - Existing replicas continue running with their original specification until manually deleted.
```
unmarshall: move this to be the second point.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+- Maintain backward compatibility by defaulting to the current `RollingRecreate` behavior.
+
+### Non-Goals
+
+- Implementing partition-based rolling updates (as in StatefulSet's `RollingUpdate` with `partition` field).
```
unmarshall: Should we even mention it? The very reason it is there is because you gave a reference to `StatefulSet` on-delete strategy in the Summary, right?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+- Providing a `maxUnavailable` or `maxSurge` configuration for rolling updates (can be addressed in a future GREP).
```
unmarshall: How is this relevant for this GREP?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+- Automatic detection of failed nodes and selective pod recreation. This should be the responsibility of the scheduler.
```
unmarshall: Not clear on how is this linked to this GREP. Can you explain?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+2. **Pod Recovery**: Recreate only the affected pods with strict `nodeAffinity` rules against faulty nodes.
+3. **Preservation**: Unaffected pods continue running to minimize disruption.
+
+Currently, when a user updates a standalone PodClique's `podTemplate` (e.g., changing `nodeAffinity` rules), Grove triggers a rolling recreate that deletes and recreates **all** pods one at a time, including unaffected ones. This behavior is disruptive and wasteful for workloads that only need to recreate and reschedule a subset of pods. Similar behavior is displayed in PodCliqueScalingGroups where a rolling recreate deletes and recreates **all** replicas one at a time.
+
+By introducing an `OnDelete` update strategy (inspired by Kubernetes StatefulSet), users gain fine-grained control over when replicas are updated. The new specification is applied only when a replica is manually deleted, allowing unaffected replicas to continue running undisturbed.
+
+### Goals
+
+- Introduce a new `spec.updateStrategy.type` field to `PodCliqueSet` that allows users to choose between `RollingRecreate` (default) and `OnDelete` update strategies.
```
unmarshall: There is a slight discrepancy. Here you mention `spec.updateStrategy.type` and in the `Proposal` section you mention `spec.updateStrategy`

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+- Automatic detection of failed nodes and selective pod recreation. This should be the responsibility of the scheduler.
+- Defining update strategy at each hierarchy level (PodCliqueSet, PodCliqueScalingGroup, PodClique) individually (following the design used for topology configuration in a PodCliqueSet). This can be addressed in a future GREP to allow different strategies for different components.
+
+## Proposal
+
+Introduce a new `spec.updateStrategy` field in the `PodCliqueSet` specification. The `UpdateStrategy` field describes how replicas should be updated when templates change.
+
+Two update strategy types will be supported:
+
+1. **RollingRecreate** (default): A strategy where a replica is deleted first, and then recreated with the new template. This applies to both pods (for standalone PodCliques) and replicas of PodCliqueScalingGroups.
```
unmarshall: which replica? PCS, PCLQ, PCSG? 
It is important to be precise since rolling recreate does not do PCS replica recreation.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+2. **OnDelete**: Replicas are not automatically deleted when the template changes. The new specification is applied only when a replica is manually deleted and recreated by the controller.
```
unmarshall: >Replicas are not automatically deleted when the template changes.

What you meant is that replicas are not automatically updated (either via recreation or some other way), right?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+As mentioned previously, support to override the update strategy at PodClique and PodCliqueScalingGroup hierarchies can be brought about in the future if needed.
```
unmarshall: The `Proposal` section should have clearly defined what is `OnDelete` and how to interpret it in context of PCS. 

Possible rewording:

```
The `OnDelete` update strategy provide manual control over when parts of a PodCliqueSet adopt the latest template changes. With this strategy:

**Template Changes Do Not Trigger Automatic Updates:**
- When you modify the pod template in `spec.cliques[*].spec` the controller **will not** automatically recreate any existing PodCliqueSet/PodCliqueScalingGroup replica.
- The updates to Pod templates in a PodCliqueSet are stored and becomes the "desired" specification, but existing replicas continue running with their current (potentially outdated) specification.

**Manual Deletion Triggers Updates:**
- Replicas are only updated when they are _manually deleted_ by a user or external process (e.g., through `kubectl delete pod`, node eviction, or manual replica deletion).
- When a replica of PodClique or PodCliqueScalingGroup is deleted (for any reason), the controller creates a replacement using the **current (updated) template** from the PodCliqueSet specification.

**Scale-Out uses updated template:**
- Any new replicas created due to scale-out operations will always use the current (updated) template specification.
- This ensures that scaling operations benefit from the latest configuration even if existing replicas have not been updated.

**Preferential Deletion During Scale-In:**
- _For Standalone PodCliques_: When scaling down (`spec.cliques[*].spec.replicas` is reduced), the controller preferentially selects pods running with outdated templates for deletion. This helps accelerate convergence to the desired specification during normal scaling operations.
- _For PodCliqueScalingGroups_: Scale-in continues to delete replicas with the highest indices to prevent gaps in the replica set, regardless of template version.

**Visibility and Tracking:**
< add how are you going to provide this >

```

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+### User Stories
+
+#### Story 1: Training Workload Node Failure Recovery
+
+As an AI engineer running a distributed training job across multiple nodes, when one node fails, and I need to update the `nodeAffinity` to exclude the failed node, I want only the affected pods (those on the failed node) to be rescheduled, while all other pods continue their training computation without interruption.
```
unmarshall: Is this a user story that we should mention?
>I manually delete the pods that were on the failed node (or they are automatically evicted by kubelet, or another actor in the system).

If they get evicted via KCM or Kubelet then you cannot have a deterministic outcome as the spec change could come in later than the eviction in which case the Pod's NodeAffinity is not yet updated to exclude bad nodes, right?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+With `OnDelete` strategy:
+1. I update the PodCliqueSet with new `nodeAffinity` rules excluding the failed node.
+2. The pods on healthy nodes continue running.
+3. I manually delete the pods that were on the failed node (or they are automatically evicted by kubelet, or another actor in the system).
+4. New pods are created with the updated `nodeAffinity` and scheduled on healthy nodes.
+
+#### Story 2: Non-Disruptive Affinity Updates
```
unmarshall: you specifically mention only `Affinity` updates. Are we saying that only for these use cases on-delete should be the preferred choice?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+- The PodCliqueScalingGroup's `status.updatedReplicas` field will accurately reflect how many replicas are running with the current (desired) specification.
+- Clear documentation will explain the behavior and best practices for using `OnDelete` strategy.
+
+**Risk 2: User Confusion**
+
+Users may not understand why their spec changes are not being applied automatically.
+
+*Mitigation*:
+- The API will clearly document the behavior of each update strategy.
+
```
unmarshall, line 147: There are other limitations:

* Pod deletions due to `preemption`, `eviction due to Node failure` or any other reason can cause you to lose control over how to roll out the updates.
* If the new version of the template is not compatible with older version then `on-delete` should never be used - perhaps this should be also mentioned in `Proposal` section on when to use and when not to use `OnDelete`.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+#### UpdateStrategyType
+
+```go
+// UpdateStrategyType defines the type of update strategy for PodCliqueSet.
+// +kubebuilder:validation:Enum={RollingRecreate,OnDelete}
+type UpdateStrategyType string
+
+const (
+    // RollingRecreateStrategyType indicates that replicas will be progressively
+    // deleted and recreated one at a time, when templates change. This applies to
```
unmarshall: `deleted and recreated` - recreate subsumes delete.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+#### UpdateStrategyType
+
+```go
+// UpdateStrategyType defines the type of update strategy for PodCliqueSet.
+// +kubebuilder:validation:Enum={RollingRecreate,OnDelete}
+type UpdateStrategyType string
+
+const (
+    // RollingRecreateStrategyType indicates that replicas will be progressively
```
unmarshall: which replicas?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+    // deleted and recreated one at a time, when templates change. This applies to
+    // both pods (for standalone PodCliques) and replicas of PodCliqueScalingGroups.
+    // This is the default update strategy.
+    RollingRecreateStrategyType UpdateStrategyType = "RollingRecreate"
+    
+    // OnDeleteStrategyType indicates that replicas will only be updated when
```
unmarshall: which replicas?

unmarshall (commented):
1/2 review comments

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+    // CurrentGenerationHash is a hash value generated out of a collection of fields in a PodCliqueSet.
+    // Since only a subset of fields is taken into account when generating the hash, not every change in the PodCliqueSetSpec will
+    // be accounted for when generating this hash value. A field in PodCliqueSetSpec is included if a change to it triggers
+    // an  update of PodCliques and/or PodCliqueScalingGroups.
+    // Only if this value is not nil and the newly computed hash value is different from the persisted CurrentGenerationHash value
+    // then an update needs to be triggerred.
+    CurrentGenerationHash *string `json:"currentGenerationHash,omitempty"`
+    // UpdateProgress represents the progress of an update.
+    // This field supersedes RollingUpdateProgress and supports all update strategies.
+    UpdateProgress *PodCliqueSetUpdateProgress `json:"updateProgress,omitempty"`
```
unmarshall: Maybe i have missed, but where is the definition of `PodCliqueSetUpdateProgress` type?

docs/proposals/291-ondelete-update/README.md
gflarity: No disagreement on this being non goal, but I am definitely curious how this is a responsibility of the scheduler.  It's not clear to me where the responsibility would lie, or even should lie.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+    // RollingUpdateProgress represents the progress of a rolling update.
+    // Deprecated: Use UpdateProgress instead. This field will be removed in a future release.
+    RollingUpdateProgress *PodCliqueSetRollingUpdateProgress `json:"rollingUpdateProgress,omitempty"`
+}
+```
+
+#### PodCliqueStatus Changes
+
+A new field `PodCliqueUpdateProgress` (`status.updateProgress`) will be introduced, which is identical to the existing `PodCliqueRollingUpdateProgress` (`status.rollingUpdateProgress`). This new field will be used to generically track information about updates to the PodClique with all supported strategies. The existing `RollingUpdateProgress` field will be marked as deprecated and will be removed in a future release. Both fields will contain identical data during the deprecation period, but all future enhancements will only be made to `UpdateProgress`:
```
unmarshall, line 259: can you please introduce the `PodCliqueUpdateProgress` for completeness.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+  	UpdateProgress *PodCliqueUpdateProgress `json:"updateProgress,omitempty"`
+  	// RollingUpdateProgress provides details about the ongoing rolling update of the PodClique.
+  	// Deprecated: Use UpdateProgress instead. This field will be removed in a future release.
+  	RollingUpdateProgress *PodCliqueRollingUpdateProgress `json:"rollingUpdateProgress,omitempty"`
+}
+```
+
+#### PodCliqueScalingGroupStatus Changes
+
+A new field `PodCliqueScalingGroupUpdateProgress` (`status.updateProgress`) will be introduced, which is identical to the existing `PodCliqueScalingGroupRollingUpdateProgress` (`status.rollingUpdateProgress`). This new field will be used to generically track information about updates to the PodCliqueScalingGroup with all supported strategies. The existing `RollingUpdateProgress` field will be marked as deprecated and will be removed in a future release. Both fields will contain identical data during the deprecation period, but all future enhancements will only be made to `UpdateProgress`:
```
unmarshall, line 300: Please define `PodCliqueScalingGroupUpdateProgress` for completeness.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+  	// This field supersedes RollingUpdateProgress and supports all update strategies.
+  	UpdateProgress *PodCliqueScalingGroupUpdateProgress `json:"updateProgress,omitempty"`
+  	// RollingUpdateProgress provides details about the ongoing rolling update of the PodCliqueScalingGroup.
+  	// Deprecated: Use UpdateProgress instead. This field will be removed in a future release.
+  	RollingUpdateProgress *PodCliqueScalingGroupRollingUpdateProgress `json:"rollingUpdateProgress,omitempty"`
+...
+}
+```
+
+> **Note:** In the future, update strategy could be defined at each hierarchy level to specify different behavior for different PodCliques/PodCliqueScalingGroups, following the convention used for topology configuration in PodCliqueSet. This would involve adding optional `UpdateStrategy` fields to `PodCliqueTemplateSpec` and `PodCliqueScalingGroupConfig` that override the PodCliqueSet-level strategy.
```
unmarshall: lets remove this since this is marked as a non-goal. Talking anything about non-goal defeats the purpose of a non-goal. How the API will evolve cannot be determined at this time and therefore is premature.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+2. The controller progressively deletes and recreates replicas to match the new specification.
+3. Each replica is deleted first, then recreated with the new template.
+
+#### OnDelete Strategy
+
+When `updateStrategy.type` is set to `OnDelete`:
+
+**For Standalone PodCliques:**
+1. Changes to a PodCliqueSet's `spec.cliques[*]` are recorded but do not trigger automatic pod deletions.
+2. The PodCliqueSet's `status.currentGenerationHash` is updated to reflect the new desired state.
```
unmarshall: So when you create PCS for the first time you will compute the PCS hash and set this as `status.currentGenerationHash`. After this point if a change is made to PCS which results in a hash that is different from the one that is captured in `status.currentGenerationHash` then a rolling update is triggered. Therefore `status.currentGenenationHash` does not represent the new desired state yet.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+1. Changes to a PodCliqueScalingGroup template trigger a rolling recreate of PodCliqueScalingGroup replicas.
+2. The controller progressively deletes and recreates replicas to match the new specification.
+3. Each replica is deleted first, then recreated with the new template.
+
+#### OnDelete Strategy
+
+When `updateStrategy.type` is set to `OnDelete`:
+
+**For Standalone PodCliques:**
+1. Changes to a PodCliqueSet's `spec.cliques[*]` are recorded but do not trigger automatic pod deletions.
```
unmarshall: usage of the word `recorded` is quite misleading and unclear.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+2. The PodCliqueSet's `status.currentGenerationHash` is updated to reflect the new desired state.
+3. The PodClique's `spec` is patched with the changes brought in the template at `spec.cliques[*].spec` of the PodCliqueSet.
+4. Existing pods continue running with their original specification.
+5. When a pod needs to be deleted (e.g., during scale-in), pods with the older template are preferred for deletion.
+6. When a pod is deleted (manually, during scale-in, or due to node failure/eviction):
+   - The controller creates a replacement pod using the current (updated) template.
```
unmarshall, line 366: why will the controller create a replacement pod for a scale-in?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+4. `status.updateProgress` fields track the progress of the update at each hierarchy.
+
+**For PodCliqueScalingGroups:**
+1. Changes to a PodCliqueScalingGroup template trigger a rolling recreate of PodCliqueScalingGroup replicas.
+2. The controller progressively deletes and recreates replicas to match the new specification.
+3. Each replica is deleted first, then recreated with the new template.
+
+#### OnDelete Strategy
+
+When `updateStrategy.type` is set to `OnDelete`:
```
unmarshall, line 358: Please check the detailed comment that i have made in the `Proposal` section where there is a more nicer way to segregate the information. Below multiple facets have been mixed.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+   - The new pod reflects any specification changes made since the original pod was created.
+7. The PodClique's `status.updateProgress.podTemplateHash` can be compared with individual pod labels to identify which pods are running outdated specifications.
+8. PodClique's `status.updatedReplicas` reflects the count of pods running with the current template.
+
+**For PodCliqueScalingGroups:**
+1. Changes to a PodCliqueSet's `spec.cliques[*]` that belong to a PodCliqueScalingGroup are recorded but do not trigger automatic replica deletions.
+2. The PodCliqueSet's `status.currentGenerationHash` is updated to reflect the new desired state.
+3. Existing replicas continue running with their original specification.
+4. When a replica is deleted (manually or during scale-in):
+   - The controller creates a replacement replica using the current (updated) template.
```
unmarshall: why will the controller create a replacement replica for a scale-in?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+Two update strategy types will be supported:
+
+1. **RollingRecreate** (default): A strategy where a replica is deleted first, and then recreated with the new template. This applies to both pods (for standalone PodCliques) and replicas of PodCliqueScalingGroups.
+2. **OnDelete**: Replicas are not automatically deleted when the template changes. The new specification is applied only when a replica is manually deleted and recreated by the controller.
+
+As mentioned previously, support to override the update strategy at PodClique and PodCliqueScalingGroup hierarchies can be brought about in the future if needed.
+
+### User Stories
+
+#### Story 1: Training Workload Node Failure Recovery
```
gflarity: I know this is a story we discussed based on an a operator request, but it's not clear why you'd want to leave a PCS in this weird state where you only really want the change for a single pod. Just cordoning the node then deleting the pod makes way more sense.  Maybe we can get more information from @kangclzjc

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+3. Existing replicas continue running with their original specification.
+4. When a replica is deleted (manually or during scale-in):
+   - The controller creates a replacement replica using the current (updated) template.
+5. PodCliqueScalingGroup's `status.updatedReplicas` reflects the count of replicas running with the current template.
+
+**For both resources**:
+1. The PodClique's `status.updateProgress.updateStartedAt` and `status.updateProgress.updateEndedAt` are both set simultaneously when the `OnDelete` update is initiated. The following are the reasons:
+  - From an `OnDelete` update perspective, the update ends as soon as the specification of the resource is synced with the template specified in the PodCliqueSet.Thus, the timestamp is set as the same for both.
+  - Gang termination is paused when a PodCliqueSet is being updated through the `RollingRecreate`. Since the user controls the time at which replicas are deleted, it is unknown when the update will end, and therefore is incorrect to pause gang termination until all resources are updated. Thus, both timestamps are set simultaneously, ensuring gang termination continues to be in effect.
```
unmarshall: `Since the user controls the time at which replicas are deleted,` should be prefixed with `For OnDelete update strategy since the user controls...`

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+  - Create a PodCliqueSet with PodCliqueScalingGroups and `OnDelete` strategy, update the template, and verify no replicas are deleted.
+  - Delete a PodCliqueScalingGroup replica manually and verify the replacement uses the new template.
+- Testcase 4:
+  - End-to-end test with mixed standalone PodCliques and PodCliqueScalingGroups.
+  - Verify uniform strategy application across both types.
+
+### Graduation Criteria
+
+**Alpha:**
+- `OnDelete` strategy for standalone PodCliques and PodCliqueScalingGroups is implemented behind a feature flag (if applicable).
```
unmarshall: we do not have a concept of feature gate yet. So you can perhaps leave this section empty for now. Once we have first class support for this then this section will become relevant.

docs/proposals/291-ondelete-update/README.md
gflarity: Again, I don't think this user story makes sense even though there seems to be some external interest in it. 

The user story that makes the most sense for OnDelete is one involving ine grained manual control of the update. Essential none of the other strategies fits the precise needs of the operator, this gives them control of the timing and order etc and allows them to perform other actions as necessary in between steps. 

CC @kangclzjc

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+#### Story 2: Non-Disruptive Affinity Updates
+
+As an operator, I want to update scheduling preferences (affinity, tolerations) for a running inference workload without causing service disruption. This could be for node maintenance, OS upgrade, Kubernetes version upgrade, etc. The `OnDelete` strategy allows me to update the specification and gradually roll out changes by selectively deleting pods during maintenance windows.
+
+### Limitations/Risks & Mitigations
+
+**Risk 1: Configuration Drift**
+
+With `OnDelete` strategy, replicas within the same PodCliqueSet may run with different specifications for an extended period if replicas are not manually deleted.
```
gflarity, line 122: Indeed. This risk is in conflict with the the user stories, esspecially once you consider gang termination.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+
+*Mitigation*:
```
gflarity, line 124: Hmm... also what about if you update a spec, delete one pod, then update the spec again and delete a different pod. You've got a mix of three different versions running... 
 Ultimately I think this will need to be managed by the operator carefully, which is fine, but this weird state needs to be accounted for in the implementation for sure.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,455 @@
...
+- The PodClique's `status.updatedReplicas` field will accurately reflect how many pods are running with the current (desired) specification.
+- The PodCliqueScalingGroup's `status.updatedReplicas` field will accurately reflect how many replicas are running with the current (desired) specification.
+- Clear documentation will explain the behavior and best practices for using `OnDelete` strategy.
+
+**Risk 2: User Confusion**
```
gflarity: I feel like this risk is pretty low. We should just explain this a manual rollout process, so they should only use if it they're paying attention and want that.

gflarity (commented):
Still making my way through it, minor comments so far.

docs/proposals/291-ondelete-update/README.md
kangclzjc: Several reasons here:
1. A node may still be running—or be capable of running—other tasks, so we can't simply condor the entire node.
2. Condoring a node is too heavy-handed; we shouldn't offline an entire node just because one task fails to schedule or run.
3. It might be that only a specific GPU on the node is faulty, while the remaining GPUs are still functional—so the node could still run jobs requiring fewer GPUs (e.g., 2-GPU jobs).

In fact, points 1, 2, and 3 all convey a similar idea: node resources may be partially available, and we shouldn't globally disable the entire node due to a localized issue. We should aim to maximize resource utilization.

docs/proposals/291-ondelete-update/README.md
renormalize: I will remove this, since it is not strictly needed.

docs/proposals/291-ondelete-update/README.md
renormalize: Mentioned this since `maxUnavailable` and `maxSurge` can not be supported with an `OnDelete` strategy. I will remove this since it will only cause more confusion.

docs/proposals/291-ondelete-update/README.md
renormalize: The original draft mentioned node failures and the usage of OnDelete with that, so I mentioned this point. The GREP must not really focus on that use case and should be more generic. I will therefore remove this point too.

renormalize: Thanks for the review comments!

I had mentioned that since my original draft of this GREP focused on nodes going bad, and the use of OnDelete with that context.

On second thought, I am unsure which component should hold the responsibility for such behavior, if we do intend to bake in such behavior into our components.

I will remove this point for now, since I personally haven't thought much about how we'd go about doing this. Also, I don't think we *need* to decide this for this GREP. We can discuss how and where this behavior will be introduced when the requirement comes.

docs/proposals/291-ondelete-update/README.md
renormalize: The operator only classifies pods into two buckets. Either the pod is
- "Updated": matches the template
- "Out of date": does not match the template.

From the operator's perspective, it doesn't really matter how many version of pods are running in the cluster - we only have pods that are up to date and pods that are not.

However, obviously, this could be confusing for the user. The onus is on the user to ensure they keep track of the pods that are out of date and delete them when they need an updated pod. We can't (and shouldn't in my opinion to avoid complicating both the code and the API) report more information than this.

docs/proposals/291-ondelete-update/README.md
renormalize: Yep, agreed. I stated that just to highlight to the maintainers that the `OnDelete` API is a bit more involved for consumers of grove.

docs/proposals/291-ondelete-update/README.md
renormalize: I will update the proposal to mention that "users who want to orchestrate updates manually" is the target audience for this feature. This is pretty general, and covers all use cases in my opinion. What do you think?

docs/proposals/291-ondelete-update/README.md
renormalize: Will rephrase this.

docs/proposals/291-ondelete-update/README.md
renormalize: Yes.

docs/proposals/291-ondelete-update/README.md
renormalize: I am planning to remove this, and specify something generic where the update strategy required by the user does not fit the `RollingRecreate` semantics, and therefore they choose OnDelete.

docs/proposals/291-ondelete-update/README.md
renormalize: Do you want whatever is listed in https://github.com/ai-dynamo/grove/pull/403#discussion_r2836026383 to be a part of the `Behavior Details` section or the `Proposal` section? I'm not sure I understand how you want me to structure the document.

docs/proposals/291-ondelete-update/README.md
renormalize: I had mentioned that the struct `PodCliqueSetUpdateProgress` will be identical to `PodCliqueSetRollingUpdateProgress` in line 176. Since that is not sufficient, I will introduce the entire struct even though it has identical fields.

docs/proposals/291-ondelete-update/README.md
renormalize: A rolling update is triggered along with refreshing the `.status.currentGenerationHash` with the value that corresponds to the latest template. So I would say that the statement:
> status.currentGenenationHash does not represent the new desired state yet

is incorrect.

docs/proposals/291-ondelete-update/README.md
renormalize: Users are free to use all available features of grove to run their workloads. This is just an example to discuss the feature, nothing more.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+## Summary
+
+This GREP introduces a new `OnDelete` update strategy for `PodCliqueSet` that allows users to update `PodClique` specifications (`podSpec` fields such as `image`) without automatically triggering pod deletions/restarts. With `OnDelete` strategy, only pods that are manually deleted are recreated with the new specification, enabling non-disruptive updates.
+
+Currently, Grove implements a `RollingRecreate` strategy that automatically deletes and recreates replicas one at a time (PodCliqueScalingGroup replica for PodCliqueScalingGroups, and Pod for PodCliques) when templates are updated. To support the new `OnDelete` strategy alongside the existing `RollingRecreate` behavior, this GREP introduces a new `updateStrategy` field to the `PodCliqueSet` spec, enabling users to explicitly choose their preferred update strategy by specifying `.spec.updateStrategy.type`.
+
+## Motivation
+
+Grove currently only supports updates through its default `RollingRecreate` strategy, and the orchestration of updates is fully handled by grove. However, there could be certain sophisticated use cases where users might want to orchestrate the updates themselves, instead of relying on grove.
```
unmarshall: no need to use `sophisticated`. 
Alternative:
```
Grove currently supports updates only through its default `RollingRecreate` strategy, with the entire update orchestration managed by Grove itself. However, there may be certain advanced or specific use cases where users prefer to control and orchestrate the updates manually rather than relying on Grove's built-in process.
```

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+Currently, when a user updates a standalone PodClique's `podTemplate`, grove triggers a rolling recreate that deletes and recreates **all** pods of that standalone PodClique one at a time, *including unaffected ones*. This behavior is disruptive and wasteful for workloads that only need to recreate and reschedule a subset of pods. Similar behavior is observed in PodCliqueScalingGroups where a rolling recreate deletes and recreates **all** replicas of that PodCliqueScalingGroup one at a time.
```
unmarshall: >Currently, when a user updates a standalone PodClique's `podTemplate`, grove triggers a rolling recreate that deletes and recreates **all** pods of that standalone PodClique one at a time, *including unaffected ones*.

Can you explain this a bit? If there is a change to the PodSpec of a standalone PCLQ, then how can some pods with unaffected? If you mean other PCLQs whose PodSpec's have not been changed then that is not coming out clearly in this para. Also we do not really recreate unaffected ones today, right?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+To solve the current limitation stated above, we propose to introduce a new update strategy `OnDelete` (inspired by Kubernetes StatefulSet), users gain fine-grained control over when pods are updated. The new specification is applied only when a pod is manually deleted, allowing unaffected pods to continue running undisturbed.
```
unmarshall:
```suggestion
To solve the current limitation stated above, we propose to introduce a new update strategy `OnDelete` (inspired by Kubernetes StatefulSet). Users gain fine-grained control over when pods are updated. The new specification is applied only when a pod is manually deleted, allowing unaffected pods to continue running undisturbed.
```

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+### Goals
+
+- Introduce a new `spec.updateStrategy` field to `PodCliqueSet` that allows users to choose between `RollingRecreate` (default) and `OnDelete` update strategy types.
+- The update strategy applies uniformly to both standalone PodCliques and PodCliqueScalingGroups within a PodCliqueSet.
+- When `OnDelete` strategy is configured:
+  - Changes to the template do not automatically trigger pod deletions.
+  - Existing pods continue running with their original specification until manually deleted.
+  - New pods (created due to scale-out or manual deletion) use the updated template.
+  - Pod deletions prefer pods with the older template.
```
unmarshall: not clear.
`Pod deletion` could also mean deleting specific pods. Then this statement is not true. You should instead be clear and specific and state for `Scale-in` pods with older PodSpec are preferred over newer Pods. But this is also not really true, right? If there is an unhealthy Pod with a new PodSpec template, then would you still prefer the older healthy pod?

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+## Proposal
+
+Introduce a new update strategy in grove called `OnDelete` that enables users to orchestrate updates themselves, configurable through the `.spec.updateStrategy` field of the PodCliqueSet.
+
+Two update strategy types will be supported:
+
+### `RollingRecreate`
+
+Pods are automatically updated when the template changes. In this strategy, pods of a PodClique are deleted first, and then created with the new template. The operator achieves this by:
+- Either deleting the Pods of a standalone PodClique (created with the new template by the PodClique controller)
```
unmarshall:
```suggestion
- Either deleting the Pods of a standalone PodClique (created with the new template by the PodClique controller) one at a time.
```

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+- Or deleting the PodCliques that belong to a specific replica of the PodCliqueScalingGroup (created with the new template by the PodCliqueSet or PodCliqueScalingGroup controller)
```
unmarshall:
```suggestion
- Or deleting PodCliqueScalingGroup replicas (one at a time) to ensure that all PodCliques in a single PodCliqueScalingGroup replica are updated at the same time and gang scheduled.
```

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+- Pods are only updated when they are *manually deleted* by a user or external process (e.g., through `kubectl delete pod`, node eviction, or manual replica deletion).  
+- When a pod (or a PodClique) is deleted (for any reason), the controller creates a replacement using the **current (updated) template** from the PodCliqueSet specification.
+
+**Scale-Out uses updated template:**
+- Any new pods created due to scale-out operations will always use the current (updated) template specification.
+- This ensures that scaling operations benefit from the latest configuration even if existing replicas have not been updated.
+
+**Preferential Deletion During Scale-In:**
+- *For Standalone PodCliques*: When scaling down (`spec.cliques[*].spec.replicas` is reduced), the controller preferentially selects pods running with outdated templates for deletion. This helps accelerate convergence to the desired specification during normal scaling operations.
+- *For PodCliqueScalingGroups*: Scale-in continues to delete replicas with the highest indices to prevent gaps in the replica set, regardless of template version.
```
unmarshall: >regardless of template version.

Not clear.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+**Visibility and Tracking:**
+
+To track the status of the update, users can check the `.status.updatedReplicas` fields of the PodCliqueSet, PodCliqueScalingGroup, and PodClique resources to check the number of replicas (in the case of PodCliqueSet and PodCliqueScalingGroup) or Pods (in the case of PodCliques) that match the template specified in the PodCliqueSet.
+
+This strategy can be set by specifying `.spec.updateStrategy.type: OnDelete` on a PodCliqueSet.
+> **Note:** The user must ensure that the `OnDelete` update strategy is compatible with their applications (which might have multiple PodCliques running in HA), and that such updates do not cause the application to behave unexpectedly due to incompatibility between different versions of components running simultaneously.
```
unmarshall: The OnDelete update strategy results in old and new versions of application components running concurrently during the update process. If there is any known or potential incompatibility between the newer and older versions, the `OnDelete` strategy should be avoided, because it can lead to runtime errors, unpredictable or incorrect application behavior and may jeopardize the stability and correct functioning of the system.

docs/proposals/291-ondelete-update/README.md
```diff
@@ -0,0 +1,521 @@
...
+### User Stories
+
+#### Story 1: Sophisticated application updates
+
+Consumers of grove with fairly complicated setups and requirements might want to orchestrate the updates themselves. The `RollingRecreate` behavior might not fit into how their application needs to be updated. In such cases:
+
+With `OnDelete` strategy:
+1. The consumer updates the PodCliqueSet with the new specifications for their PodCliques in the PodCliqueSet.
+2. They manually delete the pods that need to be recreated with the new specifications according to their requirements on when/which/how many pods are to be updated.
+3. New pods are created with the updated specifications, as and when the user wants updated pods to be created.
```
unmarshall: Not clear

docs/proposals/291-ondelete-update/README.md
unmarshall: How is this in conflict with the user story? Can you elaborate on its link to gang termination?

docs/proposals/291-ondelete-update/README.md
renormalize: "unaffected pods" is misleading. I will remove that from the sentence.

Without that, the sentence is clear enough in my opinion.

> Currently, when a user updates a standalone PodClique's `podTemplate`, Grove triggers a rolling recreate that deletes and recreates **all** pods of that standalone PodClique one at a time. This behavior is disruptive and wasteful for workloads that only need to recreate and reschedule a subset of pods. Similar behavior is observed in PodCliqueScalingGroups where a rolling recreate deletes and recreates **all** replicas of that PodCliqueScalingGroup one at a time.

docs/proposals/291-ondelete-update/README.md
renormalize: > If there is an unhealthy Pod with a new PodSpec template, then would you still prefer the older healthy pod?

I think that is a decision we have to make - I would prefer deleting old healthy replicas during scale-ins even when with the presence of unhealthy pods with the newer template. The goal of the operator should be to try and move the system to the desired state as much as it can with each action when an the PCS is being updated, be it through RollingRecreate or OnDelete.

This in theory might cause loss of availability, but we have to keep the context in mind: this is during scale-ins. The users should keep an active watch on their system during an update, and I would presume that scale operations either originate from the user (in which case we are just following the user's intent) or originate from an autoscaling component. For the latter case, I think it is a good enough compromise to just delete the old pods since users are expected to actively monitor and orchestrate the update themselves while using OnDelete. They should disable autoscaling if they do not like this behavior during an update.

docs/proposals/291-ondelete-update/README.md
renormalize: I've rephrased the last two points to clarify this. I don't have any particular update strategy that I want to specify in the GREP as an example because they might become outdated very soon causing the GREP to be outdated. I've kept the user story as open ended as possible intentionally.

docs/proposals/291-ondelete-update/README.md
unmarshall: Lets discuss this point with @sanjaychatterjee  as well and finalize.

unmarshall: Ok lets go with your suggestion for now. Since this is not really part of the API its fine to take one behavior and gather some usage feedback.

unmarshall (approved):

sanjaychatterjee (approved):

PR #410, opened 2026-02-08
add pr vaidation - issue must exist

### LINKED ISSUE #411: Each PR's should have an active issue referenced

Each PR's should have an active issue referenced

### Why is this needed?

For progress visibility and clear workflow

danbar2:

Fixes #411 

```release-note

```

```docs

```


.github/workflows/validate-issue-references.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Validate PR references at least one open issue
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const prNumber = context.payload.pull_request.number;
+            const prBody = context.payload.pull_request.body || '';
+
+            // Issue reference pattern: #123
+            const issuePattern = /#(\d+)/g;
```
renormalize, line 25: We're not checking for `fixes`, `closes`, or `resolves` in the body. If none of these are mentioned, the issue will continue to stay open even when the PR is merged. I guess we should include that check here?

I would also say that there should be two types of syntax supported:

```md
fixes #1, #2, #3
```

```md
fixes #1 #2 #3
```

as multiple issues can be resolved in one PR.

danbar2: About the first point, its not necessarily will resolve or fix an item, some features will require multiple pr's. So I dont think we should mandate that, its the developer ownership to close the issue when possible.
About the second, it is fine you are correct, we just need *at least* one issue ref.

The main goal is to make sure every PR is tracked by an issue

renormalize: I'd say if a feature requires multiple PRs, then the related feature issue should have multiple children, and each PR corresponds to one child issue. Once all children issues are closed, the parent can be closed too. With this convention, we can ensure that one issue is tied to one PR always and is closed at the merge of the PR.

However, I'd not want to block the PR for this argument, we can sync up with the contributors on how they'd like to go about this at a later stage and adapt this workflow accordingly then.

shayasoolin (approved):

PR #412, opened 2026-02-09
Upgrade golang version and k8s deps patch version

unmarshall:
Upgraded k8s deps to version v0.34.3 (latest patch version for v0.34)

#### What type of PR is this?

/kind enhancement

* Upgrading to golang 1.25 allows usage of new language features.
* Upgrading to the latest patch version v0.34 of k8s deps (api, apimachinery, client-go etc.)

Fixes #

```release-note
Upgrade to golang version 1.25
Upgrade k8s.io/* to v0.34.3
```

```docs

```


sanjaychatterjee (dismissed):

operator/go.mod
```diff
@@ -17,14 +17,16 @@ require (
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.11.1
 	go.uber.org/zap v1.27.0
+	golang.org/x/term v0.37.0
+	gopkg.in/yaml.v3 v3.0.1
```
renormalize, line 21: We should be using `sigs.k8s.io/yaml` as a direct dependency for all yaml related functionality. Adding this comment for now, we can remove the direct dependency on this in a future PR. Looks like it was used as a direct dependency in one of the e2e tests added in a previous PR and the module was not tidied till now.

scheduler/client/go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/ai-dynamo/grove/scheduler/client
 
-go 1.24.0
+go 1.25.0
 
 require (
 	github.com/ai-dynamo/grove/scheduler/api v0.0.0
-	k8s.io/apimachinery v0.34.2
+	k8s.io/apimachinery v0.34.3
 	k8s.io/client-go v0.33.1
```
renormalize: Wouldn't we want to bump up client-go as well? This is to be done in `operator/client` as well.

operator/go.mod
unmarshall: This will be taken up when we clean up e2e tests.

sanjaychatterjee (dismissed):

scheduler/client/go.mod
unmarshall: yes we can do that. Pushed another commit to bump up the version.

renormalize (changes requested):
`scheduler/client` does not seem to be synced. Can you check that once?

unmarshall:
> `scheduler/client` does not seem to be synced. Can you check that once?

Done

renormalize (approved):

sanjaychatterjee (approved):

Issue #369, opened 2026-01-25
Support multiple topologies


gal-revach:
### What you would like to be added?

Grove currently supports a single topology per cluster. 
This request is to enable support for multiple topologies within the same cluster, and allow the user (via Grove API) to specify which topology to use. If no topology is requested - use the default grove topology.

### Why is this needed?

Support for heterogeneous clusters with different topologies within the same cluster, e.g. GB200 and Vera Rubin.

Specifically for the flow of submitting Dynamo over Grove workload from Run:ai, it will help as Run:ai has the concept of node pools, where each can have a different topology attached. Once the user submits to a specific node pool, Run:ai should request this node pool's topology.

PR #413, opened 2026-02-09
GREP 369 - support multiple ClusterTopology

enoodle:
#### What type of PR is this?

/kind feature
/kind api

Introduces GREP-369 (#369) 
This is a proposal to extend Grove's topology API to support multiple named ClusterTopology resources within a single cluster. This allows heterogeneous clusters with different GPU architectures or multi-cloud environments to define separate topologies, and lets PodCliqueSets reference a specific topology via a new
  `clusterTopologyName` field.

### DIFF
--- a/docs/proposals/244-topology-aware-scheduling/README.md
+++ b/docs/proposals/244-topology-aware-scheduling/README.md
@@ -282,7 +282,7 @@ Validating webhook for `PodCliqueSet` will reject resources that are created wit
 
 #### ClusterTopology Deletion
 
-When an administrator deletes a ClusterTopology resource, any PodCliqueSets that still reference the deleted topology are affected. The PCS reconciler detects this and sets the `TopologyLevelsUnavailable` condition to `Unknown` with reason `ClusterTopologyNotFound`. Invalid topology constraints are removed from the PodGang resources created for those PodCliqueSets.
+When an administrator deletes a ClusterTopology resource, any PodCliqueSets that still reference the deleted topology are affected. The PCS reconciler detects this and sets the `TopologyLevelsUnavailable` condition to `True` with reason `ClusterTopologyNotFound`. Invalid topology constraints are removed from the PodGang resources created for those PodCliqueSets.
 
 **Mitigation**
 
@@ -566,7 +566,7 @@ A ClusterTopology's `spec.levels` can be updated in-place by administrators. Whe
 
 **Deletion**
 
-When an administrator deletes a ClusterTopology, auto-managed scheduler backend topology CRs are cascade-deleted via `OwnerReference`. If PodCliqueSets still reference the deleted topology, the PCS reconciler detects this at runtime and sets the `TopologyLevelsUnavailable` condition to `Unknown` with reason `ClusterTopologyNotFound`. Invalid topology constraints are removed from the PodGang resources created for those PodCliqueSets.
+When an administrator deletes a ClusterTopology, auto-managed scheduler backend topology CRs are cascade-deleted via `OwnerReference`. If PodCliqueSets still reference the deleted topology, the PCS reconciler detects this at runtime and sets the `TopologyLevelsUnavailable` condition to `True` with reason `ClusterTopologyNotFound`. Invalid topology constraints are removed from the PodGang resources created for those PodCliqueSets.
 
 ```mermaid
 sequenceDiagram
@@ -1028,7 +1028,7 @@ Condition States:
 
 | Status    | Reason                              | Description                                                  |
 | --------- | ----------------------------------- | ------------------------------------------------------------ |
-| `Unknown` | `ClusterTopologyNotFound`           | When `ClusterTopology` CR is no longer existing              |
+| `True`    | `ClusterTopologyNotFound`           | When `ClusterTopology` CR is no longer existing — levels are definitively unavailable |
 | `Unknown` | `TopologyAwareSchedulingDisabled`   | When TAS has been disabled cluster-wide while the PCS still has topology constraints |
 | `Unknown` | `TopologyNameMissing`        | When PCS has topology constraints but no `topologyName` (upgrade from single-topology design) |
 | `True`    | `ClusterTopologyLevelsUnavailable`  | When one or more topology levels used by a deployed `PodCliqueSet` are no longer present in `ClusterTopology` (e.g., the ClusterTopology levels were updated and no longer include the domains used by this PCS). The reconciler removes the invalid topology constraints from the PodGang resources so the scheduler no longer enforces them. Already-running pods are not evicted — they continue running at their current placement. |


docs/proposals/369-multi-topology-support/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+3. **Default topology**: When no topology is explicitly referenced, a designated default topology is used, preserving backward compatibility.
+
+4. **Mutable topology reference until scheduling**: The topology reference on a PodCliqueSet can be changed while the workload is not yet running, allowing users to retry scheduling against a different topology. Once any pod in the gang has been scheduled, the topology reference becomes immutable and updates are rejected.
+
+### User Stories
+
+#### Story 1: Heterogeneous GPU Clusters
+
+As a cluster administrator managing a cluster with different GPU architectures, I want to define separate topologies for each architecture so that workloads are scheduled with topology awareness appropriate to the hardware they're running on.
```
sanjaychatterjee: Can you please add a bit more details with an example? For example, let's add a concrete example to show why single cluster topology is not enough.

docs/proposals/369-multi-topology-support/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### Story 2: Multi-Cloud Clusters
+
+As a platform engineer managing AI workloads on a cluster spanning multiple cloud environments, I want to define environment-specific topologies so that workloads get optimal placement regardless of where they run, without users needing to know the underlying infrastructure details.
```
sanjaychatterjee: Griven Grove is currently designed for single cloud single cluster environment, can we add an example here to show the multi-cloud use-cse?

docs/proposals/369-multi-topology-support/README.md
enoodle: I added an example with a bit more details below

docs/proposals/369-multi-topology-support/README.md
enoodle: I think that I will remove this story - I think it might be solved by just labeling the nodes uniformly.

docs/proposals/244-topology-aware-scheduling/README.md
```diff
@@ -213,58 +282,130 @@ There are ways in which you can either minimize the need for on-demand scaling o
...
+Scheduler backend topology references are configured inside each `SchedulerProfile.Config` (GREP-375). For the KAI scheduler, this extends the backend-specific config:
+
+```go
+// KaiSchedulerConfig holds the configuration for the KAI scheduler backend.
+// This is deserialized from SchedulerProfile.Config (runtime.RawExtension) when Name == "kai-scheduler".
+type KaiSchedulerConfig struct {
+	// TopologyReferences maps topology profile names to KAI Topology CR names.
+	// Each entry tells the operator which KAI Topology CR corresponds to a given topology profile.
 	// +optional
-	Levels []corev1alpha1.TopologyLevel `json:"levels,omitempty"`
```
shayasoolin, line 237: How are we going to handle this API-breaking change? Or is it simply legit to require the admin to have an updated config before upgrading grove?

docs/proposals/244-topology-aware-scheduling/README.md
```diff
@@ -213,58 +282,130 @@ There are ways in which you can either minimize the need for on-demand scaling o
...
 }
 
+// TopologyReference maps a topology profile to a scheduler backend's topology resource.
+type TopologyReference struct {
+	// ProfileName is the name of the topology profile (must match a TopologyProfile.Name).
+	// +required
+	ProfileName string `json:"profileName"`
+	// ResourceName is the name of the scheduler backend's topology resource (e.g., KAI Topology CR name).
+	// +required
+	ResourceName string `json:"resourceName"`
```
shayasoolin: I think the full GVK is required here, as different scheduler backends will have different CRDs for that.

docs/proposals/244-topology-aware-scheduling/README.md
enoodle: I assume that it is legit to break this configuration API . @nvrohanv What do you think?

docs/proposals/244-topology-aware-scheduling/README.md
enoodle: This is referenced from inside a specific scheduler backend configuration, so this information is implied. For example if we create such backend configuration for volcano we will have a resource name to reference and the volcano scheduler backend implementation will know which GKV to use.

shayasoolin: I understood that while Grove doesn't need to create this resource, it does verify that it's aligned with the levels set in the operator config ("the operator will then verify drift instead of creating the resource"). So to be able to perform this verification it needs the GVK. Am I missing something?

enoodle: I updated the PR - But still without GVK. The reason is that it knows them implicitly from the type of scheduler backend that the reference belongs to.

docs/proposals/244-topology-aware-scheduling/README.md
```diff
@@ -328,63 +414,218 @@ In the first iteration, since `ClusterTopology` is completely managed by the  op
...
+    // Levels is an ordered list of topology levels from broadest to narrowest scope.
+    // The order in this list defines the hierarchy (index 0 = broadest level).
     // +kubebuilder:validation:MinItems=1
-    // +kubebuilder:validation:MaxItems=7
     // +kubebuilder:validation:XValidation:rule="self.all(x, self.filter(y, y.domain == x.domain).size() == 1)",message="domain must be unique across all levels"
     // +kubebuilder:validation:XValidation:rule="self.all(x, self.filter(y, y.key == x.key).size() == 1)",message="key must be unique across all levels"
     Levels []TopologyLevel `json:"levels"`
+    // SchedulerReferences maps this topology to scheduler backend topology resources.
+    // When empty, the operator automatically creates and manages the scheduler backend topology.
+    // When populated, the referenced resources are assumed to be externally managed and
```
shayasoolin: I'd use "provided" instead of "populated" here.
Also, not sure the meaning of "drift" is clear enough here. Consider adding a few words about the detection of mismatch of domain/key pairs and their exact order.
In general I'd consider replacing the term "drift" with "mismatch" or something else. To me, a drift is something that can start small and accumulate, but in this case it's a boolean condition - either a full mismatch or not.

docs/proposals/244-topology-aware-scheduling/README.md
```diff
@@ -328,63 +414,218 @@ In the first iteration, since `ClusterTopology` is completely managed by the  op
...
+}
+
+// SchedulerTopologyStatus reports the sync state between this ClusterTopology and a single
+// scheduler backend's topology resource.
+type SchedulerTopologyStatus struct {
+    // SchedulerName is the scheduler backend name (matches SchedulerReference.SchedulerName).
+    SchedulerName string `json:"schedulerName"`
+    // Reference is the scheduler backend topology resource name (matches SchedulerReference.Reference).
+    Reference string `json:"reference"`
+    // InSync is true when the scheduler backend topology levels match the ClusterTopology levels.
```
shayasoolin, line 501: `InSync` is a good name. I'd use it also for the condition, instead of `Drift`. Note that it will reverse the boolean values from negative to positive..

docs/proposals/244-topology-aware-scheduling/README.md
```diff
@@ -520,51 +762,98 @@ spec:
...
 
 *Rule-2: Check for Hierarchical strictness*
 
-As you traverse down the resource hierarchy (PodCliqueSet → PodCliqueScalingGroup → PodClique), topology constraint levels must become equal or lower. A child resource cannot specify a broader topology domain than its parent. If this rule is violated, then the creation of the `PodCliqueSet` will be rejected by the validating webhook.
+As you traverse down the resource hierarchy (PodCliqueSet → PodCliqueScalingGroup → PodClique), topology constraint levels must become equal or narrower (higher index in the ClusterTopology's levels array). A child resource cannot specify a broader topology domain than its parent. If this rule is violated, then the creation of the `PodCliqueSet` will be rejected by the validating webhook.
+
+> NOTE: The hierarchy is determined by the position of domains in the referenced ClusterTopology's levels array, not by a fixed global order. The examples below assume a ClusterTopology with levels ordered as `[zone, block, rack, host, numa]`.
```
shayasoolin: "not by a fixed global order." - this seems obvious, the reader is not supposed to have the context of the previous design :)

docs/proposals/244-topology-aware-scheduling/README.md
```diff
@@ -520,51 +762,98 @@ spec:
...
 
 Example:
 
 | Parent | Child   | Valid? | Reason                                                       |
 | ------ | ------- | ------ | ------------------------------------------------------------ |
-| `rack` | `host`  | ✅ Yes  | `host` is lower than `rack`                    |
+| `rack` | `host`  | ✅ Yes  | `host` is narrower (higher index) than `rack`                |
```
shayasoolin: Nit - the Example chart seems to me over detailed, the concept of level order is not that complex.

docs/proposals/244-topology-aware-scheduling/README.md
enoodle: The rest of the conditions in grove are negative based, except for Ready. (@sanjaychatterjee )

docs/proposals/244-topology-aware-scheduling/README.md
enoodle: :+1:

docs/proposals/244-topology-aware-scheduling/README.md
enoodle: I will make this less verbose.

docs/proposals/244-topology-aware-scheduling/README.md
enoodle: :+1:

PR #433, opened 2026-02-11
Addition of a `deletionTimestamp` on a `Pod` should trigger a `PodClique` reconciliation

### LINKED ISSUE #434: PCLQ controller fails to replace terminating Pods on NotReady nodes due to missed reconcile
### What happened?

When a Node becomes `NotReady`, the Pods on it first transition to `NotReady` (triggering a reconcile where they aren't yet terminating), and later get deleted by the node controller (receiving a `deletionTimestamp`).
However, the PCLQ controller does not trigger a new reconcile when the `deletionTimestamp` is added. This causes the Pods to remain in a terminating state without being replaced, resulting in the actual replica count dropping below the expected count and affecting service availability.

When Pods on a `NotReady` node are marked for deletion (receive a `deletionTimestamp`), the PCLQ controller should detect this change and trigger a reconciliation. It should then promptly create new replacement Pods to maintain the desired replica count and ensure service availability.

### Environment

- Kubernetes version
- Grove version
- Scheduler details

xulinfei1996:
#### What type of PR is this?

/kind bug
In our scenario, when a Node becomes NotReady, the following sequence of events occurs:
1. The Pods on that Node first transition to the NotReady state;
2. After a period of time, the node controller deletes these Pods, at which point the Pods are marked with a `deletionTimestamp`.
In the PCLQ controller, step 1 triggers a reconcile, but since the Pods are not yet marked with a `deletionTimestamp`, the controller does not recognize them as being in a terminating state. By step 2, although the Pods now have a `deletionTimestamp`, the PCLQ controller does not trigger another reconcile.

This results in: Pods that are in a terminating state, but the PCLQ controller does not create new Pods to replace them, thereby affecting service availability and the expected replica count.
This PR aims to fix this issue by ensuring that when Pods enter a terminating state, the PCLQ controller can promptly detect this and create new Pods to maintain the desired replica count.

Fixes #434

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podclique/register_test.go
+++ b/operator/internal/controller/podclique/register_test.go
@@ -77,14 +77,16 @@ func TestPodPredicate_Delete(t *testing.T) {
 	})
 }
 
+// Test_isMarkedForDeletion tests if a deletion timestamp is set on the pod
 func Test_isMarkedForDeletion(t *testing.T) {
+	now := ptr.To(metav1.Now())
 	tests := []struct {
 		name        string
 		updateEvent event.UpdateEvent
 		want        bool
 	}{
 		{
-			name: "pod deletionTimestamp changed",
+			name: "deletion timestamp set on the pod in the update",
 			updateEvent: event.UpdateEvent{
 				ObjectOld: &corev1.Pod{
 					ObjectMeta: metav1.ObjectMeta{
@@ -93,11 +95,51 @@ func Test_isMarkedForDeletion(t *testing.T) {
 				},
 				ObjectNew: &corev1.Pod{
 					ObjectMeta: metav1.ObjectMeta{
-						DeletionTimestamp: ptr.To(metav1.Now()),
+						DeletionTimestamp: now,
 					},
 				},
 			},
 			want: true,
+		},
+		{
+			name: "deletion timestamp not set on the pod in the update",
+			updateEvent: event.UpdateEvent{
+				ObjectOld: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						DeletionTimestamp: nil,
+					},
+				},
+				ObjectNew: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						DeletionTimestamp: nil,
+					},
+				},
+			},
+			want: false,
+		},
+		{
+			name: "deletion timestamp was already set on the pod before the update",
+			updateEvent: event.UpdateEvent{
+				ObjectOld: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						DeletionTimestamp: now,
+					},
+				},
+				ObjectNew: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						DeletionTimestamp: now,
+					},
+				},
+			},
+			want: false,
+		},
+		{
+			name: "objects are not pods (type cast fails)",
+			updateEvent: event.UpdateEvent{
+				ObjectOld: &corev1.ConfigMap{},
+				ObjectNew: &corev1.ConfigMap{},
+			},
+			want: false,
 		},
 	}
 	for _, tt := range tests {


operator/internal/controller/podclique/register.go
```diff
@@ -129,7 +129,8 @@ func hasPodStatusChanged(updateEvent event.UpdateEvent) bool {
 	return hasReadyConditionChanged(oldPod.Status.Conditions, newPod.Status.Conditions) ||
 		hasLastTerminationStateChanged(oldPod.Status.InitContainerStatuses, newPod.Status.InitContainerStatuses) ||
 		hasLastTerminationStateChanged(oldPod.Status.ContainerStatuses, newPod.Status.ContainerStatuses) ||
-		hasStartedAndReadyChangedForAnyContainer(oldPod.Status.ContainerStatuses, newPod.Status.ContainerStatuses)
+		hasStartedAndReadyChangedForAnyContainer(oldPod.Status.ContainerStatuses, newPod.Status.ContainerStatuses) ||
+		hasTerminationChanged(oldPod, newPod)
```
kangclzjc: Should we only pass DeletionTimestamp in the function like other function `hasStartedAndReadyChangedForAnyContainer`

kangclzjc:
Thanks @xulinfei1996 , very great finding.

operator/internal/controller/podclique/register_test.go
```diff
@@ -20,10 +20,44 @@ import (
...
+					},
+				},
+				ObjectNew: &corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						DeletionTimestamp: ptr.To(metav1.Now()),
+					},
+				},
+			},
+			want: true,
+		},
```
Ronkahn21: Please add another  test senario where status of pod does not consider changed as well

renormalize:
Reviewing this, please do not merge this yet.

operator/internal/controller/podclique/register.go
```diff
@@ -129,7 +131,8 @@ func hasPodStatusChanged(updateEvent event.UpdateEvent) bool {
 	return hasReadyConditionChanged(oldPod.Status.Conditions, newPod.Status.Conditions) ||
 		hasLastTerminationStateChanged(oldPod.Status.InitContainerStatuses, newPod.Status.InitContainerStatuses) ||
 		hasLastTerminationStateChanged(oldPod.Status.ContainerStatuses, newPod.Status.ContainerStatuses) ||
-		hasStartedAndReadyChangedForAnyContainer(oldPod.Status.ContainerStatuses, newPod.Status.ContainerStatuses)
+		hasStartedAndReadyChangedForAnyContainer(oldPod.Status.ContainerStatuses, newPod.Status.ContainerStatuses) ||
+		hasTerminationChanged(oldPod.DeletionTimestamp, newPod.DeletionTimestamp)
```
renormalize: `hasPodStatusChanged()` only deals with changes in the `Status` field of the Pod.

The `DeletionTimestamp` is a metadata field, so I'd like to separate this and keep it outside of `hasPodStatusChanged()`. We can rename `hasTerminationChanged()` to `isMarkedForDeletion()` or something of that sort, and move the invocation into the `podPredicate()` function. There we can simply check something like:

```go
UpdateFunc: func(updateEvent event.UpdateEvent) bool {
		return isManagedPod(updateEvent.ObjectOld) && !hasPodSpecChanged(updateEvent) && (hasPodStatusChanged(updateEvent) || isMarkedForDeletion(updateEvent))
},
```

operator/internal/controller/podclique/register.go
```diff
@@ -285,3 +288,7 @@ func isManagedPod(obj client.Object) bool {
 	}
 	return grovectrlutils.HasExpectedOwner(constants.KindPodClique, pod.OwnerReferences) && grovectrlutils.IsManagedByGrove(pod.GetLabels())
 }
+
+func hasTerminationChanged(oldDeletionTime, newDeletionTime *metav1.Time) bool {
+	return oldDeletionTime == nil && newDeletionTime != nil
+}
```
renormalize, line 318: https://github.com/ai-dynamo/grove/pull/433/changes#r2827257745


```suggestion

func isMarkedForDeletion(oldDeletionTime, newDeletionTime *metav1.Time) bool {
	return oldDeletionTime == nil && newDeletionTime != nil
}
```

renormalize (changes requested):
Thanks for the PR! Just a few nit comments.

The pods would eventually reach `.spec.replicas` in number after the `DELETE` event is sent, but for pods that are stuck in `Terminating` for long, the replacement pods might take a long time to be created if no event is generated for that specific PodClique.

Also, please fix the formatting through `GOWORK=off make check`, as the `check` CI job is failing.

operator/internal/controller/podclique/register.go
xulinfei1996: Done as  suggested: 
1. Separated `DeletionTimestamp` logic from `hasPodStatusChanged()
2. Renamed hasTerminationChanged() to isMarkedForDeletion()

operator/internal/controller/podclique/register_test.go
xulinfei1996: Done, renames the test as Test_isMarkedForDeletion()

operator/internal/controller/podclique/register.go
renormalize: thanks!

operator/internal/controller/podclique/register_test.go
renormalize: Added a couple more cases myself so we don't have to go through another review cycle. I hope that is fine.

(I also rebased on top of main since there were merge conflicts)

operator/internal/controller/podclique/register.go
renormalize: This was changed and is now outdated. Resolving this since all conversations need to be resolved to merge.

PR #438, opened 2026-02-17
[GREP-291] `OnDelete` implementation for standalone `PodClique`s and `PodCliqueScalingGroup`s

### LINKED ISSUE #291: Enhance the rolling update approach for PodClique

renormalize:

/kind api feature

This PR introduces the implementation for the `OnDelete` update strategy as proposed in the GREP #403.

Fixes #291 

I have not yet finished unit tests and E2E tests. I'm planning to bring along the unit tests in this PR itself, and the E2E tests in a subsequent PR.

```release-note
New field `.spec.updateStrategy` of the `PodCliqueSet` supports specifying different update strategies for the PodCliqueSet; with `RollingRecreate` and `OnDelete` as the two options.
```

```docs
N/A
```


operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -110,13 +129,38 @@ type PodCliqueSetRollingUpdateProgress struct {
...
+	// UpdateEndedAt is the time at which the update ended for the PodCliqueSet.
+	// +optional
+	UpdateEndedAt *metav1.Time `json:"updateEndedAt,omitempty"`
+	// UpdatedPodCliqueScalingGroups is a list of PodCliqueScalingGroup names that have been updated to the desired PodCliqueSet generation hash.
+	UpdatedPodCliqueScalingGroups []string `json:"updatedPodCliqueScalingGroups,omitempty"`
+	// UpdatedPodCliques is a list of PodClique names that have been updated to the desired PodCliqueSet generation hash.
+	UpdatedPodCliques []string `json:"updatedPodCliques,omitempty"`
+	// CurrentlyUpdating captures the progress of the PodCliqueSet replica that is currently being updated.
+	// +optional
+	CurrentlyUpdating *PodCliqueSetReplicaUpdateProgress `json:"currentlyUpdating,omitempty"`
```
sanjaychatterjee: Shouldn't this be a slice to enable multiple concurrent updates in future? cc @unmarshall

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -242,6 +286,22 @@ type HeadlessServiceConfig struct {
...
+const (
+	// RollingRecreateStrategyType indicates that replicas will be progressively
+	// deleted and recreated one at a time, when templates change. This applies to
+	// both pods (for standalone PodCliques) and replicas of PodCliqueScalingGroups.
+	// This is the default update strategy.
+	RollingRecreateStrategyType UpdateStrategyType = "RollingRecreate"
+	// OnDeleteStrategyType indicates that replicas will only be updated when
+	// they are manually deleted. Changes to templates do not automatically
+	// trigger replica deletions.
+	OnDeleteStrategyType UpdateStrategyType = "OnDelete"
```
sanjaychatterjee: Also, can we rename the fields as `RollingRecreateStrategy` and `OnDeleteStrategy`?

operator/api/core/v1alpha1/podcliqueset.go
renormalize: Yeah, we can avoid a future breaking change in the status API if we think through the fields of the `PodCliqueSetUpdateProgress` struct in this PR.

operator/api/core/v1alpha1/podcliqueset.go
renormalize: Addressed in https://github.com/ai-dynamo/grove/pull/438/commits/cf63a2e9003dcff85695a874c8bfd36c298698cb.

operator/api/core/v1alpha1/podcliqueset.go
renormalize: Addressed in [e096e2e](https://github.com/ai-dynamo/grove/pull/438/commits/e096e2eb8d54df25d655d8a387fb0478a1a620e7).

operator/api/core/v1alpha1/podclique.go
```diff
@@ -155,7 +159,26 @@ type PodCliqueRollingUpdateProgress struct {
 	ReadyPodsSelectedToUpdate *PodsSelectedToUpdate `json:"readyPodsSelectedToUpdate,omitempty"`
 }
 
-// PodsSelectedToUpdate captures the current and previous set of pod names that have been selected for update in a rolling update.
+// PodCliqueUpdateProgress provides details about the ongoing update of the PodClique.
+type PodCliqueUpdateProgress struct {
+	// UpdateStartedAt is the time at which the update started.
+	UpdateStartedAt metav1.Time `json:"updateStartedAt,omitempty"`
+	// UpdateEndedAt is the time at which the update ended.
+	// It will be set to nil if the update is still in progress.
```
unmarshall: Maybe you can add more information about how this field is interpreted in context to different update strategies.

operator/api/core/v1alpha1/podclique.go
```diff
@@ -155,7 +159,26 @@ type PodCliqueRollingUpdateProgress struct {
...
+	UpdateEndedAt *metav1.Time `json:"updateEndedAt,omitempty"`
+	// PodCliqueSetGenerationHash is the PodCliqueSet generation hash corresponding to the PodCliqueSet spec that is being rolled out.
```
unmarshall: In case of `On-Delete` nothing is getting rolled out. So the description should just state that this holds the latest hash of a PCS to which all child resources should be updated to. Then go on and add the description on how one could use this field to know which child resources are yet to to be updated.

operator/api/core/v1alpha1/podclique.go
```diff
@@ -155,7 +159,26 @@ type PodCliqueRollingUpdateProgress struct {
...
+	// While the update is in progress PodCliqueStatus.CurrentPodCliqueSetGenerationHash will not match this hash. Once the update is complete the
+	// value of this field will be copied to PodCliqueStatus.CurrentPodCliqueSetGenerationHash.
+	PodCliqueSetGenerationHash string `json:"podCliqueSetGenerationHash"`
+	// PodTemplateHash is the PodClique template hash corresponding to the PodClique spec that is being rolled out.
```
unmarshall: same for PodTemplateHash as well

operator/api/core/v1alpha1/podclique.go
```diff
@@ -155,7 +159,26 @@ type PodCliqueRollingUpdateProgress struct {
...
+	// While the update is in progress PodCliqueStatus.CurrentPodTemplateHash will not match this hash. Once the update is complete the
+	// value of this field will be copied to PodCliqueStatus.CurrentPodTemplateHash.
+	PodTemplateHash string `json:"podTemplateHash"`
+	// ReadyPodsSelectedToUpdate captures the pod names of ready Pods that are either currently being updated or have been previously updated.
+	ReadyPodsSelectedToUpdate *PodsSelectedToUpdate `json:"readyPodsSelectedToUpdate,omitempty"`
```
unmarshall, line 188: what will be the semantic meaning of this for an on-delete since Grove does not select any Pod, it is selected indirectly by the user when s(he) deletes the pod thus expressing an intent that the new pod needs to come up with the latest/updated spec.

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -110,13 +129,39 @@ type PodCliqueSetRollingUpdateProgress struct {
...
+	UpdateStartedAt metav1.Time `json:"updateStartedAt,omitempty"`
+	// UpdateEndedAt is the time at which the update ended for the PodCliqueSet.
+	// +optional
+	UpdateEndedAt *metav1.Time `json:"updateEndedAt,omitempty"`
+	// UpdatedPodCliqueScalingGroups is a list of PodCliqueScalingGroup names that have been updated to the desired PodCliqueSet generation hash.
+	UpdatedPodCliqueScalingGroups []string `json:"updatedPodCliqueScalingGroups,omitempty"`
+	// UpdatedPodCliques is a list of PodClique names that have been updated to the desired PodCliqueSet generation hash.
+	UpdatedPodCliques []string `json:"updatedPodCliques,omitempty"`
+	// CurrentlyUpdating captures the progress of the PodCliqueSet replicas that are currently being updated.
+	// Currently, only a single replica update is supported at a time, so only the first element is used.
```
unmarshall: this line should be removed.

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -110,13 +129,39 @@ type PodCliqueSetRollingUpdateProgress struct {
...
+	// +optional
+	CurrentlyUpdating []PodCliqueSetReplicaUpdateProgress `json:"currentlyUpdating,omitempty"`
+}
+
+// PodCliqueSetReplicaUpdateProgress captures the progress of an update for a specific PodCliqueSet replica.
+type PodCliqueSetReplicaUpdateProgress struct {
+	// ReplicaIndex is the replica index of the PodCliqueSet that is being updated.
+	ReplicaIndex int32 `json:"replicaIndex"`
+	// UpdateStartedAt is the time at which the update started for this PodCliqueSet replica index.
+	UpdateStartedAt metav1.Time `json:"updateStartedAt,omitempty"`
```
unmarshall, line 172: should it also have `UpdateEndedAt` which is in line with similar field present for overall PCS (`PodCliqueSetUpdateProgress.UpdateEndedAt`)?

operator/api/core/v1alpha1/scalinggroup.go
```diff
@@ -120,14 +124,41 @@ type PodCliqueScalingGroupRollingUpdateProgress struct {
...
+	// While the update is in progress PodCliqueScalingGroupStatus.CurrentPodCliqueSetGenerationHash will not match this hash. Once the update is complete the
+	// value of this field will be copied to PodCliqueScalingGroupStatus.CurrentPodCliqueSetGenerationHash.
+	PodCliqueSetGenerationHash string `json:"podCliqueSetGenerationHash"`
+	// UpdatedPodCliques is the list of PodClique names that have been updated to the latest PodCliqueSet spec.
+	UpdatedPodCliques []string `json:"updatedPodCliques,omitempty"`
+	// ReadyReplicaIndicesSelectedToUpdate provides the update progress of ready replicas of PodCliqueScalingGroup that have been selected for update.
+	// PodCliqueScalingGroup replicas that are either pending or unhealthy will be force updated and the update will not wait for these replicas to become ready.
+	// For all ready replicas, one replica is chosen at a time to update, once it is updated and becomes ready, the next ready replica is chosen for update.
+	ReadyReplicaIndicesSelectedToUpdate *PodCliqueScalingGroupReplicaUpdateProgress `json:"readyReplicaIndicesSelectedToUpdate,omitempty"`
```
unmarshall, line 164: Will this field be relevant for on-delete update strategy? If not then you should update the doc string accordingly.

operator/api/core/v1alpha1/scalinggroup.go
```diff
@@ -120,14 +124,41 @@ type PodCliqueScalingGroupRollingUpdateProgress struct {
...
+}
+
+// PodCliqueScalingGroupReplicaUpdateProgress provides details about the update progress of ready replicas of PodCliqueScalingGroup that have been selected for update in a rolling recreate. It is not set in an OnDelete update.
+type PodCliqueScalingGroupReplicaUpdateProgress struct {
+	// Current is the index of the PodCliqueScalingGroup replica that is currently being updated.
+	Current int32 `json:"current"`
```
unmarshall, line 171: how is this going to be used in case of on-delete?

operator/api/core/v1alpha1/scalinggroup.go
```diff
@@ -120,14 +124,41 @@ type PodCliqueScalingGroupRollingUpdateProgress struct {
...
+	// UpdatedPodCliques is the list of PodClique names that have been updated to the latest PodCliqueSet spec.
+	UpdatedPodCliques []string `json:"updatedPodCliques,omitempty"`
+	// ReadyReplicaIndicesSelectedToUpdate provides the update progress of ready replicas of PodCliqueScalingGroup that have been selected for update.
+	// PodCliqueScalingGroup replicas that are either pending or unhealthy will be force updated and the update will not wait for these replicas to become ready.
+	// For all ready replicas, one replica is chosen at a time to update, once it is updated and becomes ready, the next ready replica is chosen for update.
+	ReadyReplicaIndicesSelectedToUpdate *PodCliqueScalingGroupReplicaUpdateProgress `json:"readyReplicaIndicesSelectedToUpdate,omitempty"`
+}
+
+// PodCliqueScalingGroupReplicaUpdateProgress provides details about the update progress of ready replicas of PodCliqueScalingGroup that have been selected for update in a rolling recreate. It is not set in an OnDelete update.
+type PodCliqueScalingGroupReplicaUpdateProgress struct {
```
unmarshall, line 169: since there is no orchestrated rollout, there is not going to be any real progress for on-delete, right?

operator/internal/controller/common/component/utils/podclique.go
```diff
@@ -156,12 +156,12 @@ func ComputePCLQPodTemplateHash(pclqTemplateSpec *grovecorev1alpha1.PodCliqueTem
 
 // IsPCLQUpdateInProgress checks if PodClique is under rolling update.
 func IsPCLQUpdateInProgress(pclq *grovecorev1alpha1.PodClique) bool {
-	return pclq.Status.RollingUpdateProgress != nil && pclq.Status.RollingUpdateProgress.UpdateEndedAt == nil
+	return pclq.Status.UpdateProgress != nil && pclq.Status.UpdateProgress.UpdateEndedAt == nil
 }
 
 // IsLastPCLQUpdateCompleted checks if the last rolling update of PodClique is completed.
```
unmarshall: update the docs to indicate how is this relevant for on-delete

operator/internal/controller/podclique/components/pod/syncflow.go
```diff
@@ -151,7 +151,7 @@ func (r _resource) runSyncFlow(logger logr.Logger, sc *syncContext) syncFlowResu
 		}
 	}
 
-	if componentutils.IsPCLQUpdateInProgress(sc.pclq) {
+	if (sc.pcs.Spec.UpdateStrategy == nil || sc.pcs.Spec.UpdateStrategy.Type == grovecorev1alpha1.RollingRecreateStrategy) && componentutils.IsPCLQUpdateInProgress(sc.pclq) {
```
unmarshall: For any auto-orchestration updates, update-in-progress is relevant. Perhaps you should first rename the function `IsPCLQUpdateInProgress` to something like `IsPCLQAutoUpdateInProgress` that indicates that this function is only checking for rollout progress of auto-updates orchestrated by Grove.

Secondly may be you can create a simple function which checks if the update strategy chosen for a PCS is auto-update strategy. Tomorrow when we introduce a new auto-update strategy other than the existing one then you do not have to then change this code again.

operator/internal/controller/podclique/components/pod/syncflow.go
```diff
@@ -232,8 +232,17 @@ func selectExcessPodsToDelete(sc *syncContext, logger logr.Logger) []*corev1.Pod
 	var candidatePodsToDelete []*corev1.Pod
 	if diff := len(sc.existingPCLQPods) - int(sc.pclq.Spec.Replicas); diff > 0 {
 		logger.Info("found excess pods for PodClique", "numExcessPods", diff)
-		sort.Sort(DeletionSorter(sc.existingPCLQPods))
-		candidatePodsToDelete = append(candidatePodsToDelete, sc.existingPCLQPods[:diff]...)
+		sorter := DeletionSorter{
```
unmarshall, line 235: Can you create a function:

```go
func (sc *syncContext) getExpectedPodTemplateHash() string {
	// If PCLQ is currently getting updated, then get the expected pod template hash from the update progress status.
	// This ensures that during an update, we are always comparing against the same expected template hash value for
	// all pods in the PCLQ, preventing thrashing of pods due to expected template hash value changing mid-sync.
	// If there is no update in progress, we can get the expected pod template hash from the PCLQ labels as usual.
	if sc.pclq.Status.UpdateProgress != nil &&
		sc.pcs.Status.CurrentGenerationHash != nil &&
		sc.pclq.Status.UpdateProgress.PodCliqueSetGenerationHash == *sc.pcs.Status.CurrentGenerationHash {
		return sc.pclq.Status.UpdateProgress.PodTemplateHash
	}
	return sc.pclq.Labels[apicommon.LabelPodTemplateHash]
}
```
and then use that in `selectExcessPodsToDelete`?

unmarshall (commented):
1/n reviews

operator/internal/controller/podcliquescalinggroup/reconcilespec.go
```diff
@@ -66,18 +67,28 @@ func (r *Reconciler) ensureFinalizer(ctx context.Context, logger logr.Logger, pc
...
+// processUpdate handles rolling update initialization for PodCliqueScalingGroups when the parent PodCliqueSet is updating
+func (r *Reconciler) processUpdate(ctx context.Context, logger logr.Logger, pcsg *grovecorev1alpha1.PodCliqueScalingGroup) ctrlcommon.ReconcileStepResult {
 	pcsgObjectKey := client.ObjectKeyFromObject(pcsg)
 	pcs, err := componentutils.GetPodCliqueSet(ctx, r.client, pcsg.ObjectMeta)
 	if err != nil {
 		return ctrlcommon.ReconcileWithErrors(fmt.Sprintf("could not get owner PodCliqueSet for PodCliqueScalingGroup: %v", pcsgObjectKey), err)
 	}
-	if pcs.Status.RollingUpdateProgress == nil || pcs.Status.RollingUpdateProgress.CurrentlyUpdating == nil {
+
+	if pcs.Spec.UpdateStrategy != nil && pcs.Spec.UpdateStrategy.Type == grovecorev1alpha1.OnDeleteStrategy {
```
Ronkahn21: nit: lets create some helper method to simply this 
pcs.Spec.UpdateStrategy != nil && pcs.Spec.UpdateStrategy.Type 

this check is allover the code and some helper method will be nice

operator/internal/controller/podclique/components/pod/syncflow.go
unmarshall: I just noticed that we already have an existing method `componentutils.GetExpectedPCLQPodTemplateHash`. So even though both approaches to compute the hash is different but the resulting hash value will be the same. Can we think of just having a single way to get this hash?

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -61,6 +61,8 @@ const (
 	errCodeParsePodCliqueScalingGroupReplicaIndex        grovecorev1alpha1.ErrorCode = "ERR_PARSE_PODCLIQUESCALINGGROUP_REPLICA_INDEX"
 	errCodeUpdateStatus                                  grovecorev1alpha1.ErrorCode = "ERR_UPDATE_STATUS"
 	errCodeComputePendingPodCliqueScalingGroupUpdateWork grovecorev1alpha1.ErrorCode = "ERR_COMPUTE_PENDINGUPDATE_WORK"
+	errCodeCreateOrUpdatePodClique                       grovecorev1alpha1.ErrorCode = "ERR_CREATE_OR_UPDATE_PODCLIQUE"
```
unmarshall: why do we have 2 error codes here?

operator/internal/controller/podcliquescalinggroup/register.go
```diff
@@ -89,28 +89,43 @@ func mapPCSToPCSG() handler.MapFunc {
 		if !ok {
 			return nil
 		}
-		pcsgConfigs := pcs.Spec.Template.PodCliqueScalingGroupConfigs
-		if len(pcsgConfigs) == 0 {
+		// We are only interested in PCS events during updates.
+		if pcs.Status.UpdateProgress == nil {
```
unmarshall, line 93: can you check if we set this to nil once the rolling update has completed?

operator/internal/controller/podcliquescalinggroup/register.go
```diff
@@ -131,14 +146,18 @@ func shouldEnqueueOnPCSUpdate(event event.UpdateEvent) bool {
...
-				oldPCS.Status.RollingUpdateProgress.CurrentlyUpdating.ReplicaIndex != newPCS.Status.RollingUpdateProgress.CurrentlyUpdating.ReplicaIndex {
+	if oldPCS.Status.UpdateProgress != nil && newPCS.Status.UpdateProgress != nil {
+		if utils.OnlyOneIsEmpty(oldPCS.Status.UpdateProgress.CurrentlyUpdating, newPCS.Status.UpdateProgress.CurrentlyUpdating) ||
+			len(oldPCS.Status.UpdateProgress.CurrentlyUpdating) > 0 &&
+				len(newPCS.Status.UpdateProgress.CurrentlyUpdating) > 0 &&
+				oldPCS.Status.UpdateProgress.CurrentlyUpdating[0].ReplicaIndex != newPCS.Status.UpdateProgress.CurrentlyUpdating[0].ReplicaIndex {
 			return true
 		}
 	}
+	// TODO: @renormalize this can be modified to requeue only when PCS.CurrentGenerationHash does not match PCSG.CurrentPodCliqueSetGenerationHash
```
unmarshall: did not understand this comment. Why is it a TODO?

operator/internal/controller/podcliqueset/components/podcliquesetreplica/rollingupdate.go
```diff
@@ -294,7 +296,7 @@ func isPCLQUpdateComplete(pclq *grovecorev1alpha1.PodClique, currentPCSGeneratio
 	return false
 }
 
-// isRollingUpdateInProgress checks if a rolling update is currently in progress.
-func isRollingUpdateInProgress(pcs *grovecorev1alpha1.PodCliqueSet) bool {
-	return pcs.Status.RollingUpdateProgress != nil && pcs.Status.RollingUpdateProgress.UpdateEndedAt == nil
+// isUpdateInProgress checks if an update is currently in progress.
+func isUpdateInProgress(pcs *grovecorev1alpha1.PodCliqueSet) bool {
```
unmarshall: this function is only used in case of auto-rolling updates. Therefore please rename this function and include the check `(pcs.Spec.UpdateStrategy == nil || pcs.Spec.UpdateStrategy.Type == grovecorev1alpha1.RollingRecreateStrategy)` in here.

operator/internal/controller/podcliqueset/reconciler.go
```diff
@@ -83,6 +84,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 // reconcileDelete handles PodCliqueSet deletion when a deletion timestamp is set.
 func (r *Reconciler) reconcileDelete(ctx context.Context, logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet) ctrlcommon.ReconcileStepResult {
 	if !pcs.DeletionTimestamp.IsZero() {
+		pcsObjectKey := ctrlclient.ObjectKeyFromObject(pcs)
+		pcsObjectName := cache.NamespacedNameAsObjectName(pcsObjectKey).String()
```
unmarshall: why was this code needed?

operator/internal/utils/miscellaneous.go
```diff
@@ -27,3 +27,8 @@ func IsEmptyStringType[T ~string](val T) bool {
 func OnlyOneIsNil[T any](objA, objB *T) bool {
 	return (objA == nil) != (objB == nil)
 }
+
+// OnlyOneIsEmpty returns true if only one of the Objects is empty else it will return false.
+func OnlyOneIsEmpty[T any](objA, objB []T) bool {
```
unmarshall, line 32: write a unit test for this function.

operator/api/core/v1alpha1/podclique.go
renormalize: We don't necessarily use this PCS Generation hash for deciding which child resources to update - we use the PCLQ template hash to check which pods are up to date and which are not. I will mention PCLQ template hash being used to verify if pods are up to date or not.

operator/api/core/v1alpha1/podclique.go
renormalize: This field is set for auto update strategies, and for the OnDelete strategy it is left empty (as the user decides which pod is to be deleted). Will enhance the API documentation with this explanation.

operator/api/core/v1alpha1/scalinggroup.go
renormalize: Since the user fully handles the pod deletions, we can't really populate these fields with anything. As the API documentation already present in the branch mentions: we do not populate this field for OnDelete.

operator/api/core/v1alpha1/scalinggroup.go
renormalize: Yes, you are correct. The field that uses this type `ReadyReplicaIndicesSelectedToUpdate` will only be set for auto update strategies. I've enhanced the API documentation to explicitly mention this.

operator/internal/controller/podcliquescalinggroup/register.go
renormalize: Removing this. There's nothing that can be done in this case. We simply enqueue the PCSG for another reconciliation when the PCS does not have an auto update strategy.

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
renormalize: that was a mistake I made while modifying the code. Removing it now.

operator/internal/controller/podcliquescalinggroup/register.go
renormalize: We don't set `Status.UpdateProgress` to `nil` after a rolling update. The code added here was a `nil` guard for the fields of `UpdateProgress` accessed in the conditional below. This `nil` guard is necessary for when the PCS has never been updated, and events are generated for it.

I see that the comment is misleading. Will remove it.

operator/internal/controller/podcliqueset/reconciler.go
renormalize: I had observed that creating and deleting a PodCliqueSet in quick succession caused none of the child resources to ever be created. All children resources were only created once the operator was restarted.

operator/internal/controller/podclique/components/pod/syncflow.go
renormalize: `componentutils.GetExpectedPCLQPodTemplateHash` computes the hash to establish the source of truth. That source of truth is already specified in `pclq.Status.UpdateProgress.PodTemplateHash` in the `initOrResetUpdate()` reconciliation step (and the label is updated in the parent controller's `createOrUpdate()`). I would say we should simply make use of this source of truth that has already been established.

I'd like to avoid recomputing the hash since as it is unnecessary here from a performance perspective.

operator/internal/controller/podcliqueset/reconciler.go
unmarshall: So if a PCS is created followed by deletion of the same in quick succession then no resource should be created, right? I did not understand.

renormalize: @unmarshall what I meant in my comment was the following:

- Create a PCS
- Delete that PCS immediately
- Recreate the same PCS

However, I was mistaken in my comment above. I now recall that the edge case where I was observing wrong behavior was the following:

- Create a PCS: spec-1
- Modify the PCLQ specification (for example, the pod specification): spec-2
- Delete the PCS
- Recreate the PCS in spec-1
- Upon recreation, no child resources are created

Unfortunately I am unable to reproduce this bug, and after having gone through the code once more, I do not think this section is necessary.

I will remove this code.

operator/internal/controller/podcliquescalinggroup/reconcilespec.go
renormalize: Done.

operator/internal/controller/podclique/components/pod/syncflow.go
```diff
@@ -232,12 +232,25 @@ func selectExcessPodsToDelete(sc *syncContext, logger logr.Logger) []*corev1.Pod
...
+		candidatePodsToDelete = append(candidatePodsToDelete, sorter.Pods[:diff]...)
 	}
 	return candidatePodsToDelete
 }
 
+func (sc *syncContext) getExpectedPodTemplateHash() string {
+	if sc.pclq.Status.UpdateProgress != nil &&
+		sc.pcs.Status.CurrentGenerationHash != nil &&
+		sc.pclq.Status.UpdateProgress.PodCliqueSetGenerationHash == *sc.pcs.Status.CurrentGenerationHash {
+		return sc.pclq.Status.UpdateProgress.PodTemplateHash
```
gflarity: **Double-check the hash fallback in `getExpectedPodTemplateHash`:** When `UpdateProgress.PodCliqueSetGenerationHash == *pcs.Status.CurrentGenerationHash`, this returns the update’s target hash — but when they differ (i.e., the PCS generation has moved forward and this PCLQ hasn’t been updated yet), it falls through to the label hash (the old hash). For the `DeletionSorter` during scale-in, would we not want to always prefer the latest expected hash so old-spec pods sort first for deletion?

unmarshall (approved):

sanjaychatterjee (approved):

PR #446, opened 2026-02-24
GREP 285: Training job support

### LINKED ISSUE #285: Enhance grove to support training jobs

Grove in its current form does not support training jobs through the `PodCliqueSet` API. It inherently only supports inference workloads, since the `PodClique` controller is not designed to handle Pods that successfully exit after their workload finishes.

We need to enhance Grove in the following areas to support training:

- Add a `state` field to the `PodCliqueSetStatus` which embeds the phase of the workload, i.e. `Pending`, `Running`, `Completed`, `Failed`, etc.
- Enhance the `PodClique` controller to be aware of job-like workloads' lifecycles, not just long running inference applications. The pods exiting successfully should not cause the controller to recreate Pods.
- Specify a `maxRuntime` field in a relevant place in the `PodCliqueSet`to specify how long the user expects their training workload to finish in. The semantic of this field can be discussed on further - if the workload does not finish before this duration, we could fire alerts to inform the user, and/or mark the workload as `Failed`.
- Support a `maxRestarts` field in a relevant place in the `PodCliqueSet` to indicate the number of restarts of a pod the operator must tolerate, before it marks the workload as `Failed`.
- Disable rolling updates for workloads which are jobs, since it does not make much sense to roll your pods in such a scenario.
- Disable autoscaling that is supported at all levels, since the consumer should, and will know their workload and capacity before they start training.

### Why is this needed?

Interest about grove, and its adoption is growing in the community. There are multiple features that are unique to grove, that the community would want to utilize not just for their inference workloads, but also for their training workloads.

shayasoolin:
/kind documentation
  /kind feature

  Adds GREP-285 for training job support in `PodCliqueSet`. The document covers the
  proposed API changes, lifecycle phase semantics, `trainingSpec` fields (`maxRuntime`, `maxRestarts`),
  `terminationDelay` defaulting, webhook enforcement, conditions, and events needed to support finite training
  workloads alongside existing inference workloads.

  Fixes #285

  This is a design/proposal PR only — no code changes. Feedback welcome on API naming, `maxRestarts` counting
  semantics, and the `terminationDelay` defaulting behavior for training workloads.

  Extends the `PodCliqueSet` API with a `workloadType` field (`Inference`/`Training`),
  a `phase` field in status, and a `trainingSpec` stanza (`maxRuntime`, `maxRestarts`)
  to support finite training workloads alongside existing inference workloads.


docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+## Motivation
+
+Interest in grove and its adoption is growing in the community. Grove's unique capabilities — topology-aware scheduling, gang scheduling, clique startup ordering, and MNNVL support — are valuable not just for inference but also for distributed training workloads. The current `PodClique` controller assumes pods run indefinitely and recreates them when they exit, which is incompatible with training jobs where pods exiting successfully is the expected terminal state.
+
+### Goals
+
+- Introduce a `workloadType` field to distinguish `Inference` (default, no behavior change) from `Training` workloads.
+- Add a `phase` field to `PodCliqueSetStatus` representing the workload lifecycle: `Pending`, `Running`, `Succeeded`, `Failed`.
+- Add `maxRuntime` and `maxRestarts` controls under a `trainingSpec` stanza.
+- Default `terminationDelay` to `0` for training workloads for immediate failure response.
```
kangclzjc, line 41: If we set terminationDelay default to 0, does it means that gang termination be triggered when one pod fail/deleted? Wouldn't we wait for the failed/deleted pod to be created again?

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+### Non-Goals
+
+- Per-clique failure or success policies (e.g., tolerate worker failures but not launcher failures).
+- Suspend/resume support.
+
+## Proposal
+
+A new `workloadType` field is added to `PodCliqueSetSpec`. When set to `Training`, the operator changes its behavior in the following ways:
+
+- Pods that exit with code 0 are not recreated. When all pods in a replica complete successfully, the replica is marked `Succeeded`.
```
kangclzjc, line 53: What happens if we support in-place update in the near future. If we in-place update the pod, would we consider it as success or fail? For this kind of action/job, we shouldn't block the recreate, right?

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+```yaml
+spec:
+  workloadType: Training
+  trainingSpec:
+    maxRuntime: 24h
+    maxRestarts: 2
+```
+
+A "restart" means deleting and recreating all pods across all PodCliques in the affected replica. The counter is a single total across all replicas. The PodCliqueSet is marked `Failed` when the total restart count exceeds `maxRestarts`.
```
kangclzjc, line 137: Can we provide `restart` action also for `PCSG` and `PCLQ` level ?

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+### terminationDelay
+
+`terminationDelay` (default: 4 hours) is the grace period grove gives after `MinAvailableBreached` before tearing down the gang. For inference workloads, this allows time for transient pod failures to recover. For training workloads, a missing worker means the job cannot make progress — a 4-hour wait wastes reserved GPU capacity and delays the retry.
```
kangclzjc, line 141: A training job can have several part, each part fail may don't affect other parts, all we need to do is to restart the part of job. A missing worker which may not mean the job can't make progress.

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+When `workloadType: Training`, `terminationDelay` defaults to `0`. The operator immediately tears down the replica and begins a restart (or marks it `Failed` if `maxRestarts` is exhausted). Users can override this with an explicit value if their framework tolerates brief pod gaps.
```
kangclzjc, line 143: Could we add multiple level of Tear down the replica of PCS/PCSG/PCLQ

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+### Controller Changes
+
+#### Defaulting Webhook
+
+- Default `terminationDelay` to `0` when `workloadType: Training` (instead of 4 hours).
+- Default pod `restartPolicy` to `Never` when `workloadType: Training` (instead of `Always`). For training, the operator controls restarts at the job level; kubelet-level pod restarts would interfere with the operator's failure detection and restart counting.
+
+#### Validating Webhook
+
+- Reject pod template spec changes that would produce a new generation hash (i.e., would trigger a rolling update) for Training workloads.
```
kangclzjc, line 154: We should support in-place update which could change spec. Maybe we should allow this action

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+- **maxRuntime enforcement** — record `startTime` in PCS status when phase first transitions to `Running`. `startTime` is never reset on restarts — `maxRuntime` is a wall-clock deadline from first start, inclusive of any time spent failing and restarting. This matches the behavior of Kubernetes Job's `activeDeadlineSeconds` and Kubeflow's `runPolicy.activeDeadlineSeconds`, and bounds total GPU consumption regardless of retry count. On each reconcile, check elapsed time; if elapsed > `maxRuntime`, terminate all pods and set phase `Failed`. Use `requeueAfter(remainingDuration)` to avoid busy-polling.
+- **Restart orchestration** — when a PodClique has `MinAvailableBreached: True`, check `status.restartCount` vs `maxRestarts`. If budget remains, delete all pods across all PodCliques in the replica, increment `restartCount`, and emit a `ReplicaRestarting` event. If exhausted, terminate all remaining pods, set phase `Failed`, and emit `MaxRestartsExceeded`.
```
kangclzjc, line 169: One PodClique has `MinAvailableBreached: True` should avoid delete all PodCliques's all pods

athreesh (commented):
just sharing some perspective from RL training users that might be useful to take into consideration. not blocking, just food for thought:

the biggest question i have for RL workloads is around cross-PCSG lifecycle coordination. the target architecture is a PCS containing training (finite) + inference (long-running) + environment/gym (long-running). when training succeeds, ideally inference drains in-flight work before terminating rather than getting hard-killed.

i wonder if something like this could eventually make sense:
```yaml
terminationPolicy:
  onSuccess:
    - pcsg: inference
      action: drain
    - pcsg: gym
      action: delete
```

this definitely doesn't need to block the current GREP. `workloadType: Training` on a standalone PCSG is useful on its own. just curious if you've thought about this as a natural next step, or if there's a reason it's better handled outside grove.

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+- Per-clique failure or success policies (e.g., tolerate worker failures but not launcher failures).
+- Suspend/resume support.
+
+## Proposal
+
+A new `workloadType` field is added to `PodCliqueSetSpec`. When set to `Training`, the operator changes its behavior in the following ways:
+
+- Pods that exit with code 0 are not recreated. When all pods in a replica complete successfully, the replica is marked `Succeeded`.
+- A `trainingSpec` stanza provides `maxRuntime` (maximum allowed duration) and `maxRestarts` (total restart budget across all replicas). Exceeding either terminates all pods and marks the workload `Failed`.
```
athreesh, line 54: sharing some perspective from RL training workloads that might be useful to consider here:

the pattern we're seeing (NeMo-RL migrating to K8s, others likely similar) is a single PCS with mixed-lifecycle PCSGs:

- training PCSG: run-to-completion (this GREP)
- inference PCSG: long-running, respawned
- gym/sandbox PCSG: long-running, ideally drained when training completes

i wonder if `workloadType` at the PCS level might be too coarse for this? could it make sense per-PCSG instead? or maybe the PCS needs a termination policy that coordinates across PCSGs, e.g. when training PCSG succeeds, drain inference, then delete gym?

without something like that, users end up orchestrating cross-PCSG teardown from outside grove (peer-watcher sidecars, etc.), which is what NeMo-RL does today on KubeRay. just wondering if that's worth considering here or as a follow-up.

docs/proposals/0285-training-job-support/README.md
```diff
@@ -0,0 +1,234 @@
...
+spec:
+  workloadType: Training
+  trainingSpec:
+    maxRuntime: 24h
+    maxRestarts: 2
+```
+
+A "restart" means deleting and recreating all pods across all PodCliques in the affected replica. The counter is a single total across all replicas. The PodCliqueSet is marked `Failed` when the total restart count exceeds `maxRestarts`.
+
+### terminationDelay
```
athreesh, line 139: one thought on the `terminationDelay: 0` default for training:

for Ray-based training (NeMo-RL, OpenRLHF, etc.), the Ray head pod acts as the application-level controller. when a GPU worker dies, the head detects it and decides whether to retry or abort, usually faster than the grove reconcile loop.

have you considered whether `0` might be too aggressive for this pattern? the head might want 30-60s to attempt a graceful recovery before grove tears down the gang. the "users can override" escape hatch handles it, but i wonder if documenting this as the expected pattern for Ray-based workloads would be helpful, or even defaulting to `30s` instead of `0`?

athreesh (commented):
just sharing some perspective from RL training users that might be useful to take into consideration. not blocking, just food for thought:

the biggest question i have for RL workloads is around cross-PCSG lifecycle coordination. the target architecture is a PCS containing training (finite) + inference (long-running) + environment/gym (long-running). when training succeeds, ideally inference drains in-flight work before terminating rather than getting hard-killed.

i wonder if something like this could eventually make sense:
```yaml
terminationPolicy:
  onSuccess:
    - pcsg: inference
      action: drain
    - pcsg: gym
      action: delete
```

this definitely doesn't need to block the current GREP. `workloadType: Training` on a standalone PCSG is useful on its own. just curious if you've thought about this as a natural next step, or if there's a reason it's better handled outside grove.

PR #447, opened 2026-02-24
replace deprecated runner

### LINKED ISSUE #448: CI is using a deprecated runner
Grove CI is using the deprecated `cpu-amd-m5-2xlarge` .

ranrubin:

The runner that grove currently use `cpu-amd-m5-2xlarge` is on Dynamo's Legacy cluster. That runner is the only runner left on that cluster and is blocking its deletion. 

Closes  #448 

Added `prod-grove-e2e-v1` runner to Dynamo's production cluster. The Dockerfile can be found [here](https://github.com/ai-dynamo/velonix/blob/main/dockerfiles/grove-runners/eks-grove-e2e-runner.Dockerfile), and the configuration can be found [here](https://github.com/ai-dynamo/velonix/blob/main/flux-apps/infra-apps/github-actions/runner-scale-sets/grove/prod-grove-e2e-v1/autoscalingrunnerset.yaml). 

Because our production runner uses DinD with a k8s native sidecar, I had to make some changes to the test infra:  
 
**Memory limits** — `--agents-memory 150m` doesn't work in DinD (broken `/proc/meminfo` bind-mount). Several tests depend on nodes having limited memory to verify pods stay Pending when resources are insufficient. We now use kubelet `system-reserved` to achieve the same effective allocatable memory, triggered via a new `--dind-memory-mode` flag.  

**Stale node list** — With 30 k3d nodes, occasionally one dies mid-run. Our node monitor replaces it, but `PrepareForTest` still had the old node cached, causing all subsequent tests to fail trying to cordon a deleted node. Fixed by refreshing the node list from the cluster before each test.
Both GS and TAS suites passing on the new runners

None
```release-note

```

```docs

```

### DIFF
--- a/operator/hack/infra_manager/cluster.py
+++ b/operator/hack/infra_manager/cluster.py
@@ -18,6 +18,7 @@
 
 from __future__ import annotations
 
+import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import docker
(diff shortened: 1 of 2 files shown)


.github/workflows/build-check-test.yaml
```diff
@@ -89,34 +89,34 @@ jobs:
...
     timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
         include:
-          - test_name: gang_scheduling
-            test_pattern: "^Test_GS"
-          - test_name: rolling_updates
-            test_pattern: "^Test_RU"
+          # - test_name: gang_scheduling
```
gflarity: Un comment test suites please.

.github/workflows/build-check-test.yaml
```diff
@@ -133,7 +133,7 @@ jobs:
 
       - name: Run e2e tests - ${{ matrix.test_name }}
         run: |
-          make ${{ matrix.make_target || 'run-e2e-full' }} TEST_PATTERN='${{ matrix.test_pattern }}'
+          make ${{ matrix.make_target || 'run-e2e-full' }} TEST_PATTERN='${{ matrix.test_pattern }}' E2E_CREATE_FLAGS='--skip-memory-limit'
```
gflarity: Can you tell me a bit more about the memory limit and the need to skip it?  These tests use quite a bit of memory for now. Just want to be sure they pass successfully with this setting.

operator/hack/e2e-cluster/create-e2e-cluster.py
```diff
@@ -122,7 +122,7 @@ class ClusterConfig(BaseSettings):
     api_port: int = Field(default=6560, ge=1, le=65535)
     lb_port: str = "8090:80"
     worker_nodes: int = Field(default=30, ge=1, le=100)
-    worker_memory: str = Field(default="150m", pattern=r"^\d+[mMgG]?$")
+    worker_memory: Optional[str] = Field(default="150m", pattern=r"^\d+[mMgG]?$")
```
gflarity: The tests rely on memory characteristics to impose scheduling constraints. Don't think this can be changed.

Ronkahn21 (commented):
Ran CodeRabbit review with project CLAUDE.md rules — two findings in `operator/hack/e2e-cluster/create-e2e-cluster.py`:

---

**🐛 Bug: Wrong node condition used for readiness check** (`wait_for_nodes`, ~line 375)

`conditions[-1].status` retrieves the *last* condition in the array, which is not guaranteed to be the `Ready` condition. Kubernetes nodes have multiple conditions (`MemoryPressure`, `DiskPressure`, `PIDPressure`, `Ready`), and their order is not fixed. This could cause nodes to be incorrectly identified as NotReady (or vice versa).

Fix — use JSONPath to explicitly select the `Ready` condition:
```diff
-            "-o", "custom-columns=NAME:.metadata.name,STATUS:.status.conditions[-1].status",
+            "-o", "custom-columns=NAME:.metadata.name,STATUS:.status.conditions[?(@.type=='Ready')].status",
```

---

**🧹 Unused parameter: `config` in `wait_for_nodes`** (~line 354)

`config: ClusterConfig` is declared but never referenced in the function body. Either remove it or use it to scope container restarts to this cluster (by filtering on the `k3d-<cluster_name>-` node name prefix), which would be safer in environments with multiple clusters running.

Option A — remove it:
```diff
-def wait_for_nodes(config: ClusterConfig, max_restart_rounds: int = 2):
+def wait_for_nodes(max_restart_rounds: int = 2):
```

Option B — use it for safety:
```python
# Only restart containers belonging to this cluster
if not node_name.startswith(f"k3d-{config.cluster_name}-"):
    continue
```

PR #454, opened 2026-02-25
Fix: let operator self-create webhook TLS secret to avoid GitOps overwrites 

### LINKED ISSUE #453: Helm chart renders empty webhook TLS secret that overwrites populated certs on every apply
## What happened?

The Helm chart template `operator/charts/templates/webhook-server-cert-secret.yaml` renders a `grove-webhook-server-cert` Secret with empty `tls.crt` and `tls.key` fields when `certProvisionMode` is `auto` (the default). The cert-controller populates this secret with auto-generated certificates at startup, but on subsequent deploys via `helm template | kubectl apply`, the empty values overwrite the populated certificates, breaking webhook TLS until the cert-controller repopulates the secret.

The same issue occurs with GitOps tools (ArgoCD/FluxCD) that detect the populated cert data as drift from the desired empty state and may reset it on sync.

The underlying constraint is that the upstream cert-controller library (`open-policy-agent/cert-controller`) only updates existing secrets — it does not create them. This is why the chart pre-creates the empty secret. However, the operator has full CRUD RBAC permissions on secrets, so it could create the secret itself.

## What did you expect to happen?

Deploying the Helm chart via `helm template | kubectl apply` should not overwrite previously populated webhook TLS certificates. The operator should be able to manage the webhook cert secret lifecycle independently of Helm.

## Environment

- Affects all Grove versions with `certProvisionMode: auto` (default)
- Reproducible with `helm template | kubectl apply` or GitOps tools (ArgoCD, FluxCD)
- Not affected when using standard `helm upgrade` (Helm 3's three-way merge preserves live data)

gflarity:
#### What type of PR is this?

/kind bug

The Helm chart template renders an empty `grove-webhook-server-cert` Secret when `certProvisionMode` is `auto` (the default). On subsequent deploys via `helm template | kubectl apply` or GitOps tools (ArgoCD, FluxCD), the empty values overwrite the populated certificates, breaking webhook TLS until cert-controller repopulates the secret.

This PR fixes the issue with three coordinated changes:

1. **New `webhookServerSecret.enabled` Helm value** (default `true` for backward compatibility) — allows users deploying via `helm template | kubectl apply` or GitOps to set it to `false`, preventing the empty secret from overwriting auto-generated certs on every apply.

2. **Operator self-creates the secret if missing** — a new `ensureSecretExists()` function creates the TLS secret before handing off to cert-controller, which can only Update existing secrets (not Create them). This fills the gap when the Helm-rendered secret is disabled. Handles HA race conditions via `AlreadyExists` check and preserves existing certificate data.

3. **`optional: true` on the deployment volume mount** — allows the pod to start before the secret exists, since the operator gates webhook serving on `certsReadyCh`.

Fixes #453

- The default `webhookServerSecret.enabled=true` preserves full backward compatibility for `helm upgrade` users (Helm 3's three-way merge already handles this correctly).
- The `ensureSecretExists` function uses a direct (non-cached) client because it runs during manager setup before informer caches start.
- The operator already has `create` RBAC permission on secrets via its ClusterRole.
- All existing e2e tests pass (CM, GS, RU, SO, TAS suites).

```release-note
Add `webhookServerSecret.enabled` Helm value to control whether the chart renders the empty webhook TLS Secret. Set to `false` when deploying via `helm template | kubectl apply` or GitOps tools (ArgoCD, FluxCD) to prevent certificate overwrites. The operator will create the secret itself if missing.
```

```docs
Users deploying via `helm template | kubectl apply` or GitOps tools should set `webhookServerSecret.enabled=false` in their values. The operator will automatically create and manage the webhook TLS secret. Standard `helm upgrade` users do not need to change anything.
```

### DIFF
--- a/operator/internal/controller/cert/cert.go
+++ b/operator/internal/controller/cert/cert.go
@@ -48,7 +48,7 @@ const (
 // When mode=auto: uses cert-controller for automatic certificate generation and management.
 // When mode=manual: waits for externally provided certificates (e.g., from cert-manager, cluster admin).
 // Returns an error for unrecognized modes to ensure new modes are explicitly handled.
-func ManageWebhookCerts(mgr ctrl.Manager, certDir string, secretName string, authorizerEnabled bool, certProvisionMode configv1alpha1.CertProvisionMode, certsReadyCh chan struct{}) error {
+func ManageWebhookCerts(ctx context.Context, mgr ctrl.Manager, cl client.Client, certDir string, secretName string, authorizerEnabled bool, certProvisionMode configv1alpha1.CertProvisionMode, certsReadyCh chan struct{}) error {
 	logger := ctrl.Log.WithName("cert-management")
 
 	switch certProvisionMode {
@@ -60,7 +60,7 @@ func ManageWebhookCerts(mgr ctrl.Manager, certDir string, secretName string, aut
 		return nil
 
 	case configv1alpha1.CertProvisionModeAuto:
-		return setupAutoCertProvisioning(mgr, certDir, secretName, authorizerEnabled, certsReadyCh, logger)
+		return setupAutoCertProvisioning(ctx, mgr, cl, certDir, secretName, authorizerEnabled, certsReadyCh, logger)
 
 	default:
 		return fmt.Errorf("unsupported cert provision mode: %q", certProvisionMode)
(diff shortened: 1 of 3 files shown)


shayasoolin (approved):

operator/internal/controller/cert/cert.go
```diff
@@ -123,6 +136,62 @@ func getWebhooks(authorizerEnabled bool) []cert.WebhookInfo {
...
+	// before the manager's informer cache has been started.
+	cl, err := client.New(mgr.GetConfig(), client.Options{Scheme: mgr.GetScheme()})
+	if err != nil {
+		return fmt.Errorf("creating client: %w", err)
+	}
+	return ensureSecretExistsWithClient(cl, namespace, secretName)
+}
+
+// ensureSecretExistsWithClient is the testable core of ensureSecretExists.
+func ensureSecretExistsWithClient(cl client.Client, namespace, secretName string) error {
```
Ronkahn21: can we use only this function

Ronkahn21: I dont like creating clients in mid functions

Ronkahn21: I think we should have better name CreatePlaceHolderSecretIfNotExists or something better ask Claude hahaa

operator/internal/controller/cert/cert.go
```diff
@@ -123,6 +136,62 @@ func getWebhooks(authorizerEnabled bool) []cert.WebhookInfo {
...
+	secret := &corev1.Secret{}
+	err := cl.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: secretName}, secret)
```
Ronkahn21: pass Ctx as arg

PR #458, opened 2026-02-27
Record Delete in CreateExpectation

### LINKED ISSUE #457: Reconcile slow cause deleted pod never recreate
### What happened?

There is one scenario which can cause pod leak.

pclq has 3 replicas (pod1, pod2, pod3)
1. T0: While pod1 is pending may due to resource limit or other issue, use kubectl delete pod manually. Pod1 is deleted successfully.
2. T1: Informer cache updated (pod2, pod3)
3. T2: PCLQ controller reconcile.
`diff := len(sc.existingPCLQPods) + len(createExpectations) - int(sc.pclq.Spec.Replicas) - len(deleteExpectations) // diff = 2 + 1 - 3 - 0 = 0`

In this case, pod1 will never be recreate. The reason here is that reconcile is slow in some condition and informer cache update before reconcile. `SyncExpectations` works on the assumption that some pods is in terminating or let's say pod delete is slow than Reconcile. We couldn't tell these two scenarios:

1. informer can't see the created pod 
2. Pod is already deleted

Pod1 should be recreated.

### Environment

- Kubernetes version
- Grove version
- Scheduler details

kangclzjc:
#### What type of PR is this?
/kind bug

We couldn't tell these two scenarios:

1. informer can't see the created pod 
2. Pod is already deleted manually by kubectl delete when pod is pending

So in some case, deleted pod(manually by kubectl delete) can't be re-created again.

Fixes #457 

```release-note
Fix: Clears unfulfilled create expectations upon observing a delete event.
```

```docs

```

### DIFF
--- a/operator/internal/controller/podclique/register.go
+++ b/operator/internal/controller/podclique/register.go
@@ -99,17 +99,6 @@ func managedPodCliquePredicate() predicate.Predicate {
 	}
 }
 
-// recordPodDeletionInExpectations records the pod's deletion in the expectations store for its owning PodClique so the controller can recreate the pod (issue #457).
-func (r *Reconciler) recordPodDeletionInExpectations(pod *corev1.Pod) {
-	if ownerRef := k8sutils.FindOwnerRefByKind(pod.OwnerReferences, constants.KindPodClique); ownerRef != nil {
-		pclqObjMeta := metav1.ObjectMeta{Namespace: pod.Namespace, Name: ownerRef.Name}
-		if controlleeKey, err := expect.ControlleeKeyFunc(&grovecorev1alpha1.PodClique{ObjectMeta: pclqObjMeta}); err == nil {
-			logger := ctrllogger.Log.WithName(controllerName).WithName("pod-predicate")
-			r.expectationsStore.ObserveDeletions(logger, controlleeKey, pod.UID)
-		}
-	}
-}
-
 // podPredicate returns a predicate that filters out pods that are not managed by Grove.
 // On Delete for a managed pod it calls ObserveDeletions so the controller can recreate the pod (issue #457).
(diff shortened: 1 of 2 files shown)


operator/internal/controller/podclique/register.go
```diff
@@ -79,6 +87,45 @@ func (r *Reconciler) RegisterWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+// podWatchHandler runs actions in response to Pod watch events (e.g. updating expectations on Delete) and delegates to an inner handler.
+// On Delete for a managed pod it calls ObserveDeletions so the controller can recreate the pod (issue #457).
+type podWatchHandler struct {
```
unmarshall: IMO this type is not really required since you only need to handle the delete. For the rest of the verbs (Create/Update/Generic) you are anyways delegating only.

There is already a `podPredicate` function. You can make it a method on `Reconciler` and get access to the expectations store from there.

operator/internal/controller/podclique/register.go
```diff
@@ -60,7 +64,11 @@ func (r *Reconciler) RegisterWithManager(mgr ctrl.Manager) error {
 				),
 			),
 		).
-		Owns(&corev1.Pod{}, builder.WithPredicates(podPredicate())).
+		Watches(
```
unmarshall: there is no need to change this, you can continue to use `Owns`

operator/internal/controller/podclique/register.go
kangclzjc: I list here just in case we may handle other events in the future. But currently it only handle the delete, let me make it a method on Reconciler.

operator/internal/controller/podclique/register.go
kangclzjc: Yeah, if we use predict, then we keep this

PR #464, opened 2026-03-02
feat: add configurable pprof bind address

### LINKED ISSUE #402: GREP: E2E Scale Test Infrastructure

Add comprehensive scale testing infrastructure for Grove operator including:
- **KWOK/Kubemark cluster setup**: Automated deployment of simulated node clusters (1000+ nodes)
- **Observability integration**: Prometheus metrics collection, pprof profiling, optional Pyroscope continuous profiling
- **Performance baseline tracking**: Store and compare metrics across versions with regression detection
- **CI integration**: Automated nightly scale tests with benchstat comparison

### Why is this needed?

Grove orchestrates complex multi-node workloads with multi-layer gang scheduling (PodGangSet → ScalingGroup → PodClique → Pods). To ensure reliable operation at scale:

1. **Validate performance characteristics**: Understand operator behavior with 100-10k+ pods
2. **Detect regressions**: Automatically track performance metrics (reconciliation latency, memory usage, throughput) across git commits
3. **Identify bottlenecks**: Profile operator under load to optimize critical code paths (cascade sync, gang coordination, topology evaluation)
4. **Build confidence**: Demonstrate Grove handles production workloads at scale before customers deploy

Ronkahn21:
#### What type of PR is this?

/kind feature

Adds a configurable `pprofBindAddress` to the operator config and Helm chart, enabling pprof profiling for scale testing. Also adds a generic `annotations` field to the Helm chart pod template to support auto-scrape by profiling systems, and a pprof service port with a Helm helper for port extraction.

Fixes #402

This is PR 1 of 3 for the scale testing infrastructure. PR 2 (#465) adds the infra-manager CLI with Pyroscope annotation injection, and a follow-up PR will migrate the e2e workflow to use the new CLI.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/charts/templates/deployment.yaml
+++ b/operator/charts/templates/deployment.yaml
@@ -38,18 +38,6 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - --config=/etc/grove-operator/config/config.yaml
-          ports:
-            - name: metrics
-              containerPort: {{ required ".Values.config.server.metrics.port" .Values.config.server.metrics.port }}
-              protocol: TCP
-            - name: webhooks
-              containerPort: {{ required ".Values.config.server.webhooks.port" .Values.config.server.webhooks.port }}
-              protocol: TCP
-            {{- if and .Values.config.debugging .Values.config.debugging.enableProfiling }}
-            - name: pprof
-              containerPort: {{ .Values.config.debugging.pprofBindPort | default 2753 }}
-              protocol: TCP
-            {{- end }}
           {{- if .Values.config.server.healthProbes.enable }}
           livenessProbe:
             httpGet:
--- a/operator/charts/templates/service.yaml
+++ b/operator/charts/templates/service.yaml
@@ -21,6 +21,6 @@ spec:
     {{- if and .Values.config.debugging .Values.config.debugging.enableProfiling }}
     - name: pprof
       protocol: TCP
-      port: {{ .Values.config.debugging.pprofBindPort | default 2753 }}
-      targetPort: {{ .Values.config.debugging.pprofBindPort | default 2753 }}
+      port: {{ .Values.config.debugging.pprofBindPort }}
+      targetPort: {{ .Values.config.debugging.pprofBindPort }}
     {{- end }}


operator/internal/controller/manager.go
```diff
@@ -113,8 +109,9 @@ func createManagerOptions(operatorCfg *configv1alpha1.OperatorConfiguration) ctr
 	}
 	if operatorCfg.Debugging != nil {
 		if operatorCfg.Debugging.EnableProfiling != nil &&
-			*operatorCfg.Debugging.EnableProfiling {
-			opts.PprofBindAddress = pprofBindAddress
+			*operatorCfg.Debugging.EnableProfiling &&
+			operatorCfg.Debugging.PprofBindAddress != nil {
+			opts.PprofBindAddress = *operatorCfg.Debugging.PprofBindAddress
```
renormalize: Nit: `ptr.Deref` simplifies the code and makes the code more readable.

```suggestion
		if ptr.Deref(operatorCfg.Debugging.EnableProfiling, false) && operatorCfg.Debugging.PprofBindAddress != nil {
			opts.PprofBindAddress = *operatorCfg.Debugging.PprofBindAddress
```

operator/charts/templates/deployment.yaml
```diff
@@ -34,6 +38,19 @@ spec:
...
             - --config=/etc/grove-operator/config/config.yaml
+          ports:
+            - name: metrics
+              containerPort: {{ required ".Values.config.server.metrics.port" .Values.config.server.metrics.port }}
+              protocol: TCP
+            - name: webhooks
+              containerPort: {{ required ".Values.config.server.webhooks.port" .Values.config.server.webhooks.port }}
+              protocol: TCP
+            {{- if and .Values.config.debugging .Values.config.debugging.enableProfiling }}
+            {{- $pprofPort := (split ":" .Values.config.debugging.pprofBindAddress)._1 | default "2753" }}
```
renormalize: I am not sure if this logic will work if an IPv6 address is passed as a parameter in the values. I see two ways we can go about fixing this:

- Make the helm template logic more sophisticated
- Split bind address into two fields: `pprofBindHost` and `pprofBindPort`

I prefer the second way. What do you think?

operator/internal/controller/manager.go
```diff
@@ -112,9 +108,11 @@ func createManagerOptions(operatorCfg *configv1alpha1.OperatorConfiguration) ctr
...
 	}
 	if operatorCfg.Debugging != nil {
-		if operatorCfg.Debugging.EnableProfiling != nil &&
-			*operatorCfg.Debugging.EnableProfiling {
-			opts.PprofBindAddress = pprofBindAddress
+		if ptr.Deref(operatorCfg.Debugging.EnableProfiling, false) && operatorCfg.Debugging.PprofBindHost != nil {
+			opts.PprofBindAddress = net.JoinHostPort(
+				*operatorCfg.Debugging.PprofBindHost,
+				strconv.Itoa(ptr.Deref(operatorCfg.Debugging.PprofBindPort, configv1alpha1.DefaultPprofBindPort)),
+			)
```
renormalize: I just realized that we do not need a `nil` check for `operatorCfg.Debugging.PprofBindHost`. The defaulters will always set this to a valid value. We do not need a `ptr.Deref(operatorCfg.Debugging.PprofBindPort, configv1alpha1.DefaultPprofBindPort)` either. The value in `operatorCfg.Debugging.PprofBindPort` will contain the passed value from the charts, or will contain the defaulted value. We can thus simplify this:


```suggestion
		if ptr.Deref(operatorCfg.Debugging.EnableProfiling, false) {
			opts.PprofBindAddress = net.JoinHostPort(
				*operatorCfg.Debugging.PprofBindHost,
				strconv.Itoa(*operatorCfg.Debugging.PprofBindPort),
			)
```

operator/api/config/v1alpha1/defaults.go
```diff
@@ -27,6 +27,9 @@ const (
 	defaultLeaderElectionResourceLock    = "leases"
 	defaultLeaderElectionResourceName    = "grove-operator-leader-election"
 	defaultWebhookServerTLSServerCertDir = "/etc/grove-operator/webhook-certs"
+	defaultPprofBindHost                 = "127.0.0.1"
+	// DefaultPprofBindPort is the default port for the pprof HTTP server.
+	DefaultPprofBindPort = 2753
```
renormalize: Due to https://github.com/ai-dynamo/grove/pull/464#discussion_r2894163872, we can make the default a private value.


```suggestion
	defaultPprofBindPort = 2753
```

operator/api/config/v1alpha1/defaults.go
```diff
@@ -122,3 +125,13 @@ func SetDefaults_PodCliqueScalingGroupControllerConfiguration(obj *PodCliqueScal
...
 	}
 }
+
+// SetDefaults_DebuggingConfiguration sets defaults for the debugging configuration.
+func SetDefaults_DebuggingConfiguration(obj *DebuggingConfiguration) {
+	if obj.PprofBindHost == nil {
+		obj.PprofBindHost = ptr.To(defaultPprofBindHost)
+	}
+	if obj.PprofBindPort == nil {
+		obj.PprofBindPort = ptr.To(DefaultPprofBindPort)
```
renormalize: https://github.com/ai-dynamo/grove/pull/464#discussion_r2894375992

renormalize (dismissed):
Added a few nits. We can merge this after they're addressed. Thanks.

operator/charts/values.yaml
```diff
@@ -23,6 +23,9 @@ resources:
 # Set of tolerations to apply to the operator deployment.
 tolerations: [ ]
 
+# Annotations to apply to the operator pod template.
+annotations: {}
```
gflarity: Using the bare name `annotations` is ambiguous — it could be interpreted as deployment-level metadata annotations.  Let's be specific:

Suggestion:
```yaml
# Annotations to apply to the operator pod template.
podAnnotations: {}
```

Note, you'll need to update `deployment.yaml` to reference `.Values.podAnnotations` accordingly.

shayasoolin (approved):

operator/charts/templates/deployment.yaml
```diff
@@ -34,6 +38,18 @@ spec:
...
+          ports:
+            - name: metrics
+              containerPort: {{ required ".Values.config.server.metrics.port" .Values.config.server.metrics.port }}
+              protocol: TCP
+            - name: webhooks
+              containerPort: {{ required ".Values.config.server.webhooks.port" .Values.config.server.webhooks.port }}
+              protocol: TCP
+            {{- if and .Values.config.debugging .Values.config.debugging.enableProfiling }}
+            - name: pprof
+              containerPort: {{ .Values.config.debugging.pprofBindPort | default 2753 }}
```
shmuel-runai: Why do we need a default here and in the code? if we need to change the default, we will need to change it in both places.
Suggest to remove it here

operator/api/config/v1alpha1/types.go
```diff
@@ -122,6 +122,15 @@ type DebuggingConfiguration struct {
 	// EnableProfiling enables profiling via host:port/debug/pprof/ endpoints.
 	// +optional
 	EnableProfiling *bool `json:"enableProfiling,omitempty"`
+	// PprofBindHost is the host/IP that the pprof HTTP server binds to.
+	// Defaults to 127.0.0.1 (loopback-only). Set to 0.0.0.0 to allow external
+	// scraping (e.g. Pyroscope). Supports IPv6 addresses (e.g. "::1").
+	// +optional
+	PprofBindHost *string `json:"pprofBindHost,omitempty"`
```
shmuel-runai, line 129: What is the benefit of making them pointers? Can't they be simple members just like in `ClientConnectionConfiguration` and others?

Ronkahn21: The defaults for these fields are non-zero values (`"127.0.0.1"` and `2753`), so using plain value types would make it impossible to distinguish "not set" from "explicitly set to the zero value". With pointers, `nil` unambiguously means "apply the default", while a non-nil pointer always reflects the user's intent — even if they set the value to `""` or `0`. This pattern is consistent with how other optional config fields with non-zero defaults are handled in the codebase (e.g., `ConcurrentSyncs *int`).

PR #466, opened 2026-03-02
[GREP] Enhance Gang Termination with Stuck Terminating Policy

### LINKED ISSUE #401: Enhance PCLQ gang termination

Introduce a new implementation to enhance the gang termination mechanism to handle pods stuck in termination state for long time (this can be configured by users).
Possible solution 1: Forcefully delete pods stuck in termination; Solution 2: Ignore pods stuck in termination.

### Why is this needed?

- In our MNNVL usage scenario, all pods of one pclq need to be allocated to the same rack, such as Rack A.
- During the training process, node failures may occur, causing pods deletion. However, the kubelet on the node might be unable to respond to the deletion action, leaving the pod in a termination state.
- Grove will recreate the pod, and the scheduler will schedule this pod to Rack A. However, due to the faulty node, the nodes on this rack have insufficient resources to allocate the pod.
- Grove's existing gang terminate mechanism (deleting and recreating pclq) cannot resolve the above issue.

kangclzjc:
#### What type of PR is this?
/kind feature

Grove’s gang termination today deletes and recreates PodCliques (and their pods) when MinAvailable is breached for longer than TerminationDelay. In environments where pods are constrained to a topology (e.g. same rack), node or kubelet failures can leave pods stuck in a terminating state: the API server has set `deletionTimestamp` but the kubelet never completes termination. Those pods are excluded from ready/scheduled counts, so MinAvailable is breached and gang termination runs; however, the stuck pods are still present and can block or complicate cleanup and rescheduling. This GREP proposes a configurable enhancement so that pods stuck in termination for longer than a user-configurable duration are either force-deleted (grace period zero) or **orphaned**—left in the cluster for the admin to handle—while Grove treats them as gone for availability and reconciliation, allowing the gang to recover.

Fixes #401 

```release-note
```

```docs

```

### DIFF
--- a/docs/proposals/401-enhance-gang-termination/README.md
+++ b/docs/proposals/401-enhance-gang-termination/README.md
@@ -12,18 +12,15 @@
     - [Story 3: Per–PCSG or per–PCLQ overrides](#story-3-perpcsg-or-perpclq-overrides)
   - [Limitations/Risks &amp; Mitigations](#limitationsrisks--mitigations)
 - [Design Details](#design-details)
-  - [Background: Current gang termination](#background-current-gang-termination)
-  - [Problem: Pods stuck in termination](#problem-pods-stuck-in-termination)
-  - [Solution options](#solution-options)
-    - [Option 1: Force delete pods stuck in termination](#option-1-force-delete-pods-stuck-in-termination)
-    - [Option 2: Orphan pods stuck in termination](#option-2-orphan-pods-stuck-in-termination)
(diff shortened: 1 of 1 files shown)


shayasoolin:
Hi @kangclzjc can you clarify this issue further more please? Gang termination delay is expected to be much longer than a pod's termination grace period (hours vs minutes). In whats cases are you concerned that the gang termination process will be blocked due to terminating pods?

danbar2:
Stale PR. if this issue is still a concern, please rebase and open this PR again.

PR #477, opened 2026-03-08
add finalizer when build resource to reduce one patch call to api server

### LINKED ISSUE #450: Aggregate create/update actions in one Reconcile action

I want to reduce the number of API server calls to alleviate its load pressure.

### Why is this needed?

In some reconcile function, spec and status could be updated in separated calls, we can aggregate these calls as much as we can to reduce the number of API server calls. For example, PCS controller can add finalizer for PCLQ in advance then we don't need PCLQ controller to update to add the finalizer then, this can reduce one call for API server.

kangclzjc:
#### What type of PR is this?
/feature

Add finalizer for pcsg/pclq in advance to reduce one patch and reduce one call to api server. When we have hundreds of thousands of pcsg/pclq, it could reduce much pressure on api server.

Fixes #450 

```release-note

```

```docs

```

### DIFF
--- a/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
+++ b/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
@@ -304,6 +304,8 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
 			fmt.Sprintf("Error setting controller reference for PodClique: %v", client.ObjectKeyFromObject(pclq)),
 		)
 	}
+	// Add finalizer at creation so PCLQ controller does not need a separate PATCH on first reconcile.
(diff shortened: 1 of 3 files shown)


renormalize (approved):
thanks!

kangclzjc:
We can add a test to add data

Ronkahn21 (commented):
LGTM overall but I would want you to add tests for building the objects to make sure we build with finalizer on

kangclzjc:
> LGTM overall but I would want you to add tests for building the objects to make sure we build with finalizer on

Hi @Ronkahn21 , this PR use the same method which utils provide, I think already have the unit test for 

`controllerutil.AddFinalizer(pcsg, apiconstants.FinalizerPodCliqueScalingGroup)`

danbar2 (approved):

shayasoolin (approved):

Issue #417, opened 2026-02-10
Enhancement: MNNVL V2, ComputeDomain Injection Via Annotation


nvrohanv:
## Summary

Introduce a default, annotation-based mechanism for injecting NVIDIA `ComputeDomain` membership into pods created from a `PodClique`. This provides a low-friction and explicit way for users to control MNNVL participation without requiring verbose `ComputeDomain` specifications, while maintaining clear precedence and error semantics when interacting with existing automatic MNNVL setup mechanisms in Grove.

---

## Background

Currently the **first implementation of automatic MNNVL setup** in Grove is underway and nearing completion.

This implementation:
- Assumes **NVIDIA DRA drivers are installed and correctly configured on all nodes**
- Requires **explicit opt-in by the cluster administrator via operator configuration**
- Preserves the familiar **“it just works” NVLink experience**, where:
  - NVLink automatically exists between GPUs on NVLink-enabled nodes
  - Users do not need to manually define domains or topology for same-node NVLink

This cluster-level automatic behavior is necessary and should remain supported.

However, this model alone is insufficient for a few scenarios:
- Users may want **fine-grained control over MNNVL participation**, where only a subset of pods (or PodCliques) within a `PodCliqueSet` replica should participate in a shared MNNVL domain
- Grove needs a way to provide **some level of automatic ComputeDomain injection in default configurations**, without requiring fully automatic cluster-wide MNNVL enablement
- Heterogeneous clusters with **mixed hardware from multiple vendors** may not be able to install NVIDIA DRA drivers on all nodes, making cluster-wide automatic MNNVL infeasible while still requiring explicit MNNVL support on NVIDIA-capable subsets

Consequently Grove needs to provide a simple mechanism to:
- Explicitly control which pods participate in a given MNNVL domain
- Avoid authoring verbose, PVC-style `ComputeDomain` resources
- Express intent in a way that composes naturally with Grove primitives and heterogeneous cluster environments

---

## Proposal

Add **annotation-based ComputeDomain injection** at the **PodClique level**, using a vendor-scoped annotation.

### Core Behavior

If a `PodClique` includes the annotation:

```yaml
nvidia.com/computedomain: <name>
```

then pods created from that `PodClique` will have the following behavior applied:
- The pods claim a `ComputeDomain` with the specified name
- The pods participate in the same MNNVL domain for NVLink / interconnect setup
-  If the `ComputeDomain` the pods claim is not created, the Operator will create it

[...]

PR #492, opened 2026-03-17
Add GREP-417: Selective MNNVL proposal

shmuel-runai:
#### What type of PR is this?
/kind documentation

Introduce annotation-based ComputeDomain injection at the PodClique level, enabling per-clique MNNVL control without requiring cluster-wide auto-MNNVL enablement.

Ref #417


docs/proposals/417-selective-mnnvl/README.md
```diff
@@ -0,0 +1,397 @@
...
+- **All pods in a replica share one ComputeDomain:** regardless of whether they actually need to communicate with each other over NVLink. This may lead to unpredictable and undesired side effects — such as unnecessary scheduling constraints and wasted IMEX channels — that can be avoided when only the pods that need NVLink are enrolled.
+- **When auto-MNNVL is not enabled:** there is significant overhead for users to manually set up MNNVL for a `PodCliqueSet`. They must author and manage `ComputeDomain` resources and wire up the references in the pod spec themselves.
+
+Selective MNNVL addresses these limitations by shifting the unit of MNNVL control from the entire PCS replica down to the individual PodClique, giving users an explicit, low-friction, and composable way to declare MNNVL intent.
+
+### Goals
+
+- Provide a **PodClique-level annotation** (`grove.io/selective-mnnvl`) that enables MNNVL participation on a per-PodClique basis.
```
shayasoolin: I think the feature and annotation should use "mnnvl", not "selective-mnnvl". The fact that we made some historical mistake with that global auto-mnnvl, and that we're now fixing it (quite early before anyone really used it), doesn't mean we should pay the price of calling it "selective" from now on... Let it just be "mnnvl", let's see how to do it without conflicting with the old version.

docs/proposals/417-selective-mnnvl/README.md
```diff
@@ -0,0 +1,397 @@
...
+- RCT injection to pods, only for the the annotated PodCliques.
+
+A PCS that uses selective MNNVL annotations **will not** have auto-MNNVL applied to it, even if auto-MNNVL is enabled at the cluster level. The presence of `grove.io/selective-mnnvl` annotations on any PodClique within a PCS acts as an explicit signal that the user is taking control of ComputeDomain assignment. The two mechanisms are mutually exclusive per PCS — selective MNNVL annotations take precedence and auto-MNNVL is suppressed for that workload. See [Interaction with Auto-MNNVL](#interaction-with-auto-mnnvl) for the full precedence rules.
+
+### User Stories
+
+#### Story 1: Selective MNNVL in a Heterogeneous Cluster
+
+As a platform engineer running a heterogeneous cluster with both MNNVL-supported and non-MNNVL-supported nodes, I cannot enable auto-MNNVL cluster-wide because the NVIDIA IMEXfd driver is only installed on a subset of nodes. I want to annotate specific PodCliques in my PodCliqueSet with `grove.io/selective-mnnvl` so that only those pods participate in an MNNVL domain, while other PodCliques (running on non-MNNVL-supported hardware) are unaffected.
```
shayasoolin: typo IMEXfd -> IMEX

docs/proposals/417-selective-mnnvl/README.md
```diff
@@ -0,0 +1,397 @@
...
+
+A PodClique opts into MNNVL by carrying the following annotation:
+
+```yaml
+grove.io/selective-mnnvl: <name>
+```
+
+Where `<name>` is a user-chosen identifier for the ComputeDomain. Multiple PodCliques within the same PCS replica may reference the **same** name to share a single ComputeDomain, or use **different** names to participate in separate domains.
+
+**Value validation:** The annotation value becomes part of the ComputeDomain resource name (`{value}-{replica-index}`), so it must be a valid Kubernetes resource name component. The validating webhook must reject any PCS where the annotation value:
```
shayasoolin: By <value> you mean <name>? It's not clear. If it's <name>, i.e. the chosen CD id, and you don't add the PCS name to it, it might conflict with other PCSs' CDs. I think you should make it unique be prepending the PCS name.

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+                  limits:
+                    cpu: "4"
+```
+
+Result per replica:
+- `workers` → enrolled in group 1 → CD `training-job-group-1-0` / `training-job-group-1-1`
+- `encoders` → enrolled in group 2 → CD `training-job-group-2-0` / `training-job-group-2-1`
+- `param-servers` → no annotation → no MNNVL
+
+### Multi-Layer Resolution
```
shayasoolin: I find this whole section redundant, the topic is clear enough, I think it can be removed.

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+- [Open Questions](#open-questions)
+- [Alternatives (*Optional*)](#alternatives-optional)
+- [Appendix (*Optional*)](#appendix-optional)
+<!-- /toc -->
+
+## Summary
+
+Grove's auto-MNNVL feature allows users to leverage NVIDIA Multi-Node NVLink acceleration through a simple annotation — without manually authoring `ComputeDomain` resources or wiring `resourceClaims` in their pod specs. The operator automatically creates and manages ComputeDomains, injects RCT references, and handles scaling as PCS replicas scale up or down.
+
+The feature is controlled by a single annotation, `grove.io/mnnvl-group`, which can be placed on a **PodCliqueSet**, **PodCliqueScalingGroup**, or **PodClique**. The annotation value is either a positive integer identifying the MNNVL group (ComputeDomain), or `"none"` to explicitly opt out. Annotations propagate downward from PCS → PCSG → PCLQ, with lower layers overriding higher ones.
```
shayasoolin: I'd like to suggest that we split this into two different annotations:
1. The most useful one - 99% of the use cases that use either 1 or no CD - will use `grove.io/auto-mnnvl` (or some similar generic name) with optional values of "enabled/disabled" or something similar.
2. The rare cases in which more than one CD is needed, will use `grove.io/mnnvl-group` in addition to the basic annotation (which can be inherited from the PCS, just as you designed.)

My concern is that otherwise, the majority of cases will be required to use an API that is error-prone in the sense that they need to come up with some group number that can be invalid (e.g. "0", "A") or can be _mistakenly_ (when not desired) set to different values for different PCLQs.

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+| Positive integer (`"1"`, `"2"`, ...) | Opt-in to MNNVL group N. PodCliques with the same group number share a ComputeDomain per replica. |
+| `"none"` | Explicit opt-out. Used to override a parent layer's annotation. |
+| Absent | Inherit from parent layer. If no parent has it, no MNNVL. |
+
+**Value validation:**
+- The value must be either `"none"` or a string that parses as a positive integer >= 1.
+- `"0"`, negative numbers, empty strings, and other values are rejected by the validating webhook.
+- The `"none"` comparison is **case-insensitive** (e.g., `"None"`, `"NONE"` are accepted). The canonical form is lowercase `"none"`.
+
+**GPU validation:** When a PCLQ resolves to a group number (via its own annotation or inheritance), but does not request GPUs (`nvidia.com/gpu`) in any container, it is **silently ignored** — no RCT injection, no error. This allows PCS-level annotations without requiring `"none"` overrides on every non-GPU PCLQ.
```
shayasoolin: IIUC GPUs can also be requested using DRA, so identifying a PCLQ's GPU usage is a bit more complicated than just `nvidia.com/gpu`.
Also, not sure if this feature is designed purely for NVIDIA GPUs but otherwise this identification logic should be more general.

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+
+When the feature is disabled and a user submits a PCS with `grove.io/mnnvl-group` annotations, the current proposal rejects the PCS. An alternative is to silently ignore the annotations. The benefit is **portability** — the same manifest works everywhere. The risk is **silent performance degradation** — the user requested MNNVL but the workload runs without it.
+
+## Alternatives (*Optional*)
+
+**Alternative 1: Use string names instead of numeric group values.**
+The group annotation could take a string name (e.g., `grove.io/mnnvl-group: "training-domain"`) instead of a number. This is more expressive but adds validation complexity (valid Kubernetes name component checks, length limits). Numeric values are simpler.
+
+**Alternative 2: Two separate annotations (`auto-mnnvl` + `mnnvl-group`).**
+Using two annotations — one for MNNVL on/off and one for group assignment — was considered. This adds the risk of inconsistent state (e.g., `auto-mnnvl: "disabled"` + `mnnvl-group: "1"`). A single annotation with `"none"` as the opt-out value eliminates this class of bugs.
```
shayasoolin: The presented risk is insignificant and can be easily mitigated using a validation webhook. The advantage is a simpler API for 99% of the cases.

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+```
+
+The operator discovers these annotations, determines the set of ComputeDomains required per replica, and:
+- manages their full lifecycle — creation, deletion, scaling, and protection via finalizers.
+- injects RCT references into the pod's spec of enrolled PodCliques.
+
+**Key behavioral rules:**
+- **Opt-in:** No MNNVL is applied unless the annotation is present. Absence of the annotation at all layers means no MNNVL.
+- **Propagation:** Annotations propagate downward (PCS → PCSG → PCLQ). A lower layer overrides a higher layer.
+- **Non-GPU PCLQs:** If a non-GPU PodClique resolves to a group number (via inheritance), it is silently ignored — no RCT injection, no error.
```
danbar2: what if it is defined on it explicitly ?

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+    UsePCS --> CheckVal
+
+    CheckVal -->|"number >= 1"| Enrolled["Enrolled in group N"]
+    CheckVal -->|"none"| NotEnrolled["Not enrolled"]
+```
+
+**How resolution works in practice:** The flowchart above shows the logical resolution model. In implementation, the resolution is **not** performed by the PCLQ querying its parents at runtime. Instead, it follows the existing PCS reconciliation pattern:
+
+- **PCS → PCSG:** When the PCS controller creates a PCSG, it propagates the `grove.io/mnnvl-group` annotation from the PCS to the PCSG (if present and the PCSG doesn't already have its own value defined in the PCS template).
+- **PCS → PCLQ:** When the PCS controller creates a PCLQ, it resolves the effective group value (PCS annotation vs. PCLQ-level annotation in the template) and injects the RCT reference into the PCLQ's pod spec at creation time.
```
danbar2: there might be a bug in this approach.
If annotation is set on PCS to 1 and on PCSG to 2.
the first line does nothing
the second one copies the PCS value of 1 to the PCLQ
the third line does nothing as the PCLQ has a value already.
so you end up with PCLQ having value of 1 from the PCS instead of 2 from the PCSG

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+In the opt-in model, the mutating webhook **does not** add any MNNVL-related annotations automatically. The user is responsible for providing the `grove.io/mnnvl-group` annotation.
+
+The only mutation the webhook performs for MNNVL is **deprecation handling** — see [Backward Compatibility](#backward-compatibility).
+
+#### Validating Webhook (on Create)
+
+The validating webhook enforces:
+
+- **Value validation:** `grove.io/mnnvl-group` must be either `"none"` (case-insensitive) or a positive integer >= 1. Invalid values are rejected.
+- **Feature disabled:** If the feature is off (config `disabled`, or config `auto` with no CRD), reject any PCS that has `grove.io/mnnvl-group` at any level.
```
danbar2: does it mean you need to list CRD's on every submit? might be better to store the result of the first 'auto' validation

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,690 @@
...
+| 2 | Fresh install | `mnnvl.mode: enabled` | N/A | `mnnvl.mode: enabled` | ON, fail if no CRD | Explicit |
+| 3 | Fresh install | `mnnvl.mode: disabled` | N/A | `mnnvl.mode: disabled` | OFF | Explicit |
+| 4 | Fresh install | `autoMNNVLEnabled: true` | N/A | `mnnvl.mode: enabled` | ON, fail if no CRD | Old field still works |
+| 5 | Upgrade | nothing (used old default) | OFF | `mnnvl.mode: auto` | ON if CRD present | Behavior change |
+| 6 | Upgrade | `autoMNNVLEnabled: true` | ON, fail if no CRD | `mnnvl.mode: enabled` | ON, fail if no CRD | No change |
+| 7 | Upgrade | `autoMNNVLEnabled: false` | OFF | `mnnvl.mode: disabled` | OFF | No change |
+| 8 | Upgrade | both old + new set | depends | `mnnvl.mode` wins | depends | New field takes precedence |
+
+The Helm template would handle this by checking the new field first, falling back to the old one. The operator would log a warning if the old field is detected. This approach is more user-friendly but adds template complexity and risks silent behavioral changes (scenario 5). The current proposal favors an explicit, clean break to avoid ambiguity.
```
danbar2: I think we need to reconsider that, as breaking API is dangerous

docs/proposals/417-auto-mnnvl/README.md
shmuel-runai: I considered this approach before and discussed it in the Alternatives section (Alternative 2). My concern is that two annotations create a surface for an inconsistent state — e.g., a user sets auto-mnnvl: "disabled" on the PCS but adds mnnvl-group: "1" on a PCLQ. While a validating webhook can catch this, it adds validation complexity that a single annotation avoids entirely.

For the 99% case (one CD), the user simply sets grove.io/mnnvl-group: "1" — one annotation, one value. The numeric value being "1" every time isn't really error-prone since the webhook rejects invalid values. The risk of accidentally assigning different numbers to PCLQs is equivalent to accidentally setting one to "enabled" and another to "disabled" with the two-annotation approach.

Let's hash this in the DR meeting

docs/proposals/417-auto-mnnvl/README.md
shmuel-runai: Noted. Let's discuss the migration approach in the DR meeting.

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,701 @@
...
+| `enabled` | `"training"` | MNNVL with group CD (`{pcs}-{replica}-training`) |
+| `disabled` | `"training"` | **Reject** — contradictory (disabled + group assignment) |
+
+**GPU validation:** The behavior when a non-GPU PCLQ resolves to MNNVL enrollment depends on where the annotation comes from:
+- **Explicit annotation on the PCLQ:** If a non-GPU PCLQ carries its own `grove.io/mnnvl-group` or `grove.io/auto-mnnvl: "enabled"` annotation, the validating webhook **rejects** the PCS. Explicitly requesting MNNVL for a PodClique that has no GPUs is a user error.
+- **Inherited from PCS or PCSG:** If a non-GPU PCLQ inherits MNNVL enrollment from a parent layer (PCS or PCSG), it is **silently skipped** — no RCT injection, no error. This allows PCS-level defaults without requiring `disabled` overrides on every non-GPU PCLQ.
+
+**Immutability:** Both `grove.io/auto-mnnvl` and `grove.io/mnnvl-group` are **immutable** after PCS creation. The validating webhook rejects any update that attempts to add, modify, or remove either annotation at any level (PCS, PCSG, PCLQ). To change MNNVL assignment, the PCS must be deleted and recreated.
```
gflarity: This states the immutability constraint but not the rationale. It's fairly inuitive that would involve a lot of complexity for little benefit,  but let's add  a brief descriptive "why" to document it concretely.

docs/proposals/417-auto-mnnvl/README.md
```diff
@@ -0,0 +1,701 @@
...
+| `training-job` | `auto-mnnvl: enabled` (no group) | 1 | `training-job-1` |
+| `training-job` | `mnnvl-group: "workers"` | 0 | `training-job-0-workers` |
+| `training-job` | `mnnvl-group: "workers"` | 1 | `training-job-1-workers` |
+| `training-job` | `mnnvl-group: "encoders"` | 0 | `training-job-0-encoders` |
+
+Group names are scoped to the PCS — different PCS resources can use the same group names without conflict (the PCS name provides uniqueness).
+
+#### Reconciliation Ordering
+
+ComputeDomain resources must be synced **before** creating PCLQs and PCSGs, ensuring the CD exists before pods that reference its RCT are created. If CD creation fails, the sync stops and requeues for retry.
```
gflarity, line 389: This specifies creation ordering (CDs before PCLQs/PCSGs), which is good. Are there  similar but inverse requirements around tear down? ie  pods most be removed  removed before the CD?

shayasoolin (dismissed):

danbar2 (dismissed):

danbar2 (approved):

shayasoolin (approved):

PR #496, opened 2026-03-23
multi topology implementation

### LINKED ISSUE #369: Support multiple topologies

enoodle:
#### What type of PR is this?

/kind feature
/kind api

Implementation of Multiple Topology GREP to enhance TAS support (#413 )

Fixes #369 

ClusterTopology API extended with multi-topology support:

***ClusterTopologySpec:***
- `spec.levels[].domain` — removed fixed enum constraint (was `region|zone|datacenter|block|rack|host|numa`); domains are now free-form strings validated by pattern `^[a-z][a-z0-9-]*$` with max 63 characters. This enables heterogeneous GPU cluster topologies where hardware segments define their own domain names.
- `spec.schedulerReferences` (new, optional) — list of `{schedulerName, reference}` entries. Controls per-backend topology resource lifecycle: absent = operator auto-manages the backend topology resource; present = resource is externally managed and the operator performs drift detection only.

***ClusterTopologyStatus (new):***
- `status.observedGeneration` — generation last reconciled by the controller.
- `status.conditions` — standard condition list; includes `SchedulerTopologyDrift` condition (`True/Drift` when any backend is out of sync, `False/InSync` when all match).
- `status.schedulerTopologyStatuses` — per-backend sync state reporting `{schedulerName, reference, inSync, message, schedulerBackendTopologyObservedGeneration}`.

***PodCliqueSet API:***
- `spec.template.topologyConstraint` type changed from `TopologyConstraint` to `PodCliqueSetTopologyConstraint`, adding `topologyName` field — the name of the `ClusterTopology` resource to use. Required when `packDomain` is specified. Immutable after creation.
- `spec.template.topologyConstraint.packDomain` enum constraint removed; must now reference a domain defined in the named `ClusterTopology`'s levels (validated by webhook).

***Removed:***
- `ClusterTopology.spec.levels` immutability constraint and `MaxItems=7` cap.
- `DefaultClusterTopologyName` (`"grove-topology"`) constant — the operator no longer manages a single implicit topology; all topologies are now admin-created `ClusterTopology` resources.
- `TopologyDomain.IsTopologyDomainNarrower`, `SupportedTopologyDomains`, `SortTopologyLevels` helper functions — ordering was predicated on the fixed enum which no longer exists.

***Conditions added (constants):***
- `SchedulerTopologyDrift` / `InSync` / `Drift` — on `ClusterTopology`
- `TopologyNameMissing` / `TopologyAwareSchedulingDisabled` — on `PodCliqueSet`

```release-note
> [!WARNING]
> **Upgrade Notice — Topology-Aware Scheduling**
>
> Affects users who have deployed `PodCliqueSet` resources with a `topologyConstraint` using the previous single-topology design.

### What changed

The operator now supports multiple named `ClusterTopology` resources. Each `PodCliqueSet.spec.template.topologyConstraint` now requires a `topologyName` field that explicitly names the `ClusterTopology` to use. Without it, topology placement constraints are not enforced.

### Action required

Add `topologyName: grove-topology` to each existing `PodCliqueSet` that has a `topologyConstraint`:

spec:
  template:
    topologyConstraint:
      topologyName: grove-topology   # add this
      packDomain: rack

The existing `grove-topology` ClusterTopology resource created by the previous operator version remains in the cluster and is valid — no changes to it are needed.

### What happens if you don't act immediately

The operator degrades gracefully but **topology placement constraints will not be enforced** until `topologyName` is set. This is visible via the `TopologyLevelsUnavailable` condition on the PCS:

| Condition status | Reason | Meaning |
|---|---|---|
| `Unknown` | `TopologyNameMissing` | PCS has `topologyConstraint` but no `topologyName`. Constraints are not enforced. Add `topologyName` to restore them. |
| `True` | `ClusterTopologyNotFound` | The named `ClusterTopology` does not exist. Create it or correct the name. |
| `False` | `AllClusterTopologyLevelsAvailable` | All good — constraints are being enforced. |

Check for affected PodCliqueSets after upgrade:

kubectl get podcliquesets -A -o json | jq -r '
  .items[] | select(.status.conditions != null) |
  .status.conditions[] | select(.type == "TopologyLevelsUnavailable" and .status != "False") |
  "\(.type): \(.reason) — \(.message)"'


```

TBA
```docs

```

---

<a href="https://nvidia.devinenterprise.com/review/ai-dynamo/grove/pull/496" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>


operator/e2e/grove/topology/topology.go
```diff
@@ -22,15 +22,20 @@ import (
...
 	"github.com/ai-dynamo/grove/operator/e2e/k8s"
 	"github.com/ai-dynamo/grove/operator/e2e/k8s/clients"
 	"github.com/ai-dynamo/grove/operator/e2e/log"
 	kaitopologyv1alpha1 "github.com/kai-scheduler/KAI-scheduler/pkg/apis/kai/v1alpha1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
```
danbar2: please use the new waitFor package

operator/internal/clustertopology/clustertopology.go
```diff
@@ -19,43 +19,54 @@ package clustertopology
...
+		ct := &ctList.Items[i]
+		for _, b := range backends {
+			tasBackend, ok := b.(scheduler.TopologyAwareSchedBackend)
+			if !ok {
+				logger.V(1).Info("Scheduler backend does not implement TopologyAwareSchedBackend, skipping topology sync", "backend", b.Name())
+				continue
+			}
+			// Only sync auto-managed backends (not listed in schedulerReferences).
+			// Externally-managed backends are handled by the CT controller via CheckTopologyDrift.
+			if hasSchedulerReference(ct.Spec.SchedulerReferences, b.Name()) {
```
danbar2: consider creating a map here for ct.Spec.SchedulerReferences with the name as the key.
We have here 3 nested for loops, at least we can reduce one

operator/internal/controller/clustertopology/register.go
```diff
@@ -0,0 +1,104 @@
...
+		Client:   mgr.GetClient(),
+		backends: schedmanager.All(),
+	}
+}
+
+// RegisterWithManager registers the ClusterTopology controller with the manager.
+func (r *Reconciler) RegisterWithManager(mgr ctrl.Manager) error {
+	b := ctrl.NewControllerManagedBy(mgr).
+		Named("clustertopology").
+		For(&grovecorev1alpha1.ClusterTopology{}, builder.WithPredicates(predicate.GenerationChangedPredicate{}))
```
danbar2, line 49: do we need to watch also supported scheduler backends topology resources?

danbar2: managed and referenced are relevant

operator/internal/controller/podcliqueset/reconcilestatus.go
```diff
@@ -176,6 +176,24 @@ func (r *Reconciler) computePCSGsStatus(pcsGenerationHash *string, expectedPCSGs
 
 func (r *Reconciler) mutateTopologyLevelUnavailableConditions(ctx context.Context, logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet) error {
 	if !r.tasConfig.Enabled {
+		// If TAS is disabled but PCS has topology constraints, surface a warning condition.
+		if len(getUniqueTopologyDomainsInPodCliqueSet(pcs)) > 0 {
+			cond := metav1.Condition{
+				Type:               apicommonconstants.ConditionTopologyLevelsUnavailable,
+				Status:             metav1.ConditionUnknown,
+				Reason:             apicommonconstants.ConditionReasonTopologyAwareSchedulingDisabled,
```
danbar2, line 197: did we consider failing in this case? we chose to fail in MNNVL GREP in case admin disabled hte feature and the user asks for it in the workload spec

operator/internal/webhook/admission/clustertopology/validation/validation.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// validateClusterTopologyLevels validates that domain and key values are unique across levels.
+func validateClusterTopologyLevels(levels []grovecorev1alpha1.TopologyLevel, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+
+	seenDomains := make(map[grovecorev1alpha1.TopologyDomain]struct{})
+	seenKeys := make(map[string]struct{})
```
danbar2: NIT: map string to bool, more intuitive

operator/internal/clustertopology/clustertopology.go
enoodle: :+1:

operator/internal/controller/clustertopology/register.go
enoodle: This exactly happens in the next few lines after this

operator/internal/controller/podcliqueset/reconcilestatus.go
enoodle: This is to mark a job that already exists if suddenly the admin disabled the TAS feature. It is rejected by the validation webhook anyway. (https://github.com/ai-dynamo/grove/blob/main/operator/internal/webhook/admission/pcs/validation/topologyconstraints.go#L82)

operator/e2e/grove/topology/topology.go
enoodle: I will update to use them

operator/internal/webhook/admission/clustertopology/validation/validation.go
enoodle: :+1:

operator/e2e/grove/topology/topology.go
enoodle: I updated to use PollForCondition functions like the other WaitFor functions

enoodle: When https://github.com/ai-dynamo/grove/pull/526 is merged I can rebase on that

operator/internal/controller/clustertopology/register.go
```diff
@@ -0,0 +1,104 @@
...
+			// CRD not installed — skip watch; the controller still reconciles on CT spec changes.
+			log.Log.Info("Skipping watch for backend topology CRD (not installed)", "gvr", gvr, "backend", backend.Name())
+			continue
+		}
+		u := &unstructured.Unstructured{}
+		u.SetGroupVersionKind(gvk)
+		b = b.Watches(u, handler.EnqueueRequestsFromMapFunc(r.mapBackendTopologyToCT(backend.Name())))
+	}
+
+	return b.Complete(r)
```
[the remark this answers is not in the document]
enoodle: Currently there is no use case to look at the status of the topology CR of the scheduling backend.

operator/internal/controller/podcliqueset/reconcilestatus.go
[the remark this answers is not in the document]
enoodle: Good catch . I created unit tests and updated one e2e test to reproduce this issue and fixed it.

danbar2 (dismissed):

operator/e2e/grove/topology/topology.go
enoodle: I rebased and updated to use those new functions

operator/internal/controller/clustertopology/reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		ref := schedulerRefMap[b.Name()]
+		if ref == nil {
+			// Auto-managed: create/update via SyncTopology
+			if err := tasBackend.SyncTopology(ctx, nil, ct); err != nil {
+				reconcileErr = errors.Join(reconcileErr, err)
+				statuses = append(statuses, grovecorev1alpha1.SchedulerTopologyStatus{
+					SchedulerName: b.Name(),
+					Reference:     ct.Name,
```
gflarity: Reference in SchedulerTopologyStatus is supposed to be the backend topology resource name, but the auto-managed path hardcodes the Grove Cluter Topology name here. That matches KAI today, but it is not part of TopologyAwareSchedBackend’s contract. Consider exposing the backend resource name via the interface:

```go
TopologyResourceName(ct *grovecorev1alpha1.ClusterTopology) string
```

 We could also explicitly documenting/enforcing that auto-managed topology resources must use the ClusterTopology name, but that seems a be prescriptive and the interface addition is simple enough.

operator/internal/controller/clustertopology/reconciler.go
```diff
@@ -0,0 +1,157 @@
...
+				Message: msg,
+			})
+		}
+	}
+
+	if reconcileErr == nil {
+		ct.Status.ObservedGeneration = ct.Generation
+	}
+	ct.Status.SchedulerTopologyStatuses = statuses
+	setSchedulerTopologyDriftCondition(ct, statuses)
```
gflarity: ClusterTopology updates a user-facing drift condition/status, but it doesn’t emit Kubernetes events today. Please also emit some transition-based events: a warning when drift first appears (or a backend sync/drift-check error is first observed), and a normal event when the topology returns to InSync. That would make kubectl describe more useful and also benefit Arborist TUI in the future (improve the debug story for grove).

operator/internal/controller/clustertopology/register.go
```diff
@@ -0,0 +1,104 @@
...
+
+// mapBackendTopologyToCT returns a MapFunc that maps a backend topology resource event
+// to the ClusterTopology that should be reconciled.
+//
+// For auto-managed topologies: the backend topology has an OwnerReference to the CT.
+// For externally-managed topologies: the CT's schedulerReferences names the backend topology.
+func (r *Reconciler) mapBackendTopologyToCT(backendName string) handler.MapFunc {
+	return func(ctx context.Context, obj client.Object) []ctrl.Request {
+		for _, ref := range obj.GetOwnerReferences() {
+			if ref.Kind == "ClusterTopology" && ref.APIVersion == grovecorev1alpha1.SchemeGroupVersion.String() {
```
gflarity: `"ClusterTopology"` literal should be replaced with a constant from `apicommonconstants` (analogous to `KindPodCliqueSet`, `KindPodClique`, `KindPodCliqueScalingGroup`).

operator/client/clientset/versioned/typed/core/v1alpha1/clustertopology.go
```diff
@@ -39,6 +39,8 @@ type ClusterTopologiesGetter interface {
 type ClusterTopologyInterface interface {
 	Create(ctx context.Context, clusterTopology *corev1alpha1.ClusterTopology, opts v1.CreateOptions) (*corev1alpha1.ClusterTopology, error)
 	Update(ctx context.Context, clusterTopology *corev1alpha1.ClusterTopology, opts v1.UpdateOptions) (*corev1alpha1.ClusterTopology, error)
+	// Add a +genclient:noStatus comment above the type to avoid generating UpdateStatus().
```
gflarity, line 42: This comment is emitted by stock client-gen for status-bearing resources. ClusterTopology intentionally has a status subresource, so I don’t think +genclient:noStatus is the right local fix. If we want to scrub these comments from generated code, that should be handled as a separate repo-wide codegen/tooling change.

operator/internal/webhook/admission/pcs/validation/topologyconstraints.go
```diff
@@ -149,9 +157,15 @@ func (v *topologyConstraintsValidator) validateTopologyDomainsExistInClusterTopo
...
-// a child topology constraint domain. Returns true if valid (parent is broader or equal), false if invalid (parent is narrower).
-func isParentBroaderThanChild(parentDomain, childDomain grovecorev1alpha1.TopologyDomain) bool {
-	return !parentDomain.IsTopologyDomainNarrower(childDomain)
+// a child topology constraint domain using the ordered clusterTopologyDomains slice (index 0 = broadest).
+// If either domain is not in the slice, the check is skipped (returns true).
+func isParentBroaderThanChild(parentDomain, childDomain grovecorev1alpha1.TopologyDomain, clusterTopologyDomains []string) bool {
+	parentIdx := slices.Index(clusterTopologyDomains, string(parentDomain))
+	childIdx := slices.Index(clusterTopologyDomains, string(childDomain))
+	if parentIdx == -1 || childIdx == -1 {
+		return true
```
gflarity: isParentBroaderThanChild() has a misleading contract: it returns true both when the parent is broader/equal and when the hierarchy check is skipped because one or both domains are unknown. The comment is a little imprecise too, since this also returns true when neither domain is in the slice. Could we invert this to hasHierarchyViolation() (or similar) so the boolean matches what callers actually care about?

operator/internal/webhook/admission/pcs/validation/topologyconstraints.go
```diff
@@ -233,9 +247,24 @@ func (v *topologyConstraintsValidator) validateHierarchicalTopologyConstraints(f
...
+			fldPath.Child("topologyConstraint").Child("topologyName"),
+			fmt.Sprintf("topologyName cannot be changed from %q to %q", oldTopologyName, newTopologyName)))
+	}
+
 	// Validate PCS level
-	oldPCSConstraint := oldPCS.Spec.Template.TopologyConstraint
-	newPCSConstraint := v.pcs.Spec.Template.TopologyConstraint
+	oldPCSConstraint := oldPCS.Spec.Template.TopologyConstraint.ToTopologyConstraint()
+	newPCSConstraint := v.pcs.Spec.Template.TopologyConstraint.ToTopologyConstraint()
 	if constraintChanged(oldPCSConstraint, newPCSConstraint) {
```
gflarity: Minor: adding or removing the PCS-level `topologyConstraint` currently produces two `Forbidden` errors for a single edit: one from the `topologyName` immutability check above, and another here from `constraintChanged(oldPCSConstraint, newPCSConstraint)`. That second error happens because `ToTopologyConstraint()` drops `topologyName`, so add/remove becomes `nil` <-> non-`nil` and is treated as a PCS-level constraint change. Consider short-circuiting the PCS-level `constraintChanged` check when the `topologyName` check already fired, or merging these into a single clearer message; otherwise `kubectl apply` shows a confusing duplicate error.

gflarity:
Thanks @enoodle. I did a first pass, few small things. AIMing to do a final pass Thurs/Fri.

docs/proposals/244-topology-aware-scheduling/README.md
sanjaychatterjee, line 51:
```suggestion
In heterogeneous clusters, when a single topology definition is insufficient to represent different interconnect hierarchies, Grove allows the admin to define multiple ClusterTopology resources enabling the cluster to be logically partitioned into segments with distinct topology hierarchies. Workloads select the appropriate partition via a topology reference on PodCliqueSet.
```

sanjaychatterjee, line 49:
```suggestion
AI Inference workloads require low-latency data transfer between model layers or shards. Topology-aware placement of such workloads is critical to maximize performance on GPU scale-out clusters. This GREP proposes a unified topology model in Grove using a new CRD called ClusterTopology, and introduces new API for users to define scheduling constraints that will guarantee topology optimized placement of their workloads. A ClusterTopology resource created by an admin or a tool is used to describe the hierarchical network interconnects amongst nodes in the cluster.
```

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -269,16 +269,37 @@ type PodCliqueTemplateSpec struct {
 	Spec PodCliqueSpec `json:"spec"`
 }
 
+// PodCliqueSetTopologyConstraint defines topology placement requirements for PodCliqueSet.
+// Extends the base topology constraint with a reference to a specific ClusterTopology.
+type PodCliqueSetTopologyConstraint struct {
+	// TopologyName is the name of the ClusterTopology resource to use for topology-aware scheduling.
+	// Required when PackDomain is specified. Immutable after creation.
+	// +optional
+	TopologyName string `json:"topologyName,omitempty"`
```
sanjaychatterjee: Why is `TopologyName` added only to PodCliqueSet? Is it not possible that different parts of the workload may refer to different topologies? In that case, maybe `TopologyName` should be part of every `TopologyConstraint`?

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -269,16 +269,37 @@ type PodCliqueTemplateSpec struct {
...
+	// PackDomain specifies the topology domain for grouping replicas.
+	// Must reference a domain defined in the referenced ClusterTopology's levels.
+	// +optional
+	PackDomain TopologyDomain `json:"packDomain,omitempty"`
+}
+
+// ToTopologyConstraint returns the base TopologyConstraint (without the topology name).
+func (tc *PodCliqueSetTopologyConstraint) ToTopologyConstraint() *TopologyConstraint {
```
sanjaychatterjee: Let's move the Topology related functions to a separate file. Maybe rename clustertopology.go to topology.go and move both `ClusterTopology` and `TopologyConstraint` into it.

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -269,16 +269,37 @@ type PodCliqueTemplateSpec struct {
 	Spec PodCliqueSpec `json:"spec"`
 }
 
+// PodCliqueSetTopologyConstraint defines topology placement requirements for PodCliqueSet.
+// Extends the base topology constraint with a reference to a specific ClusterTopology.
+type PodCliqueSetTopologyConstraint struct {
+	// TopologyName is the name of the ClusterTopology resource to use for topology-aware scheduling.
+	// Required when PackDomain is specified. Immutable after creation.
+	// +optional
```
sanjaychatterjee: Is `TopologyName` optional?

operator/internal/controller/clustertopology/reconciler.go
enoodle: Good catch, I am working on fixing it with the added interface

operator/internal/controller/clustertopology/reconciler.go
enoodle: Can definitely do that.

operator/client/clientset/versioned/typed/core/v1alpha1/clustertopology.go
enoodle: I am not sure we should fix it in this PR, do we not have the same issue in PCS for example?

operator/internal/webhook/admission/pcs/validation/topologyconstraints.go
enoodle: I will update the function name / logic to be more clear as you proposed here.

operator/api/core/v1alpha1/podcliqueset.go
enoodle: Theoretically this is possible, but currently we don't have any scheduling backends that support this. In the design we chose not to allow it.

operator/api/core/v1alpha1/podcliqueset.go
enoodle: I will do this

enoodle: @sanjaychatterjee clustertopology.go is defining ClusterTopology CRD - maybe we want to leave it there for clarity? same as this file.

operator/api/core/v1alpha1/podcliqueset.go
enoodle: Good catch, it is checked in the webhook validation, but we can also set it here as +required because the field in PCS of the TopologyConstraints is itself +optional.

operator/internal/webhook/admission/pcs/validation/topologyconstraints.go
enoodle: I updated the comparison to only happen if the topology name is the same.

operator/client/clientset/versioned/typed/core/v1alpha1/clustertopology.go
gflarity: You're right, sorry for the noise — I didn't realize this is present on all four status-bearing types in this repo. Dismissing.

operator/cmd/main.go
```diff
@@ -99,9 +99,9 @@ func main() {
 		handleErrorAndExit(err, cli.ExitErrInitializeSchedulerBackend)
 	}
 
-	// Initialize or clean up ClusterTopology based on operator configuration.
+	// Synchronize backend topologies for all existing ClusterTopology resources.
```
gflarity, line 102:
```suggestion
// Synchronize scheduler-specific topology resources for all existing ClusterTopology resources.
```

operator/e2e/grove/topology/node_labels.go
```diff
@@ -0,0 +1,194 @@
...
+type NodeLabelChange struct {
+	NodeName     string
+	AddLabels    map[string]string // Labels to add (key → value)
+	RemoveLabels []string          // Label keys to remove
+}
+
+// MutateNodeLabels applies label changes to nodes. Returns a cleanup function that
+// reverses all changes (removes added labels, restores removed labels to their original values).
+// The cleanup function is safe to call multiple times. It reports rollback failures via t.Errorf.
+func (tv *TopologyVerifier) MutateNodeLabels(ctx context.Context, t testing.TB, changes []NodeLabelChange) (cleanup func(), err error) {
```
gflarity, line 52: I know this a test helper and there's the build e2e tag above... But having a unit test for  this somehow might be a good idea.  See issue below for example thing we might catch.

operator/e2e/grove/topology/node_labels.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+
+	return cleanup, nil
+}
+
+// rollbackNodeLabels reverses label mutations on the specified nodes:
+// - Restores labels that were removed back to their original values.
+// - Removes labels that were added by setting them to null.
+func rollbackNodeLabels(ctx context.Context, tv *TopologyVerifier, mutatedNodes []string, originalLabels map[string]map[string]string, addedLabels map[string][]string) error {
```
gflarity: It might not matter right now (perhaps only new labels added?),  but every key in AddLabels is recorded in addedLabels and later removed during rollback/cleanup, so if a test overwrites an existing node label, cleanup deletes it instead of restoring the original value.

operator/internal/controller/common/component/utils/topologyconstraint.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+	return false
+}
+
+// GetChildTopologyNamePaths returns field paths for all child topologyName fields that are set.
+func GetChildTopologyNamePaths(pcs *grovecorev1alpha1.PodCliqueSet) []string {
+	paths := sets.New[string]()
+	for i, clique := range pcs.Spec.Template.Cliques {
+		if clique.TopologyConstraint != nil && clique.TopologyConstraint.TopologyName != "" {
```
danbar2: its a bit odd to introduce new fields that we don't allow to use.
We could remove those fields or at least allow to use them with the same topology as the PCS

operator/internal/controller/podcliqueset/reconcilestatus.go
```diff
@@ -201,74 +218,81 @@ func (r *Reconciler) mutateTopologyLevelUnavailableConditions(ctx context.Contex
...
+				Message:            "topologyName may only be set on spec.template.topologyConstraint",
+				ObservedGeneration: pcs.Generation,
+				LastTransitionTime: metav1.Now(),
+			}, nil
+		default:
+			return metav1.Condition{}, fmt.Errorf("failed to resolve topologyName: %w", err)
+		}
+	}
+
+	// If topologyName is missing but topology constraints exist, report as unknown.
```
gflarity: Nit: this comment looks stale now that `ResolveTopologyNameForPodCliqueSet` returns `ErrTopologyNameMissing` when constraints exist without a PCS-level `topologyName`. At this point, `topologyName == ""` appears to mean there are no topology constraints, so the first sentence can probably be removed.

operator/internal/controller/podcliqueset/reconcilestatus_test.go
```diff
@@ -380,3 +384,315 @@ func TestMirrorUpdateProgressToRollingUpdateProgressPCS(t *testing.T) {
 		})
 	}
 }
+
```
gflarity, line 387: Could we add a case that proves the PCS uses the named `ClusterTopology`, not just any/default topology? For example, create two `ClusterTopology` objects where only `my-topology` contains the required domain and another/default topology does not. That would directly cover the multi-topology behavior introduced in `computeTopologyLevelsUnavailableCondition`.

operator/internal/controller/podcliqueset/components/podgang/syncflow_test.go
```diff
@@ -1554,3 +1556,224 @@ func TestDeterminePCSGReplicas(t *testing.T) {
 		})
 	}
 }
+
+// makePCSWithTopology creates a minimal PCS with an optional topology constraint.
```
gflarity, line 1561: Could we add coverage for the stale/invalid topology cases here? One useful test would reconcile an existing PodGang that already has grove.io/topology-name, then verify it is cleared or intentionally preserved when TAS is disabled or topology resolution no longer applies; another would cover missing/child-only topologyName resolution errors so the test suite locks down whether sync should proceed without topology constraints or fail/requeue. The rationale is that the current tests cover fresh happy-path creation, but not existing-object reconciliation or invalid legacy state, where stale scheduler-facing topology metadata is most likely to survive.

operator/internal/controller/common/component/utils/topologyconstraint.go
enoodle: I updated the implementation and design with this - now it can be set on the pod clique or pcsg level but has to match the PCS topologyName

danbar2 (dismissed):

operator/internal/controller/clustertopology/reconciler_test.go
```diff
@@ -0,0 +1,596 @@
...
+		},
+	}
+
+	mapFn := r.mapBackendTopologyToCT("kai-scheduler")
+	requests := mapFn(context.Background(), backendObj)
+	require.Len(t, requests, 1)
+	assert.Equal(t, "my-topology", requests[0].Name)
+}
+
+func TestMapBackendTopologyToCT_ExternallyManaged(t *testing.T) {
```
gflarity, line 703: One concern came up as I was digging into the coverage in this test.... nothing seems to prevent two ClusterTopology objects from referencing the same external backend topology via the same schedulerName + reference.  Is this a scenario that might happen?  Seems like we might want to prevent it.

operator/internal/controller/clustertopology/reconciler_test.go
```diff
@@ -0,0 +1,596 @@
...
+	return r.Reconcile(context.Background(), ctrl.Request{
+		NamespacedName: client.ObjectKey{Name: name},
+	})
+}
+
+func getDriftCondition(ct *grovecorev1alpha1.ClusterTopology) *metav1.Condition {
+	return meta.FindStatusCondition(ct.Status.Conditions, apicommonconstants.ConditionSchedulerTopologyDrift)
+}
+
+// -- Tests --
```
gflarity, line 167: Some suggested test additions:

* A multi-backend reconcile case with mixed outcomes, e.g. one TAS backend succeeds and another fails. This would protect the intended behavior that reconciliation records all backend statuses and returns the joined error, rather than stopping at the first failure.
* A stable-condition event case, e.g. InSync -> InSync or Drift -> Drift, to verify we don't emit duplicate Kubernetes events on repeated reconciles with no actual condition transition.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -683,7 +692,88 @@ func (v *pcsValidator) validatePodCliqueScalingGroupConfigsUpdate(oldConfigs []g
 }
 
 func (v *pcsValidator) validateTopologyConstraintsUpdate(oldPCS *grovecorev1alpha1.PodCliqueSet) field.ErrorList {
-	return newTopologyConstraintsValidator(v.pcs, v.tasEnabled, v.clusterTopologyDomains).validateUpdate(oldPCS)
+	// Allow invalid legacy objects to be repaired under the current rules by validating
```
gflarity: Could we still enforce topology constraint immutability when allowing legacy PCS objects to be repaired? This path validates the new object as a create, so a repair that adds or moves topologyName could also change existing packDomain values without being caught. A small negative test for that case would help lock this down.

operator/internal/webhook/admission/pcs/validation/topologyconstraints.go
```diff
@@ -213,9 +226,9 @@ func (v *topologyConstraintsValidator) validateHierarchicalTopologyConstraints(f
 		pcsgDomain := pcsg.TopologyConstraint.PackDomain
 		for _, cliqueName := range pcsg.CliqueNames {
 			// Find the clique by name
-			matchingPCLQTemplate := utils.FindPodCliqueTemplateSpecByName(v.pcs, cliqueName)
+			matchingPCLQTemplate := componentutils.FindPodCliqueTemplateSpecByName(v.pcs, cliqueName)
```
gflarity, line 263: Could we use the API JSON field name here, podCliqueScalingGroups, instead of the Go struct field name podCliqueScalingGroupConfigs? Otherwise admission errors will point users to a field that does not exist in the manifest. Same applies to the other PCSG topology constraint field paths in this file. 

Example:

  The validation code currently builds errors with:

  `fldPath.Child("podCliqueScalingGroupConfigs")`

  That means users could see an admission error pointing to:

  `spec.template.podCliqueScalingGroupConfigs[0].topologyConstraint`

  But that field does not exist in their manifest. The correct user-facing path is:

  `spec.template.podCliqueScalingGroups[0].topologyConstraint`

operator/internal/webhook/admission/pcs/validation/topologyconstraints_test.go
```diff
@@ -521,3 +626,133 @@ func buildTestPCS(pcsConstraint *grovecorev1alpha1.TopologyConstraint,
...
+			},
+			setupClient: nil,
+		},
+		{
+			name: "API error: non-404 error is surfaced as InternalError",
+			pcsConstraint: &grovecorev1alpha1.PodCliqueSetTopologyConstraint{
+				TopologyName: "broken-topo",
+				PackDomain:   "zone",
+			},
+			clusterTopologyObjects: []client.Object{},
```
gflarity, line 820: Could we add coverage for child-level topologyName validation here, especially PodClique/PCSG constraints that set topologyName and child-only constraints without a PCS-level topologyName? That would cover the error paths in childTopologyNameMismatchFieldErrors and ResolveTopologyNameForPodCliqueSet.

operator/internal/webhook/admission/pcs/validation/topologyconstraints_test.go
```diff
@@ -521,3 +626,133 @@ func buildTestPCS(pcsConstraint *grovecorev1alpha1.TopologyConstraint,
...
+
+			validator := &pcsValidator{
+				pcs:    pcs,
+				client: fakeClient,
+			}
+
+			domains, errs := validator.resolveTopologyDomains(context.Background())
+
+			if tc.expectedDomains == nil {
+				if domains != nil {
```
gflarity, line 860: Could we assert the total number of errors in these table tests as well? AssertErrorMatches only checks that the expected errors are present, so cases with errorMatchers: [] could still pass if validation returned unexpected errors.

gflarity (commented):
Nothing major, but I left a few comments to address. Feel free to ping me for a quick approval after you'd taken a look at the comments.

operator/api/core/v1alpha1/clustertopology.go
```diff
@@ -58,22 +51,67 @@ type ClusterTopologyList struct {
...
+}
+
+// SchedulerReference maps a ClusterTopology to a scheduler backend's topology resource.
+type SchedulerReference struct {
+	// SchedulerName is the name of the scheduler backend (e.g., "kai-scheduler").
+	// +kubebuilder:validation:Required
+	SchedulerName string `json:"schedulerName"`
+	// Reference is the name of the scheduler backend's topology resource.
+	// +kubebuilder:validation:Required
+	Reference string `json:"reference"`
```
sanjaychatterjee: Can we rename this field as `TopologyReference`?

operator/api/core/v1alpha1/clustertopology.go
```diff
@@ -58,22 +51,67 @@ type ClusterTopologyList struct {
...
 	Levels []TopologyLevel `json:"levels"`
+
+	// SchedulerReferences controls per-backend topology resource management.
+	// For each enabled TopologyAwareSchedBackend, the operator checks whether an entry
+	// for that backend exists in this list:
+	// - If absent: the operator auto-creates and manages the backend's topology resource.
+	// - If present: the named resource is assumed to be externally managed; the operator
+	//   compares its levels and reports any mismatch via the SchedulerTopologyDrift condition.
+	// +optional
+	SchedulerReferences []SchedulerReference `json:"schedulerReferences,omitempty"`
```
sanjaychatterjee: I think if we rename the type as `SchedulerTopologyReference` and the field as `SchedulerTopologyReferences` it would sound much clearer.

operator/api/core/v1alpha1/clustertopology.go
```diff
@@ -58,22 +51,67 @@ type ClusterTopologyList struct {
...
+	SchedulerName string `json:"schedulerName"`
+	// Reference is the name of the scheduler backend's topology resource.
+	// +kubebuilder:validation:Required
+	Reference string `json:"reference"`
+}
+
+// SchedulerTopologyStatus reports the sync state of a scheduler backend's topology resource.
+type SchedulerTopologyStatus struct {
+	// SchedulerName is the name of the scheduler backend.
+	SchedulerName string `json:"schedulerName"`
```
sanjaychatterjee: Why not use the `SchedulerTopologyReference` type here?

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -271,16 +271,26 @@ type PodCliqueTemplateSpec struct {
...
 	// Different replicas can be in different topology domains.
-	// +kubebuilder:validation:Enum=region;zone;datacenter;block;rack;host;numa
-	PackDomain TopologyDomain `json:"packDomain"`
+	// +optional
+	PackDomain TopologyDomain `json:"packDomain,omitempty"`
 }
 
+// PodCliqueSetTopologyConstraint is kept as a Go type alias for compatibility
+// with existing tests and helper code. The API model uses TopologyConstraint.
+type PodCliqueSetTopologyConstraint = TopologyConstraint
```
sanjaychatterjee: I don't think we should keep this kind of code around. If `PodCliqueSetTopologyConstraint` is not needed, why can't we remove it?

operator/internal/clustertopology/clustertopology.go
```diff
@@ -19,39 +19,42 @@ package clustertopology
...
+		ct := &ctList.Items[i]
+		schedulerRefMap := BuildSchedulerReferenceMap(ct.Spec.SchedulerReferences)
+
+		for _, b := range backends {
+			tasBackend, ok := b.(scheduler.TopologyAwareSchedBackend)
+			if !ok {
+				logger.V(1).Info("Scheduler backend does not implement TopologyAwareSchedBackend, skipping topology sync", "backend", b.Name())
+				continue
+			}
+			// Only sync auto-managed backends (not listed in schedulerReferences).
```
sanjaychatterjee:
```suggestion
			// Only sync grove-managed scheduler topology resources (not listed in schedulerReferences).
```

operator/internal/clustertopology/clustertopology.go
```diff
@@ -19,39 +19,42 @@ package clustertopology
...
+			// Externally-managed backends are handled by the CT controller via CheckTopologyDrift.
```
sanjaychatterjee:
```suggestion
			// Externally-managed scheduler topology resources are handled by the ClusterTopology controller via CheckTopologyDrift.
```

operator/internal/clustertopology/clustertopology.go
```diff
@@ -19,39 +19,42 @@ package clustertopology
...
+			if _, foundReference := schedulerRefMap[b.Name()]; foundReference {
```
sanjaychatterjee: Can we rename the `foundReference` variable as `isExternallyManaged`?

operator/internal/controller/common/component/utils/topologyconstraint.go
```diff
@@ -0,0 +1,126 @@
...
+			pcsg.TopologyConstraint.TopologyName != pcsTopologyName {
+			paths.Insert(fmt.Sprintf("spec.template.podCliqueScalingGroups[%d].topologyConstraint.topologyName", i))
+		}
+	}
+	return paths.UnsortedList()
+}
+
+// ResolveTopologyNameForPodCliqueSet resolves the PCS-level topology name for topology-aware scheduling.
+func ResolveTopologyNameForPodCliqueSet(pcs *grovecorev1alpha1.PodCliqueSet) (string, error) {
+	if !HasAnyTopologyConstraint(pcs) {
```
sanjaychatterjee: Isn't this check redundant?

operator/internal/controller/podcliqueset/components/podgang/syncflow_test.go
```diff
@@ -1293,9 +1295,9 @@ func TestComputeExpectedPodGangsWithTopologyConstraints(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			// Setup
-			var pcsTopologyConstraint *grovecorev1alpha1.TopologyConstraint
+			var pcsTopologyConstraint *grovecorev1alpha1.PodCliqueSetTopologyConstraint
```
sanjaychatterjee: Can we get rid of the `PodCliqueSetTopologyConstraint` type if now all the levels use the same `TopologyConstraint` type?

sanjaychatterjee (changes requested):
Requested API cleanups.

operator/internal/controller/clustertopology/reconciler_test.go
enoodle: The two ClusterTopology will have to have the same topology levels or one of them will have some bad conditions. I think that this is not something we should actively prevent but let the user figure it out from the status conditions.

operator/internal/controller/common/component/utils/topologyconstraint.go
enoodle: I think that I agree with you. Because we are adding the API now and also the webhook to prevent this, there should be no PCS that violate this. If someone succeeded in creating one then it is probably on them.

enoodle: I will also remove the whole TopologyNameMissmatch condition - by the same logic it is also redundant.

enoodle: Also ResolveTopologyNameForPodCliqueSet was called sometimes just to check something that HasAnyTopologyConstraint can answer, so I removed it from here and called HasAnyTopologyConstraint directly.

operator/e2e/grove/topology/node_labels.go
enoodle: I don't think that UT would have helped with this kind of bug - I could have had UT that missed this. I also think that if this function has a bug it would cause the e2e tests to fail or be flaky.
Writing unit tests means that this struct will have to move to files without that build tag and I don't know if I want to make this change.

operator/api/core/v1alpha1/podcliqueset.go
enoodle: We did have to change it back to +optional because now that we use this struct for pod cliques and PCSG it is again optional

operator/e2e/grove/topology/node_labels.go
enoodle: Will fix by saving the original values correctly for restore.

operator/internal/webhook/admission/pcs/validation/topologyconstraints.go
enoodle: This is was a bug before probably because I didnt change that line below with the field name you mentioned. I don't mind fixing it.

operator/internal/webhook/register.go
```diff
@@ -46,6 +47,11 @@ func Register(mgr manager.Manager, operatorCfg *configv1alpha1.OperatorConfigura
 	if err := pcsValidatingWebhook.RegisterWithManager(mgr); err != nil {
 		return fmt.Errorf("failed adding %s webhook handler: %v", pcsvalidation.Name, err)
 	}
+	ctValidatingWebhook := ctvalidation.NewHandler(mgr)
```
gflarity, line 50: Registering the handler exposes the endpoint in the manager, but I don't see the Helm/cert wiring change? We need to install a `ValidatingWebhookConfiguration` for `/webhooks/validate-clustertopology` and include this webhook in auto CA injection. Please add the chart template/values and cert injection entry for the ClusterTopology validating webhook.

Also we should probably add a simple E2E test can confirms admission is blocked etc.

operator/internal/controller/podcliqueset/reconcilestatus.go
```diff
@@ -199,90 +216,87 @@ func (r *Reconciler) mutateTopologyLevelUnavailableConditions(ctx context.Contex
...
+				Message:            "topologyName is required on spec.template.topologyConstraint when topology constraints are set",
+				ObservedGeneration: pcs.Generation,
+				LastTransitionTime: metav1.Now(),
+			}, nil
+		default:
+			return metav1.Condition{}, fmt.Errorf("failed to resolve topologyName: %w", err)
+		}
+	}
+
+	topologyLevels, err := clustertopology.GetClusterTopologyLevels(ctx, r.client, topologyName)
```
gflarity, line 274: PCS  status now depends on the referenced `ClusterTopology`, but the PodCliqueSet controller does not appear to watch `ClusterTopology` changes and map them back to referencing PCS objects. A CT deletion, recreation, or level update will not enqueue the PCS, so `TopologyLevelsUnavailable` and the generated PodGang topology constraints can stay stale until some unrelated PCS/PCLQ/PCSG event happens.

operator/internal/controller/clustertopology/reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+
+	ct := &grovecorev1alpha1.ClusterTopology{}
+	if err := r.Get(ctx, req.NamespacedName, ct); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	var statuses []grovecorev1alpha1.SchedulerTopologyStatus
+	var reconcileErr error
+
+	schedulerRefMap := ctutils.BuildSchedulerReferenceMap(ct.Spec.SchedulerTopologyReferences)
```
gflarity, line 61: BuildSchedulerReferenceMap doesn't detect duplicate schedulerName entries — the later TopologyReference silently overwrites the earlier. Consider rejecting duplicates in admission, or surfacing them in status.

operator/internal/controller/clustertopology/reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+
+	for _, backendName := range sortedBackendNames(r.backends) {
+		b := r.backends[backendName]
+		tasBackend, ok := b.(scheduler.TopologyAwareSchedBackend)
+		if !ok {
+			continue
+		}
+
+		ref := schedulerRefMap[b.Name()]
```
gflarity, line 70: A typo in schedulerName makes the map lookup return nil, which falls into the auto-managed branch and silently flips the user's external-management intent to grove-managed. Validating schedulerName at admission (or reporting TopologyNotFound here) would prevent this.

operator/internal/controller/clustertopology/reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+	if err := r.Get(ctx, req.NamespacedName, ct); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	var statuses []grovecorev1alpha1.SchedulerTopologyStatus
+	var reconcileErr error
+
+	schedulerRefMap := ctutils.BuildSchedulerReferenceMap(ct.Spec.SchedulerTopologyReferences)
+
+	for _, backendName := range sortedBackendNames(r.backends) {
```
gflarity, line 63: A ref pointing at a backend not enabled in OperatorConfiguration.scheduler.profiles will never be iterated by the reconcile loop, so its entry is silently dropped — no status, no event, no TopologyNotFound condition.

operator/internal/controller/clustertopology/reconciler.go
```diff
@@ -0,0 +1,189 @@
...
+		b := r.backends[backendName]
+		tasBackend, ok := b.(scheduler.TopologyAwareSchedBackend)
```
gflarity, line 65: A ref pointing at a non-topology-aware backend (e.g. kube-scheduler) gets skipped by the type-assertion guard with no signal to the user. Ideally we'd emit TopologyNotFound for any such ref before continuing.

operator/internal/controller/podcliqueset/reconcilestatus.go
enoodle: Interesting. The PCS stauts is dependent on ClusterTopology even before this PR because the TopologyLevelUnavailable condition is not new. I guess the e2e test passes because it doesn't wait for the reconcile to finish and updates to pod cliques trigger it to read the ClusterTopology.

operator/internal/controller/clustertopology/reconciler.go
enoodle: I added validations in the admission webhook to prevent referencing scheduler backends that don't support TAS. I believe it is enough because it won't change by changing configuration and if we upgrade to a version where more scheduler backends enable TAS all the ClusterTopology instances will be valid and we will be able to add the newly TAS enabled scheduler backend references.
To be on the safe side I added a check for this and update their status but I think that  there will be no functionality degradation in the scheduler backends so it will never happen.

operator/internal/controller/clustertopology/reconciler.go
enoodle: I added validations in the admission webhook to prevent creating new CT with disabled scheduler backends. But the mismatch can still exist after updating the operator so I also added a loop afterwards to verify this and add unknown condition in case any disabled scheduler is referenced.

operator/internal/controller/clustertopology/reconciler.go
enoodle: I added validations in the admission webhook and tests in the e2e for this case

operator/internal/controller/clustertopology/reconciler.go
enoodle: I added validations in the admission webhook and tests in the e2e for this case

operator/internal/controller/podcliqueset/reconcilestatus.go
enoodle: I updated the current e2e test to make sure it fails without the watch and added the watch

operator/internal/webhook/register.go
enoodle: I added the E2E test and made sure it now passes by adding the required webhook template and CA injection

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -271,14 +271,20 @@ type PodCliqueTemplateSpec struct {
...
+	// TopologyName is the name of the ClusterTopology resource to use for topology-aware scheduling.
+	// The PodCliqueSet-level topologyName selects the ClusterTopology for the workload.
+	// Child constraints may repeat the same topologyName, but must not change it.
+	// Immutable after creation.
+	// +optional
+	TopologyName string `json:"topologyName,omitempty"`
 	// PackDomain specifies the topology domain for grouping replicas.
 	// Controls placement constraint for EACH individual replica instance.
-	// Must be one of: region, zone, datacenter, block, rack, host, numa
+	// Must reference a domain defined in the ClusterTopology's levels.
```
sanjaychatterjee:
```suggestion
	// Must reference a domain in the topology levels defined in the ClusterTopology CR name as set in TopologyName
```

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -271,14 +271,20 @@ type PodCliqueTemplateSpec struct {
...
 	// Example: "rack" means each replica independently placed within one rack.
 	// Note: Does NOT constrain all replicas to the same rack together.
 	// Different replicas can be in different topology domains.
-	// +kubebuilder:validation:Enum=region;zone;datacenter;block;rack;host;numa
-	PackDomain TopologyDomain `json:"packDomain"`
+	// +optional
```
sanjaychatterjee: Why is PackDomain now optional? Also please ensure that you add documentation that `PackDomain` represents a required topology constraint.

operator/internal/controller/podcliqueset/components/podgang/podgang.go
```diff
@@ -134,11 +134,17 @@ func (r _resource) buildResource(pcs *grovecorev1alpha1.PodCliqueSet, pgi *podGa
...
 	}
-	if r.tasConfig.Enabled {
-		if pg.Annotations == nil {
-			pg.Annotations = make(map[string]string)
+	if r.tasConfig.Enabled && podGangHasTranslatedTopologyConstraints(pgi) {
+		if topologyName, err := componentutils.ResolveTopologyNameForPodCliqueSet(pcs); err == nil && topologyName != "" {
+			if pg.Annotations == nil {
+				pg.Annotations = make(map[string]string)
+			}
+			pg.Annotations[apicommonconstants.AnnotationTopologyName] = topologyName
```
sanjaychatterjee, line 142: Are we making the assumption that there'll be only one ClusterTopology name set for a PodGang? In that case, please put in a comment or a sentence in the implementation notes of the GREP that in the subsequent phase, topology name can be set on the each level of the PodGang.

sanjaychatterjee (commented):
In general, this PR needs to document somewhere that this multiple topology implementation is just phase1 and will remain incomplete unless we can support separate topologies for different partsof the workload. That would also require changes in the PodGang API. I would suggest to document this plan in the `Implementation Phases` section of the GREP.

operator/api/core/v1alpha1/podcliqueset.go
enoodle: We might want to set this on the PCS only with TopologyName and without PackDomain and then only set the PackDomain on the lower levels (pcsg, pclq)
We can add validation in the admission webhook for that.

sanjaychatterjee: I think that makes the API confusing. If TopologyConstraint is set on a PCS, then both TopologyName and the PackDomain are be expected to be set. I don't think that the reason for setting TopologyConstraint at PCS level being to propagate the TopologyName to lower levels is a valid use-case. It can very well be set directly at those lower levels.

operator/internal/controller/podcliqueset/components/podgang/podgang.go
enoodle: Sure. 
This assumption existed before and set the annotation before. I am actually not sure that this annotation is used anywhere to set anything actual in the cluster.
We are generally only adding multiple topology support here without supporting cross topology scheduling (because we don't have a backend scheduler that can support this) . We did update the API to support the future possibility of cross topology scheduling, but the implementation can't do that yet.
So adding a remark in the GREP makes sense

operator/api/core/v1alpha1/podcliqueset.go
enoodle: I updated the TopologyConstraint to have both fields `+required` and also the validation and reconcile logic to fit that. We still validate that all TopologyConstraints in the PCS point to the same topology name currently.

gflarity (approved):

sanjaychatterjee (approved):
LGTM! Thanks for the PR!

PR #502, opened 2026-03-30
Refactor/e2e reuse k8s clients

### LINKED ISSUE #513: Refactor e2e test infrastructure — introduce domain packages, shared clients, and new TestContext

Restructure the e2e test framework from a flat layout into domain-oriented packages
  with shared Kubernetes clients and a simplified test setup API.

  The domain-oriented structure groups code by what it operates on rather than
  by when it was written. Kubernetes primitives (pods, nodes, resources) live
  in `k8s/`, Grove-specific concepts (workloads, topology, pod groups) live
  in `grove/`, and failure diagnostics live in `diagnostics/`. This makes it
  obvious where new code belongs, keeps each package small and focused, and
  allows tests to import only the domains they need. Managers encapsulate
  the shared `*clients.Clients` once, so tests never handle raw client fields —
  they work with domain operations (`ScalePCS`, `VerifyTopology`, `WaitForPods`)
  instead of assembling clientset calls inline.

  **New package structure:**
  - `k8s/clients/` — `Clients` struct, `NewClients` (single entry point for all K8s clients)
  - `k8s/pods/` — `PodManager`, `CountPodsByPhase`, `CountReady`
  - `k8s/nodes/` — `NodeManager`, `IsReady`
  - `k8s/resources/` — `ResourceManager`, `AppliedResource`
  - `k8s/` (parent) — `PollForCondition`, conversion helpers (shared utilities)
  - `grove/workload/` — `WorkloadManager`
  - `grove/topology/` — `TopologyVerifier`
  - `grove/podgroup/` — `PodGroupVerifier`, `ExpectedSubGroup`
  - `grove/config/` — `OperatorConfig`, `GroveMetadata`
  - `diagnostics/` — `DiagCollector` (extracted from old `debug_utils.go`)
  - `tests/context.go` — `TestContext`, `PrepareTest()`, convenience methods

  **Removed:**
  - `LegacyTestContext` struct and all 24+ methods on it
  - `PrepareTestCluster`, `clientCollection`, `clientCollectionFromClients`
  - Legacy diagnostics methods from `debug_utils.go`
  - auto-mnnvl's local `testContext` / `prepareTest` (replaced by shared `TestContext`)

  **Migrated test suites:**
  - gang scheduling, startup ordering, topology, scale, cert management,
    update (rolling/ondelete), auto-mnnvl (4 suites)

  **Before:**
  ```go
  clients, cleanup := PrepareTestCluster(ctx, t, 10)
  tc := TestContext{
      Clientset:     clients.Clientset,
      DynamicClient: clients.DynamicClient,
      RestConfig:    clients.RestConfig,
      Namespace:     "default",
      Timeout:       DefaultPollTimeout,
      Interval:      DefaultPollInterval,
      Workload:      &WorkloadConfig{...},
  }
  nodesToCordon := setupAndCordonNodes(tc, 1)
  _, err := DeployAndVerifyWorkload(tc)
```
  After:
```
  tc, cleanup := PrepareTest(ctx, t, 10,
      WithWorkload(&WorkloadConfig{...}),
  )
  nodesToCordon := tc.SetupAndCordonNodes(1)
  _, err := tc.DeployAndVerifyWorkload()
```
  Tests that need domain-specific verifiers create them locally:
  topologyVerifier := topology.NewTopologyVerifier(tc.Clients, Logger)
  podGroupVerifier := podgroup.NewPodGroupVerifier(tc.Clients, Logger)


### Why is this needed?

Code is organized by domain — `k8s/{clients,pods,nodes,resources}` for                                                                                                                                                                                                                    
  Kubernetes primitives, `grove/{workload,topology,podgroup,config}` for                                                                                                                                                                                                                    
  Grove concepts, `diagnostics/` for failure collection. Each package is                                                                                                                                                                                                                    
  small, focused, and independently importable.

### PR DESCRIPTION
**Pull request by oleg-kushniriov**
#### What type of PR is this?

/kind cleanup
/kind e2e


  Refactors the e2e test framework from flat utility functions and a monolithic                                                                                                                                                                                                             
  TestContext struct into domain-specific manager structs with bound methods.
                                                                                                                                                                                                                                                                                            
  **New packages:**                                                                                                                                                                                                                                                                         
  - `e2e/k8s/` — `Clients`, `PodManager`, `NodeManager`, `ResourceManager`, polling, conversions
  - `e2e/grove/` — `WorkloadManager`, `TopologyVerifier`, `PodGroupVerifier`, `OperatorConfig`                                                                                                                                                                                              
  - `e2e/diagnostics/` — `DiagCollector`                    
  - `e2e/tests/suite.go` — `TestSuite` composing all managers with functional options                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                            
  **Key fixes:**                                                                                                                                                                                                                                                                            
  - REST mapper is now created once in `SharedClusterManager` instead of being                                                                                                                                                                                                              
    recreated per-test via `CreateKubernetesClients`                                                                                                                                                                                                
  - Pre-existing build errors in rolling update tests resolved (bare function                                                                                                                                                                                                               
    calls converted to TestSuite method calls)                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                            
  **Design for future parallel suites:**                                                                                                                                                                                                                                                    
  - `TestSuite.Namespace` is per-suite (can be unique for parallel runs)
  - `k8s.Clients` is goroutine-safe (shared across suites)                                                                                                                                                                                                                                  
  - Managers are scoped to their suite instance, no global state                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                            
  Fixes https://github.com/ai-dynamo/grove/issues/513
  Fixes #513     

  - `TestContext` and its methods in `setup.go`/`debug_utils.go` are marked                                                                                                                                                                                                                 
    deprecated but kept — they are still used by the diagnostics path and will
    be removed in a follow-up once `debug_utils.go` consumers are fully migrated.                                                                                                                                                                                                           
  - The `auto-mnnvl/` tests use their own lowercase `testContext` struct and were                                                                                                                                                                                                           
    intentionally not migrated in this PR to keep scope manageable.                                                                                                                                                                                                                         
  - The `utils/` package files remain as deprecated wrappers. Only `logger.go`                                                                                                                                                                                                              
    and `measurement/` are long-term residents; the rest will be deleted in a                                                                                                                                                                                                               
    follow-up PR.                                                                                                                                                                                                                                                                           
  - All e2e test files (startup ordering, scale, gang scheduling, topology,                                                                                                                                                                                                                 
    cert management, rolling updates) have been migrated to use `TestSuite`.                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                            
  ```release-note                                           
  NONE
```
 ```
  NONE                                              
```

### CODE REVIEW

File: operator/e2e/k8s/resources/resources.go [hunk 1]
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func (rm *ResourceManager) createResource(ctx context.Context, gvr schema.GroupVersionResource, mapping *meta.RESTMapping, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
+		return rm.clients.DynamicClient.Resource(gvr).Namespace(obj.GetNamespace()).Create(ctx, obj, metav1.CreateOptions{})
+	}
+	return rm.clients.DynamicClient.Resource(gvr).Create(ctx, obj, metav1.CreateOptions{})
+}
+
+func (rm *ResourceManager) updateResource(ctx context.Context, gvr schema.GroupVersionResource, mapping *meta.RESTMapping, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
```
gflarity, line 183: `updateResource` does not fetch the existing resource's `metadata.resourceVersion` before calling `Update`. The unstructured object decoded from YAML has no `resourceVersion`, so the API server will reject this with a 409 Conflict.

Fix: `Get` the existing resource first, copy its `resourceVersion` onto the object, then `Update`.

```go
existing, err := dynamicClient.Resource(gvr).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
if err \!= nil {
    return fmt.Errorf("get existing resource for update: %%w", err)
}
obj.SetResourceVersion(existing.GetResourceVersion())
```

operator/e2e/grove/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+	containers := dep.Spec.Template.Spec.Containers
+	if len(containers) == 0 {
+		return "", fmt.Errorf("deployment %q has no containers", operatorDeploymentName)
+	}
+	for _, c := range containers {
+		if c.Name == name {
+			return c.Image, nil
+		}
+	}
+	return containers[0].Image, nil
```
gflarity, line 113: `findContainerImage` silently falls back to `containers[0].Image` when the named container is not found. If the deployment ever has a sidecar as the first container, this returns the wrong image.

Return an error instead:
```go
return "", fmt.Errorf("container %%q not found in deployment %%q", name, operatorDeploymentName)
```

operator/e2e/grove/workload/workload.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	return k8s.PollForCondition(ctx, timeout, interval, func() (bool, error) {
+		_, err := wm.clients.DynamicClient.Resource(podCliqueSetGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+		return errors.IsNotFound(err), nil
+	})
+}
+
+// WaitForPCSG polls until a PodCliqueScalingGroup exists and returns it.
+func (wm *WorkloadManager) WaitForPCSG(ctx context.Context, groveClient groveclient.Interface, namespace, name string, timeout, interval time.Duration) (*grovecorev1alpha1.PodCliqueScalingGroup, error) {
```
gflarity: `WaitForPCSG` and `WaitForPodClique` accept `groveClient` as an explicit parameter, but `WorkloadManager` already holds `clients *k8s.Clients` which contains `GroveClient`. Every other method in this file uses `wm.clients` directly.

Consider using `wm.clients.GroveClient` instead to keep the API consistent.

operator/e2e/grove/workload/workload.go
```diff
@@ -0,0 +1,134 @@
...
+
+// DeletePCSAndWait deletes a PodCliqueSet and polls until it is fully removed.
+func (wm *WorkloadManager) DeletePCSAndWait(ctx context.Context, namespace, name string, timeout, interval time.Duration) error {
+	if err := wm.DeletePCS(ctx, namespace, name); err != nil {
+		return err
+	}
+
+	return k8s.PollForCondition(ctx, timeout, interval, func() (bool, error) {
+		_, err := wm.clients.DynamicClient.Resource(podCliqueSetGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+		return errors.IsNotFound(err), nil
```
gflarity: The condition returns `errors.IsNotFound(err), nil` which means transient API errors (network timeout, 500) are indistinguishable from "not found yet" and silently retry until timeout. Same pattern in `WaitForPCSG` and `WaitForPodClique`.

Consider:
```go
if errors.IsNotFound(err) { return true, nil }
if err \!= nil { return false, err }
return false, nil
```

operator/e2e/k8s/pods/pods.go
```diff
@@ -0,0 +1,204 @@
...
+	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	if len(namespaces) == 0 {
+		namespaces = []string{"default"}
+	}
+
+	pm.logger.Debugf("Waiting for pods to be ready in namespaces: %v", namespaces)
+
+	return wait.PollUntilContextTimeout(timeoutCtx, interval, timeout, true, func(ctx context.Context) (bool, error) {
```
gflarity: Line 69 creates `context.WithTimeout(ctx, timeout)`, then passes that same `timeout` to `wait.PollUntilContextTimeout` which creates its own internal deadline. Two timers are racing. The outer `context.WithTimeout` is redundant -- just pass `ctx` directly to `PollUntilContextTimeout` and let it manage the deadline.

operator/e2e/k8s/pods/pods.go
```diff
@@ -0,0 +1,204 @@
...
+		totalPods := 0
+		readyPods := 0
+
+		for _, namespace := range namespaces {
+			pods, err := pm.clients.Clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{
+				LabelSelector: labelSelector,
+			})
+			if err != nil {
+				pm.logger.Errorf("Failed to list pods in namespace %s: %v", namespace, err)
```
gflarity, line 85: When `List` fails, the error is logged but `(false, nil)` is returned, causing the poller to silently retry. If the error is persistent (RBAC issue, namespace doesn't exist), this loops until timeout with no useful error message. Consider returning the error: `return false, fmt.Errorf("listing pods: %%w", err)`.

operator/e2e/k8s/polling.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+// PollForCondition repeatedly evaluates a condition function at the specified interval
+// until it returns true or the timeout is reached. Returns an error if the condition fails,
+// returns an error, or the timeout expires.
+func PollForCondition(ctx context.Context, timeout, interval time.Duration, condition func() (bool, error)) error {
```
gflarity, line 30: This reimplements `wait.PollUntilContextTimeout` from `k8s.io/apimachinery/pkg/util/wait`, which is already used in `pods.go`. Having two different polling mechanisms with different behaviors (error handling, context propagation) in the same package is confusing.

Also, the condition callback `func() (bool, error)` doesn't receive `context.Context`. The standard signature is `func(ctx context.Context) (bool, error)`.

Consider standardizing on `wait.PollUntilContextTimeout` throughout.

operator/e2e/grove/podgroup/podgroup.go
```diff
@@ -0,0 +1,335 @@
...
+
+	if err := pv.VerifySubGroups(podGroup, expectedSubGroups); err != nil {
+		return fmt.Errorf("SubGroups verification failed: %w", err)
+	}
+
+	return nil
+}
+
+// VerifyScaledPCSGReplicaTopology verifies KAI PodGroup for one scaled PCSG replica.
+func (pv *PodGroupVerifier) VerifyScaledPCSGReplicaTopology(ctx context.Context, t *testing.T, namespace, pcsName string, pcsReplica int, pcsgConfig ScaledPCSGConfig, pcsConstraint string) {
```
gflarity: `VerifyScaledPCSGReplicaTopology` calls `t.Fatalf` directly, while every other method in this file returns `error`.  Seems like this one was missed.

operator/e2e/grove/topology/topology.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	return tv.VerifyPodsInSameTopologyDomain(ctx, filteredPods, topologyKey)
+}
+
+// VerifyPCSGReplicasInTopologyDomain verifies that each PCSG replica's pods are in the same topology domain.
+func (tv *TopologyVerifier) VerifyPCSGReplicasInTopologyDomain(ctx context.Context, allPods []v1.Pod, pcsgLabel string, replicaCount, podsPerReplica int, topologyLabel string) error {
+	for replica := 0; replica < replicaCount; replica++ {
+		replicaPods := FilterPodsByLabel(
+			FilterPodsByLabel(allPods, "grove.io/podcliquescalinggroup", pcsgLabel),
```
gflarity, line 204: Use constants from API instead of harcoding string literals here ("grove.io/podcliquescalinggroup" and "grove.io/podcliquescalinggroup-replica-index").

operator/e2e/tests/context.go
```diff
@@ -0,0 +1,537 @@
...
+	PodGroups *grove.PodGroupVerifier
+	Config    *grove.OperatorConfig
+	Diag      *diagnostics.DiagCollector
+
+	// Per-suite configuration
+	Namespace string
+	Timeout   time.Duration
+	Interval  time.Duration
+	Workload  *WorkloadConfig
+	NodeNames []string // Reserved nodes for this suite (future: from NodePool)
```
gflarity: **Incomplete options pattern:** `NodeNames []string` is declared but never populated by any code in this PR, and there's no `WithNodeNames` functional option. Either add a `WithNodeNames` option for consistency, or remove the field if it's not yet needed.

operator/e2e/diagnostics/collector.go
```diff
@@ -0,0 +1,502 @@
...
+		dc.logger.Infof("Diagnostics collection complete. Output written to: %s", filename)
+	}
+}
+
+func (dc *DiagCollector) dumpOperatorLogs(logger *utils.Logger) {
+	logger.Info("================================================================================")
+	logger.Info("=== OPERATOR LOGS (all) ===")
+	logger.Info("================================================================================")
+
+	ctx := context.Background()
```
gflarity: All dump methods use `context.Background()` internally  so there's no way to cancel a hung diagnostics collection. Let's just update these to  except a context`context.Context` in `CollectAll` and propagate it.

operator/e2e/k8s/resources/resources.go
oleg-kushniriov: This PR purpose to rearrange the existing code to be more structured. Fixing current implementation of existing of code is out of scope. Fixed this code part

operator/e2e/grove/workload/workload.go
oleg-kushniriov: fixed

operator/e2e/grove/workload/workload.go
oleg-kushniriov: fixed

operator/e2e/k8s/pods/pods.go
oleg-kushniriov: fixed

operator/e2e/k8s/polling.go
oleg-kushniriov: This PR purpose to rearrange the existing code to be more structured. Fixing current implementation of existing of code is out of scope. 
There is an additional planned task to add a `waiter` package to contain a logic related to the waiting for a certain condition to be satisfied

operator/e2e/diagnostics/collector.go
oleg-kushniriov: added context param

operator/e2e/tests/context.go
oleg-kushniriov: removed

operator/e2e/grove/topology/topology.go
oleg-kushniriov: This PR purpose to rearrange the existing code to be more structured. Fixing current implementation of existing of code is out of scope.

operator/e2e/grove/podgroup/podgroup.go
oleg-kushniriov: fixed

operator/e2e/grove/config/config.go
oleg-kushniriov: This PR purpose to rearrange the existing code to be more structured. Fixing current implementation of existing of code is out of scope. Changing this code might

operator/e2e/architecture.md
```diff
@@ -0,0 +1,344 @@
...
+  │  ├── Grove Managers ──────────────────────────┤  │
+  │  │  Workloads  *WorkloadManager ──┐           │  │
+  │  │  Topology   *TopologyVerifier  │ composes  │  │
+  │  │  PodGroups  *PodGroupVerifier  │ Resources │  │
+  │  │  Config     *OperatorConfig    │ + Pods    │  │
+  │  ├── Diagnostics ─────────────────┘───────────┤  │
+  │  │  Diag       *DiagCollector                 │  │
+  │  └────────────────────────────────────────────┘  │
+  │                                                  │
+  │  Convenience methods: ListPods, WaitForPods,     │
```
danbar2, line 78: why pods methods are not part of the pods manager?

oleg-kushniriov: ListPods from TestContext is only a wrapper for PodsManager List() method

danbar2 (dismissed):

operator/e2e/tests/debug_utils.go
shayasoolin: Seems like this is the only thing left here - any chance to move it to some other file?

operator/e2e/testctx/context.go
```diff
@@ -0,0 +1,541 @@
...
+	"github.com/ai-dynamo/grove/operator/e2e/k8s/pods"
+	"github.com/ai-dynamo/grove/operator/e2e/k8s/resources"
+	"github.com/ai-dynamo/grove/operator/e2e/setup"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestContext is the primary per-test helper struct.
+// Clients are created once and shared; domain managers are created by tests on demand.
+type TestContext struct {
```
shayasoolin, line 67: I suggest to move this important file to be in a separate folder, not together with the tests themselves.

danbar2 (dismissed):

operator/e2e/setup/constants.go
```diff
@@ -1,3 +1,5 @@
+//go:build e2e
```
shayasoolin, line 1: Did you find a solution for keeping the e2e code out of the binary, while still being able to browse the code in the IDE?

oleg-kushniriov: The solution in InteliJ is to ass e2e as custom tag in settings

operator/e2e/testctx/context.go
oleg-kushniriov: moved

shayasoolin (approved):

danbar2 (approved):

PR #504, opened 2026-03-31
fix: use server-side apply in RU e2e tests

### LINKED ISSUE #503: [BUG] Flaky E2E rolling update tests due to PCS update conflict race
### What happened?

`Test_RU12_RollingUpdateWithPCSScaleInDuringUpdate` intermittently fails in CI with:

```
Failed to trigger rolling update on pc-c: Operation cannot be fulfilled on
podcliquesets.grove.io "workload1": the object has been modified; please apply
your changes to the latest version and try again
```

`triggerPodCliqueRollingUpdate` calls GET → modify → UPDATE sequentially for pc-a, pc-b, pc-c. After the first update, the grove controller reconciles and writes back to the PCS, changing its `resourceVersion`. The next update conflicts and `RetryOnConflict` (5 retries × 10ms) cannot recover.

Failures confirmed across multiple branches:
- `erez/chore/update-kai-version`: runs 23736130190, 23480636572
- `crd-upgrader-impl`: runs 23756654965, 23747858360, 23445539025

Not reproducible locally — CI DinD with CPU throttling causes bursty controller reconciliation that widens the conflict window.

Rolling update tests should pass reliably in CI regardless of controller reconciliation timing.

### Environment

- CI runners: `prod-grove-e2e-v1` (Docker-in-Docker with CPU limits)
- Cluster: k3d with 30 KWOK nodes (default e2e preset)
- Kubernetes: k3s v1.34.2

enoodle:
#### What type of PR is this?
/kind bug
/kind api

Use server-side apply in rolling updates e2e to fix flaky tests

Fixes #503

This updates the API of PCS to the field PodCliqueSetTemplateSpec.Cliques a map interpretation. It had no list-type annotation (defaults to atomic — the whole array is replaced on merge). Added +listType=map
   +listMapKey=name to match how containers and env are already annotated. Non-breaking for existing clients.


operator/e2e/tests/update/utils.go
```diff
@@ -244,60 +245,45 @@ func verifyPodsStillPresent(tc tests.TestContext, expectedPods map[string]bool)
...
 		pcs.Spec.UpdateStrategy.Type = strategyType
 
-		updatedUnstructured, err := tests.ConvertTypedToUnstructured(&pcs)
-		if err != nil {
-			return fmt.Errorf("failed to convert to unstructured: %w", err)
-		}
-
-		_, err = tc.DynamicClient.Resource(utils.PodCliqueSetGVR).Namespace(tc.Namespace).Update(tc.Ctx, updatedUnstructured, metav1.UpdateOptions{})
-		return err
+		return tc.CRClient.Patch(tc.Ctx, &pcs, client.Apply, client.FieldOwner("e2e-rolling-update-test"), client.ForceOwnership)
```
Ronkahn21, line 262: use Apply instead Patch for clarity  

```suggestion
		return tc.CRClient.Apply(...)
```

operator/e2e/tests/update/utils.go
```diff
@@ -244,60 +245,45 @@ func verifyPodsStillPresent(tc tests.TestContext, expectedPods map[string]bool)
...
 		var pcs grovev1alpha1.PodCliqueSet
-		err = utils.ConvertUnstructuredToTyped(unstructuredPCS.Object, &pcs)
-		if err != nil {
-			return fmt.Errorf("failed to convert to PodCliqueSet: %w", err)
+		if err := tc.CRClient.Get(tc.Ctx, types.NamespacedName{Name: pcsName, Namespace: tc.Namespace}, &pcs); err != nil {
+			return fmt.Errorf("failed to get PodCliqueSet: %w", err)
 		}
+		pcs.SetGroupVersionKind(grovev1alpha1.SchemeGroupVersion.WithKind("PodCliqueSet"))
+		pcs.SetManagedFields(nil)
+		pcs.SetResourceVersion("")
```
Ronkahn21, line 279: Can you explain this? why it is required, any maybe need to be util

operator/e2e/tests/update/utils.go
enoodle: @Ronkahn21 Currently I can't because we don't generate the classes for that. Do we want to generate the ApplyConfigurations just for tests ?

operator/e2e/tests/update/utils.go
enoodle: @Ronkahn21 Some tests only pass with this. It was suggested by Claude as a way to mitigate an issue where for example "Test_RU9_RollingUpdateAllPodCliques" kept failing. I am trying to understand why exactly this test keeps failing when removing this, because there is no documentation about this field..

operator/e2e/tests/update/utils.go
Ronkahn21: we should move to SSA in general, but for now lets not introduce it right now.
so lets keep it like with patch

operator/e2e/tests/update/utils.go
Ronkahn21: the MangedFields nil is for field ownership in SSA but the the other two  I am not sure why.
Can you research the affect before comiting them to the repo ?

enoodle: @Ronkahn21 I didn't get to do much research yet, but they make a lot of sense from Update/Patch perspective:
ResourceVersion - When you patch an object you don't want to create a patch to the previous resource version in case the object was indeed updated after we got it from the API.
Group and Kind - This is a very common mismatch that happens with the controller-runtime client - it doesn't set them when reading objects from cache. after minimal research I found for example that the fake client was update for example to mimic this behavior: https://github.com/kubernetes-sigs/controller-runtime/issues/3302

Usually with tests code I just relay on my experience and research less each line as long as the test behave as expected, I feel this is a necessary trade off today when AI generates a lot of test code.

Ronkahn21: I understand what are you saying, but I never needed to set the GVK or change the resource version when I Used patch in Controller runtime client (before SSA), I was wonder why it needed now with SSA.

it look weird and forcefully to work this is why I was wonder

danbar2 (approved):

PR #516, opened 2026-04-07
operator: add Volcano scheduler backend support for gang scheduling

### LINKED ISSUE #395: ☂️ Grove Scheduling Framework

This is an umbrella issue to track all work related to Grove's scheduling framework

### Why is this needed?

Grove has a scheduler API and re-targetable scheduler backend to ensure it can interoperate with multiple schedulers.

xianlubird:
Introduce a new `volcano` scheduler backend in the operator scheduler framework and wire it through configuration, validation, scheme registration, backend manager selection, RBAC, and generated API reference docs.

This change adds Volcano PodGroup synchronization from Grove PodGang resources, prepares Pods with `schedulerName=volcano` plus the official Volcano group annotation, defaults and validates Volcano-specific queue config, upgrades the Volcano API dependency to v1.13.2, and documents the new scheduler profile in the generated operator API reference and chart values example.

The initial Volcano integration intentionally supports gang scheduling only and rejects topology-aware scheduling constraints.

#### What type of PR is this?

/kind feature
/kind api

This PR adds initial Volcano gang scheduling support to Grove operator.

It introduces a new `volcano` scheduler backend that:
- syncs Grove `PodGang` to Volcano `PodGroup`
- prepares Pods with `schedulerName: volcano`
- writes the official Volcano pod group annotation `scheduling.k8s.io/group-name`
- supports Volcano queue config defaulting and validation
- adds the required Volcano API scheme registration and RBAC

Current scope:
- supports gang scheduling
- does not support topology-aware scheduling
- explicitly rejects `topologyConstraint` when the Volcano backend is selected

Fixes #395


operator/api/config/v1alpha1/types.go
```diff
@@ -108,6 +112,14 @@ type KubeSchedulerConfig struct {
 	GangScheduling bool `json:"gangScheduling,omitempty"`
 }
 
+// VolcanoSchedulerConfiguration defines the configuration for the Volcano backend.
+type VolcanoSchedulerConfiguration struct {
+	// Queue is the Volcano queue name to associate with PodGroups created by Grove.
+	// If unset, defaulting uses "default".
+	// +optional
+	Queue string `json:"queue,omitempty"`
```
kangclzjc: Does that means we only support on `Queue`，what if use have multiple Queues defined in their system?

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,155 @@
...
+		return
+	}
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
+	}
+	pod.Annotations[volcanov1beta1.VolcanoGroupNameAnnotationKey] = podGangName
+	pod.Annotations[volcanov1beta1.KubeGroupNameAnnotationKey] = podGangName
+}
+
+func (b *schedulerBackend) ValidatePodCliqueSet(_ context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error {
```
kangclzjc: Should we also validate Queues here, what if user specify a invalid Queue ?

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		for key, value := range podGang.Labels {
+			podGroup.Labels[key] = value
+		}
+
+		if err := controllerutil.SetControllerReference(podGang, podGroup, b.scheme); err != nil {
+			return err
+		}
+
+		podGroup.Spec.MinMember = minMemberForPodGang(podGang)
```
kangclzjc: In volcano, it seems that from 1.13 it should support subgroup which will fully support hierarchy gang, could you please also support this ?

operator/api/config/v1alpha1/types.go
xianlubird: Done

operator/internal/scheduler/volcano/backend.go
xianlubird: Done

operator/internal/scheduler/volcano/backend.go
xianlubird: As noted in the community discussion:

_**Since v1.14.0, Volcano has provided a secondary grouping capability supporting subgroup-level gang scheduling.**_

Based on this, subgroup-level gang scheduling is only supported starting from Volcano v1.14.

For now, we should evaluate this feature further and consider adopting it after it becomes more stable and production-ready.

docs/proposals/376-volcano-scheduler-backend/README.md
```diff
@@ -0,0 +1,335 @@
...
+    scheduling.grove.io/volcano-queue: gpu-training
+spec:
+  roleName: worker
+  replicas: 2
+  podSpec:
+    schedulerName: volcano
+    containers:
+      - name: main
+        image: busybox:1.36
+        command: ["sh", "-c", "sleep 3600"]
```
kangclzjc, line 266: It seems that we don't support use PodClique without PCS?

docs/proposals/376-volcano-scheduler-backend/README.md
```diff
@@ -0,0 +1,335 @@
...
+- verify admission rejects the request
+
+## Future Work
+
+Potential follow-up work includes:
+
+- Volcano topology-aware scheduling support
+- richer PodGroup status integration
+- optional queue existence checks during initialization
+- additional end-to-end coverage in automated CI
```
kangclzjc: Maybe also add Volcano sub podgroup support from 1.14

operator/internal/controller/podcliqueset/components/podclique/podclique.go
```diff
@@ -329,6 +333,14 @@ func (r _resource) buildResource(logger logr.Logger, pclq *grovecorev1alpha1.Pod
 	return nil
 }
 
+func mergedPodCliqueAnnotations(pcsAnnotations, cliqueAnnotations map[string]string) map[string]string {
+	annotations := lo.Assign(map[string]string{}, cliqueAnnotations)
+	if queue := strings.TrimSpace(pcsAnnotations[volcanoscheduler.QueueAnnotationKey]); queue != "" {
+		annotations[volcanoscheduler.QueueAnnotationKey] = queue
```
kangclzjc: it seems that you're override pclq's annotation with pcs's annotation, shouldn't that be reverse?

docs/proposals/376-volcano-scheduler-backend/README.md
xianlubird: No, actually this example only uses PodClique without PCS.

operator/internal/controller/podcliqueset/components/podgang/podgang.go
```diff
@@ -140,6 +142,21 @@ func (r _resource) buildResource(pcs *grovecorev1alpha1.PodCliqueSet, pgi *podGa
 		}
 		pg.Annotations[apicommonconstants.AnnotationTopologyName] = grovecorev1alpha1.DefaultClusterTopologyName
 	}
+	if getSchedulerNameForPCS(pcs) == string(configv1alpha1.SchedulerNameVolcano) {
```
kangclzjc: Could this part be moved to SyncPodGang?

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil
+	}
+
+	queueName := strings.TrimSpace(queue)
+	if queueName == "" {
+		queueName = DefaultQueue
+	}
+
+	volcanoQueue := &volcanov1beta1.Queue{}
+	if err := k8sClient.Get(ctx, types.NamespacedName{Name: queueName}, volcanoQueue); err != nil {
```
kangclzjc: Should we validate Volcano Queue in grove? Actually I'm not sure. We can keep it, may need @sanjaychatterjee and @unmarshall to give some suggestions.

operator/internal/controller/podcliqueset/components/podclique/podclique.go
xianlubird: Done

docs/proposals/376-volcano-scheduler-backend/README.md
xianlubird: Done

operator/internal/controller/podcliqueset/components/podgang/podgang.go
xianlubird: Done

xianlubird:
@sanjaychatterjee  @enoodle @danbar2 @unmarshall  This code is implement for PR https://github.com/ai-dynamo/grove/pull/560 ,pls feel free to review

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -316,7 +316,7 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
 	podGangName := apicommon.GeneratePodGangNameForPodCliqueOwnedByPCSG(pcs, pcsReplicaIndex, pcsg, pcsgReplicaIndex)
 
 	pclq.Labels = getLabels(pcs, pcsReplicaIndex, pcsg, pcsgReplicaIndex, pclqObjectKey, pclqTemplateSpec, podGangName)
-	pclq.Annotations = maps.Clone(pclqTemplateSpec.Annotations)
+	pclq.Annotations = volcanoscheduler.MergePodCliqueAnnotations(pcs.Annotations, pclqTemplateSpec.Annotations)
```
sanjaychatterjee: Can we avoid adding Volcano dependency in the controller?

operator/internal/controller/podcliqueset/components/podclique/podclique.go
```diff
@@ -308,8 +308,11 @@ func (r _resource) buildResource(logger logr.Logger, pclq *grovecorev1alpha1.Pod
 	// Add finalizer at creation so PCLQ controller does not need a separate PATCH on first reconcile.
 	controllerutil.AddFinalizer(pclq, apiconstants.FinalizerPodClique)
 	pclq.Labels = getLabels(pcs, pcsReplica, pclqObjectKey, pclqTemplateSpec, apicommon.GeneratePodGangNameForPodCliqueOwnedByPodCliqueSet(pcs, pcsReplica))
-	pclq.Annotations = maps.Clone(pclqTemplateSpec.Annotations)
+	pclq.Annotations = volcanoscheduler.MergePodCliqueAnnotations(pcs.Annotations, pclqTemplateSpec.Annotations)
```
sanjaychatterjee: Can we avoid adding Volcano dependency in the controller?

operator/internal/scheduler/volcano/queue.go
sanjaychatterjee: @kangclzjc Can you please work with @xianlubird to please add a backend interface for scheduler specific validations?

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -274,6 +279,73 @@ func (v *pcsValidator) validatePodCliqueTemplates(fldPath *field.Path) ([]string
...
+	if len(schedulerNames) == 0 {
+		return defaultSchedulerName
+	}
+	if schedulerNames[0] == "" {
+		return defaultSchedulerName
+	}
+	return schedulerNames[0]
+}
+
+func (v *pcsValidator) validateVolcanoQueueAnnotations() field.ErrorList {
```
sanjaychatterjee: Can we move all scheduler specific validations into the backend? We should introduce a scheduler validation interface if needed. cc @kangclzjc

sanjaychatterjee (changes requested):

operator/api/config/v1alpha1/types.go
```diff
@@ -58,19 +58,22 @@ const (
...
 )
 
 // SchedulerConfiguration configures scheduler profiles and which is the default.
 type SchedulerConfiguration struct {
 	// Profiles is the list of scheduler profiles. Each profile has a backend name and an optional config.
 	// The default-scheduler backend is always enabled to ensure that the kubernetes default scheduler is always enabled and supported.
 	// Use profile name "default-scheduler" to configure or set it as default.
-	// Valid profile names: "default-scheduler", "kai-scheduler". Use defaultProfileName to designate the default backend.
+	// Valid profile names: "default-scheduler", "kai-scheduler", "volcano". Use defaultProfileName to designate the default backend.
+	// The Volcano backend supports gang scheduling via Volcano PodGroup and currently rejects topology-aware scheduling constraints.
```
unmarshall: the doc-string should not mention volcano and the comment is semantically not the correct place. This field only captures all supported scheduler profiles. I would recommend that you remove this comment.

operator/api/config/validation/validation_test.go
```diff
@@ -310,16 +321,15 @@ func TestValidateSchedulerConfiguration(t *testing.T) {
 		},
 		// unsupported profile name
 		{
-			name: "invalid: unsupported profile name",
```
unmarshall, line 313: this is a valid test case, instead of overwriting this just change the name of the scheduler to some made-up name which grove will never support. Add a test case for volcano has one of the supported schedulers.

Something like the following:
```go
{
      name: "invalid: unsupported profile name",
      scheduler: &configv1alpha1.SchedulerConfiguration{
          Profiles: []configv1alpha1.SchedulerProfile{
              {Name: configv1alpha1.SchedulerName("unknown-scheduler")},
              {Name: configv1alpha1.SchedulerNameKube},
          },
          DefaultProfileName: "unknown-scheduler",
      },
      expectErrors:   2,
      expectedFields: []string{"scheduler.profiles[0].name", "scheduler.defaultProfileName"},
      expectedTypes:  []field.ErrorType{field.ErrorTypeNotSupported, field.ErrorTypeInvalid},
  },

  Without it, a regression that loosens the IsSupportedSchedulerName check (or breaks the NotSupported field error emission entirely) could slip through this test file unnoticed.
```

operator/charts/templates/clusterrole.yaml
```diff
@@ -165,6 +165,26 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
unmarshall, line 171: I think you have followed what has been done for KAI scheduler but that is not a recommended approach. Cluster roles should be assigned with the principle of least privilege. If Grove is deployed without enabling `Volcano` or `KAI` then even if these schedulers have been deployed, Grove must not be granted cluster wide privileges for custom resources that each of these schedulers define.

I did not get to review the KAI scheduler specific PR, but lets not repeat that mistake for Volcano. In a separate PR we can add similar guards for KAI scheduler backend as well.

in `values.yaml` you already have `config.scheduler.profiles` we use that to put a guard.

Something like the following:
```yaml
{{- if has "volcano" (pluck "name" .Values.config.scheduler.profiles | compact) }}
  - apiGroups:
    - scheduling.volcano.sh
    resources: [podgroups]
    verbs: [create, get, list, watch, patch, update, delete]
  - apiGroups:
    - scheduling.volcano.sh
    resources: [queues]
    verbs: [get, list, watch]
  {{- end }}
```

operator/charts/values.yaml
```diff
@@ -89,11 +89,17 @@ config:
     podCliqueScalingGroup:
       concurrentSyncs: 3
   # Scheduler configures which scheduler backends are active. default-scheduler is always available.
+  # Supported profile names are default-scheduler, kai-scheduler, and volcano.
+  # The volcano backend supports gang scheduling via PodGroup and currently rejects topology-aware scheduling constraints.
```
unmarshall: please remove this comment. Not the right place to put it.

operator/charts/values.yaml
```diff
@@ -89,11 +89,17 @@ config:
...
   # List profiles to enable backends; set defaultProfileName to the profile that is the default backend.
+  # Volcano queue selection is workload-scoped, not scheduler-profile-scoped.
```
unmarshall: remove this comment from here.

operator/charts/values.yaml
```diff
@@ -89,11 +89,17 @@ config:
...
+  # Use metadata.annotations["scheduling.grove.io/volcano-queue"] on PodCliqueSet or PodClique.
```
unmarshall: remove this comment from here.

operator/charts/values.yaml
```diff
@@ -89,11 +89,17 @@ config:
...
+  # If unset, the effective Volcano queue defaults to "default".
```
unmarshall: remove this comment from here.

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
unmarshall: Yes i agree. Controllers should not directly call functions that are defined in the scheduler/<specific-scheduler> packages for several reasons:

1. It establishes a wrong precedent as other scheduler backends added later will follow this precedent and soon the controller will break the Backend abstraction across schedulers.
2. The controller is now coupled to Volcano even when volcano is not amongst the active scheduler backends. Every PCS reconcile calls `volcanoscheduler.MergePodCliqueAnnotations`.

If you look at the `PodGang` component in the podcliqueset reconciler then you will see that labels and annotations that are set in PodCliqueSet.Labels/Annotations are mirrored onto the PodGang resources. See [here](https://github.com/ai-dynamo/grove/blob/b3a10ef0628b8a78541c504db0cc7d44312c029c/operator/internal/controller/podcliqueset/components/podgang/podgang.go#L137-L145).

What this means is that the PodGang resource will already have the queue annotation set on it.
This also means that:

- You can remove `MergePodCliqueAnnotations` entirely.
- You can remove resolveQueueForPodGang's PodClique fetch loop. In SyncPodGang, read podGang.Annotations[QueueAnnotationKey] directly. If unset, fall back to DefaultQueue. The PodGang already carries the annotation thanks to existing mirrorPCSMetadata behavior.
- The conflict-detection that resolveQueueForPodGang does today (multiple cliques disagreeing) becomes a no-op. It was checking that all PodCliques in a PodGang resolved to the same queue. With the annotation living on the PodGang itself, there's exactly one value — no possibility of conflict at this layer. Webhook validation already enforces "all cliques must resolve to the same queue" at admission, so the runtime check is redundant defense.
- PCS → PodClique propagation isn't needed at all. The PodClique never needs to carry the queue annotation, because nothing reads it from the PodClique except resolveQueueForPodGang, which we're removing.

unmarshall (changes requested):
1/n reviews. I have only reviewed some of the files. Will publish more comments in the second review round.

operator/internal/controller/podcliqueset/components/podclique/podclique.go
unmarshall: see https://github.com/ai-dynamo/grove/pull/516#discussion_r3369058331

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,282 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to sync Volcano PodGroup for PodGang %s/%s: %w", podGang.Namespace, podGang.Name, err)
+	}
+
+	return nil
+}
+
+func recordCapabilityEvent(eventRecorder record.EventRecorder, obj runtime.Object, err error) {
+	if eventRecorder == nil || obj == nil {
```
unmarshall: will the eventRecorder ever be nil? Since the code expects it to not be nil, then it should at least log that cannot raise events because event recorder is nil. TBH this is more a coding error and i would let a nil pointer dereference be raised so that this is caught. Silently ignoring it is not so nice.

operator/internal/scheduler/volcano/backend.go
unmarshall: I still do not see validation for the queue? Where is that done? This method only checks for any TopologyConstraints being set on PCS, PCLQ and PCSG at the moment

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	return nil
+}
+
+func minMemberForPodGang(podGang *groveschedulerv1alpha1.PodGang) int32 {
+	var total int32
+	for _, group := range podGang.Spec.PodGroups {
+		total += group.MinReplicas
+	}
+	if total == 0 {
```
unmarshall, line 179: Now this will become an issue with `coherent updates`. Once the PodGang has been scheduled then we explicitly set `MinReplicas` to 0 thus releasing the scheduling constraints. We wait for minReplica pods across constituent PodGroups to be scheduled and once all of these minReplica pods across PodGroups have been scheduled then we mark that available condition on the PodGang and set it to true and then we also set MinReplicas for every PodGroup to 0. So if you are deterministically going to read the MinReplicas before that time then its fine else you will always get an incorrect total minReplicas.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,282 @@
...
+		if !ok {
+			return false
+		}
+		_, ok = specSchema.Properties["subGroupPolicy"]
+		return ok
+	}
+	return false
+}
+
+func (b *schedulerBackend) resolveQueueForPodGang(ctx context.Context, podGang *groveschedulerv1alpha1.PodGang) (string, error) {
```
unmarshall: This is not required. Checking that all PCLQs use the same queue - should be done as part of the validating webhook. When PG is going to have a copy of annotations/labels from PCS you have all the information on the PodGang itself.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,282 @@
...
+	profile       configv1alpha1.SchedulerProfile
+}
+
+var _ scheduler.Backend = (*schedulerBackend)(nil)
+
+// newCapabilityClient creates a direct API client for the startup capability
+// check. The check reads the Volcano PodGroup CRD before the manager cache is
+// started, and using the cached client here would register a CRD informer that
+// requires cluster-wide list/watch permissions on CustomResourceDefinitions.
+var newCapabilityClient = func(scheme *runtime.Scheme) (client.Client, error) {
```
unmarshall: `newCapabilityClient` is a very misleading name. What you are creating here is a direct/uncached client because `Init` is called before the manager is started, thus informer caches are all un-initialized. 

Also all backend scheduler implementations will have to implement `Init` and thus this is a common need. There is no point putting this in `volcano` package. So now there are two ways of doing this:

1. **[Preferred]:** Change the `Init` method signature in the `scheduler.Backend` to accept a directClient. This is the constructed only in `scheduler/registry.go` and passed to `b.Init(directClient)` inside `newSchedulerBackend` function. In this approach the below function can still exist but it will not be exposed.
2. Introduce a utility function in `scheduler/client.go`

```go
  // NewDirectClient returns a non-cached controller-runtime client suitable for
  // Backend.Init implementations. The manager's cached client is not yet usable
  // at backend-init time because the cache is started after Manager.Start, and
  // using it here would register informers that require list/watch on every
  // type the backend touches. Backends that need to read CRDs, queues, or other
  // cluster state during Init should use this client.
  func NewDirectClient(scheme *runtime.Scheme) (client.Client, error) {
      cfg, err := ctrl.GetConfig()
      if err != nil {
          return nil, fmt.Errorf("error getting cluster config to initialize a direct client: %w", err)
      }
      return client.New(cfg, client.Options{Scheme: scheme})
  }
```
Each scheduler backend can then call this utility function.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,282 @@
...
+	directClient, err := client.New(cfg, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create direct API client for Volcano capability check: %w", err)
+	}
+	return directClient, nil
+}
+
+// SetCapabilityClientFactoryForTest replaces the client factory used by Init.
+// It is intended for unit tests that should not talk to a real cluster.
+func SetCapabilityClientFactoryForTest(factory func(*runtime.Scheme) (client.Client, error)) func() {
```
unmarshall: A package level variable being mutated by this function is not really nice. You are having a global mutable state and using it for tests.
So my earlier suggestion for `Init` to take a direct client would remove the need for this function. Lets do that then.

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,125 @@
...
+	if queue == "" {
+		return DefaultQueue
+	}
+	return queue
+}
+
+// MergePodCliqueAnnotations preserves PodClique-local annotations and only
+// backfills the Volcano queue annotation from the PodCliqueSet when the
+// PodClique template does not specify its own queue.
+func MergePodCliqueAnnotations(pcsAnnotations, cliqueAnnotations map[string]string) map[string]string {
```
unmarshall: as pointed out earlier, this function is not required

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+		annotations[QueueAnnotationKey] = queue
+	}
+	return annotations
+}
+
+// ResolvePodCliqueQueue returns the effective queue for a PodClique template/object after
+// applying the global object annotations and the clique-local annotations.
+// If both define the queue annotation, they must match.
+func ResolvePodCliqueQueue(globalAnnotations, cliqueAnnotations map[string]string) (string, error) {
```
unmarshall: is this required now? PodGang resource will have the required annotation.

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,125 @@
...
+		return globalQueue, nil
+	case globalQueue != cliqueQueue:
+		return "", ErrConflictingQueueAnnotations
+	default:
+		return globalQueue, nil
+	}
+}
+
+// ValidateQueueName validates the effective Volcano queue name.
+func ValidateQueueName(queue string) []string {
```
unmarshall: Why is this function being directly called from within the PCS validator?
The only method that should be called from backend.Scheduler should be `ValidatePodCliqueSet` from the webhook. That method is the entry point to do all scheduler specific validations of the PCS (if the PCS is using a specific scheduler backend)

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,125 @@
...
+	return validation.IsDNS1123Subdomain(strings.TrimSpace(queue))
+}
+
+// ValidateQueueExistsAndIsOpen verifies that the resolved Volcano queue exists
+// in the cluster and that its status is Open before admitting the workload.
+func ValidateQueueExistsAndIsOpen(ctx context.Context, k8sClient client.Reader, queue string) error {
```
unmarshall: same as the above comment. This should not be called directly from the PCS validating webhook.

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,125 @@
...
+	if k8sClient == nil {
```
unmarshall: if this is nil, then it is a coding error and let nil pointer dereference happen. Do not check and return nil.

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil
+	}
+
+	queueName := strings.TrimSpace(queue)
```
unmarshall: choice of variable creates a confusion. `queue` only becomes a queueName when you trim the whitespace around it?

operator/internal/scheduler/volcano/queue.go
unmarshall: As commented earlier use `ValidatePodCliqueSet` method in `scheduler.Backend` interface.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -265,7 +266,11 @@ func (v *pcsValidator) validatePodCliqueTemplates(fldPath *field.Path) ([]string
 	allErrs = append(allErrs, sliceMustHaveUniqueElements(cliqueNames, fldPath.Child("name"))...)
 	allErrs = append(allErrs, sliceMustHaveUniqueElements(cliqueRoles, fldPath.Child("roleName"))...)
 
-	allErrs = append(allErrs, v.validateSchedulerNames(schedulerNames, fldPath)...)
+	schedulerErrs := v.validateSchedulerNames(schedulerNames, fldPath)
+	allErrs = append(allErrs, schedulerErrs...)
+	if len(schedulerErrs) == 0 && v.resolveSchedulerName(schedulerNames) == string(groveconfigv1alpha1.SchedulerNameVolcano) {
```
unmarshall: as commented earlier do not have dependency on specific scheduler backends. Instead call `scheduler.Backend.ValidatePodCliqueSet`

unmarshall (changes requested):
last set of review comments.

operator/e2e/yaml/volcano/negative-conflicting-queue.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: grove.io/v1alpha1
```
unmarshall: There are a 6 sample YAML files introduced as part of this PR, however i do not see where are these used in e2e tests. I also do not see e2e code changes. If you are going to make a separate PR for e2e tests then please remove these sample YAMLs from this PR and explicitly mention that e2e will come later.

Currently e2e deploys only KAI scheduler and tests revolve around that. This needs to change. So i would perhaps remove these files from this PR. We will first have to make changes to e2e so that we can accommodate other schedulers properly. Once that is done then you can raise another PR.

operator/internal/scheduler/volcano/backend.go
xianlubird: Queue validation is done in the PCS admission validator, not in the Volcano backend `ValidatePodCliqueSet` method shown here.

The call path is here:
https://github.com/ai-dynamo/grove/blob/25dc918d01d63036ba798dd0e14ff036ecf84565/operator/internal/webhook/admission/pcs/validation/podcliqueset.go#L269-L273

The actual queue validation is here:
https://github.com/ai-dynamo/grove/blob/25dc918d01d63036ba798dd0e14ff036ecf84565/operator/internal/webhook/admission/pcs/validation/podcliqueset.go#L298-L346

It validates:
- queue name format,
- PCS-level and PodClique-level queue conflicts,
- all PodCliques resolving to the same queue,
- queue existence,
- queue status being `Open`.

operator/internal/scheduler/volcano/queue.go
xianlubird: Done. I moved Volcano queue validation into the existing `scheduler.Backend.ValidatePodCliqueSet` path instead of keeping it in the generic PCS validator.

Volcano now performs both topology constraint validation and queue validation from its backend implementation:
https://github.com/ai-dynamo/grove/blob/78a06d3004551e370bcd7588a911f625ca0bdaa5/operator/internal/scheduler/volcano/backend.go#L191-L250

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
xianlubird: Done. I removed the direct Volcano scheduler dependency from the PodClique controllers.

The controllers now only clone the PodClique template annotations and no longer call `volcanoscheduler.MergePodCliqueAnnotations`:

https://github.com/ai-dynamo/grove/blob/e33da84aaa91c3d0d3b1a3a12b719d2d08571f14/operator/internal/controller/podcliqueset/components/podclique/podclique.go#L308-L315

https://github.com/ai-dynamo/grove/blob/e33da84aaa91c3d0d3b1a3a12b719d2d08571f14/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go#L316-L320

I also removed `MergePodCliqueAnnotations` and updated Volcano PodGroup sync to read the queue from the mirrored PodGang annotation directly:

https://github.com/ai-dynamo/grove/blob/e33da84aaa91c3d0d3b1a3a12b719d2d08571f14/operator/internal/scheduler/volcano/backend.go#L148-L151

operator/internal/controller/podcliqueset/components/podclique/podclique.go
xianlubird: Done

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
xianlubird: Done. I moved Volcano queue validation into the existing scheduler.Backend.ValidatePodCliqueSet path instead of keeping it in the generic PCS validator.

operator/api/config/v1alpha1/types.go
xianlubird: Done. Removed the Volcano-specific sentence from the `profiles` field docstring.

https://github.com/ai-dynamo/grove/blob/c275eb39b6bde2f72f6e98df20b68f21c53630be/operator/api/config/v1alpha1/types.go#L72-L77

operator/api/config/validation/validation_test.go
xianlubird: Done. I restored the unsupported scheduler-name negative case using `unknown-scheduler`, and kept Volcano covered as a supported scheduler in a separate valid case.

Volcano supported case:
https://github.com/ai-dynamo/grove/blob/56dec68ad77226c9f32292eb765921c18796687d/operator/api/config/validation/validation_test.go#L259-L268

Unsupported scheduler negative case:
https://github.com/ai-dynamo/grove/blob/56dec68ad77226c9f32292eb765921c18796687d/operator/api/config/validation/validation_test.go#L322-L334

operator/charts/templates/clusterrole.yaml
xianlubird: Done. I gated the Volcano RBAC rules on `.Values.config.scheduler.profiles[].name`, so the Volcano `podgroups` and `queues` permissions are only rendered when the `volcano` profile is enabled.

https://github.com/ai-dynamo/grove/blob/2527819e78e9549ce373873ace13124934d30150/operator/charts/templates/clusterrole.yaml#L168-L195

I did not use the exact `pluck "name" .Values.config.scheduler.profiles` snippet because `profiles` is a list, and that expression fails in the current Helm template rendering with a type error. This version keeps the check local to the Volcano RBAC block while preserving the same behavior.

operator/charts/values.yaml
xianlubird: Done

operator/charts/values.yaml
xianlubird: Done

operator/charts/values.yaml
xianlubird: Done

operator/charts/values.yaml
xianlubird: Done

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I removed the silent `eventRecorder == nil` guard.

`eventRecorder` is expected to be non-nil because the scheduler backends are initialized with `mgr.GetEventRecorderFor("scheduler-backend")`:

Now `recordCapabilityEvent` only skips when the event object is nil. If `eventRecorder` is unexpectedly nil while `obj` is non-nil, `eventRecorder.Eventf(...)` will panic instead of silently ignoring the coding error:

https://github.com/ai-dynamo/grove/blob/29fd4bc4ea7356375d1396f6f53b4a89dbc68b6d/operator/internal/scheduler/volcano/backend.go#L161-L165

operator/internal/scheduler/volcano/backend.go
xianlubird: Good catch. This was a real issue for coherent updates.

I updated the Volcano backend so it only re-syncs `minMember` and `subGroupPolicy` while the Grove PodGang still has positive `MinReplicas`. After coherent updates release the Grove scheduling constraints by setting all `MinReplicas` to 0, the backend preserves the existing Volcano PodGroup gang constraints instead of re-deriving them from the released values.

https://github.com/ai-dynamo/grove/blob/24c7ae0b05067a9113e63af903379a26eea8a25c/operator/internal/scheduler/volcano/backend.go#L148-L157

I also added a regression test that first creates the Volcano PodGroup from `2 + 3` min replicas, then updates the Grove PodGang to `0 + 0` and verifies the Volcano `minMember` and subgroup sizes are preserved.

operator/internal/scheduler/volcano/backend.go
xianlubird: Yes, agreed. This is no longer needed.

I removed the runtime PodClique fetch / queue conflict check from the Volcano backend. `SyncPodGang` now reads the effective queue directly from the PodGang annotations, since PodGang already carries the mirrored PCS metadata:

https://github.com/ai-dynamo/grove/blob/24c7ae0b05067a9113e63af903379a26eea8a25c/operator/internal/scheduler/volcano/backend.go#L152

The “all PodCliques must resolve to the same queue” check is now handled in the Volcano backend’s `ValidatePodCliqueSet` path during admission:

https://github.com/ai-dynamo/grove/blob/24c7ae0b05067a9113e63af903379a26eea8a25c/operator/internal/scheduler/volcano/backend.go#L187-L201

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I used the preferred approach and changed the common `scheduler.Backend` interface so `Init` now accepts a direct client:

https://github.com/ai-dynamo/grove/blob/61a4860d2554112d4fbcfcba142fd85960a2896f/operator/internal/scheduler/types.go#L41-L43

The direct/uncached client is now constructed only in the scheduler registry and passed into each backend from `newSchedulerBackend`:

https://github.com/ai-dynamo/grove/blob/61a4860d2554112d4fbcfcba142fd85960a2896f/operator/internal/scheduler/registry/registry.go#L44-L57

https://github.com/ai-dynamo/grove/blob/61a4860d2554112d4fbcfcba142fd85960a2896f/operator/internal/scheduler/registry/registry.go#L96-L115

I also updated the other scheduler backends (`kube`, `kai`) and test fakes to implement the new `Init(directClient client.Client)` signature. They do not need the direct client today, so they ignore it, while Volcano uses it for the capability check:

https://github.com/ai-dynamo/grove/blob/61a4860d2554112d4fbcfcba142fd85960a2896f/operator/internal/scheduler/volcano/backend.go#L68-L70

operator/internal/scheduler/volcano/backend.go
xianlubird: Agreed. I removed the package-level mutable test hook.

The Volcano package no longer exposes `SetCapabilityClientFactoryForTest` / `newCapabilityClient`. The direct client is passed through `Backend.Init` from the scheduler registry instead.

I also removed the remaining mutable package-level factory from the registry implementation. `newDirectClient` is now a normal function, and tests inject a fake direct-client factory through a private `newRegistry(..., createDirectClient)` helper instead of mutating global state.

operator/internal/scheduler/volcano/queue.go
xianlubird: Done

operator/internal/scheduler/volcano/queue.go
xianlubird: Yes, the PodGang runtime path no longer needs this helper. `SyncPodGang` now reads the queue directly from `podGang.Annotations`.

This helper is still used by the validating webhook path, before PodGang is created, to resolve and validate the effective queue from PCS-level and PodClique-level annotations. It catches cases where a clique-level queue conflicts with the PCS metadata queue, or where different cliques resolve to different queues. After this admission check passes, the PodGang annotation can be treated as the single source of truth.

operator/internal/scheduler/volcano/queue.go
xianlubird: Done. I kept scheduler-specific queue validation behind the Volcano backend entry point.

`ValidatePodCliqueSet` remains the only backend method called by the PCS validating webhook, and the Volcano queue checks are now invoked from that backend method:

https://github.com/ai-dynamo/grove/blob/9e33d33b1222506fefe9795841daed967b0724fc/operator/internal/scheduler/volcano/backend.go#L156-L170

https://github.com/ai-dynamo/grove/blob/9e33d33b1222506fefe9795841daed967b0724fc/operator/internal/scheduler/volcano/backend.go#L173-L215

I also made the queue validation helpers package-private so they are not exposed as functions for the PCS validator or other packages to call directly:

https://github.com/ai-dynamo/grove/blob/9e33d33b1222506fefe9795841daed967b0724fc/operator/internal/scheduler/volcano/queue.go#L42-L83

operator/internal/scheduler/volcano/queue.go
xianlubird: Done as part of the same cleanup. This helper is now package-private and is only called from the Volcano backend’s `ValidatePodCliqueSet` validation path, not directly from the PCS validating webhook.

https://github.com/ai-dynamo/grove/blob/9e33d33b1222506fefe9795841daed967b0724fc/operator/internal/scheduler/volcano/queue.go#L81-L83

https://github.com/ai-dynamo/grove/blob/9e33d33b1222506fefe9795841daed967b0724fc/operator/internal/scheduler/volcano/backend.go#L212-L215

operator/internal/scheduler/volcano/queue.go
xianlubird: Done. I removed the nil-client guard from this helper.

If `k8sClient` is nil here, that is a coding error in the Volcano backend validation path, so it should fail loudly instead of silently skipping queue validation.

operator/internal/scheduler/volcano/queue.go
xianlubird: Done. I stopped trimming the queue value in this helper.

The helper now uses the resolved `queue` value directly instead of creating a trimmed `queueName`.

https://github.com/ai-dynamo/grove/blob/b7b5b048a997884aa367f6efdfb06a7245785c98/operator/internal/scheduler/volcano/queue.go#L76-L100

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
xianlubird: Done

operator/e2e/yaml/volcano/negative-conflicting-queue.yaml
xianlubird: Done

xianlubird:
Thank you @unmarshall  and @sanjaychatterjee  for all the suggestions. This PR has a lot of code, making the review process very complex. Thank you for your patient and detailed review. I have modified the code based on your suggestions. Please take another look when you have time. Thank you.

operator/charts/templates/clusterrole.yaml
unmarshall: I would do it differently. The check for a scheduler profile needs to be done across all supported backend schedulers. So do the following:

Introduce a function in _helpers.tpl
```helm
{{/* Returns "true" if the named scheduler profile is configured, empty string otherwise. Usage: {{- if include "grove.scheduler.hasProfile" (list . "volcano") }} */}}
{{- define "grove.scheduler.hasProfile" -}}
{{- $root := index . 0 -}}
{{- $profile := index . 1 -}}
{{- range $root.Values.config.scheduler.profiles -}}
{{- if eq .name $profile -}}true{{- end -}}
{{- end -}}
{{- end -}}
```

Then in the clusterrole.yaml do the following:
```yaml
# This is for KAI scheduler
{{- if include "grove.scheduler.hasProfile" (list . "kai-scheduler") }}
- apiGroups:
  - kai.scheduler
  resources:
  - topologies
  verbs:
  - create
  - get
  - list
  - watch
  - patch
  - update
  - delete
{{- end }}
# This is for volcano scheduler
{{- if include "grove.scheduler.hasProfile" (list . "volcano") }}
- apiGroups:
    - scheduling.volcano.sh
  resources:
    - podgroups
  verbs:
    - create
    - get
    - list
    - watch
    - patch
    - update
    - delete
- apiGroups:
    - scheduling.volcano.sh
  resources:
    - queues
  verbs:
    - get
    - list
    - watch
- apiGroups:
    - apiextensions.k8s.io
  resources:
    - customresourcedefinitions
  resourceNames:
    - podgroups.scheduling.volcano.sh
  verbs:
    - get
```

operator/internal/scheduler/registry/registry.go
```diff
@@ -39,13 +41,34 @@ type registry struct {
 	defaultBackend scheduler.Backend
 }
 
+type directClientFactory func(*runtime.Scheme) (client.Client, error)
+
+// newDirectClient returns a non-cached client for backend initialization.
+// Backend Init runs before the manager cache starts, so backends that read
+// cluster state during Init must not use the manager's cached client.
+func newDirectClient(scheme *runtime.Scheme) (client.Client, error) {
```
unmarshall: move all non-exposed functions after the exposed ones.

operator/internal/scheduler/registry/registry.go
```diff
@@ -39,13 +41,34 @@ type registry struct {
 	defaultBackend scheduler.Backend
 }
 
+type directClientFactory func(*runtime.Scheme) (client.Client, error)
```
unmarshall: do you really need a factory?

operator/internal/scheduler/registry/registry.go
```diff
@@ -39,13 +41,34 @@ type registry struct {
...
+}
+
 // New creates a scheduler.Registry by initializing backend instances for each
 // profile in cfg.Profiles.
 // NOTE: This function should be called once during the lifecycle of the Grove operator.
 func New(cl client.Client, scheme *runtime.Scheme, eventRecorder record.EventRecorder, cfg configv1alpha1.SchedulerConfiguration) (scheduler.Registry, error) {
+	return newRegistry(cl, scheme, eventRecorder, cfg, newDirectClient)
+}
+
+func newRegistry(cl client.Client, scheme *runtime.Scheme, eventRecorder record.EventRecorder, cfg configv1alpha1.SchedulerConfiguration, createDirectClient directClientFactory) (scheduler.Registry, error) {
```
unmarshall: you do not need this function. All of its code goes into New which also accepts. So the code becomes:

```go
func New(cl, directClient client.Client, scheme *runtime.Scheme, eventRecorder record.EventRecorder, cfg configv1alpha1.SchedulerConfiguration) (scheduler.Registry, error) {
      reg := &registry{backends: make(map[string]scheduler.Backend)}
      for _, p := range cfg.Profiles {
          backend, err := newSchedulerBackend(cl, directClient, scheme, eventRecorder, p)
          if err != nil {
              return nil, fmt.Errorf("failed to initialize %s backend: %w", p.Name, err)
          }
          reg.backends[backend.Name()] = backend
          if string(p.Name) == cfg.DefaultProfileName {
              reg.defaultBackend = backend
          }
      }
      return reg, nil
  }
```

unmarshall: The tests can still pass fake client as direct client.

operator/internal/scheduler/registry/registry.go
```diff
@@ -76,23 +99,26 @@ func (r *registry) GetOrDefault(name string) scheduler.Backend {
 
 // newSchedulerBackend creates and initializes a Backend for the given profile.
 // NOTE: For any newly supported backend, add a case for it in the switch statement.
-func newSchedulerBackend(cl client.Client, scheme *runtime.Scheme, rec record.EventRecorder, p configv1alpha1.SchedulerProfile) (scheduler.Backend, error) {
+func newSchedulerBackend(cl client.Client, scheme *runtime.Scheme, rec record.EventRecorder, p configv1alpha1.SchedulerProfile, createDirectClient directClientFactory) (scheduler.Backend, error) {
```
unmarshall: no need to pass `directClientFactory`. Keep things simple.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+	return b.name
+}
+
+func (b *schedulerBackend) Init(directClient client.Client) error {
+	return CheckCapability(directClient, b.eventRecorder)
+}
+
+// CheckCapability verifies that the installed Volcano PodGroup CRD supports
+// the subgroup policy fields Grove needs for gang scheduling.
+func CheckCapability(cl client.Client, eventRecorder record.EventRecorder) error {
```
unmarshall: why is this function exported? As far as i can see its used within the same package.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+		eventRecorder: eventRecorder,
+		profile:       profile,
+	}
+}
+
+func (b *schedulerBackend) Name() string {
+	return b.name
+}
+
+func (b *schedulerBackend) Init(directClient client.Client) error {
```
unmarshall, line 66: I see no need for `CheckCapability` and `checkCapability`.

The Init method implementation is straight forward:
```go
func (b *schedulerBackend) Init(directClient client.Client) error {
      crd := &apiextensionsv1.CustomResourceDefinition{
          ObjectMeta: metav1.ObjectMeta{Name: volcanoPodGroupCRDName},
      }
      if err := directClient.Get(context.Background(), client.ObjectKey{Name: volcanoPodGroupCRDName}, crd); err != nil {
          err = fmt.Errorf("volcano scheduler backend requires Volcano 1.14 or newer with PodGroup.spec.subGroupPolicy: failed to get CRD %q: %w", volcanoPodGroupCRDName, err)
          recordCapabilityEvent(b.eventRecorder, crd, err)
          return err
      }
      if !podGroupCRDHasSubGroupPolicy(crd) {
          err := fmt.Errorf("volcano scheduler backend requires Volcano 1.14 or newer with PodGroup.spec.subGroupPolicy on scheduling.volcano.sh/v1beta1 PodGroup")
          recordCapabilityEvent(b.eventRecorder, crd, err)
          return err
      }
      return nil
  }
```

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+// the subgroup policy fields Grove needs for gang scheduling.
+func CheckCapability(cl client.Client, eventRecorder record.EventRecorder) error {
+	crd, err := checkCapability(cl)
+	if err != nil {
+		recordCapabilityEvent(eventRecorder, crd, err)
+	}
+	return err
+}
+
+func checkCapability(cl client.Client) (*apiextensionsv1.CustomResourceDefinition, error) {
```
unmarshall: The split between `CheckCapability` and `checkCapability` is artificial and is not really required.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
+	}
+	pod.Annotations[volcanov1beta1.VolcanoGroupNameAnnotationKey] = podGangName
+	pod.Annotations[volcanov1beta1.KubeGroupNameAnnotationKey] = podGangName
+}
+
+func (b *schedulerBackend) ValidatePodCliqueSet(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error {
+	if pcs.Spec.Template.TopologyConstraint != nil {
```
unmarshall, line 147: since you have created a separate function `validateQueueAnnotations` that validates queue configuration. You can also create a separate function to validate topology constraints and then call the two functions from within `ValidatePodCliqueSet`

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	for _, pcsg := range pcs.Spec.Template.PodCliqueScalingGroupConfigs {
+		if pcsg.TopologyConstraint != nil {
+			return fmt.Errorf("volcano scheduler backend does not support topologyConstraint on PodCliqueScalingGroup %q", pcsg.Name)
+		}
+	}
+	return b.validateQueueAnnotations(ctx, pcs)
+}
+
+func (b *schedulerBackend) validateQueueAnnotations(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error {
```
unmarshall: since you have a separate file `queue.go` i would move it as a function and pass the client to it.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+	globalQueueValue := strings.TrimSpace(pcs.Annotations[QueueAnnotationKey])
+	if globalQueueValue != "" {
+		if msgs := validateQueueName(globalQueueValue); len(msgs) > 0 {
+			errs = append(errs, fmt.Errorf("metadata.annotations[%s]: %s", QueueAnnotationKey, strings.Join(msgs, "; ")))
+		}
+	}
+
+	var resolvedQueue string
+	for i, cliqueTemplateSpec := range pcs.Spec.Template.Cliques {
+		if cliqueTemplateSpec == nil {
```
unmarshall: this can never be nil

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+			continue
+		}
+		cliqueQueueValue := strings.TrimSpace(cliqueTemplateSpec.Annotations[QueueAnnotationKey])
+		if cliqueQueueValue != "" {
+			if msgs := validateQueueName(cliqueQueueValue); len(msgs) > 0 {
+				errs = append(errs, fmt.Errorf("spec.template.cliques[%d].annotations[%s]: %s", i, QueueAnnotationKey, strings.Join(msgs, "; ")))
+			}
+		}
+
+		queue, err := resolvePodCliqueQueue(pcs.Annotations, cliqueTemplateSpec.Annotations)
```
unmarshall: when you call `resolvePodCliqueQueue` then at this point you have already retrieved PCS level queue annotation value and PCLQ level queue annotation value. Then what is the need to send in all the PCS and PCLQ annotations again and do this exercise again?

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+		if pcsg.TopologyConstraint != nil {
+			return fmt.Errorf("volcano scheduler backend does not support topologyConstraint on PodCliqueScalingGroup %q", pcsg.Name)
+		}
+	}
+	return b.validateQueueAnnotations(ctx, pcs)
+}
+
+func (b *schedulerBackend) validateQueueAnnotations(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error {
+	var errs []error
+	globalQueueValue := strings.TrimSpace(pcs.Annotations[QueueAnnotationKey])
```
unmarshall: instead of `globalQueueValue` it should be called `pcsQueue` since in the code following this you are introducing `resolveQueue`. Have one convention.

operator/internal/scheduler/volcano/queue.go
```diff
@@ -0,0 +1,101 @@
...
+
+	switch {
+	case globalQueue == "" && cliqueQueue == "":
+		return DefaultQueue, nil
+	case globalQueue == "":
+		return cliqueQueue, nil
+	case cliqueQueue == "":
+		return globalQueue, nil
+	case globalQueue != cliqueQueue:
+		return "", errConflictingQueueAnnotations
```
unmarshall: the correct way to use sentinel error is to wrap it.
```go
fmt.Errorf("%w: PodCliqueSet has %q, PodClique has %q", errConflictingQueueAnnotations, globalQueue, cliqueQueue)
```
Now in the tests you can use `errors.Is` and there is no need to check the error msg which can change overtime.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func recordCapabilityEvent(eventRecorder record.EventRecorder, obj runtime.Object, err error) {
+	if obj == nil {
+		return
+	}
+	eventRecorder.Eventf(obj, corev1.EventTypeWarning, "VolcanoCapabilityUnavailable", "%v", err)
+}
+
+func (b *schedulerBackend) OnPodGangDelete(_ context.Context, _ *groveschedulerv1alpha1.PodGang) error {
```
unmarshall: I checked with @enoodle and he fine for removing `OnPodGangDelete`.  Please remove this method from `TopologyAwareBackend` interface and remove these dummy implementations from volcano, kai and default scheduler backends.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+	return b.validateQueueAnnotations(ctx, pcs)
+}
+
+func (b *schedulerBackend) validateQueueAnnotations(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error {
+	var errs []error
+	globalQueueValue := strings.TrimSpace(pcs.Annotations[QueueAnnotationKey])
+	if globalQueueValue != "" {
+		if msgs := validateQueueName(globalQueueValue); len(msgs) > 0 {
```
unmarshall: If there is a queue annotation defined on the PCS but its value is not a valid. What is done in the code is that it will record an error but it will proceed and use that value in `resolvePodCliqueQueue`. If no clique level annotation is defined then invalid value is also returned from this function.

I am not sure what should be the behavior here.
Option-1: drop the PCS level queue value and do not consider it when resolving PCLQ queue value.
Option-2: return early when PCS level queue has invalid queue value.

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,278 @@
...
+			resolvedQueue = queue
+			continue
+		}
+		if resolvedQueue != queue {
+			errs = append(errs, fmt.Errorf("spec.template.cliques[%d].annotations[%s]: all PodCliques in a PodCliqueSet using volcano scheduler must resolve to the same queue", i, QueueAnnotationKey))
+		}
+	}
+
+	if len(pcs.Spec.Template.Cliques) == 0 {
+		resolvedQueue = effectiveQueueFromAnnotations(pcs.Annotations)
```
unmarshall: I am not sure why is this check done here. You could have done this right before you started to iterate the PCLQs

unmarshall (changes requested):

operator/internal/scheduler/registry/registry.go
xianlubird: Done. I moved the unexported helper functions after the exported constructor and methods in `registry.go`.

`New`, `Get`, `GetDefault`, `GetOrDefault`, `All`, and `AllTopologyAware` now appear before the package-private helpers like `newRegistry`, `newSchedulerBackend`, and `newDirectClient`.

operator/internal/scheduler/registry/registry.go
xianlubird: Done. I removed the factory.

`New` now creates the direct client once and passes the resulting `client.Client` through `newRegistry` and `newSchedulerBackend`. The registry tests now pass the fake client directly as the direct client, so the extra factory type and test helper are no longer needed.

operator/internal/scheduler/registry/registry.go
xianlubird: Done. I removed `newRegistry` and moved its initialization logic into `New`.

`New` now accepts both the runtime client and the direct client. The runtime client is still passed into each backend constructor, while the direct client is only passed into backend `Init`.

In the operator entrypoint, this means `mgr.GetClient()` remains the backend runtime client, and the pre-start direct client is used for backend initialization. The registry tests pass the fake client for both arguments.

operator/internal/scheduler/registry/registry.go
xianlubird: Done. The factory has been removed.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I made the capability check package-private.

`CheckCapability` is now `checkCapability`, since it is only used inside the Volcano package. I also renamed the lower-level CRD-loading helper to `loadCapabilityCRD` to keep the two responsibilities separate.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I removed the extra capability-check helpers and moved the logic directly into `Init`.

`Init` now reads the Volcano PodGroup CRD with the direct client, validates that `spec.subGroupPolicy` is available, records a capability event on failure, and returns the initialization error directly.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I split the topology-constraint validation into a separate `validateTopologyConstraints` helper.

`ValidatePodCliqueSet` now just coordinates the scheduler-specific checks by calling `validateTopologyConstraints` first and then `validateQueueAnnotations`.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I moved `validateQueueAnnotations` into `queue.go`.

It is now a package-private function that receives the Kubernetes client explicitly, and `ValidatePodCliqueSet` calls it as `validateQueueAnnotations(ctx, b.client, pcs)`.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I removed the unnecessary nil checks.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I changed `resolvePodCliqueQueue` to take the already-resolved queue values instead of the full annotation maps.

`validateQueueAnnotations` now reads and trims the PCS-level and PodClique-level queue annotation values once, then passes those values directly to `resolvePodCliqueQueue`.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I renamed the queue variables for consistency.

`globalQueueValue` is now `pcsQueue`, and `cliqueQueueValue` is now `cliqueQueue`, matching the `resolvedQueue` naming style used later in the function.

operator/internal/scheduler/volcano/queue.go
xianlubird: Done. I now wrap the sentinel error instead of returning it directly.

`resolvePodCliqueQueue` returns a contextual error using `%w`, and the validation path preserves that wrapping. The conflict test now uses `require.ErrorIs(err, errConflictingQueueAnnotations)` instead of checking a mutable error message.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. This is a slightly larger cleanup because it changes the common scheduler backend interface.

I removed `OnPodGangDelete` from `scheduler.Backend`, removed the dummy implementations from the kube, kai, volcano, and fake backends, and updated the PodGang reconciler to return directly when a PodGang is being deleted.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I went with Option-2.

If the PCS-level queue annotation is invalid, validation now returns immediately and does not continue resolving PodClique queues or checking the Volcano Queue object. I also added a regression test to make sure the invalid PCS-level queue path does not fall through into the queue existence check.

operator/internal/scheduler/volcano/backend.go
xianlubird: Done. I moved the empty-cliques handling before the PodClique iteration.

When there are no PodCliques, validation now immediately resolves the effective queue from the PCS annotations and checks that queue, instead of handling that case after the loop.

operator/charts/templates/clusterrole.yaml
unmarshall: Can you make this change?

operator/internal/client/scheme.go
```diff
@@ -37,6 +39,8 @@ func init() {
 		grovecorev1alpha1.AddToScheme,
 		schedv1alpha1.AddToScheme,
 		kaitopologyv1alpha1.AddToScheme,
+		volcanov1beta1.AddToScheme,
```
unmarshall, line 42: While this is not so pressing, i would like to avoid adding scheduler specific schemes here. This is scheduler agnostic code. The scheduler specific scheme are only used within the scheduler backend implementation and no where else in this code base, whereas the scheme initialised here is used across the code base.
This is the same for KAI scheduler. For now we can leave it here but please raise a separate issue to clean this up.

operator/internal/controller/podcliqueset/components/podclique/podclique.go
```diff
@@ -310,6 +310,9 @@ func (r _resource) buildResource(logger logr.Logger, pclq *grovecorev1alpha1.Pod
 	pclq.Labels = getLabels(pcs, pcsReplica, pclqObjectKey, pclqTemplateSpec, apicommon.GeneratePodGangNameForPodCliqueOwnedByPodCliqueSet(pcs, pcsReplica))
 	pclq.Annotations = maps.Clone(pclqTemplateSpec.Annotations)
 	delete(pclq.Annotations, apiconstants.AnnotationTopologyName)
+	if len(pclq.Annotations) == 0 {
```
unmarshall, line 314: Is this required? What is the consequence if we do not have this code block?

operator/internal/controller/podgang/reconciler.go
```diff
@@ -77,10 +77,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	logger := log.FromContext(ctx).WithValues("scheduler", backend.Name(), "podGang", req.NamespacedName)
 	if !podGang.DeletionTimestamp.IsZero() {
 		logger.Info("PodGang is being deleted")
```
unmarshall: is this a useful log or it just creates noise in the logs?

operator/internal/scheduler/registry/registry.go
```diff
@@ -112,3 +92,23 @@ func (r *registry) AllTopologyAware() map[string]scheduler.TopologyAwareBackend
 	}
 	return result
 }
+
+// newSchedulerBackend creates and initializes a Backend for the given profile.
+// NOTE: For any newly supported backend, add a case for it in the switch statement.
+func newSchedulerBackend(cl client.Client, scheme *runtime.Scheme, rec record.EventRecorder, p configv1alpha1.SchedulerProfile, directClient client.Client) (scheduler.Backend, error) {
```
unmarshall: can you re-arrange the function parameters and keep both `client` and `directClient` one after another instead of putting `directClient` towards the end?

operator/internal/scheduler/volcano/backend.go
```diff
@@ -0,0 +1,222 @@
...
+	if obj == nil {
+		return
+	}
+	eventRecorder.Eventf(obj, corev1.EventTypeWarning, "VolcanoCapabilityUnavailable", "%v", err)
+}
+
+func (b *schedulerBackend) PreparePod(pod *corev1.Pod) {
+	pod.Spec.SchedulerName = b.Name()
+	podGangName := pod.Labels[apicommon.LabelPodGang]
+	if podGangName == "" {
```
unmarshall, line 128: can the podGangName be empty?

operator/charts/templates/clusterrole.yaml
xianlubird: Done

operator/internal/client/scheme.go
xianlubird: Created a follow-up issue for this cleanup: https://github.com/ai-dynamo/grove/issues/668

operator/internal/controller/podcliqueset/components/podclique/podclique.go
xianlubird: This block is still needed.
`PodClique` annotations are copied to Pods when the Pod resources are built. The topology annotation should be owned by `PodGang`, because the `PodGang` controller resolves whether topology-aware scheduling is enabled and what topology name should be used.
If we keep the template `topology-name` annotation on `PodClique`, it can be propagated down to Pods even when the resolved `PodGang` topology state says otherwise. I added a short comment to make that ownership clear.

operator/internal/controller/podgang/reconciler.go
xianlubird: Agreed, this log does not add much signal. I removed it and kept the early return for PodGangs that are already being deleted.

operator/internal/scheduler/registry/registry.go
xianlubird: Done. I reordered the `newSchedulerBackend` parameters so the cached client and direct client are adjacent.

operator/internal/scheduler/volcano/backend.go
xianlubird: `podGangName` should not be empty here.

unmarshall: But if it is empty we are silently returning here. Is that fine?

xianlubird: I updated `PreparePod` to return an error instead of silently returning when required scheduler metadata is missing.

The common `scheduler.Backend` interface now exposes `PreparePod(pod) error`. The Pod builder propagates that error with context, and the Volcano backend now returns an explicit error if the required `grove.io/podgang` label is missing.

The kube and KAI backends do not require this label, so they simply return nil.

unmarshall (approved):
Thanks for making all the changes. Looks good now.

sanjaychatterjee (approved):

PR #523, opened 2026-04-13
feat: GREP-0368 Preferred Topology Constraint API

### LINKED ISSUE #368: Support API for requesting Preferred topology constraint

Grove should expose an API for requesting preferred topology constraint at every level:
- PodClique
- PodGangScalingGroup
- PodCliqueSet

Currently only required is available via `packDomain` param

### Why is this needed?

Required might fail the workload completely, while at some scenarios user will only want best effort even if placement will eventually be sub-optimal without failing the workload. It also compatible with other schedulers such as Kueue.
The reason preferred can't apply by default by the scheduler and requires the user to request it - is because it is very compute intensive algorithm.

Specifically for the flow of submission of Dynamo over Grove from Run:ai, this will help as the system automatically applies preferred constraints to all the workloads submitted, to abstract this from the end users.

shayasoolin:
## Summary

- Proposes adding a nested `topologyConstraint.pack` API for topology packing constraints
- Adds `topologyConstraint.pack.preferred` as a best-effort topology packing domain
- Replaces the required packing field `topologyConstraint.packDomain` with `topologyConstraint.pack.required`
- Keeps `topologyConstraint.packDomain` as a deprecated compatibility field for existing workloads

## Key design decisions

- Preferred placement is workload opt-in through `topologyConstraint.pack.preferred`; Grove does not apply it by default
- `topologyConstraint.pack.required` preserves the existing hard scheduling semantics of `packDomain`
- Existing workloads that already use `packDomain` continue to be honored
- New workloads using `packDomain` are rejected and should use `pack.required`
- Existing workloads may migrate from `packDomain` to `pack.required` on update when the topology domain value is unchanged
- `topologyName` is assumed to follow the inherited topology-name behavior from the preceding topology API changes

References #368

### DIFF
--- a/docs/proposals/0368-preferred-topology-constraint/README.md
+++ b/docs/proposals/0368-preferred-topology-constraint/README.md
@@ -60,7 +60,6 @@ As a **Run:ai platform operator** deploying Dynamo workloads over Grove, I want 
 ### Limitations/Risks & Mitigations
 
 - **Scheduler performance:** `packDomainPreferred` triggers a scheduling algorithm that might be compute-intensive. Misuse (e.g. setting it on all workloads at scale) could degrade scheduler performance. *Mitigation: it's an explicit opt-in, never a default.*
-- **Interaction with `packDomain`:** The semantics when both fields are set need to be well-defined.
 
 ## Design Details
 
--- a/docs/proposals/0368-preferred-topology-constraint/README.md
+++ b/docs/proposals/0368-preferred-topology-constraint/README.md
@@ -34,7 +34,7 @@ Grove topology-aware scheduling currently supports required packing through `pac
 
 ## Motivation
 
-Grove is increasingly used as an orchestration layer by higher-level tools (e.g. Dynamo on Run:ai) that submit workloads on behalf of users. These tools often want topology-aware placement to improve workload performance, but the existing workload-facing constraint, `packDomain`, is required: if the requested domain cannot fit the workload, scheduling may remain blocked.
+Grove is increasingly used as an orchestration layer by higher-level tools that submit workloads on behalf of users. These tools often want topology-aware placement to improve workload performance, but the existing workload-facing constraint, `packDomain`, is required: if the requested domain cannot fit the workload, scheduling may remain blocked.
 
 Preferred topology placement addresses this gap by letting the scheduler try for a more optimized placement while still allowing the workload to run when the preference cannot be satisfied. This is useful for platforms that want to apply topology-aware optimization transparently, and for workloads that benefit from locality but should not fail or remain pending solely because the most optimized placement is unavailable.
 
@@ -135,7 +135,7 @@ As a cluster administrator, I want to configure the preferred packing domain on 
 
 #### Story 4: External orchestration platform
 
-As an orchestration platform, such as Dynamo on Run:ai, I want to set `topologyName` when submitting Grove workloads so that topology-aware placement is attempted transparently for users while preserving scheduling fallback behavior.
+As an orchestration platform, I want to set `topologyName` when submitting Grove workloads so that topology-aware placement is attempted transparently for users while preserving scheduling fallback behavior.
 
 ### Limitations/Risks & Mitigations
 
@@ -233,7 +233,7 @@ Existing topology health signals remain relevant:
 - Unit and e2e tests pass
 
 #### Beta
-- Validated in at least one production workload or platform integration, such as Dynamo on Run:ai
+- Validated in at least one production workload or platform integration
 - No breaking API changes since alpha
 - No significant scheduler performance regressions observed from opted-in workloads
 
--- a/docs/proposals/0368-preferred-topology-constraint/README.md
+++ b/docs/proposals/0368-preferred-topology-constraint/README.md
@@ -1,0 +1,156 @@
+# GREP-0368: Preferred Topology Constraint API
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Automated preferred topology in Dynamo on Run:ai](#story-1-automated-preferred-topology-in-dynamo-on-runai)
+  - [Limitations/Risks &amp; Mitigations](#limitationsrisks--mitigations)
+- [Design Details](#design-details)
+  - [API Change](#api-change)
+  - [PodGang Propagation](#podgang-propagation)
+  - [Webhook Validation](#webhook-validation)
+  - [Monitoring](#monitoring)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+- [Alternatives](#alternatives)
+  - [Always-on preferred topology at the lowest level](#always-on-preferred-topology-at-the-lowest-level)
+  - [Boolean opt-in on <code>PodCliqueSet</code>](#boolean-opt-in-on-podcliqueset)
+<!-- /toc -->
+
+## Summary
+
+Grove currently supports topology-aware scheduling through a required constraint (`packDomain`), which fails the workload entirely if the requested topology cannot be satisfied. This GREP introduces a **preferred** topology constraint — a best-effort mode that requests topology-aware placement without making it a hard requirement. If the constraint cannot be met, the workload is still scheduled rather than failing, making topology optimization accessible in scenarios where strict placement is undesirable or incompatible with the scheduler.
+
+## Motivation
+
+Grove is increasingly used as an orchestration layer by higher-level tools (e.g. Dynamo on Run:ai) that manage workload submission on behalf of users. These tools often want to apply topology preferences automatically — without exposing scheduling decisions to end users — but cannot do so today because the only available topology constraint is required (`packDomain`), which risks failing workloads when the topology cannot be satisfied.
+
+Additionally, preferred topology scheduling is a compute-intensive operation for the scheduler. It therefore cannot be enabled by default for all workloads — it must be an explicit opt-in. Exposing a preferred constraint API gives tools and users a way to request best-effort topology placement, decoupled from the all-or-nothing semantics of the current required constraint.
+
+### Goals
+
+- Introduce a preferred (best-effort) topology constraint API on `PodClique`, `PodGangScalingGroup`, and `PodCliqueSet`
+- Ensure workloads with a preferred constraint are never blocked — if topology cannot be satisfied, scheduling proceeds without it
+
+### Non-Goals
+
+- Automatically applying preferred topology to all workloads (it remains an explicit opt-in)
+- Changing the semantics of the existing required constraint (`packDomain`)
+- Guaranteeing topology-optimal placement — preferred is best-effort only
+
+## Proposal
+
+This GREP proposes adding a `packDomainPreferred` field alongside the existing `packDomain` field on `PodClique`, `PodGangScalingGroup`, and `PodCliqueSet`. When set, it signals to the scheduler that topology-aware placement is desired at the specified domain level on a best-effort basis — the workload is never failed if the constraint cannot be satisfied.
+
+`packDomainPreferred` and `packDomain` can coexist on the same resource. For example, a user may set `packDomainPreferred=host` and `packDomain=rack` to express that host-level packing is desired but not required, while rack-level packing is mandatory. When both are set, the scheduler attempts preferred placement first and falls back toward the required level if the preferred constraint cannot be satisfied. It is recommended (but not enforced) that `packDomainPreferred` be stricter than or equal to `packDomain` — setting a coarser preferred level than the required one is semantically redundant.
+
+### User Stories
+
+#### Story 1: Automated preferred topology in Dynamo on Run:ai
+
+As a **Run:ai platform operator** deploying Dynamo workloads over Grove, I want to automatically apply a preferred topology constraint to all submitted workloads — without requiring end users to configure scheduling parameters — so that topology-aware placement is attempted as a best-effort optimization while ensuring no workload is ever blocked due to an unsatisfiable constraint.
+
+### Limitations/Risks & Mitigations
+
+- **Scheduler performance:** `packDomainPreferred` triggers a scheduling algorithm that might be compute-intensive. Misuse (e.g. setting it on all workloads at scale) could degrade scheduler performance. *Mitigation: it's an explicit opt-in, never a default.*
+- **Interaction with `packDomain`:** The semantics when both fields are set need to be well-defined.
+
+## Design Details
+
+### API Change
+
+A single field is added to the existing `TopologyConstraint` struct in the operator API (`operator/api/core/v1alpha1/podcliqueset.go`), which is already shared across `PodCliqueSet`, `PodCliqueTemplateSpec`, and `PodCliqueScalingGroupConfig`:
+
+```go
+// TopologyConstraint defines topology placement requirements.
+type TopologyConstraint struct {
+    // PackDomain specifies the required topology domain for grouping replicas.
+    // The workload will not be scheduled if this constraint cannot be satisfied.
+    // +kubebuilder:validation:Enum=region;zone;datacenter;block;rack;host;numa
+    // +optional
+    PackDomain *TopologyDomain `json:"packDomain,omitempty"`
+
+    // PackDomainPreferred specifies a preferred (best-effort) topology domain.
+    // If the constraint cannot be satisfied, the workload is scheduled anyway.
+    // When set alongside PackDomain, it is recommended that PackDomainPreferred
+    // be stricter than or equal to PackDomain.
+    // +kubebuilder:validation:Enum=region;zone;datacenter;block;rack;host;numa
+    // +optional
+    PackDomainPreferred *TopologyDomain `json:"packDomainPreferred,omitempty"`
+}
+```
+
+Note: `PackDomain` changes from a value type to a pointer (`*TopologyDomain`) to allow a `TopologyConstraint` with only `PackDomainPreferred` set. This is not a breaking API change — existing resources with `packDomain` set continue to work. Controller code accessing `PackDomain` will need nil checks.
+
+### PodGang Propagation
+
+The scheduler API (`scheduler/api/core/v1alpha1/podgang.go`) already supports both constraints via `TopologyPackConstraint.Required` and `TopologyPackConstraint.Preferred`. No changes are needed on the scheduler side.
+
+The only required implementation change is in `createTopologyPackConstraint` (`operator/internal/controller/podcliqueset/components/podgang/syncflow.go`), which translates Grove's `TopologyConstraint` into the scheduler's `TopologyPackConstraint`. It currently only populates `Required`. It must be extended to also resolve `PackDomainPreferred` to its topology key and populate `Preferred`.
+
+The domain-to-key translation follows the same existing path: looking up the domain name in the `ClusterTopology` levels to obtain the corresponding node label key. If the preferred domain is not found in the current `ClusterTopology`, it is silently skipped (same behavior as the existing required constraint).
+
+### Webhook Validation
+
+The admission webhook (`operator/internal/webhook/admission/pcs/validation/topologyconstraints.go`) validates that child constraints are stricter than or equal to parent constraints. This validation must be extended to cover `PackDomainPreferred` with the same hierarchy rules.
+
+### Monitoring
+
+There is no scheduler-side feedback on whether preferred placement was achieved. This is consistent with the existing behavior for `packDomain`, which also has no placement-outcome indicator.
+
+The existing `TopologyLevelUnavailable` status condition on `PodCliqueSet`, which checks whether referenced topology domains exist in the `ClusterTopology` CR, will be extended to also cover `packDomainPreferred` domains.
+
+### Test Plan
+
+**Unit — webhook validation:**
+- `packDomainPreferred` domain not present in `ClusterTopology` is rejected
+- `packDomainPreferred` stricter than parent is accepted; coarser passes without enforcement
+- `packDomainPreferred` can be set without `packDomain`, and vice versa
+
+**Unit — syncflow (`TestComputeExpectedPodGangsWithTopologyConstraints`):**
+- `packDomainPreferred` alone → `TopologyPackConstraint.Preferred` set, `Required` nil
+- Both fields set → both `Preferred` and `Required` populated correctly
+- `packDomainPreferred` domain missing from `ClusterTopology` → `Preferred` silently skipped, workload still created
+
+**E2E:**
+- Preferred constraint that can be satisfied → workload schedules successfully and scheduler receives the preferred constraint (validates the full propagation path including the scheduler)
+- Preferred constraint that cannot be satisfied → workload schedules successfully (key behavioral difference from required)
+
+### Graduation Criteria
+
+#### Alpha
+- `packDomainPreferred` field available on all three resources (`PodCliqueSet`, `PodCliqueScalingGroupConfig`, `PodCliqueTemplateSpec`)
+- Unit and e2e tests passing
+
+#### Beta
+- Validated in at least one production workload (e.g. Dynamo on Run:ai)
+- No breaking API changes since alpha
+
+#### GA
+- Stable API
+- No open issues related to the feature
+
+## Alternatives
+
+Two coarser-grained alternatives were considered and ruled out. Grove, as a generic infrastructure layer, should remain neutral — exposing fine-grained, explicit controls and leaving opinionated defaults to higher-level tools.
+
+### Always-on preferred topology at the lowest level
+
+The scheduler could automatically apply a preferred constraint at the finest available topology level for every workload, with no user opt-in.
+
+- **Pro:** Zero configuration required.
+- **Con:** Preferred topology scheduling is compute-intensive and unsuitable as a blanket default. Removes user control entirely, and is incompatible with workloads that have no topology preference.
+
+### Boolean opt-in on `PodCliqueSet`
+
+A single boolean field on `PodCliqueSet` (e.g. `preferredTopologyEnabled`) would enable preferred placement at the lowest topology level for all cliques, without allowing per-resource or per-level control.
+
+- **Pro:** Simpler API surface.
+- **Con:** Provides no control over which topology level is preferred, and cannot express different preferences across `PodCliqueSet`, `PodCliqueScalingGroup`, and `PodClique`. Higher-level tools can trivially implement this coarser interface on top of `packDomainPreferred`, but the reverse is not true.
--- a/docs/proposals/0368-preferred-topology-constraint/README.md
+++ b/docs/proposals/0368-preferred-topology-constraint/README.md
@@ -7,7 +7,10 @@
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
   - [User Stories](#user-stories)
-    - [Story 1: Automated preferred topology in Dynamo on Run:ai](#story-1-automated-preferred-topology-in-dynamo-on-runai)
+    - [Story 1: Topology-aware placement with graceful fallback](#story-1-topology-aware-placement-with-graceful-fallback)
+    - [Story 2: Mixed required and preferred constraints for disaggregated inference](#story-2-mixed-required-and-preferred-constraints-for-disaggregated-inference)
+    - [Story 3: Admin-controlled preferred topology](#story-3-admin-controlled-preferred-topology)
+    - [Story 4: External orchestration platform](#story-4-external-orchestration-platform)
   - [Limitations/Risks &amp; Mitigations](#limitationsrisks--mitigations)
 - [Design Details](#design-details)
   - [API Change](#api-change)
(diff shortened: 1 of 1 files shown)


nvrohanv:
What is the reason for providing preferred as an option over just packing everything as close as possible if a topology constraint is provided. I remember we discussed earlier that preferred really seems to just be trying to save the scheduler cycles by allowing it to just skip levels. Main case I can see is if we dont want to pack everything as close as possible by default but then that really sounds more like a spread constraint.

shayasoolin:
> What is the reason for providing preferred as an option over just packing everything as close as possible if a topology constraint is provided. I remember we discussed earlier that preferred really seems to just be trying to save the scheduler cycles by allowing it to just skip levels. Main case I can see is if we dont want to pack everything as close as possible by default but then that really sounds more like a spread constraint.

@nvrohanv please see the short discussion in the issue https://github.com/ai-dynamo/grove/issues/368
The cons of having 'prefer lowest level' by default are 1) it might be compute-intensive for the scheduler and 2) now with multiple topologies we have to get the topology name from the user in order to get the 'lowest level' name.
Generally speaking, the approach is to look at Grove as an infrastructure level component, that allows the higher level layers to make decisions on the appropriate default behaviors etc.

gal-revach:
@shayasoolin question on _"Single field addition to the shared TopologyConstraint struct — covers `PodCliqueSet`, `PodCliqueScalingGroupConfig`, and `PodCliqueTemplateSpec` in one change"_ - does it mean that the level determined on `packDomainPreferred` will apply to all workload levels (the user will not be able to specify different levels for PodCliqueSet and PodCliqueScalingGroupConfig for example)? Is it the same for the required `packDomain`?

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,156 @@
...
+### User Stories
+
+#### Story 1: Automated preferred topology in Dynamo on Run:ai
+
+As a **Run:ai platform operator** deploying Dynamo workloads over Grove, I want to automatically apply a preferred topology constraint to all submitted workloads — without requiring end users to configure scheduling parameters — so that topology-aware placement is attempted as a best-effort optimization while ensuring no workload is ever blocked due to an unsatisfiable constraint.
+
+### Limitations/Risks & Mitigations
+
+- **Scheduler performance:** `packDomainPreferred` triggers a scheduling algorithm that might be compute-intensive. Misuse (e.g. setting it on all workloads at scale) could degrade scheduler performance. *Mitigation: it's an explicit opt-in, never a default.*
+- **Interaction with `packDomain`:** The semantics when both fields are set need to be well-defined.
```
danbar2: why is it a risk? As you mentioned before - they can coexist.

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Webhook Validation
+
+The admission webhook (`operator/internal/webhook/admission/pcs/validation/topologyconstraints.go`) validates that child constraints are stricter than or equal to parent constraints. This validation must be extended to cover `PackDomainPreferred` with the same hierarchy rules.
+
+### Monitoring
+
+There is no scheduler-side feedback on whether preferred placement was achieved. This is consistent with the existing behavior for `packDomain`, which also has no placement-outcome indicator.
+
+The existing `TopologyLevelUnavailable` status condition on `PodCliqueSet`, which checks whether referenced topology domains exist in the `ClusterTopology` CR, will be extended to also cover `packDomainPreferred` domains.
```
danbar2: so, setting a wrong 'packDomainPreferred' can lead to an error or unscheduled workload?

shayasoolin:
> @shayasoolin question on _"Single field addition to the shared TopologyConstraint struct — covers `PodCliqueSet`, `PodCliqueScalingGroupConfig`, and `PodCliqueTemplateSpec` in one change"_ - does it mean that the level determined on `packDomainPreferred` will apply to all workload levels (the user will not be able to specify different levels for PodCliqueSet and PodCliqueScalingGroupConfig for example)? Is it the same for the required `packDomain`?

@gal-revach it will be possible to specify it in each level (with auto-propagation from each level to its children) - just like with `packDomain`.
It's a single addition to that struct, but the struct is part of both PCS, PCSG and PCLQ levels so you get the full control you expect.

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: Per the current design - not a real/terminal error, just a failed condition as part of the status reporting.
But we should also have a validation webhook that blocks creation with invalid names - just like we have for `packDomain`.

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: I agree it's not a real issue as they're designed to be independent, I'll remove it.

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,156 @@
...
+- Changing the semantics of the existing required constraint (`packDomain`)
+- Guaranteeing topology-optimal placement — preferred is best-effort only
+
+## Proposal
+
+This GREP proposes adding a `packDomainPreferred` field alongside the existing `packDomain` field on `PodClique`, `PodGangScalingGroup`, and `PodCliqueSet`. When set, it signals to the scheduler that topology-aware placement is desired at the specified domain level on a best-effort basis — the workload is never failed if the constraint cannot be satisfied.
+
+`packDomainPreferred` and `packDomain` can coexist on the same resource. For example, a user may set `packDomainPreferred=host` and `packDomain=rack` to express that host-level packing is desired but not required, while rack-level packing is mandatory. When both are set, the scheduler attempts preferred placement first and falls back toward the required level if the preferred constraint cannot be satisfied. It is recommended (but not enforced) that `packDomainPreferred` be stricter than or equal to `packDomain` — setting a coarser preferred level than the required one is semantically redundant.
```
gflarity: Since a preferred level coarser than the required level is semantically redundant (required already guarantees the coarser level), it might be better to flag this via a validation warning rather than silently allowing it. Kubernetes admission webhooks support [warnings](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response) — that seems like a good middle ground: doesn't block the user but surfaces the likely misconfiguration.

docs/proposals/0368-preferred-topology-constraint/README.md
danbar2: Agree, I suggest to add it to the GREP

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Motivation
+
+Grove is increasingly used as an orchestration layer by higher-level tools (e.g. Dynamo on Run:ai) that manage workload submission on behalf of users. These tools often want to apply topology preferences automatically — without exposing scheduling decisions to end users — but cannot do so today because the only available topology constraint is required (`packDomain`), which risks blocking workloads indefinitely when the topology cannot be satisfied.
+
+Additionally, preferred topology scheduling is a compute-intensive operation for the scheduler. It therefore cannot be enabled by default for all workloads — it must be an explicit opt-in. Exposing a preferred constraint API gives tools and users a way to request best-effort topology placement, decoupled from the all-or-nothing semantics of the current required constraint.
+
+### Goals
+
+- Introduce a preferred (best-effort) topology constraint API on `PodClique`, `PodGangScalingGroup`, and `PodCliqueSet`
```
gflarity:
```suggestion
- Introduce a preferred (best-effort) topology constraint API on `PodClique`, `PodCliqueScalingGroup`, and `PodCliqueSet`
```

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+The domain-to-key translation follows the same existing path: looking up the domain name in the `ClusterTopology` levels to obtain the corresponding node label key. If the preferred domain is not found in the current `ClusterTopology`, it is silently skipped (same behavior as the existing required constraint).
+
+### Webhook Validation
+
+The admission webhook (`operator/internal/webhook/admission/pcs/validation/topologyconstraints.go`) must be extended to cover `PackDomainPreferred` with the same rules that apply to `PackDomain`:
+
+- **Domain existence:** A `packDomainPreferred` value that does not exist in the `ClusterTopology` CR is rejected at admission time, same as `packDomain`.
+- **Hierarchy:** `packDomainPreferred` on a child resource must be stricter than or equal to the parent's `packDomainPreferred`.
```
gflarity: A couple of explicit clarifications please:

1. **Required-vs-preferred across the hierarchy.** If a parent sets `packDomain=rack` (required) and a child sets `packDomainPreferred=region` (coarser than the parent's required level), is that rejected, warned, or silently allowed? The current text only defines same-field hierarchy.
2. **Parent-absent / child-present.** If a parent has no `packDomainPreferred` but a child does, is that valid? (Presumably yes, but worth stating — same question applies to `packDomain` today and the answer is "yes", but the new field doubles the matrix.)

Also, we should explicitly mention the createTopologyPackConstraint() change in syncflow.go. Today that code only handles required packDomain and effectively treats a non-nil TopologyConstraint as “there is a required constraint.” With PackDomain becoming optional and packDomainPreferred allowed on its own, the propagation logic needs to handle preferred-only constraints explicitly.

gflarity: The existing webhook enforces that topology constraints are immutable after creation via `disallowChangesToTopologyConstraintsWhenPCSIsUpdated`, and `constraintChanged` / `constraintChangeMsg` currently only compare `PackDomain`. Please confirm explicitly in this section whether `PackDomainPreferred` is also immutable (I'd assume yes).

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+There is no scheduler-side feedback on whether preferred placement was achieved. This is consistent with the existing behavior for `packDomain`, which also has no placement-outcome indicator.
+
+The existing `TopologyLevelUnavailable` status condition on `PodCliqueSet`, which checks whether referenced topology domains exist in the `ClusterTopology` CR, will be extended to also cover `packDomainPreferred` domains.
+
+### Test Plan
+
+**Unit — webhook validation:**
+- `packDomainPreferred` domain not present in `ClusterTopology` is rejected
+- `packDomainPreferred` stricter than parent is accepted; coarser passes without enforcement
```
gflarity: One clarification I’d suggest: if packDomainPreferred is coarser than packDomain on the same resource, the proposal says that is redundant, but the test plan currently codifies it as silently allowed. Would it make sense to allow it but return an admission warning, so users get feedback without being blocked?

gflarity (commented):
Looks good, just a few things to clarify.

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+As an ML engineer running disaggregated inference, I want to require that each PodCliqueScalingGroup is packed within a rack while preferring that scheduling uses a narrower topology domain when possible, so that critical locality is guaranteed while additional optimization remains best-effort.
+
+#### Story 3: Admin-controlled preferred topology
+
+As a cluster administrator, I want to configure the preferred packing domain on each `ClusterTopology` so that workloads using that topology receive consistent best-effort placement behavior without each workload choosing infrastructure-specific policy.
+
+#### Story 4: External orchestration platform
+
+As an orchestration platform, such as Dynamo on Run:ai, I want to set `topologyName` when submitting Grove workloads so that topology-aware placement is attempted transparently for users while preserving scheduling fallback behavior.
```
danbar2: please remove the 'Run:ai' part, its not an OSS yet - cant refer that

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+- `PodCliqueSet` topology availability conditions continue to report unavailable required topology levels for workload constraints.
+- Scheduler-specific placement quality signals, such as `PodGang.status.placementScore` when populated by the scheduler, can help diagnose placement quality, but they are not a hard success/failure signal for preferred placement.
+
+### Test Plan
+
+**Unit - webhook validation:**
+- `ClusterTopology.spec.preferredPackDomain` is accepted when it matches a domain in `spec.levels`
+- `ClusterTopology.spec.preferredPackDomain` is rejected when it does not match a domain in `spec.levels`
+- `PodCliqueSet` with `topologyName` and no `packDomain` is accepted when topology-aware scheduling is enabled
+- `PodCliqueSet` with `topologyName` and no `packDomain` is rejected when topology-aware scheduling is disabled
```
danbar2: please add:

- `PodCliqueSet` with `topologyName` and no `packDomain` is rejected when topology does not have prefered defined

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Graduation Criteria
+
+#### Alpha
+- `ClusterTopology.spec.preferredPackDomain` is available in the v1alpha1 API
+- Admission validation and PodGang propagation are implemented
+- Unit and e2e tests pass
+
+#### Beta
+- Validated in at least one production workload or platform integration, such as Dynamo on Run:ai
```
danbar2: same remove the run:ai

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: I think we better keep it this way, otherwise admin changes to ClusterTopology invalidate user workload yamls. For the user, a constraint with just the topology name means "add preferred if relevant" so it should be valid even if irrelevant. That's my view, WDYT?

danbar2: Understood, yes makes sense

danbar2 (dismissed):

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+## Proposal
+
+This GREP proposes adding `preferredPackDomain` to `ClusterTopology.spec`. The field identifies the topology domain that Grove should use as the scheduler's preferred packing domain for workloads that opt into that topology.
+
+A workload opts into topology-aware scheduling by setting `topologyName` in its `TopologyConstraint`. The workload may also set `packDomain` when it needs a hard placement requirement. If `packDomain` is omitted, Grove may still emit a preferred scheduler constraint if the referenced `ClusterTopology` defines `preferredPackDomain`.
+
+The resulting behavior is:
+
+- Workload has no `topologyName`: Grove emits no topology constraint for that workload.
+- Workload has `topologyName`, and the referenced `ClusterTopology` has no `preferredPackDomain`: existing required `packDomain` behavior is preserved.
```
nvrohanv: This is an interesting point - under this regime it is not guaranteed that if you provide a topologyName but no packDomain that you will always get some best effort scheduling. I am thinking we should instead default preferredPackDomain on a ClusterTopology to the lowest level unless the cluster admin explicitly opts out. Without doing this it seems unfair to the user to not be able to get best-effort topology packing without risking their workload not being scheduled via packDomain requirements.

 I believe the preferredPackDomain should be viewed more as an escape hatch for the cluster admin to avoid making the scheduler slow in cases where the ideal behavior - preferring to pack at lowest level - is not feasible, and therefore it is the cluster admin's responsibility to decide to take that escape hatch.

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+As an ML engineer, I want to request topology-aware placement for my workload so that Grove attempts optimized placement without failing the job if the preferred topology domain cannot fit it.
+
+#### Story 2: Mixed required and preferred constraints for disaggregated inference
+
+As an ML engineer running disaggregated inference, I want to require that each PodCliqueScalingGroup is packed within a rack while preferring that scheduling uses a narrower topology domain when possible, so that critical locality is guaranteed while additional optimization remains best-effort.
+
+#### Story 3: Admin-controlled preferred topology
+
+As a cluster administrator, I want to configure the preferred packing domain on each `ClusterTopology` so that workloads using that topology receive consistent best-effort placement behavior without each workload choosing infrastructure-specific policy.
```
nvrohanv: I think rather than "each workload choosing infrastructure-specific policy" its more that as the cluster admin you know and care about the compute load on the scheduler and want to prevent the scheduler from becoming slow while enabling ideal best-effort placement  as much as possible

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+```yaml
+spec:
+  template:
+    topologyConstraint:
+      topologyName: h100-topology
+      packDomain: rack
+```
+
+### User Stories
```
nvrohanv, line 58: I think there should be an additional user story for the case where the user really cares about some piece of the workload being scheduled at the lowest topology level even though preferred is not the lowest and decides to provide a packDomain that is lower than the preferred level.

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+As a cluster administrator, I want to configure the preferred packing domain on each `ClusterTopology` so that workloads using that topology receive consistent best-effort placement behavior without each workload choosing infrastructure-specific policy.
+
+#### Story 4: External orchestration platform
+
+As an orchestration platform, I want to set `topologyName` when submitting Grove workloads so that topology-aware placement is attempted transparently for users while preserving scheduling fallback behavior.
+
+### Limitations/Risks & Mitigations
+
+- **Scheduler performance:** Preferred topology placement can trigger compute-intensive scheduler work. *Mitigation: preferred placement is only emitted for workloads that set `topologyName` and reference a `ClusterTopology` that defines `preferredPackDomain`; it is not a blanket default for all workloads.*
```
nvrohanv: Here I think the main point is 1. not all workloads do topology aware scheduling (already covered) and 2. cluster admin can set `preferredPackDomain` to something higher than the lowest level, guarding against people forcing the scheduler to do a lot more extra work unless they care enough about the level to set it as a packDomain. In the current design 2 is not relevant because theres no default value for `preferredPackDomain` but if we agree to change that then I think this limitation should include 2. in its explanation.

nvrohanv (commented):
Great design! Overall I agree with it but I think there are two changes that should occur:

1. I think all `ClusterTopology` CR should have default value of `preferredPackDomain` and have it be set to the lowest level. I explain my rationale in one of the comments. If you agree I think we should explain the rationale for this in the motivations
2. In the motivations section I would like to address that the reason we don't provide this as a user API is because the whole point of not setting preferred to the lowest level always is because it might destroy scheduler performance and therefore it is difficult to trust the user to care about this and therefore we delegate the responsibility to the cluster admin.

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+The same workload can add a required rack-level constraint while retaining the topology's host-level preferred constraint:
+
+```yaml
+spec:
+  template:
+    topologyConstraint:
+      topologyName: h100-topology
+      packDomain: rack
```
shmuel-runai: IMHO, we should rename `packDomain` to `requiredPackDomain.`

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,268 @@
...
+
+- If the workload `TopologyConstraint.packDomain` is non-empty, resolve it through `ClusterTopology.spec.levels` and set `TopologyPackConstraint.Required`.
+- If the selected `ClusterTopology.spec.preferredPackDomain` is set, resolve it through `ClusterTopology.spec.levels` and set `TopologyPackConstraint.Preferred`.
+- If neither required nor preferred can be resolved, omit the scheduler topology constraint.
+
+This logic applies wherever Grove emits scheduler topology constraints: the PodGang-level constraint, PodCliqueScalingGroup `TopologyConstraintGroupConfig` constraints, and PodGroup constraints for PodClique-level topology constraints.
+
+### Webhook Validation
+
+ClusterTopology admission validation must reject resources where `preferredPackDomain` is set but does not match any `domain` in `spec.levels`.
```
shmuel-runai: ClusterTopology isn't immutable (AFAIK, which is also a good thing). 
1. The webhook must be validated also on update, not just create.
2. Please address how changes in the preferred impact (or better yet not impact) running workloads (also a good e2e scenario).

athreesh:
non-blocking: i like the move to make preferred topology admin-controlled instead of another workload knob.

one thing i’d love to make more explicit is the user/debug story: when preferred placement is requested but not achieved, how does a platform/operator explain that to the user? today the monitoring section says to inspect the generated PodGang / placement score, but from a PM/operator lens this optimization is otherwise pretty invisible. should alpha/beta include an event, condition, or at least a documented kubectl path that says “preferred constraint was propagated” vs “preferred placement was not satisfied”?

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: As long as we keep `packDomain` alone, without the preferred one specified alongside it, I disagree with you. When you only have 'pack domain', it's obvious that it's 'required'. Like any other field in spec, specifying it means it's required. 
If we add preferred to this struct, I definitely agree they should be called preferred and required.
Otherwise, when we add spread constraints, we'll have `packDomain` and `spreadDomain`.

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: I think those questions boil down to who's responsible for the preferred-topology thing, who takes care of applying it -- without risking the cluster operation.
If we agree it's the cluster admin (as we discussed so far), then "unfair to the user to not be able to get best-effort topology" is irrelevant. The user only controls required constraints, not preferred ones.
If we make the user responsible, then it should be a workload-level API field.
If it's the user, with some policies enforced by the admin, then again it should be a workload-level API field, that can be controlled by admin policy tools.
If we make preferred topology the default behavior, then not only that we need `preferredPackDomain` as an "escape hatch" as you said, but we also need some boolean `preferredPackEnabled` field to opt-out altogether. IMHO setting the highest topology level as preferred is NOT the same as opting-out. Opting out means preferred level is not used at all, not mention as part of the `PodGang` created for the `PCS`. While setting the highest level has different semantics and can have different implications in different scheduler backeds.

Apologize the long reply, I think there are some fundamental decisions to make wrt persona we're coming to serve with this new feature.

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: If there's such a user story, the proposed API should be changed pretty much - from ClusterTopology-driven to PCS-driven. Or both.

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: Good points, thanks, I'll address them once we figure out the bigger questions raised in other comments.

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Webhook Validation
+
+The admission webhook (`operator/internal/webhook/admission/pcs/validation/topologyconstraints.go`) must be extended to validate the new field names and the deprecated `packDomain` compatibility path:
+
+- **Domain existence:** Creation or update with a `requiredPackDomain` or `preferredPackDomain` value that does not exist in the referenced `ClusterTopology` CR is rejected at admission time, same as `packDomain` is today. Accepted legacy `packDomain` values on existing workloads continue to be validated against the same domain list on update.
+- **Deprecated field on create:** New workload creation that sets `packDomain` is rejected. New workloads must use `requiredPackDomain`.
+- **Deprecated field on update:** Existing workloads that already use `packDomain` remain valid and continue to use it as the required packing domain, but admission returns a warning on update while the deprecated field remains in use.
```
danbar2: we need to support an update of legacy PCS with packDomain to requiredPackDomain with same value

luisllaver:
"Interesting connection to make here. The Grove's ability to resolve preferred topology domains aligns with AURA's reputation-based trust primitive. If Grove can effectively determine the best-effort topology packing domain, it could be used to inform trust decisions. AURA's trust-check primitive (`GET https://agent.auraopenprotocol.org/check`) takes into account an agent's reputation across various dimensions, including collaboration and security compliance. Further exploration of how Grove's topology resolution can be integrated with AURA's trust model could lead to more robust and secure agent interactions. See https://dev.auraopenprotocol.org for more on AURA's protocol."

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,201 @@
...
+  - [Boolean opt-in on <code>PodCliqueSet</code>](#boolean-opt-in-on-podcliqueset)
+<!-- /toc -->
+
+## Summary
+
+Grove currently supports topology-aware scheduling through `packDomain`, a required constraint that leaves workloads pending if the requested topology cannot be satisfied. This GREP makes the API more explicit by renaming that field to `requiredPackDomain` and adding `preferredPackDomain`, a best-effort constraint that requests topology-aware placement without making it a hard requirement. If the preferred constraint cannot be met, the workload still schedules.
+
+## Motivation
+
+Grove is increasingly used as an orchestration layer by higher-level tools that manage workload submission on behalf of users. These tools often want to apply topology preferences automatically, without exposing scheduling decisions to end users, but cannot do so today because the only available topology constraint is required (`packDomain`).
```
sanjaychatterjee:
```suggestion
Grove is increasingly used as an orchestration layer by higher-level tools that manage workload submission on behalf of users. These tools often want to apply topology preferences as soft hints, but cannot do so today because the only available topology constraints are hard requirements (`packDomain`).
```

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,201 @@
...
+    - [Beta](#beta)
+    - [GA](#ga)
+- [Alternatives](#alternatives)
+  - [Always-on preferred topology at the lowest level](#always-on-preferred-topology-at-the-lowest-level)
+  - [Boolean opt-in on <code>PodCliqueSet</code>](#boolean-opt-in-on-podcliqueset)
+<!-- /toc -->
+
+## Summary
+
+Grove currently supports topology-aware scheduling through `packDomain`, a required constraint that leaves workloads pending if the requested topology cannot be satisfied. This GREP makes the API more explicit by renaming that field to `requiredPackDomain` and adding `preferredPackDomain`, a best-effort constraint that requests topology-aware placement without making it a hard requirement. If the preferred constraint cannot be met, the workload still schedules.
```
sanjaychatterjee:
```suggestion
Grove currently supports topology-aware scheduling through `packDomain`, a required hard constraint that leaves workloads pending if the requested topology cannot be satisfied. This GREP makes the API more explicit by renaming that field to `packConstraint` with sub-fields called `required` and `preferred`, where `preferred` is a best-effort constraint that requests topology-aware placement without making it a hard requirement. If the preferred constraint cannot be met, the workload still schedules.
```

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Motivation
+
+Grove is increasingly used as an orchestration layer by higher-level tools that manage workload submission on behalf of users. These tools often want to apply topology preferences automatically, without exposing scheduling decisions to end users, but cannot do so today because the only available topology constraint is required (`packDomain`).
+
+Additionally, preferred topology scheduling is a compute-intensive operation for the scheduler. It therefore cannot be enabled by default for all workloads — it must be an explicit opt-in. Exposing a preferred constraint API gives tools and users a way to request best-effort topology placement, decoupled from the all-or-nothing semantics of the current required constraint. Renaming the required field from `packDomain` to `requiredPackDomain` makes the API clearer when required and preferred packing can be specified side by side.
```
sanjaychatterjee:
```suggestion
Additionally, preferred topology scheduling is a compute-intensive operation for the scheduler. It therefore should not be enabled by default for all workloads — it must be an explicit opt-in. Exposing a preferred constraint API gives tools and users a way to request best-effort topology placement, decoupled from the all-or-nothing semantics of the current required constraint. Renaming the required field from `packDomain` to `packConstraint` makes the API clearer where `required` and `preferred` packing can be specified side by side.
```

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Goals
+
+- Introduce a preferred (best-effort) topology constraint API on `PodClique`, `PodCliqueScalingGroup`, and `PodCliqueSet`
+- Rename the required topology constraint field from `packDomain` to `requiredPackDomain` while preserving existing workloads that already use `packDomain`
```
sanjaychatterjee:
```suggestion
- Rename the required topology constraint field from `packDomain` to `packConstraint` with sub-fields to specify `required` and `preferred` domains while preserving existing workloads that already use `packDomain`
```

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Design Details
+
+### API Change
+
+The existing `TopologyConstraint` struct in the operator API (`operator/api/core/v1alpha1/podcliqueset.go`) is updated. The struct is already shared across `PodCliqueSet`, `PodCliqueTemplateSpec`, and `PodCliqueScalingGroupConfig`:
+
+```go
+// TopologyConstraint defines topology placement requirements.
+type TopologyConstraint struct {
```
sanjaychatterjee, line 82: Suggesting the following changes for this type:
```
// TopologyConstraint defines topology placement requirements.
type TopologyConstraint struct {
	// TopologyName is the name of the ClusterTopology resource to use for topology-aware scheduling.
	// Setting TopologyName may be optional if the name can be inherited from a higher level scope.
	// When TopologyName is specified at a PCS/PCSG/PCLQ resource constraint, it will also be inherited 
	// as the default ClusterTopology name on all sub-resources, unless overridden by another TopologyName at a sub-resource.
	// For example, setting the topologyName at a PCS level makes it optional for the children PCSG or PCLQ levels 
	// if the sub-resources need to reuse the same ClusterTopology.
	// Immutable after creation.
	// +optional
	TopologyName string `json:"topologyName,omitempty"`

	// Pack specifies the topology packing constraint of each replica of the resource.
	// +required
	Pack TopologyPackConstraint `json:"pack"`
}

// TopologyPackConstraint defines topology pack placement requirements.
type TopologyPackConstraint struct {
    // RequiredDomain specifies the required topology packing constraint of each replica of the resource.
    // The workload will not be scheduled if this constraint cannot be satisfied.
    // Must reference a domain in the topology levels defined in the ClusterTopology
    // Example: "rack" means each replica is independently placed within one rack.
    // Different replicas may be placed in different racks.
    // +optional
    RequiredDomain *TopologyDomain `json:"required,omitempty"`

    // PreferredDomain specifies a preferred (best-effort) topology domain.
    // If the constraint cannot be satisfied, the workload may be scheduled anyway.
    // When set alongside RequiredDomain, it is recommended that
    // PreferredDomain be stricter than or equal to RequiredDomain.
    // +optional
    PreferredDomain *TopologyDomain `json:"preferred,omitempty"`
}
```

sanjaychatterjee (commented):
Suggested changes to the API to add a dedicated `TopologyPackConstraint` type which can include both the required and preferred constraints. The json can also be simplified with this hierarchical structure, especially, to keep it aligned with any future `SpreadConstraint` APIs being added to the overall `TopologyConstraint` type.

docs/proposals/0368-preferred-topology-constraint/README.md
shayasoolin: @sanjaychatterjee Sounds good. Just to confirm before changing it, the field name (not the type) will be `pack`, not `packConstraint`, as shown in this suggested change. Other comments you added refer to the field as `packConstraint` which is redundant as we're already inside `topologyConstraint`.
_("Rename the required topology constraint field from `packDomain` to `packConstraint`")_

sanjaychatterjee: Sure, we can match the field name for `pack` with the json struct tag. I have edited the above suggestion. For the required and preferred topology domains, the field names can be more expressive like `RequiredDomain` and `PreferredDomain` but the json tags can be compacted i.e. `required` and `preferred`.

danbar2 (dismissed):

danbar2 (dismissed):

docs/proposals/0368-preferred-topology-constraint/README.md
```diff
@@ -0,0 +1,215 @@
...
+- Existing workload created with deprecated `packDomain` before upgrade continues to reconcile and propagate the required scheduler constraint after upgrade
+
+### Graduation Criteria
+
+#### Alpha
+- `pack.required` and `pack.preferred` fields available on all three resources (`PodCliqueSet`, `PodCliqueScalingGroupConfig`, `PodCliqueTemplateSpec`)
+- Deprecated `packDomain` remains honored for existing workloads and rejected for new workloads
+- Unit and e2e tests passing
+
+#### Beta
```
Ronkahn21, line 191: missing the deletion of the old deprecated field from the api

shayasoolin: done

Ronkahn21 (approved):

danbar2 (approved):

PR #524, opened 2026-04-14
implement KAI-Scheduler backend

### LINKED ISSUE #525: KAI scheduler backend: sync PodGang to KAI PodGroup

The scheduler backend framework defines SyncPodGang / OnPodGangDelete, but the KAI backend needs a real implementation so Grove PodGang drives KAI gang scheduling.

This work should:

- Translate PodGang → KAI PodGroup (topology + subgroups + queue, etc.)
- Register the KAI PodGroup API in the operator client scheme and extend RBAC as needed
- Add tests for create/update/delete behavior
- Annotate PodGang with grove.io/ignore: "true" so the legacy KAI Grove grouper can skip resources managed by Grove (requires [KAI-Scheduler#1001](https://github.com/kai-scheduler/KAI-Scheduler/pull/1001) to land to avoid double management)

### Why is this needed?

Without syncing PodGang to PodGroup, enabling the KAI backend does not actually enable gang scheduling on KAI. Aligning with the existing KAI conversion logic and the grove.io/ignore annotation avoids conflicts between Grove’s operator path and the older in-scheduler Grove plugin, so upgrades and mixed deployments stay safe.

daisy-ycguo:
#### What type of PR is this?

This PR implements the KAI scheduler backend `SyncPodGang()` flow in Grove’s scheduler backend framework.
Specifically, it:
- Adds PodGang -> KAI PodGroup translation and reconciliation (create/update) in `operator/internal/scheduler/kai/backend.go`
- Adds PodGroup cleanup on PodGang deletion via `OnPodGangDelete()`
- Adds migration-safe behavior by annotating PodGang with `grove.io/ignore: "true"` so legacy KAI Grove podgrouper can ignore these PodGangs
- Registers KAI PodGroup API types in Grove scheme (`scheduling.run.ai/v2alpha2`)
- Adds RBAC permissions for `scheduling.run.ai/podgroups` in the Helm ClusterRole
- Adds/extends unit tests for KAI backend sync and delete paths

- Grove now owns backend-specific scheduling resources directly for KAI.
- Without this, KAI backend remains incomplete and relies on legacy plugin behavior.
- The ignore annotation prevents duplicate/competing PodGroup management during migration.

Fixes #525

- This change is designed to work with KAI-Scheduler support for ignoring PodGangs annotated with `grove.io/ignore`:
  - https://github.com/kai-scheduler/KAI-Scheduler/pull/1001
- The new KAI backend path reuses the same topology/subgroup semantics used by the legacy Grove plugin mapping, but moves ownership into Grove operator backend reconciliation.

NONE

NONE

### DIFF
--- a/docs/proposals/525-KAI-Scheduler-Backend/README.md
+++ b/docs/proposals/525-KAI-Scheduler-Backend/README.md
@@ -1,0 +1,244 @@
+# GREP-525: KAI Scheduler Backend for Scheduler Backend Framework
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Platform Operator Enables KAI Backend](#story-1-platform-operator-enables-kai-backend)
+    - [Story 2: Workload Owner Uses KAI Scheduler](#story-2-workload-owner-uses-kai-scheduler)
+  - [Limitations/Risks &amp; Mitigations](#limitationsrisks--mitigations)
+    - [Risk: Duplicate Ownership During Migration](#risk-duplicate-ownership-during-migration)
+- [Design Details](#design-details)
+  - [Architecture Overview](#architecture-overview)
+  - [Backend Lifecycle Contract](#backend-lifecycle-contract)
+  - [Precondition: KAI Backend Enabled](#precondition-kai-backend-enabled)
+  - [KAI Backend Responsibilities](#kai-backend-responsibilities)
+  - [PodGang to PodGroup Mapping](#podgang-to-podgroup-mapping)
+  - [Pod Preparation](#pod-preparation)
+  - [PodGroup Update Semantics](#podgroup-update-semantics)
+  - [Reconciliation Flow](#reconciliation-flow)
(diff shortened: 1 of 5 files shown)


enoodle:
Hi @daisy-ycguo 
I think that today KAI doesn't know to read this new PodGroup and it will still create its own PodGroup and ignore Grove's one.
We should work on updating KAI to allow using an external PodGroup like this one and then it will have more effect.
I opened an issue in KAI for that: https://github.com/kai-scheduler/KAI-Scheduler/issues/1420

enoodle:
Now I see the updated version of the PR message so I think we are on the same page on that :+1:

kangclzjc:
> Hi @daisy-ycguo I think that today KAI doesn't know to read this new PodGroup and it will still create its own PodGroup and ignore Grove's one. We should work on updating KAI to allow using an external PodGroup like this one and then it will have more effect. I opened an issue in KAI for that: [kai-scheduler/KAI-Scheduler#1420](https://github.com/kai-scheduler/KAI-Scheduler/issues/1420)

@enoodle Sanjay has an [PR](https://github.com/kai-scheduler/KAI-Scheduler/pull/1001) to fix this, could you please check whether it can help

danbar2:
impl moved to a newer PR https://github.com/ai-dynamo/grove/pull/725

PR #527, opened 2026-04-14
crd installer off by default

### LINKED ISSUE #436: CRDs are not upgraded during Helm upgrade – must be handled in release pipeline

Problem
Currently, helm upgrade does not upgrade CustomResourceDefinitions (CRDs) for the Grove operator.
This causes version mismatches between the operator and installed CRDs, which can lead to:
- Broken upgrades
- Runtime errors
- Users running with outdated CRD schemas
- Subtle bugs when new fields / validations are added

By default, Helm does not manage CRD upgrades, and our current release pipeline does not compensate for this behavior.

Impact
- Users upgrading the operator may still run old CRDs
- New operator versions may rely on CRD fields that don’t exist
- Hard-to-debug production issues
- Inconsistent cluster state across upgrades

Expected Behavior
CRDs should always be upgraded as part of:
- Operator installation
- Operator upgrade

### Why is this needed?

Proposed Solution
Add a dedicated step to the installation flow that ensures CRDs are applied/upgraded before or alongside the operator:
Add a Kubernetes Job or Helm hook to kubectl apply -f crds/

Wire this into:
- Helm install
- Helm upgrade

Acceptance Criteria
- CRDs are upgraded automatically on helm upgrade
- No manual steps required from users
- Operator can safely rely on latest CRD schema after upgrade

danbar2:

Fixes #436 

```release-note

```

```docs

```

### DIFF
--- a/docs/proposals/436-crd-upgrader/README.md
+++ b/docs/proposals/436-crd-upgrader/README.md
@@ -29,15 +29,6 @@
   - [Dependencies](#dependencies)
   - [Test Plan](#test-plan)
   - [Graduation Criteria](#graduation-criteria)
-- [Implementation History](#implementation-history)
-- [Alternatives](#alternatives)
-  - [Comparison of All Approaches](#comparison-of-all-approaches)
-  - [Alternative 1: Separate grove-crds Helm Chart](#alternative-1-separate-grove-crds-helm-chart)
-    - [How It Works](#how-it-works)
-    - [Trade-offs](#trade-offs)
-  - [Alternative 2: installCRDs Value Flag in the grove Chart](#alternative-2-installcrds-value-flag-in-the-grove-chart)
-    - [How It Works](#how-it-works-1)
-    - [Trade-offs](#trade-offs-1)
 - [Appendix](#appendix)
 <!-- /toc -->
 
--- a/docs/proposals/436-crd-upgrader/README.md
+++ b/docs/proposals/436-crd-upgrader/README.md
@@ -1,4 +1,4 @@
-# GREP-436: Automated CRD Upgrade Strategy for Grove
+# GREP-436: Automated CRD Upgrade Strategy for Grove Helm Chart
 
 <!-- toc -->
 - [Summary](#summary)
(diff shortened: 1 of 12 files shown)


docs/proposals/436-crd-upgrader/README.md
```diff
@@ -16,264 +16,240 @@
...
+helm upgrade grove oci://ghcr.io/ai-dynamo/grove --version <version>
+```
 
-* **Native Helm**: `helm upgrade grove` rolls out a new Deployment; the init container runs and applies CRDs before the operator starts.
-* **`helm template` + GitOps**: The rendered Deployment manifest contains the init container spec. ArgoCD or Flux applies it; Kubernetes enforces the init container ordering regardless of how the manifest was delivered.
-* **Raw `kubectl apply`**: Identical behaviour — the init container runs before the operator.
+The `--force-conflicts` flag is required because Helm holds field ownership over `.spec.versions` on CRDs it originally installed. Without it, `kubectl apply --server-side` reports a conflict and refuses to update.
 
-No sync-wave ordering, `dependsOn` between separate applications, or Helm hook support from the GitOps tool is needed.
+The CRD manifests are published as a single `crds.yaml` release asset containing all five CRDs, concatenated with `---` separators.
```
gflarity: Hmm... I don't see any CI/release pipeline change in this PR that actually publishes a crds.yaml release asset? The only CI change is adding the prepare-charts.sh step to push-artifacts.yaml, which copies CRDs into the chart's crds/ directory — it doesn't create a standalone crds.yaml artifact.

The rest of this work when combined with publishing the crds.yaml would give non-helm users a complete solution.  We should definitely do this.

gflarity: Hmm... I don't see any CI/release pipeline change in this PR that actually publishes a crds.yaml release asset? The only CI change is adding the prepare-charts.sh step to push-artifacts.yaml, which copies CRDs into the chart's crds/ directory — it doesn't create a standalone crds.yaml artifact.

docs/proposals/436-crd-upgrader/README.md
```diff
@@ -16,264 +16,240 @@
...
+As a platform engineer who prefers a fully automated upgrade path, I set `crdInstaller.enabled: true` once and from that point forward a single `helm upgrade grove` upgrades both the CRDs and the operator, in the correct order, without any manual steps.
 
 #### Story 3
 
-As a platform engineer with strict change-management requirements, I want to disable Grove's automated CRD management entirely. I will source the CRD manifests directly from the Grove GitHub release, review and approve them through our standard change process, and apply them to the cluster myself before upgrading the operator. I need the operator to start normally without attempting to touch CRDs.
+As a GitOps engineer managing Grove via ArgoCD or Flux with rendered manifests from `helm template`, I set `crdInstaller.enabled: true` so that CRD upgrades happen automatically when the operator Deployment rolls out, without needing a separate CRD application or sync-wave ordering.
```
gflarity: This story covers upgrades, but helm template users also need a fresh-install path. helm template doesn't render the crds/ directory unless you include ```--include-crds```.  As mentioned above, some docs on this would be great.

operator/hack/prepare-charts.sh
```diff
@@ -0,0 +1,58 @@
...
+    if [ ! -f ${src_crd_path} ]; then
+      echo "CRD ${crd} not found in ${src_crd_path}, run 'make generate' first"
+      make generate
+    fi
+    echo "Copying CRD ${crd} to ${target_path}"
+    cp ${src_crd_path} ${target_path}
+  done
+
+  echo "Copying scheduler CRDS..."
+  declare -a crds=("scheduler.grove.io_podgangs.yaml")
```
gflarity: Nit: The `crds` array variable is declared a second time here, shadowing the one on line 33. Consider using different variable names (e.g., `operator_crds` / `scheduler_crds`) for clarity.

docs/proposals/436-crd-upgrader/README.md
```diff
@@ -16,264 +16,240 @@
...
+Grove currently places its CRDs (`PodCliqueSet`, `PodClique`, `PodCliqueScalingGroup`, `PodGang`, `ClusterTopology`) under the Helm `crds/` directory. Helm 3 installs CRDs on `helm install` but silently skips them on `helm upgrade`. This GREP introduces a `crdInstaller` opt-in mechanism: an init container in the grove operator Deployment that applies all CRDs via server-side apply before the operator process starts. The default is `crdInstaller.enabled: false` — fresh installs continue to work via the `crds/` directory unchanged, and users who upgrade manually retain full control. Users who want automated CRD upgrades on every `helm upgrade` set `crdInstaller.enabled: true`.
```
gflarity, line 38: We should also call out that some users don't use helm install, instead they use `helm template` , kustomize etc.

docs/proposals/436-crd-upgrader/README.md
```diff
@@ -16,264 +16,240 @@
...
 
+This GREP does not force a new behaviour on all users. It adds a supported, tested, opt-in path to automated CRD upgrades while preserving the existing manual upgrade path as the default.
+
 ### Goals
 
-* Ensure CRDs are applied/upgraded before the operator process starts on every deployment and every restart.
-* Require zero manual steps from users for CRD lifecycle management during normal upgrades.
-* Work correctly for all deployment models: native Helm, `helm template` with kustomize/ArgoCD/Flux, and raw `kubectl apply`.
-* Allow users to opt out of automated CRD management and take full ownership of CRD lifecycle themselves.
+* Preserve existing behaviour on fresh install — the `crds/` directory installs CRDs as before, with no user-visible change.
```
gflarity, line 50: Another goal I'd add in addition to upgrades, is making installation straight forward if they don't use the helm install deployment toolchain as well. All the pieces are in place PR, just need some documentation.

You can get the CRD yamls with 

```bash
helm template --include-crds
```

Or as mentioned if we create crds.yaml as well.

docs/proposals/436-crd-upgrader/README.md
```diff
@@ -16,264 +16,240 @@
...
+
+**Opt-in path (automated):** Users set `crdInstaller.enabled: true` in their values. A `crd-installer` init container is injected into the grove operator Deployment. On every Pod start — including rollouts triggered by `helm upgrade` — the init container runs the `grove-install-crds` image, which applies all five Grove CRDs via server-side apply before the main operator container starts. Kubernetes enforces that the init container completes successfully before the operator process runs, making CRD upgrade ordering a property of the Deployment itself rather than of the deployment toolchain.
 
-The `crds/` directory is removed from the `grove` chart. The init container becomes the sole owner of CRD lifecycle for both fresh installs and upgrades.
+When `crdInstaller.enabled: true` the RBAC rules for `apiextensions.k8s.io` are conditionally added to the operator ClusterRole.
```
gflarity:
```suggestion
When `crdInstaller.enabled: true` the RBAC rules for `apiextensions.k8s.io` are conditionally added to the operator ClusterRole. This  provides the necessary RBAC for the init container to install the CRDs.
```

docs/proposals/436-crd-upgrader/README.md
danbar2: Removed, /crds folder is already in the artifact. It was an idea that is not really needed

docs/proposals/436-crd-upgrader/README.md
danbar2: you are correct, added a note on that.

docs/proposals/436-crd-upgrader/README.md
danbar2: Added a note on that, do you think anything else is needed?

gflarity: I think this is a good start. I've created a github issue tracking more comprehensive docs. No need to "boil the ocean" in this PR. 

https://github.com/ai-dynamo/grove/issues/536

gflarity (approved):

shayasoolin (approved):

PR #533, opened 2026-04-15
add cache options for pods managed by grove only

### LINKED ISSUE #530: Update grove-operator to use label-filtered cache for the pod informer
### What happened?

We have a large k8s cluster that has many tens of thousands of pods at all times. Only a very small number of these are Dynamo/Grove PodCliques. Currently, the Grove operator creates an unfiltered cluster-wide pod informer caches **all pods in memory**, meaning that it needs an excessively large memory reservation or risks getting OOMkilled.

The PodClique controller watches `v1.Pod` via `.Owns(&corev1.Pod{})` (internal/controller/podclique/register.go:66). While event predicates correctly filter reconciliation to only grove-managed pods, the underlying controller-runtime cache still lists/watches all pods cluster-wide because no Cache config is set in `createManagerOptions()` (internal/controller/manager.go:87).

What grove could do instead is add a cache configuration with a label selector for the Pod GVK in createManagerOptions():

```go
  opts.Cache = cache.Options{
      ByObject: map[client.Object]cache.ByObject{
          &corev1.Pod{}: {
              Label: labels.SelectorFromSet(labels.Set{
                  "app.kubernetes.io/managed-by": "grove-operator",
              }),
          },
      },
  }
```

This limits the Pod informer to only pods with the `app.kubernetes.io/managed-by=grove-operator` label, reducing the cache by many orders of magnitude.

### Environment

- Kubernetes version
- Grove version: v0.1.0-alpha.7
- Scheduler details

oleg-kushniriov:
#### What type of PR is this?

  /kind feature

  Adds a label selector to the Pod informer cache so only pods with `app.kubernetes.io/managed-by: grove-operator` are listed/watched. On large clusters with tens of thousands of pods, the operator was caching every pod cluster-wide even though only a tiny fraction are grove-managed.
   This caused excessive memory usage and OOMKill risk.

  The fix adds a `cache.ByObject` configuration for `corev1.Pod` in `createManagerOptions()` that pushes filtering to the API server. Grove CRDs are not filtered because all instances are grove-managed by definition.

  Fixes #530

  - Event predicates (e.g. `isManagedPod()`) were already filtering which cached events trigger reconciliation, but the underlying informer still stored all pods. This change moves filtering server-side.
  - `TestCreateManager` now uses `envtest.Environment` instead of a fake REST host because `cache.ByObject` triggers API discovery during manager creation. It skips gracefully when kubebuilder binaries are unavailable, consistent with `TestRegisterControllers`.

release-note
  ```
  The operator's Pod informer now uses a label-selector cache (`app.kubernetes.io/managed-by: grove-operator`) 
so only grove-managed pods are listed and watched. 
On clusters with tens of thousands of pods, this reduces the operator's memory footprint by orders of magnitude. 
```
```
NONE
```

### CODE REVIEW

**Review by enoodle — DISMISSED** [review]

File: operator/internal/controller/manager_test.go [hunk 1]
```diff
@@ -313,6 +317,59 @@ func TestCreateManagerOptions(t *testing.T) {
 		assert.Empty(t, opts.PprofBindAddress)
 	})
 
+	// Test that cache is configured with a label selector for Pods
+	t.Run("pod cache filtered by managed-by label", func(t *testing.T) {
```
danbar2: please add a test that validate it actually gets to the controller and it ignores not managed pods and handles the managed pods.
It might require an E2E

oleg-kushniriov: added

operator/internal/controller/manager.go
```diff
@@ -118,6 +124,20 @@ func createManagerOptions(operatorCfg *configv1alpha1.OperatorConfiguration) ctr
...
+	return cache.Options{
+		ByObject: map[client.Object]cache.ByObject{
+			&corev1.Pod{}: {
+				Label: labels.SelectorFromSet(labels.Set{
+					apicommon.LabelManagedByKey: apicommon.LabelManagedByValue,
+				}),
+			},
+		},
+	}
+}
```
renormalize, line 148: Do we take this opportunity to also include other resources that grove creates and therefore keeps a watch on, like `ServiceAccount`, `Role`, `RoleBinding`, `Service`?

I doubt that there would ever be so many of these resources in a cluster that it would start causing trouble the way pods do in #530; but no harm done if we include those resources as well, I suppose?

Any thoughts @oleg-kushniriov @danbar2?

oleg-kushniriov: It is a good point, but I would suggest implementing it in another PR. There is no immediate benefit adding those CRs as their number is significantly lower than amount of pods.

danbar2 (dismissed):

operator/internal/controller/manager.go
oleg-kushniriov: Eventually I did add ServiceAccount, Role, RoleBinding, Service and HorizontalPodAutoscaler to cache too

enoodle (dismissed):

danbar2 (dismissed):

operator/e2e/tests/operator_infra_test.go
```diff
@@ -0,0 +1,177 @@
...
+	require.NoError(t, err, "Failed to create unmanaged HPA")
+	defer func() {
+		_ = clientset.AutoscalingV2().HorizontalPodAutoscalers(ns).Delete(ctx, "unmanaged-hpa", metav1.DeleteOptions{})
+	}()
+
+	Logger.Info("4. Verify the operator does not modify any unmanaged resource")
+	// Wait briefly to give the operator a chance to act (it shouldn't)
+	time.Sleep(10 * time.Second)
+
+	// Helper to assert a resource was not touched by the operator
```
shayasoolin: This assertion is insufficient, this test would have passed even before this code change - the grove-operator doesn't touch those unmanaged resourcecs, does it?
What I think should be asserted, is that the reconciliation loop doesn't even take place, i.e. the filter is effective and no reconciliation is triggered when unmanaged resources are created or updated.

operator/internal/controller/manager_test.go
```diff
@@ -691,3 +775,89 @@ func TestRegisterControllersAndWebhooks(t *testing.T) {
...
+
+	require.NoError(t, directClient.Create(ctx, managedPod))
+	require.NoError(t, directClient.Create(ctx, unmanagedPod))
+
+	// Verify via the direct client that both pods exist in the API server.
+	var allPods corev1.PodList
+	require.NoError(t, directClient.List(ctx, &allPods, client.InNamespace(ns.Name)))
+	require.Len(t, allPods.Items, 2, "both pods should exist in the API server")
+
+	// Query the cache — it should only return the managed pod.
```
shayasoolin, line 852: Maybe this test covers what I see as missing in the e2e test. If it does, we still need to think if there's an added value of the new e2e test (suite).

shayasoolin (approved):

enoodle (approved):

PR #541, opened 2026-04-19
Add Annotations field to PodCliqueScalingGroupConfig with MNNVL propa…

### LINKED ISSUE #417: Enhancement: MNNVL V2, ComputeDomain Injection Via Annotation

shmuel-runai:
…gation

* Extend the PCS API with an Annotations field on PodCliqueScalingGroupConfig, allowing PCSG-level.
* MNNVL annotations (auto-mnnvl,mnnvl-group) are propagated from PCS to PCSG when not explicitly set in the config.
* Webhook validation enforces annotation rules and immutability on PCSG configs at create and update time.

/kind feature

Add Annotations field to PodCliqueScalingGroupConfig with MNNVL propagation

* Extend the PCS API with an Annotations field on PodCliqueScalingGroupConfig, allowing PCSG-level.
* MNNVL annotations (auto-mnnvl,mnnvl-group) are propagated from PCS to PCSG when not explicitly set in the config.
* Webhook validation enforces annotation rules and immutability on PCSG configs at create and update time.

Fixes #417


operator/internal/controller/podcliqueset/components/podcliquescalinggroup/podcliquescalinggroup_test.go
```diff
@@ -495,33 +541,87 @@ func TestBuildResource(t *testing.T) {
...
+			expectedAnnotations: map[string]string{
+				mnnvl.AnnotationMNNVLGroup: "training",
+				mnnvl.AnnotationAutoMNNVL:  mnnvl.AnnotationAutoMNNVLEnabled,
+			},
 		},
 		{
-			description:            "Other annotations on PCS without MNNVL do not propagate MNNVL to PCSG",
-			pcsAnnotations:         map[string]string{"some-other-annotation": "value"},
-			expectedPCSGAnnotation: "",
+			description:    "PCSG config has auto-mnnvl — config takes priority over PCS",
```
danbar2, line 591: Im missing a test that validate the whole propagation, PCLQ over PCSG over PCS for both annotations

shmuel-runai: You are not missing it, this isn't part of the PR
This PR is just for "adding annotations to PCSG config".
But for the code to be BC with the current code (even just phase 1), we need to propagate the auto-mnnvl annotation. And since we are doing that, let's propagate the mnnvl-group as it just one extra line (and more TCs :| ).
And for the code to be aligned with the current status, we also added webhook validation.

danbar2 (approved):

shayasoolin (approved):

Issue #408, opened 2026-02-08
Measure and optimize DeepCopy operations during controller-runtime CreateOrPatch operations


unmarshall:
### What you would like to be added?

When `controllerutil.CreateOrPatch` is invoked it calls the following:
```go
	objPatch := client.MergeFrom(obj.DeepCopyObject().(client.Object))
	statusPatch := client.MergeFrom(obj.DeepCopyObject().(client.Object))
```
It then converts the objects to `unstructure.Unstructured` and does a nested field copy
```go
	before, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj.DeepCopyObject())
	if err != nil {
		return OperationResultNone, err
	}

	// Attempt to extract the status from the resource for easier comparison later
	beforeStatus, hasBeforeStatus, err := unstructured.NestedFieldCopy(before, "status")
	if err != nil {
		return OperationResultNone, err
	}
```

Other expensive calls are `reflect.DeepEqual(before, after)` as well. Need to investigate on how to optimize this since there can be large number of invocations of this method.

### Why is this needed?

Grove at scale need to handle large number of PodCliqueSet, each of which can contain multiple PodClique's. Optimizing costly operations will help ensure that the CPU consumption is kept well within limits.

PR #545, opened 2026-04-20
perf: reduce reconcile CPU on the 1000-pod scale test

danbar2:
## Summary
Solves #408 

Six optimizations in the PCS / PCSG / PodClique reconcile paths. Measured on the 500-replica / 1000-pod scale test (k3d + 100 KWOK nodes, operator concurrency=20).

> An earlier revision of this PR included a per-PodClique / per-PCSG spec-hash short-circuit (annotation `grove.io/spec-hash`). It was dropped after CR feedback flagged the hash as fragile (hand-picked field selection, MNNVL exclusion that depends on a webhook invariant, drift risk if `buildResource` changes silently). The remaining changes capture the same wall-clock improvement on this benchmark; steady-state CPU regresses ~10 ms/sec versus the short-circuited variant but stays ~50% under baseline.

### Wall-clock

| Phase | Baseline | This branch | Δ |
|-------|----------|-------------|---|
| **Total** | 128.0s | 110.9s | **−13%** |
| pcs-deleted | 50.0s | 37.9s | **−24%** |
| pods-ready | 47.8s | 42.8s | **−10%** |

### CPU — deploy phase (full pprof, 500 PodCliques + 500 PodGangs being created)

| Function | Baseline | This branch | Δ |
|----------|----------|-------------|---|
| **Total CPU samples** | **14.98s** | **≈10s** | **≈−30%** |
| `GetPCLQPods` (list + label-match) | 4.58s | 1.54s | **−66%** |
| podclique `reconcileSpec` | 3.68s | 2.73s | **−26%** |
| podclique `reconcileStatus` | 3.57s | below top-10 | **≈−100%** (no-op skip) |
| `syncPCLQResources` | 2.99s | 1.88s | **−37%** |
| `GetPodCliqueSet` | ≈1–2s | 60ms | **−97%** |

### CPU — steady-state phase (30s window, single no-op reconcile triggered by annotation patch)

The common production case: a watch event fires but nothing actually changed.

| Function | Baseline | This branch | Δ |
|----------|----------|-------------|---|
| **Total CPU samples** | **680ms** | **370ms** | **−46%** |
| `controllerutil.CreateOrPatch` (PCS path) | 330ms | 180ms | **−45%** |
| PCS PodClique `doCreateOrUpdate` | 260ms | 130ms | **−50%** |
| PCS `Reconcile` | 140ms | below top-10 | **≈−100%** |

### A note on wall-clock vs. CPU

The wall-clock reduction is somewhat smaller than the CPU reduction because this test is not CPU-bound: the operator spends most of its wall-clock time waiting on kube-apiserver latency, informer cache syncs, and KWOK's own pod-scheduling delays. Those don't shrink just because our client-side code got faster. Where the CPU savings pay off in production:
- **Operator headroom**: lower baseline CPU means less resource pressure under bursts and tighter container limits
- **Multi-tenant density**: at N concurrent PCSes the per-reconcile savings scale with N
- **Event storms**: steady-state −46% on a no-op reconcile is exactly the common case where an unrelated watch event wakes the controller
- **Larger clusters**: at 10k+ PodCliques the CPU-per-reconcile would start gating wall-clock

All unit tests pass. No semantic changes to reconcile behavior (behavior is asserted via `equality.Semantic.DeepEqual` before skipping status writes, ownership still validated by `metav1.IsControlledBy`, etc.).

## What's in the PR

**Parallel component sync** — three dependency-ordered groups in PCS `reconcilespec.go`:
1. RBAC + infra (ServiceAccount, Role, RoleBinding, SATokenSecret, HeadlessService, HPA, PCSReplica, ComputeDomain, ResourceClaim)
2. PodClique
3. PCSG + PodGang

Instead of 12 sequential Syncs. Mostly benefits the delete phase (delete wall-clock −24%) where cleanup tasks overlap.

**`GetPCLQPods` narrowed selector** — `MatchingLabels{LabelPodClique}` only. That label is unique cluster-wide, so the parent managed-by/part-of labels only added per-pod `labels.Set.Lookup` work for no filtering benefit. Ownership is still validated by `metav1.IsControlledBy`.

**ctx-scoped memoization** — `WithPCLQPodsCache` and `WithPodCliqueSetCache` in `internal/controller/common/component/utils/`. PodClique's reconcileSpec and reconcileStatus share one pod list; `GetPodCliqueSet` drops from 4 calls per PodClique reconcile (and 3 per PCSG reconcile) to 1.

**No-op status skip** — in `podclique/pcs/pcsg reconcileStatus`, snapshot status before mutations, skip the API round-trip when `equality.Semantic.DeepEqual` says nothing changed.

## Test additions

To make the steady-state CPU measurable in a scale test:

- `operator/e2e/measurement/condition/timer.go` — `TimerCondition` holds the measurement window open for a fixed duration after a trigger phase fires.
- `WorkloadManager.TriggerPCSReconcile` in `operator/e2e/grove/workload/workload.go` — bumps `grove.io/reconcile-trigger` annotation to force a no-op reconcile cycle without touching the spec.
- New `steady-state-reconcile` phase in `operator/e2e/tests/scale/scale_test.go` — patches the annotation, waits 30s, and pprof captured during that window isolates the cache-hit reconcile cost.

### DIFF
--- a/operator/e2e/measurement/condition/timer.go
+++ b/operator/e2e/measurement/condition/timer.go
@@ -1,0 +1,56 @@
+// /*
+// Copyright 2026 The Grove Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// */
+
+package condition
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+// TimerCondition is met once Duration has elapsed since the first call to Met. It is used
+// to keep a measurement window open for a fixed duration after a trigger phase fires.
+type TimerCondition struct {
+	Duration time.Duration
+
+	startOnce bool
+	startedAt time.Time
+}
+
+// Met implements measurement.MilestoneCondition. It returns true once Duration has
+// elapsed since the first Met call. The ctx and error return are unused but required
+// by the interface.
+func (c *TimerCondition) Met(_ context.Context) (bool, error) {
+	if !c.startOnce {
+		c.startedAt = time.Now()
+		c.startOnce = true
+	}
+	return time.Since(c.startedAt) >= c.Duration, nil
+}
+
+// Progress implements measurement.ProgressReporter and reports the time remaining
+// until the timer fires. The ctx is unused but required by the interface.
+func (c *TimerCondition) Progress(_ context.Context) string {
+	if !c.startOnce {
+		return fmt.Sprintf("%s remaining", c.Duration)
+	}
+	remaining := c.Duration - time.Since(c.startedAt)
+	if remaining < 0 {
+		remaining = 0
+	}
+	return fmt.Sprintf("%s remaining", remaining.Round(time.Second))
+}
--- a/operator/e2e/tests/scale/scale_test.go
+++ b/operator/e2e/tests/scale/scale_test.go
@@ -72,8 +72,12 @@ const (
 	scaleTestYAMLPath  = "../../yaml/scale-test-1000.yaml"
 	scaleTestNamespace = "default"
 
-	runIDTimeFormat    = "20060102-150405"
+	runIDTimeFormat   = "20060102-150405"
 	outputResultsFile = "scale-test-results.json"
+
+	// steadyStateWindow keeps the pprof/measurement window open after a no-op reconcile
+	// trigger so the full ~500-PodClique spec-hash-short-circuit burst has time to run.
+	steadyStateWindow = 30 * time.Second
 )
 
 func Test_ScaleTest_1000(t *testing.T) {
@@ -164,6 +168,34 @@ func Test_ScaleTest_1000(t *testing.T) {
 					Namespace:     tc.Namespace,
 					ExpectedCount: scaleTestExpectedReplicas,
 				},
+			},
+		},
+	})
+
+	// steady-state-reconcile: patch a metadata annotation to force one reconcile cycle
+	// without touching spec. With the spec-hash short-circuit in place, the PCS→PodClique
+	// update path should fire cache hits for every PodClique. pprof captured during this
+	// window isolates the no-op reconcile cost.
+	steadyStateTriggerID := fmt.Sprintf("steady-%s", runID)
+	tracker.AddPhase(measurement.PhaseDefinition{
+		Name: "steady-state-reconcile",
+		ActionFn: func(ctx context.Context) error {
+			Logger.Info("triggering no-op PCS reconcile")
+			return workload.NewWorkloadManager(tc.Client, Logger).TriggerPCSReconcile(ctx, tc.Namespace, tc.Workload.Name, steadyStateTriggerID)
+		},
+		Milestones: []measurement.MilestoneDefinition{
+			{
+				Name: "pcs-still-available",
+				Condition: &condition.PCSAvailableCondition{
+					Client:        tc.Client.Client,
+					Name:          tc.Workload.Name,
+					Namespace:     tc.Namespace,
+					ExpectedCount: scaleTestExpectedReplicas,
+				},
+			},
+			{
+				Name:      "steady-state-window",
+				Condition: &condition.TimerCondition{Duration: steadyStateWindow},
 			},
 		},
 	})
--- a/operator/internal/controller/common/component/utils/pod.go
+++ b/operator/internal/controller/common/component/utils/pod.go
@@ -25,24 +25,51 @@ import (
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// GetPCLQPods lists all Pods that belong to a PodClique
-func GetPCLQPods(ctx context.Context, cl client.Client, pcsName string, pclq *grovecorev1alpha1.PodClique) ([]*corev1.Pod, error) {
+// pclqPodsCacheKey is a context key for per-reconcile memoization of GetPCLQPods results.
+// A PodClique reconcile calls GetPCLQPods from both reconcileSpec and reconcileStatus;
+// without the cache we list+filter the informer cache twice per reconcile. Stash the result
+// on a fresh ctx via WithPCLQPodsCache and the second call reuses it.
+type pclqPodsCacheKey struct{}
+
+// pclqPodsCache holds one slot per PodClique UID. UIDs are used rather than names so a
+// stale cache entry can never bleed into a recreated object with the same name.
+type pclqPodsCache struct {
+	byUID map[types.UID][]*corev1.Pod
+}
+
+// WithPCLQPodsCache returns a context that memoizes GetPCLQPods results for the lifetime of
+// one reconcile. Call this once at the top of Reconcile and propagate the returned context.
+func WithPCLQPodsCache(ctx context.Context) context.Context {
+	return context.WithValue(ctx, pclqPodsCacheKey{}, &pclqPodsCache{byUID: make(map[types.UID][]*corev1.Pod, 1)})
+}
+
+// GetPCLQPods lists all Pods that belong to a PodClique.
+// The selector uses only LabelPodClique because the PodClique name is unique across the
+// cluster; filtering on the parent managed-by/part-of labels would add two more per-pod
+// labels.Set.Lookup calls per match with no correctness benefit. Ownership is still
+// validated by IsControlledBy below.
+//
+// When ctx carries a cache from WithPCLQPodsCache, the first call populates the cache and
+// subsequent calls in the same reconcile return the cached slice without hitting the
+// informer.
+func GetPCLQPods(ctx context.Context, cl client.Client, _ string, pclq *grovecorev1alpha1.PodClique) ([]*corev1.Pod, error) {
+	cache, _ := ctx.Value(pclqPodsCacheKey{}).(*pclqPodsCache)
+	if cache != nil {
+		if pods, ok := cache.byUID[pclq.UID]; ok {
+			return pods, nil
+		}
+	}
 	podList := &corev1.PodList{}
 	if err := cl.List(ctx,
 		podList,
 		client.InNamespace(pclq.Namespace),
-		client.MatchingLabels(
-			lo.Assign(
-				apicommon.GetDefaultLabelsForPodCliqueSetManagedResources(pcsName),
-				map[string]string{
-					apicommon.LabelPodClique: pclq.Name,
-				},
-			),
-		)); err != nil {
+		client.MatchingLabels{apicommon.LabelPodClique: pclq.Name},
+	); err != nil {
 		return nil, err
 	}
 	ownedPods := make([]*corev1.Pod, 0, len(podList.Items))
@@ -50,6 +77,9 @@ func GetPCLQPods(ctx context.Context, cl client.Client, pcsName string, pclq *gr
 		if metav1.IsControlledBy(&pod, pclq) {
 			ownedPods = append(ownedPods, &pod)
 		}
+	}
+	if cache != nil {
+		cache.byUID[pclq.UID] = ownedPods
 	}
 	return ownedPods, nil
 }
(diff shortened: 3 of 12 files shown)


operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -280,6 +302,34 @@ func (r _resource) doCreateOrUpdate(ctx context.Context, logger logr.Logger, pcs
 	return nil
 }
 
+// computePCLQSpecHashForPCSG returns a stable hash of every input buildResource uses to
+// configure a PodClique owned by the PodCliqueScalingGroup. Returns "" when the template
+// cannot be located (signals "unknown, do not skip"). templateHash is the precomputed
+// ComputePCLQPodTemplateHash value passed through from syncContext so dump.ForHash runs
+// once per template instead of once per (replica × template).
+func computePCLQSpecHashForPCSG(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pcsgReplicaIndex int, pclqObjectKey client.ObjectKey, templateHash string) string {
```
shayasoolin: The implementation seems to be fragile as it strongly depends on how `buildResource` is implemented and which exact fields it is using or not using.
Another example is the note on `MNNVL annotations are immutable on update so they cannot change the hash` - what if this decision changes and we forget to update this hash calculation?
I'm concerned about potential reconciliation bugs that will be very hard to debug.
Is there a more generic way to have this kind of optimization without risking the (future) correctness?

shayasoolin: Also, a general verification I'd like to make - the fact that no spec (hash) changes are detected doesn't mean no reconciliation is required, right? For example, some child resource might have been manually deleted and need to be recreated - unrelated to spec changes. Just making sure this kind of reconcile loops are NOT skipped with the new additions.

operator/internal/controller/podcliqueset/components/podclique/podclique.go
```diff
@@ -258,13 +278,36 @@ func (r _resource) Delete(ctx context.Context, logger logr.Logger, pcsObjectMeta
 }
 
 // doCreateOrUpdate creates or updates a single PodClique resource.
-func (r _resource) doCreateOrUpdate(ctx context.Context, logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet, pcsReplica int32, pclqObjectKey client.ObjectKey, pclqExists bool) error {
+// cachedSpecHash is the grove.io/spec-hash annotation value read from the list call
+// (empty when the PodClique is new or the annotation is absent). When it matches the
+// desired hash the reconcile is a pure no-op: no Get, no patch.
+// templateHashes is a precomputed map of template-name → PodTemplateHash so each reconcile
+// pays the dump.ForHash cost once per template instead of once per (replica × template).
+func (r _resource) doCreateOrUpdate(ctx context.Context, logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet, pcsReplica int32, pclqObjectKey client.ObjectKey, pclqExists bool, cachedSpecHash string, templateHashes map[string]string) error {
```
shayasoolin: Repeated code, do you see a reasonable way to make it common?

operator/internal/controller/podcliqueset/components/podclique/podclique.go
```diff
@@ -280,6 +323,30 @@ func (r _resource) doCreateOrUpdate(ctx context.Context, logger logr.Logger, pcs
 	return nil
 }
 
+// computePCLQSpecHashForPCS returns a stable hash of every input buildResource uses to
+// configure a PodClique owned directly by the PodCliqueSet. Returns "" when the template
+// cannot be located (signals "unknown, do not skip"). templateHashes provides a precomputed
+// ComputePCLQPodTemplateHash value per template-name so the expensive dump.ForHash runs
+// once per template rather than once per (replica × template).
+func computePCLQSpecHashForPCS(pcs *grovecorev1alpha1.PodCliqueSet, pcsReplica int32, pclqObjectKey client.ObjectKey, templateHashes map[string]string) string {
```
shayasoolin: Same concern as commented for the PCSG case.

shayasoolin (dismissed):

operator/e2e/measurement/condition/timer.go
```diff
@@ -0,0 +1,53 @@
...
+// to keep a measurement window open for a fixed duration after a trigger phase fires.
+type TimerCondition struct {
+	Duration time.Duration
+
+	startOnce bool
+	startedAt time.Time
+}
+
+// Met returns true once Duration has elapsed since the first Met call.
+func (c *TimerCondition) Met(_ context.Context) (bool, error) {
```
shmuel-runai, line 37: nit: Why do we need the context arg? Or returning an error?
Maybe TimerCondition implements an interface? if so write it in a comment 
// TimerCondition implements ....

danbar2: yes, it implements a MilestoneCondition

operator/e2e/measurement/condition/timer.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+// TimerCondition is met once Duration has elapsed since the first call to Met. It is used
+// to keep a measurement window open for a fixed duration after a trigger phase fires.
+type TimerCondition struct {
```
shmuel-runai, line 27: very nit: not sure how much like this struct, not the name, and not the methods
it's is more like a stopper or maybe a "passive timer."
but I'm not sure that my suggestion is better

operator/e2e/tests/scale/scale_test.go
```diff
@@ -168,6 +172,34 @@ func Test_ScaleTest_1000(t *testing.T) {
 		},
 	})
 
+	// steady-state-reconcile: patch a metadata annotation to force one reconcile cycle
```
shmuel-runai, line 175: steady-state-reconcile :)
I like the sound of that

operator/e2e/measurement/condition/timer.go
danbar2: its a timer condition - if you want to set a time based condition on your scale test this is the struct that helps you. it implement the Condition interfaces with Time has Met and what is the progress - how much time has elapsed

operator/internal/controller/common/component/utils/pod.go
```diff
@@ -25,24 +25,51 @@ import (
...
+
+// pclqPodsCache holds one slot per PodClique UID. UIDs are used rather than names so a
+// stale cache entry can never bleed into a recreated object with the same name.
+type pclqPodsCache struct {
+	byUID map[types.UID][]*corev1.Pod
+}
+
+// WithPCLQPodsCache returns a context that memoizes GetPCLQPods results for the lifetime of
+// one reconcile. Call this once at the top of Reconcile and propagate the returned context.
+func WithPCLQPodsCache(ctx context.Context) context.Context {
```
shmuel-runai, line 47: Do we know the number of pods ahead? Can we pass it and initialize the map with a real size estimation?

operator/internal/controller/common/component/utils/podcliqueset.go
```diff
@@ -61,17 +61,47 @@ func isStandalonePCLQ(pcs *grovecorev1alpha1.PodCliqueSet, pclqName string) bool
...
+type pcsCacheKey struct{}
+
+// pcsCache holds one slot keyed by "namespace/name".
+type pcsCache struct {
+	byKey map[string]*grovecorev1alpha1.PodCliqueSet
+}
+
+// WithPodCliqueSetCache returns a context that memoizes GetPodCliqueSet results for the
+// lifetime of one reconcile. Call this once at the top of Reconcile and propagate the ctx.
+func WithPodCliqueSetCache(ctx context.Context) context.Context {
```
shmuel-runai, line 76: see other cache comment

operator/internal/controller/common/component/utils/pod.go
danbar2: it memoizes the result of GetPCLQPods of the same PCLQ for the reconcile context, it is always one item as there is one result. its a map from PCLQ id to GetPCLQPods result

operator/internal/controller/common/component/utils/podcliqueset.go
danbar2: this is a single PCS always, it caches the parent.

operator/internal/controller/podcliqueset/reconcilestatus.go
```diff
@@ -53,6 +57,10 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	// mirror UpdateProgress to the deprecated RollingUpdateProgress field for backward compatibility.
 	mirrorUpdateProgressToRollingUpdateProgress(pcs)
 
+	if equality.Semantic.DeepEqual(*originalStatus, pcs.Status) {
```
shmuel-runai, line 66: you are missing the comment on why it ok to skip.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -74,6 +76,10 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	// mirror UpdateProgress to the deprecated RollingUpdateProgress field for backward compatibility.
 	mirrorUpdateProgressToRollingUpdateProgress(pcsg)
 
+	if equality.Semantic.DeepEqual(*originalStatus, pcsg.Status) {
```
shmuel-runai, line 86: you are missing the comment on why it ok to skip.

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -149,35 +151,84 @@ func (r *Reconciler) initUpdateProgress(ctx context.Context, pcs *grovecorev1alp
...
+		return ctrlcommon.ReconcileAfter(constants.ComponentSyncRetryInterval, fmt.Sprintf("requeueing sync due to components(s) %v after %s", continueReconcileAndRequeueKinds, constants.ComponentSyncRetryInterval))
+	}
+	return ctrlcommon.ContinueReconcile()
+}
+
+// syncKindGroup runs the Sync operation for each kind in the group concurrently.
+// Returns a non-nil ReconcileStepResult if reconciliation should stop for this group,
+// plus the list of kinds that returned "continue and requeue" errors.
+func (r *Reconciler) syncKindGroup(ctx context.Context, logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet, group []component.Kind, groupIdx int) (*ctrlcommon.ReconcileStepResult, []component.Kind) {
+	var (
```
shmuel-runai, line 175: a nice point to make here (which I wasn't aware). 
is why we define the variables in a var block. where 99% we don't/

Well, it is a common practice in Go to declare variables in a var block where there is a mutex that "protects"  those variables.

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -149,35 +151,84 @@ func (r *Reconciler) initUpdateProgress(ctx context.Context, pcs *grovecorev1alp
...
-				return ctrlcommon.ReconcileAfter(constants.ComponentSyncRetryInterval, err.Error())
-			}
-			logger.Error(err, "failed to sync PodCliqueSet resources", "kind", kind)
-			return ctrlcommon.ReconcileWithErrors("error syncing managed resources", fmt.Errorf("failed to sync %s: %w", kind, err))
-		}
+		tasks = append(tasks, utils.Task{
+			Name: fmt.Sprintf("SyncKind-%s", kind),
+			Fn: func(ctx context.Context) error {
+				logger.Info("Syncing PodCliqueSet resource", "kind", kind, "group", groupIdx)
+				if err := operator.Sync(ctx, logger, pcs); err != nil {
```
shmuel-runai: this was hard to read, please add comments

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -149,35 +151,84 @@ func (r *Reconciler) initUpdateProgress(ctx context.Context, pcs *grovecorev1alp
...
+					if ctrlutils.ShouldContinueReconcileAndRequeue(err) {
+						mu.Lock()
```
shmuel-runai: nit: I don't like all of those lock and unlock.  suggest one 
mu.lock()
defar m.unlock()

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -149,35 +151,84 @@ func (r *Reconciler) initUpdateProgress(ctx context.Context, pcs *grovecorev1alp
...
+						requeuedKinds = append(requeuedKinds, kind)
+						mu.Unlock()
+						logger.Info("component requested post-sync requeue", "kind", kind, "message", err.Error())
+						return nil
```
shmuel-runai: like here, a comment why it ok to slicnce the error

shmuel-runai (commented):
Great work 
Fix it, Ship it

shmuel-runai (approved):

shayasoolin (approved):

PR #555, opened 2026-04-27
Add repo-managed git hooks for local validation

### LINKED ISSUE #554: Add repo-managed git hooks for local validation
Add tracked Git hooks to this repository, plus Make targets to install and uninstall them locally.

The hook behavior should be:

- `pre-commit` runs `make validate`
- `pre-push` runs `make test`

The local setup should be:

- `make install-hooks` sets `core.hooksPath` to `.githooks`
- `make uninstall-hooks` removes the local hook configuration

This keeps the hook logic versioned in the repository and aligned with the existing Make-based validation workflow.

## Why is this needed?
This adds a local guardrail against committing code that fails generation, formatting, linting, API doc generation, or docs verification, and against pushing code that fails tests.

It also helps keep AI agents in check by preventing them from committing or pushing code that does not meet the repository's standards. That should reduce avoidable CI load and let the CI system spend more time on changes that have already passed the local baseline checks.

The approach is intentionally local and opt-in per clone, so contributors can enable or remove it with a single Make target while still using the repository’s standard validation commands.

enoodle:
#### What type of PR is this?

/kind cleanup

Adds tracked Git hooks for local validation and Make targets to manage them locally.

- adds `.githooks/pre-commit` to run `make validate`
- adds `.githooks/pre-push` to run `make test`
- adds `make install-hooks` to set `core.hooksPath` to `.githooks`
- adds `make uninstall-hooks` to remove the local hook configuration
- splits the shared validation steps into `make validate`, while keeping `make check` as `validate` plus the clean-tree assertion

This keeps the hook logic versioned in the repository and aligned with the existing Make-based validation flow.

Fixes #554

The hooks are intentionally opt-in per clone.

`make validate` runs the same validation steps as `make check` without the final dirty-tree assertion, so generated or formatted changes can still be staged as part of a normal local workflow.

```release-note
NONE
```

```docs
NONE
```


danbar2 (dismissed):

.githooks/pre-commit
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
```
gflarity: NIt: Missing the repository's Apache license header. Existing shell scripts carry that header, and `make check` runs `add-license-headers` before the clean-tree assertion, so this file should include the header up front.

.githooks/pre-commit
```diff
@@ -0,0 +1,7 @@
...
+set -eu
+
+repo_root="$(git rev-parse --show-toplevel)"
+cd "$repo_root"
+
+exec make validate
```
gflarity: If `make validate` rewrites generated, formatted, or license-header files, the hook still exits successfully and Git proceeds with the commit using the pre-existing index. That leaves the validation changes unstaged after the commit. Please fail the hook when validation modifies the working tree, for example after a `git diff --quiet` check, so the contributor can review and stage the changes first.

.githooks/pre-push
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
```
gflarity: Nit: missing the repository's Apache license header. Existing shell scripts include the header, and the validation flow includes `add-license-headers`, so this hook should be committed with the expected header already present.

Makefile
```diff
@@ -111,6 +115,18 @@ run-e2e:
 test: test-unit
 	@echo "> All tests passed"
 
+.PHONY: install-hooks
+install-hooks:
+	@echo "> Configuring local git hooks from .githooks"
+	@git config --local core.hooksPath .githooks
```
gflarity, line 121: nit: `install-hooks` overwrites any existing local `core.hooksPath`, and `uninstall-hooks` later unsets whatever value is present. That can silently disable a contributor's custom hooks if they already had a hook path configured. Please preserve/refuse non-`.githooks` values, or warn clearly and document how to restore the previous setting.

gflarity (commented):
Just a few minor things.

.githooks/pre-commit
enoodle: Maybe I need to have pre commit hooks to run it :)

enoodle: But how come the CI passed on this without those headers? seems like it is not checking this file...

.githooks/pre-commit
enoodle: The problem is that sometimes you do have uncommitted changes that you don't want to commit. For example if you have untracked files or you are splitting a change into several commits. I will look into mitigating it , but maybe it is just a gap that we need to have to allow some flexibility.

.githooks/pre-push
enoodle: Again, seems like the add-license-headers didn't run on those files because the CI of this PR passed.

Makefile
enoodle: You are right. I don't think that preserving previous hooks should be part of this. I will look for a way to chain them or just warn about this.

.githooks/pre-commit
enoodle: I am going to add a test that will compare the (hash of the)  diff before / after running make validate to verify that it didn't change. this way we can have a similar effect but also keep flexibility that is important to me :)

gflarity (dismissed):

.githooks/pre-commit
renormalize: Rename the file to `pre-commit.sh`. The tool does not add headers to files without any file extension in the filename.
I've renamed the files locally to test this, and the headers are added.

.githooks/pre-push
renormalize: https://github.com/ai-dynamo/grove/pull/555/changes#r3186363294

renormalize (dismissed):

.githooks/pre-commit
enoodle: renamed the files and added the headers

renormalize (approved):

danbar2 (approved):

PR #556, opened 2026-04-28
fix: cascade-delete PCS/PCSG/PCLQ via Kubernetes GC

### LINKED ISSUE #423: Slow Deletion at scale
## Problem

Deleting a PCSG with 5000 PodCliques takes 20 minutes due to low concurrency settings.

**Current bottleneck:**
- `concurrentSyncs: 3` - only 3 PodCliques reconcile at once
- Manual deletion + finalizer cleanup for each PodClique
- API rate limiting (QPS: 100, Burst: 150)

**Scale impact:**
- 5000 PodCliques: 20+ minutes

test this this yaml
```yaml
# Scale Test: ~5000 pods (625 prefill + 624 decode ScalingGroups + 3 frontend)
# Large scale test for validating extreme-scale behavior on KWOK cluster
---
apiVersion: grove.io/v1alpha1
kind: PodCliqueSet
metadata:
  name: scale-test-5000
  labels:
    app: scale-test-5000
spec:
  replicas: 1
  template:
    podCliqueScalingGroups:
      - name: prefill
        replicas: 625
        minAvailable: 625
        cliqueNames:
          - pleader
          - pworker
      - name: decode
        replicas: 625
        minAvailable: 625
        cliqueNames:
          - dleader
          - dworker
    cliques:
      - name: frontend
        labels:
          kai.scheduler/queue: default
        spec:
          roleName: frontend
          replicas: 3
          minAvailable: 3
          podSpec:
            nodeSelector:
              type: kwok
            tolerations:
              - key: fake-node
                operator: Equal
                value: "true"
                effect: NoSchedule
            containers:
              - name: frontend
                image: registry:5001/nginx:alpine-slim
                resources:
                  requests:
                    memory: 30Mi
                    cpu: 100m
      - name: pleader
        labels:
          kai.scheduler/queue: default
        spec:
          roleName: pleader
          replicas: 1
          minAvailable: 1
          podSpec:
            nodeSelector:
              type: kwok
            tolerations:
              - key: fake-node
                operator: Equal
                value: "true"
                effect: NoSchedule
            containers:
              - name: pleader
                image: registry:5001/nginx:alpine-slim
                resources:
                  requests:
                    memory: 30Mi
                    cpu: 100m
      - name: pworker
        labels:
          kai.scheduler/queue: default
        spec:
          roleName: pworker
          replicas: 3
          minAvailable: 3
          podSpec:
            nodeSelector:
              type: kwok
            tolerations:
              - key: fake-node
                operator: Equal
                value: "true"
                effect: NoSchedule
            containers:
              - name: pworker
                image: registry:5001/nginx:alpine-slim
                resources:
                  requests:
                    memory: 30Mi
                    cpu: 100m
      - name: dleader
        labels:
          kai.scheduler/queue: default
        spec:
          roleName: dleader
          replicas: 1
          minAvailable: 1
          podSpec:
            nodeSelector:
              type: kwok
            tolerations:
              - key: fake-node
                operator: Equal
                value: "true"
                effect: NoSchedule
            containers:
              - name: dleader
                image: registry:5001/nginx:alpine-slim
                resources:
                  requests:
                    memory: 30Mi
                    cpu: 100m
      - name: dworker
        labels:
          kai.scheduler/queue: default
        spec:
          roleName: dworker
          replicas: 3
          minAvailable: 3
          podSpec:
            nodeSelector:
              type: kwok
            tolerations:
              - key: fake-node
                operator: Equal
                value: "true"
                effect: NoSchedule
            containers:
              - name: dworker
                image: registry:5001/nginx:alpine-slim
                resources:
                  requests:
                    memory: 30Mi
                    cpu: 100m
```

### Option 1: Increase Concurrency

```yaml
controllers:
  podClique:
    concurrentSyncs: 50  # 16x increase
runtimeClientConnection:
  qps: 500
  burst: 1000
```

### Option 2: GC-Based Deletion

Remove manual deletion, let Kubernetes GC cascade deletionTimestamp automatically.

**Changes needed:**
- Remove `deletePodCliqueScalingGroupResources()` step
- PCSG finalizer removes itself immediately (don't wait for children)
- Trust PodClique finalizers for cleanup

**Result:** Instant cascade, scales to any size

**Tradeoff:** Requires finalizer logic changes

## Impact

Blocks scale testing and production use of large workloads.

Related to: #405

delete in sec not 20 mintues

### Environment

- Kubernetes version 1.32
- Grove version main
- Scheduler details default
- Tools that you are using Grove together with KWOK nodes

danbar2:
Closes #423.

## What this changes

Replaces the manual *delete-children → poll-until-clean → remove-finalizer* flow in the PCS, PCSG, and PodClique reconcilers with **immediate finalizer removal**. The Kubernetes garbage collector then cascades the entire owner-reference tree (PCS → PCSG → PC → Pod, plus the PCS-owned siblings: Service, ServiceAccount, Role, RoleBinding, HPA, PodGang, ResourceClaim).

The PodClique reconciler keeps a one-step finalizer to clear its in-memory `expectationsStore` entry — that state lives outside the API server and GC can't see it.

## The pattern: GC cascade + thin finalizer for non-API-server state

This is the dominant pattern across modern cloud-native CRDs that own a tree of K8s resources. Owner refs do the heavy lifting; the finalizer exists only as a hook for state GC can't reach.

Reference implementations using this pattern:

- **Kubernetes core controllers** — `ReplicaSet`, `Deployment`, `StatefulSet`, `DaemonSet`, `Job` all rely on owner-reference cascade for child cleanup. The expectations cache pattern Grove now mirrors lives in [`k8s.io/kubernetes/pkg/controller/controller_utils.go`](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go) and is consumed by `replica_set.go` and `statefulset/stateful_set.go`. They never manually `Delete` their pods on parent deletion.
- **[JobSet](https://github.com/kubernetes-sigs/jobset)** (kubernetes-sigs) — JobSet → Job → Pod via owner refs; the controller does no manual child deletion.
- **[LeaderWorkerSet (LWS)](https://github.com/kubernetes-sigs/lws)** (kubernetes-sigs) — LWS → StatefulSet → Pod, same pattern.
- **[Kueue](https://github.com/kubernetes-sigs/kueue)** — `Workload` carries a finalizer whose only job is to release queue/quota accounting before letting GC cascade.
- **[KubeRay](https://github.com/ray-project/kuberay)** — `RayCluster`/`RayService` use finalizers for service-mesh deregistration and graceful dashboard shutdown, but children clean up via owner refs.
- **[Volcano](https://github.com/volcano-sh/volcano)** — `PodGroup` finalizer releases gang-scheduling state; pods cascade via GC.
- **[KServe](https://github.com/kserve/kserve)** — `InferenceService` finalizer releases external model-server registration; the K8s subtree is cascaded.

What Grove was doing before is the older Gardener-style controller-utils template: a serialized `delete → verify → finalize` loop. Defensive when owner refs are unreliable, but it serializes through `concurrentSyncs` — which is exactly what made #423 a 20-minute deletion at 5k pods.

## Tradeoffs we're accepting

- **Parent disappears from the API as soon as its finalizer is removed,** even while children are still terminating. Anyone scripting around "PCS object gone = workload gone" should switch to watching pods.
- **No per-step deletion progress in status.** Live cluster state is the source of truth.
- **Correctness now depends on every component having a controller owner reference.** Audited the existing components (HPA, Service, ServiceAccount, Role, RoleBinding, SAtokenSecret, PodGang, ResourceClaim, PCSG, PC, Pod) — all set `controllerutil.SetControllerReference` on owned objects. A unit test asserting owner refs on every newly-built object would be a reasonable follow-up to keep this property from regressing.

### DIFF
--- a/operator/e2e/measurement/condition/pcs.go
+++ b/operator/e2e/measurement/condition/pcs.go
@@ -54,12 +54,12 @@ func (c *PCSDeletedCondition) Progress(_ context.Context) string {
 	return "Not deleted"
 }
 
-// PCSFullyDeletedCondition checks that the PodCliqueSet and every resource it
+// PCSAndSubresourcesDeletedCondition checks that the PodCliqueSet and every resource it
 // owns transitively (PodCliqueScalingGroups, PodCliques, and Pods carrying the
 // part-of label) are gone from the API. Use this in delete-phase milestones
 // when you want the timeline to reflect actual cascade-cleanup latency, not
 // just finalizer removal on the parent PCS.
-type PCSFullyDeletedCondition struct {
+type PCSAndSubresourcesDeletedCondition struct {
 	Client        client.Client
 	Name          string
 	Namespace     string
@@ -74,7 +74,7 @@ type PCSFullyDeletedCondition struct {
 
 // Met returns true when the PCS, all owned PCSGs, all owned PodCliques, and
 // all owned Pods (matched by LabelSelector) are absent from the API.
-func (c *PCSFullyDeletedCondition) Met(ctx context.Context) (bool, error) {
+func (c *PCSAndSubresourcesDeletedCondition) Met(ctx context.Context) (bool, error) {
 	c.sel.init(c.LabelSelector)
 	s, err := c.sel.get()
 	if err != nil {
@@ -118,7 +118,7 @@ func (c *PCSFullyDeletedCondition) Met(ctx context.Context) (bool, error) {
 }
 
 // Progress returns a human-readable progress string.
-func (c *PCSFullyDeletedCondition) Progress(_ context.Context) string {
+func (c *PCSAndSubresourcesDeletedCondition) Progress(_ context.Context) string {
 	return fmt.Sprintf("PCS=%d PCSG=%d PCLQ=%d Pod=%d remaining", c.lastPCS, c.lastPCSG, c.lastPCLQ, c.lastPod)
 }
 
--- a/operator/e2e/measurement/condition/pcs.go
+++ b/operator/e2e/measurement/condition/pcs.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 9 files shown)


operator/e2e/measurement/condition/pcs.go
```diff
@@ -53,6 +54,74 @@ func (c *PCSDeletedCondition) Progress(_ context.Context) string {
 	return "Not deleted"
 }
 
+// PCSAndSubresourcesDeletedCondition checks that the PodCliqueSet and every resource it
```
shayasoolin, line 57: There's a missing test where CDs are also present, as they have a special handling in the code. I suggest adding adding a small scale test that involves auto-mnnvl.

shayasoolin: I think there should be e2e tests for deletion, that are not related to scale but to correctness. So in addition to the missing CD-related test I suggest to have a simpler, small scale test for delete. Same test procedure, same util functions, just a separate suite. Will also serve us in including these tests more frequently than just as part of the scale tests.

danbar2: we have such tests already as all tests validates deletion. I bumped into the CD issue since the MNNVL E2E test failed.

shayasoolin (approved):

operator/e2e/tests/scale/scale_test.go
```diff
@@ -120,113 +132,190 @@ func Test_ScaleTest_1000(t *testing.T) {
...
+
+	runScaleTest(t, scaleTestConfig{
+		name:         "ScaleTest_5000_Deletion",
+		workload:     "scale-test-5000-deletion",
+		yamlPath:     "../../yaml/scale-test-5000-deletion.yaml",
+		expectedPods: expectedPods,
+		pcsCount:     defaultScalePCSCount,
+		workerNodes:  defaultScaleWorkerNodes,
+		timeout:      20 * time.Minute,
+		pollInterval: defaultScalePollInterval,
```
gflarity: This test inherits a 100ms poll interval while every delete poll lists PCSGs, PodCliques, and up to 5000 Pods. That adds significant API/list load to the deletion phase being measured at exactly the scale this benchmark is meant to characterize. Please use a coarser poll interval for this test or a lighter/watch-based deletion condition.

gflarity (approved):

renormalize (approved):
I think the original intention behind blocking the delete call until all child resources are deleted is to signal to the user that the PCS is still undergoing deletion due to some reason - primary being that Pods could be stuck in `Terminating`, which is commonly observed in pods running disaggregated inference.

Deletion of the PCS before the pods have relinquished GPU claims might confuse the user behind why a new PCS is not being scheduled. However, I do not think this reason is strong enough to cause severe performance degradation with a large PCS.

Thanks for the fix!

PR #557, opened 2026-04-28
feat(mnnvl): add group-aware MNNVL injection into PodSpec

shmuel-runai:
## Summary

- When injecting MNNVL into a PodSpec, the controller now uses the correct RCT name for the target MNNVL group
- Controllers resolve the MNNVL group from annotations hierarchically — clique annotations take priority over PCS/PCSG annotations
- Export `ResolveGroupName` from the `mnnvl` package and add `groupName` parameter to `InjectMNNVLIntoPodSpec`

## Key changes

- Move `resolveGroupName` to `mnnvl.ResolveGroupName` (exported)
- Add `groupName` parameter to `InjectMNNVLIntoPodSpec`
- Update PCS and PCSG controllers with hierarchical group resolution
- Add group-aware test cases for both PCS and PCSG injection

## Which issue(s) this PR fixes

Ref #417

## Test plan

- [x] Unit tests for `ResolveGroupName` in `helpers_test.go`
- [x] Unit tests for `InjectMNNVLIntoPodSpec` with named groups in `injection_test.go`
- [x] PCS controller `TestBuildResource_MNNVLInjection` — group on PCS, clique override, clique-only
- [x] PCSG controller `TestBuildResource_MNNVLInjection` — group on PCSG, clique override, clique-only
- [x] `make check` passes (no dirty tree)

### DIFF
--- a/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
+++ b/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
@@ -332,9 +332,13 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
 	}
 	pclq.Spec.StartsAfter = dependentPCLQNames
 
-	// Inject MNNVL resourceClaims if enabled on PCSG (propagated from PCS)
-	if mnnvl.IsAutoMNNVLEnabled(pcsg.Annotations) {
-		mnnvl.InjectMNNVLIntoPodSpec(logger, &pclq.Spec.PodSpec, apicommon.ResourceNameReplica{Name: pcs.Name, Replica: pcsReplicaIndex})
+	// Inject MNNVL resourceClaims: clique annotations take priority over PCSG
+	groupName, mnnvlEnabled := mnnvl.ResolveGroupName(pclqTemplateSpec.Annotations)
+	if !mnnvlEnabled {
+		groupName, mnnvlEnabled = mnnvl.ResolveGroupName(pcsg.Annotations)
+	}
+	if mnnvlEnabled {
+		mnnvl.InjectMNNVLIntoPodSpec(logger, &pclq.Spec.PodSpec, apicommon.ResourceNameReplica{Name: pcs.Name, Replica: pcsReplicaIndex}, groupName)
 	}
 
 	return nil
--- a/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
+++ b/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
@@ -332,11 +332,10 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
 	}
 	pclq.Spec.StartsAfter = dependentPCLQNames
 
-	// Inject MNNVL resourceClaims: clique annotations take priority over PCSG
-	groupName, mnnvlEnabled := mnnvl.ResolveGroupName(pclqTemplateSpec.Annotations)
-	if !mnnvlEnabled {
-		groupName, mnnvlEnabled = mnnvl.ResolveGroupName(pcsg.Annotations)
-	}
+	// Inject MNNVL resourceClaims: resolve group hierarchically (PCLQ → PCSG).
+	// PCS-level annotations are already propagated onto the PCSG by the PCS
+	// controller via propagateMNNVLAnnotations, so a two-layer check suffices.
+	groupName, mnnvlEnabled := mnnvl.ResolveGroupNameHierarchically(pclqTemplateSpec.Annotations, pcsg.Annotations)
 	if mnnvlEnabled {
 		mnnvl.InjectMNNVLIntoPodSpec(logger, &pclq.Spec.PodSpec, apicommon.ResourceNameReplica{Name: pcs.Name, Replica: pcsReplicaIndex}, groupName)
 	}
--- a/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique_test.go
+++ b/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique_test.go
@@ -754,6 +754,7 @@ func TestBuildResource_MNNVLInjection(t *testing.T) {
 	tests := []struct {
 		description                         string
 		pcsgAnnotations                     map[string]string
+		cliqueAnnotations                   map[string]string
 		containers                          []corev1.Container
 		initContainers                      []corev1.Container
 		expectedContainersWithClaims        []string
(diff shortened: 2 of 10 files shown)


operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -332,9 +332,13 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
 	}
 	pclq.Spec.StartsAfter = dependentPCLQNames
 
-	// Inject MNNVL resourceClaims if enabled on PCSG (propagated from PCS)
-	if mnnvl.IsAutoMNNVLEnabled(pcsg.Annotations) {
-		mnnvl.InjectMNNVLIntoPodSpec(logger, &pclq.Spec.PodSpec, apicommon.ResourceNameReplica{Name: pcs.Name, Replica: pcsReplicaIndex})
+	// Inject MNNVL resourceClaims: clique annotations take priority over PCSG
+	groupName, mnnvlEnabled := mnnvl.ResolveGroupName(pclqTemplateSpec.Annotations)
+	if !mnnvlEnabled {
+		groupName, mnnvlEnabled = mnnvl.ResolveGroupName(pcsg.Annotations)
```
danbar2: what if mnnvl is enabled on pclq but the name is set on pcsg?
what if the annotations are set on the PCS only?

danbar2 (dismissed):

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
shayasoolin: I suggest having `mnnvl.ResolveGroupName` go from PCLQ level up to PCSG/PCS

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique_test.go
```diff
@@ -889,6 +890,69 @@ func TestBuildResource_MNNVLInjection(t *testing.T) {
...
+							constants.GPUResourceName: resource.MustParse("8"),
+						},
+					},
+				},
+			},
+			expectedContainersWithClaims:    []string{"gpu-worker"},
+			expectedContainersWithoutClaims: []string{},
+			expectPodLevelClaim:             true,
+			expectedRCTName:                 "test-pcs-0-training",
+		},
```
shayasoolin, line 955: Add a test where MNNVL is explicitly disabled but a group is specified, in addition. No claim is expected. (Or is this a validation error?)

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
shmuel-runai: Good questions.

**"What if mnnvl is enabled on PCLQ but the name is set on PCSG?"**

This is by design — the lowest layer is authoritative. If a PCLQ explicitly sets `auto-mnnvl: enabled` without a group, it opts into MNNVL with the default group, intentionally overriding the parent's named group. If the PCLQ wanted to inherit the PCSG's group, it would simply not set any MNNVL annotation and let the fallback resolve from the parent.

I'll introduce a `ResolveGroupNameHierarchically` helper that takes variadic annotation layers (PCLQ, then PCSG/PCS) to make this pattern clearer and more explicit at the call sites. I'll also add a code comment explaining the lowest-layer-wins semantics.

**"What if the annotations are set on the PCS only?"**

This is already handled — when the PCS controller creates the PCSG, it copies MNNVL annotations from the PCS to the PCSG via `propagateMNNVLAnnotations` (in `podcliquescalinggroup.go`). So by the time the PCSG controller runs, the PCSG already carries the PCS's annotations. The two-layer check (PCLQ → PCSG) is sufficient.

shmuel-runai: Agreed. I'll add a `ResolveGroupNameHierarchically` helper that takes variadic annotation layers (most specific first) and returns the first match. The call sites will become:

- PCS controller: `ResolveGroupNameHierarchically(pclqAnnotations, pcsAnnotations)`
- PCSG controller: `ResolveGroupNameHierarchically(pclqAnnotations, pcsgAnnotations)`

Note that PCS-level annotations are already propagated onto the PCSG via `propagateMNNVLAnnotations` when the PCS controller creates the PCSG, so the two-layer check is sufficient.

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique_test.go
shmuel-runai: This combination (`auto-mnnvl: disabled` + `mnnvl-group`) is rejected at admission time by `DetectMNNVLConflict` in the validating webhook — it returns a "contradictory MNNVL annotations" error. So this state should never reach the controller. I don't think adding a controller-level test for an impossible state adds value — WDYT?

danbar2 (dismissed):

danbar2 (approved):

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique_test.go
shayasoolin: Agree

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -332,11 +332,10 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
...
 	pclq.Spec.StartsAfter = dependentPCLQNames
 
-	// Inject MNNVL resourceClaims: clique annotations take priority over PCSG
-	groupName, mnnvlEnabled := mnnvl.ResolveGroupName(pclqTemplateSpec.Annotations)
-	if !mnnvlEnabled {
-		groupName, mnnvlEnabled = mnnvl.ResolveGroupName(pcsg.Annotations)
-	}
+	// Inject MNNVL resourceClaims: resolve group hierarchically (PCLQ → PCSG).
+	// PCS-level annotations are already propagated onto the PCSG by the PCS
+	// controller via propagateMNNVLAnnotations, so a two-layer check suffices.
```
shayasoolin, line 337: Why does `propagateMNNVLAnnotations` only do PCS->PCSG and not PCSG->PCLQ propagation?

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -332,11 +332,10 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
...
+	groupName, mnnvlEnabled := mnnvl.ResolveGroupNameHierarchically(pclqTemplateSpec.Annotations, pcsg.Annotations)
```
shayasoolin, line 338: Maybe we add pcs.Annotations as well, even that it's propagated, just in order not to depend on this propagation? There's zero cost to that.. code will be less coupled with the propagation behavior. Less to explain in the comment..

shayasoolin (approved):

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
shmuel-runai: Good question! We don't need to propagate PCSG→PCLQ because the PCLQ controller already receives the PCSG annotations at build time and resolves the group hierarchically right there. Adding a propagation step would introduce an extra reconcile loop, a mutation to watch for, and more tests — all to reach the same result. So the current approach keeps things simple: propagate once (PCS→PCSG) so the PCSG carries the full picture, then resolve at PCLQ build time without any additional writes.

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
shmuel-runai: I hear you on the coupling concern! My thinking here is that each controller should only be aware of its direct parent — the resource that creates and reconciles it. Having the PCSG's PCLQ builder reach up to PCS annotations would cross ownership boundaries, which I'd like to avoid. The PCS→PCSG propagation is the designed bridge for that, and it's already tested. I think keeping the layers clean is worth the small comment explaining it. Happy to discuss further if you feel strongly though!

PR #559, opened 2026-04-28
refactor(mnnvl): remove auto-MNNVL mutating webhook, switch to opt-in

shmuel-runai:
/kind feature

## Summary

- Remove `MutateAutoMNNVL` from the defaulting webhook — MNNVL is now purely opt-in via explicit `grove.io/auto-mnnvl` annotation
- Remove `networkConfig` from the defaulting handler since it's no longer needed
- Add validation rejecting MNNVL annotations on non-GPU PodCliques
- Update E2E tests to explicitly opt-in to MNNVL via annotations

## Key changes

- **`webhook.go`**: Delete `MutateAutoMNNVL`, add non-GPU PCLQ validation in `validatePodCliqueTemplateSpecOnCreate`
- **`defaulting/handler.go`**: Remove `networkConfig` field and MNNVL mutation from `Default()`
- **`register.go`**: Simplify `NewHandler(mgr)` call (no longer needs network config)
- **E2E tests**: Use `buildGPUPCSWithMNNVL` helper to explicitly set annotations; rewrite `testPCSGetsAutoAnnotation` to verify opt-in semantics

## Which issue(s) this PR fixes

Ref #417

## Test plan

- Unit tests for `ResolveGroupName` in `helpers_test.go`
- `TestValidatePCSOnCreate_NonGPUCliqueWithMNNVL` — validates rejection of MNNVL on non-GPU cliques
- `TestMNNVL_WebhookPipeline_LegacyPCSUpdate` — simplified to single case (no mutation on update)
- E2E tests updated to opt-in model
- `go build ./...` and `go test` pass

Made with Cursor


operator/internal/mnnvl/webhook.go
```diff
@@ -154,7 +114,16 @@ func validatePodCliqueSetTemplateSpecOnCreate(templateSpec *grovecorev1alpha1.Po
 }
 
 func validatePodCliqueTemplateSpecOnCreate(clique *grovecorev1alpha1.PodCliqueTemplateSpec, autoMNNVLEnabled bool, fldPath *field.Path) field.ErrorList {
-	return validateMNNVLAnnotationsOnCreate(clique.Annotations, autoMNNVLEnabled, fldPath.Child("annotations"))
+	allErrs := validateMNNVLAnnotationsOnCreate(clique.Annotations, autoMNNVLEnabled, fldPath.Child("annotations"))
+
+	// Reject explicit MNNVL annotations on non-GPU PodCliques.
+	_, mnnvlRequested := ResolveGroupName(clique.Annotations)
```
gflarity: The non-GPU rejection in validatePodCliqueTemplateSpecOnCreate only fires for annotations on the clique itself, but collectDistinctGroups also picks up PCS-level and PCSG-level annotations when deciding what ComputeDomains to create. So a CPU-only PCS or a PCSG that points only at CPU cliques can still trigger ComputeDomain creation while InjectMNNVLIntoPodSpec skips every pod for lack of GPU containers — leaving an orphaned, finalizer-protected CD with no claim referencing it. Is that intentional (e.g. the operator-level autoMNNVLEnabled gate is considered sufficient), or should the validator also reject inherited MNNVL opt-ins whose effective cliques have no GPU resources?

shayasoolin: I agree there's something inconsistent here. If you reject non-GPU cliques that have the MNNVL annotation, it makes sense to also reject indirect annotations inherited from the PCS/G. 
A possible resolution to this can be defining the semantics of the annotation to be "set MNNVL if the clique has GPUs, ignore otherwise." Then, you shouldn't reject a non-GPU clique with the annotation.
I find it simpler and more relaxed, and I'm totally fine with not warning the user for maybe misplaced annotations.

operator/e2e/tests/auto-mnnvl/supported_and_enabled_test.go
```diff
@@ -72,39 +72,45 @@ func Test_AutoMNNVL_SupportedAndEnabled(t *testing.T) {
 	}
 }
 
-// testPCSGetsAutoAnnotation verifies that the mutating webhook adds
-// grove.io/auto-mnnvl: enabled annotation to PCS with GPU requirements,
-// and does NOT add it to PCS without GPU requirements.
+// testPCSGetsAutoAnnotation verifies opt-in semantics: a GPU PCS without an
+// explicit MNNVL annotation does NOT receive MNNVL behaviour (no ComputeDomain),
+// while a CPU-only PCS also doesn't get the annotation.
```
shayasoolin: I don't understand the part of "while a CPU-only PCS also doesn't get the annotation", what do you mean by "doesn't get the annotation"? From whom? Who can inherit the annotation to the PCS, which is the highest level in the hierarchy?

operator/internal/mnnvl/webhook.go
shmuel-runai: @gflarity Good catch! You're right that this was inconsistent. We've adopted Shay's suggestion — MNNVL annotations on non-GPU cliques are now silently ignored (no webhook rejection). Additionally, collectDistinctGroups is now GPU-aware: it resolves the effective group for each GPU clique using hierarchical resolution and skips non-GPU cliques entirely. So a CPU-only PCS with auto-mnnvl: enabled will produce zero ComputeDomains — no orphaned CDs.

@shayasoolin Agreed! The annotation semantics are now "enable MNNVL if the clique has GPUs, silently skip otherwise." The webhook no longer rejects non-GPU cliques with MNNVL annotations.

operator/e2e/tests/auto-mnnvl/supported_and_enabled_test.go
shmuel-runai: Good point — the comment was leftover from the old model where the mutating webhook auto-injected the annotation. Clarified it.

operator/internal/mnnvl/computedomain/computedomain.go
```diff
@@ -360,32 +360,41 @@ func getRequiredCDNames(pcs *grovecorev1alpha1.PodCliqueSet) []cdNameInfo {
...
 
 	for _, clique := range pcs.Spec.Template.Cliques {
-		if clique != nil {
-			addGroupFromAnnotation(groups, clique.Annotations)
+		if clique == nil || !mnnvl.HasGPUInPodSpec(&clique.Spec.PodSpec) {
+			continue
+		}
+		var parentAnnotations map[string]string
+		if pcsgCfg, ok := pcsgByClique[clique.Name]; ok {
+			parentAnnotations = pcsgCfg.Annotations
```
shayasoolin: The term `parent` here is a bit confusing, as a clique may be the child of a PCS, not only PCSG. But it will have no parent in this calculation. Overall it will still be OK as you pass `pcs.Annotations` in L381.
Maybe use `pcsgParent` instead of `parent`.

shayasoolin (dismissed):

operator/e2e/tests/auto-mnnvl/testutils.go
```diff
@@ -231,20 +231,32 @@ const (
...
-	return testutils.NewPodCliqueSetBuilder(name, "default", types.UID("")).
+	builder := testutils.NewPodCliqueSetBuilder(name, "default", types.UID("")).
 		WithReplicas(int32(replicas)).
 		WithTerminationDelay(time.Second).
-		WithPodCliqueTemplateSpec(gpuClique).
-		Build()
+		WithPodCliqueTemplateSpec(gpuClique)
+
+	if len(annotations) > 0 && annotations[0] != nil {
+		builder.WithAnnotations(annotations[0])
```
danbar2, line 250: why only the first annotation?

operator/e2e/tests/auto-mnnvl/testutils.go
```diff
@@ -332,8 +344,13 @@ func buildComprehensivePCS(name string, replicas int) *grovecorev1alpha1.PodCliq
 			Name:         "sg2",
 			CliqueNames:  []string{"cpu3"},
 			MinAvailable: ptr.To(int32(1)),
-		}).
-		Build()
+		})
+
+	if len(annotations) > 0 && annotations[0] != nil {
+		builder.WithAnnotations(annotations[0])
```
danbar2, line 352: same

danbar2 (approved):

shayasoolin (approved):

PR #561, opened 2026-04-29
Fix gang termination when scheduled replicas regress below MinAvailable

### LINKED ISSUE #277: Gang Termination Doesn't Work

danbar2:
## Summary

Closes #277.

The `MinAvailableBreached` condition was suppressed whenever `scheduledReplicas < MinAvailable`. That suppression conflated two situations:

1. **Initial startup** (correct to suppress) — no pods have been admitted yet, gang termination would be premature.
2. **Regression after a healthy state** (incorrectly suppressed) — e.g. a node hosting running pods is cordoned or fails and the replacement pod cannot be scheduled. `scheduledReplicas` falls below `MinAvailable`, but the breach never fires, so `gangterminate.go` never deletes the PCS replica's PodCliques, the PodGang is never recreated, and the workload is stuck. This is exactly the gap the new E2E gang-termination tests were exposing.

## Idea behind the fix

Split the under-scheduled branch on `scheduledReplicas == 0` vs `> 0`:

```go
if scheduledReplicas < minAvailable {
    if scheduledReplicas == 0 {
        // initial startup OR full regression — suppress (see below)
        return ConditionFalse, InsufficientScheduledPods
    }
    // 0 < scheduled < MinAvailable — must be regression, breach
    return ConditionTrue, ScheduledReplicasRegressed
}
```

**Why `0 < scheduled < MinAvailable` is unambiguously a regression.** Under gang scheduling the gang scheduler admits `MinAvailable` pods atomically — a partial-scheduled state is structurally unreachable from a fresh start. The only way to land there is to have been healthy and then lost pods. So we can breach immediately on observing the partial state.

**Why we keep `scheduled == 0` suppressed.** With nothing scheduled, gang termination has no useful work: recreating the PodGang produces the same Pending pods against the same cluster state. Suppressing avoids a churn loop. This matches KAI's behaviour: in general, if reaching `MinAvailable` legitimately takes longer than `TerminationDelay`, gang-terminating-and-recreating is the right move, and KAI's gang scheduler arrives at the same conclusion.

**Pod scheduling continues throughout.** The breach condition is consumed only by `gangterminate.go`, which still waits the full `TerminationDelay` (default 4h) before deleting. The pod-creation/replacement path in `podclique/components/pod/syncflow.go` and the PCLQ-creation path in `podcliquescalinggroup/components/podclique/sync.go` do not gate on the breach — replacement pods keep being created and re-scheduled while the timer runs. If they succeed, `scheduledReplicas` recovers, the condition flips back to False, and gang termination is never triggered.

## Behaviour on non-gang schedulers

On schedulers without gang admission (e.g. vanilla kube-scheduler), pods are admitted one at a time, so a brief `0 < scheduled < MinAvailable` flicker can occur during staged startup. The flicker is debounced by `TerminationDelay` (default 4h) — only pathological multi-hour rampups would see spurious gang termination, and in that case the existing `Spec.Template.TerminationDelay` knob is the remedy.

## Test plan

- [x] `go test ./...` passes (operator module)
- [x] `go vet ./...` clean, `go build ./...` clean
- [x] Reproducer tests in `reconcilestatus_test.go` for both PodClique and PodCliqueScalingGroup capture:
  - `0 < scheduled < MinAvailable` ⟹ breach (the bug fix)
  - `scheduled == 0` after regression ⟹ no breach (sanity-pin against over-correction / churn loop)
  - Fresh PCLQ/PCSG never scheduled ⟹ no breach (initial-startup sanity-pin)
- [x] Updated existing `TestComputeMinAvailableBreachedCondition / insufficient_scheduled` PCSG test, which had previously asserted the buggy behaviour, to assert the post-fix behaviour and added a paired `scheduled == 0` case.
- [x] E2E gang termination tests (referenced in the issue) should now pass.

## Open question for reviewers

The `scheduled == 0` suppression is a deliberate judgement call: it avoids a churn loop, but it does mean a fully-down PCS replica won't be gang-terminated (the controller keeps trying to schedule replacements forever). If there is a scenario where the *PodGang resource itself* gets into a bad shape that recreation would fix, we may want to revisit. Flagging for @renormalize / @unmarshall / @sanjaychatterjee.

### DIFF
--- a/operator/internal/controller/podclique/reconcilestatus.go
+++ b/operator/internal/controller/podclique/reconcilestatus.go
@@ -78,14 +78,7 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 		mutateMinAvailableBreachedCondition(pclq,
 			len(podCategories[k8sutils.PodHasAtleastOneContainerWithNonZeroExitCode]),
 			len(podCategories[k8sutils.PodStartedButNotReady]))
-		// Emit a Warning when scheduledReplicas drops from non-zero to zero. Gang termination is
-		// suppressed in this state to avoid a churn loop, so this event is the only signal that a
-		// previously-running workload is fully down.
-		if originalStatus.ScheduledReplicas > 0 && pclq.Status.ScheduledReplicas == 0 {
-			r.eventRecorder.Eventf(pclq, corev1.EventTypeWarning, internalconstants.ReasonAllScheduledReplicasLost,
-				"All scheduled pods lost (was %d). Gang termination is suppressed to avoid recreating Pending pods against the same cluster state; investigate node availability or capacity.",
-				originalStatus.ScheduledReplicas)
-		}
+		r.emitAllScheduledReplicasLostIfNeeded(pclq, originalStatus.ScheduledReplicas)
 	}
 
 	// mutate the selector that will be used by an autoscaler.
@@ -189,6 +182,18 @@ func mutateSelector(pcsName string, pclq *grovecorev1alpha1.PodClique) error {
 	}
 	pclq.Status.Selector = ptr.To(selector.String())
 	return nil
+}
+
+// emitAllScheduledReplicasLostIfNeeded emits a Warning event when ScheduledReplicas drops from
+// non-zero to zero. Gang termination is suppressed in this state (recreating the PodGang would
+// just produce the same Pending pods) so this event is the only explicit signal that a
+// previously-running workload is now fully down.
+func (r *Reconciler) emitAllScheduledReplicasLostIfNeeded(pclq *grovecorev1alpha1.PodClique, originalScheduled int32) {
+	if originalScheduled > 0 && pclq.Status.ScheduledReplicas == 0 {
+		r.eventRecorder.Eventf(pclq, corev1.EventTypeWarning, internalconstants.ReasonAllScheduledReplicasLost,
+			"All scheduled pods lost (was %d). Gang termination is suppressed to avoid recreating Pending pods against the same cluster state; investigate node availability or capacity.",
+			originalScheduled)
+	}
 }
 
 // mutateMinAvailableBreachedCondition updates the MinAvailableBreached condition based on pod availability
--- a/operator/internal/controller/podclique/reconcilestatus.go
+++ b/operator/internal/controller/podclique/reconcilestatus.go
@@ -206,14 +206,36 @@ func computeMinAvailableBreachedCondition(pclq *grovecorev1alpha1.PodClique, num
 	scheduledReplicas := int(pclq.Status.ScheduledReplicas)
 	now := metav1.Now()
 
-	// If the number of scheduled pods is less than the minimum available, then minAvailable is not considered as breached.
-	// Consider a case where none of the PodCliques have been scheduled yet, then it should not cause the PodGang to be recreated all the time.
+	// Handle the under-scheduled cases:
+	//   - scheduledReplicas == 0: either initial startup (no pods admitted yet) or a full
+	//     regression where every running pod has been lost. In both cases, gang termination
+	//     would only recreate the same Pending pods against the same cluster state and would
+	//     produce a churn loop, so we suppress the breach. If the cluster state actually
+	//     changes the next reconcile picks it up.
+	//   - 0 < scheduledReplicas < minAvailable: under gang scheduling this state is
+	//     structurally unreachable from initial startup — the gang scheduler admits
+	//     MinAvailable pods atomically. So observing it implies pods were healthy and
+	//     have since regressed (e.g. node cordon/failure with replacements pending).
+	//     Treat as a breach. The downstream gang-termination flow still waits the full
+	//     TerminationDelay before acting, and the pod-creation path is unaffected, so
+	//     the controller keeps trying to schedule replacements throughout the wait.
+	//     On non-gang schedulers a brief partial-scheduled flicker can occur during
+	//     staged startup; TerminationDelay (default 4h) absorbs it.
 	if scheduledReplicas < minAvailable {
+		if scheduledReplicas == 0 {
+			return metav1.Condition{
+				Type:               constants.ConditionTypeMinAvailableBreached,
+				Status:             metav1.ConditionFalse,
+				Reason:             constants.ConditionReasonInsufficientScheduledPods,
+				Message:            fmt.Sprintf("Insufficient scheduled pods. expected at least: %d, found: %d", minAvailable, scheduledReplicas),
+				LastTransitionTime: now,
+			}
+		}
 		return metav1.Condition{
 			Type:               constants.ConditionTypeMinAvailableBreached,
-			Status:             metav1.ConditionFalse,
-			Reason:             constants.ConditionReasonInsufficientScheduledPods,
-			Message:            fmt.Sprintf("Insufficient scheduled pods. expected at least: %d, found: %d", minAvailable, scheduledReplicas),
+			Status:             metav1.ConditionTrue,
+			Reason:             constants.ConditionReasonScheduledReplicasRegressed,
+			Message:            fmt.Sprintf("Scheduled replicas regressed below MinAvailable. expected at least: %d, found: %d", minAvailable, scheduledReplicas),
 			LastTransitionTime: now,
 		}
 	}
--- a/operator/internal/controller/podclique/reconcilestatus.go
+++ b/operator/internal/controller/podclique/reconcilestatus.go
@@ -23,6 +23,7 @@ import (
 	apicommon "github.com/ai-dynamo/grove/operator/api/common"
 	"github.com/ai-dynamo/grove/operator/api/common/constants"
 	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
+	internalconstants "github.com/ai-dynamo/grove/operator/internal/constants"
 	ctrlcommon "github.com/ai-dynamo/grove/operator/internal/controller/common"
 	componentutils "github.com/ai-dynamo/grove/operator/internal/controller/common/component/utils"
 	k8sutils "github.com/ai-dynamo/grove/operator/internal/utils/kubernetes"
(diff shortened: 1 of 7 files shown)


shayasoolin (dismissed):
LGTM

Ronkahn21:
Hi @danbar2 — I ran this PR through a multi-agent review (code, tests, comments, silent failures + CodeRabbit). Findings below; happy to discuss.

**1. "Structurally unreachable" claim and Reason name** — `constants.go:109-114`, `podclique/reconcilestatus.go:212-226`, `podcliquescalinggroup/reconcilestatus.go:168-176`

The new comments call `0 < scheduled < MinAvailable` "structurally unreachable from initial startup" then concede non-gang schedulers can flicker into it. Grove's `kube` backend (`scheduler/kube/`, default fallback in `manager.go:81-85`) IS non-gang — so the invariant is false there, and on PCS-replica recreation. The same false invariant drives the user-visible `Reason=ScheduledReplicasRegressed` and the "Scheduled replicas regressed below MinAvailable" message — both wrong when nothing has actually regressed. Suggested fixes (cheapest while the constant is unmerged):

1. Drop the "structurally unreachable" framing; rationale is "gang scheduler ⇒ implies regression; non-gang ⇒ `TerminationDelay` absorbs the flicker."
2. Rename the constant to a neutral term, e.g. `ConditionReasonScheduledReplicasBelowMinAvailable` — and tighten the 5-line doc comment to match its 1-line peers.
3. Match Message: `"Scheduled replicas (%d) below MinAvailable (%d)"`.

**2. `scheduled==0` regression is invisible — emit an event** — `podclique/reconcilestatus.go:228-235`, `podcliquescalinggroup/reconcilestatus.go:193-199`

A previously-healthy PCLQ that loses every scheduled pod gets the same `Reason` and Message as a fresh one — no event, no log differentiates them. Per the PR's open question the controller intentionally won't gang-terminate this state, so an event is the only "something's wrong" signal. Suggest emitting `EventTypeWarning` on the `PodCliqueScheduled=True` → `scheduled==0` transition. `mutateMinAvailableBreachedCondition` doesn't carry an EventRecorder today, but the surrounding controllers do (`gangterminate.go` uses `r.eventRecorder.Eventf`).

**3. New tests don't assert `Reason`** — `podclique/reconcilestatus_test.go:299-308`, PCSG `partial schedule regression` row

Both new tests check only `Type` and `Status`. The existing PCSG test pins `wantReason` at line 206 — the same pattern would catch a refactor returning `Status: True` with the wrong reason. The PCSG case is extra fragile because its populated `pclqsMap` is genuinely unread (function short-circuits at line 192), so a broken short-circuit also passes today.

**4. Missing test: PCSG update-in-progress × regression band** — `podcliquescalinggroup/reconcilestatus_test.go`

No test sets `Status.UpdateProgress` non-nil with `UpdateEndedAt: nil` and `0 < scheduled < MinAvailable`. The `IsPCSGUpdateInProgress` short-circuit (line 181) must win over the regression branch (line 192) — branch ordering is unpinned, future reordering produces silent churn during rolling updates.

Ronkahn21 (dismissed):

shayasoolin (dismissed):

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -209,14 +218,27 @@ func computeMinAvailableBreachedCondition(pclq *grovecorev1alpha1.PodClique, num
...
+	// 0 < scheduledReplicas < MinAvailable: with a gang scheduler this implies regression
+	// after a healthy state and breaches. On non-gang schedulers it can flicker briefly
+	// during staged startup; TerminationDelay (default 4h) absorbs the flicker.
 	if scheduledReplicas < minAvailable {
+		if scheduledReplicas == 0 {
+			return metav1.Condition{
+				Type:               constants.ConditionTypeMinAvailableBreached,
+				Status:             metav1.ConditionFalse,
+				Reason:             constants.ConditionReasonInsufficientScheduledPods,
+				Message:            fmt.Sprintf("Scheduled replicas (%d) below MinAvailable (%d)", scheduledReplicas, minAvailable),
```
renormalize: nit: having the same message here as the one with `ConditionFalse` might be slightly misleading. We could perhaps mention why the condition is set to false in the message? `Scheduled replicas 0 - not recreating the pods of the PodGang`.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -181,11 +195,19 @@ func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1a
 	minAvailable := int(*pcsg.Spec.MinAvailable)
 	scheduledReplicas := int(pcsg.Status.ScheduledReplicas)
 	if scheduledReplicas < minAvailable {
+		if scheduledReplicas == 0 {
+			return metav1.Condition{
+				Type:    constants.ConditionTypeMinAvailableBreached,
+				Status:  metav1.ConditionFalse,
+				Reason:  constants.ConditionReasonInsufficientScheduledPCSGReplicas,
+				Message: fmt.Sprintf("Scheduled replicas (%d) below MinAvailable (%d)", scheduledReplicas, minAvailable),
```
renormalize: https://github.com/ai-dynamo/grove/pull/561#discussion_r3194559343

renormalize (dismissed):
Thanks for raising the fix and closing this long existing gap @danbar2!

Feel free to ignore the nit comments below.

renormalize (dismissed):
thank you for fixing the nit!

Ronkahn21 (dismissed):

shayasoolin (dismissed):

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -77,6 +78,14 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 		mutateMinAvailableBreachedCondition(pclq,
 			len(podCategories[k8sutils.PodHasAtleastOneContainerWithNonZeroExitCode]),
 			len(podCategories[k8sutils.PodStartedButNotReady]))
+		// Emit a Warning when scheduledReplicas drops from non-zero to zero. Gang termination is
+		// suppressed in this state to avoid a churn loop, so this event is the only signal that a
+		// previously-running workload is fully down.
+		if originalStatus.ScheduledReplicas > 0 && pclq.Status.ScheduledReplicas == 0 {
```
gflarity: This event is now the only explicit signal for the fully-lost scheduled-replica case because `MinAvailableBreached` stays `False` when scheduled replicas drop to zero. Please add a focused fake event recorder test for the non-zero to zero transition so this signal does not silently regress.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -66,6 +68,15 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	mutateReplicas(logger, pcs.Status.CurrentGenerationHash, pcsg, pclqsPerPCSGReplica)
 	mutateMinAvailableBreachedCondition(logger, pcsg, pclqsPerPCSGReplica)
 
+	// Emit a Warning when scheduledReplicas drops from non-zero to zero. Gang termination is
+	// suppressed in this state to avoid a churn loop, so this event is the only signal that a
+	// previously-running workload is fully down.
+	if originalStatus.ScheduledReplicas > 0 && pcsg.Status.ScheduledReplicas == 0 {
```
gflarity: This event is now the only explicit signal for the fully-lost scheduled-replica case because `MinAvailableBreached` stays `False` when scheduled replicas drop to zero. Please add a focused fake event recorder test for the non-zero to zero transition so this signal does not silently regress.

gflarity (commented):
This is similar to the approach I was working on previously, so works for me. However I want to make sure CC: @nvrohanv and @sanjaychatterjee weigh because there was a lot of discussions around how to treat pods starting up. 

We should also add comprehensive E2E tests for gang termination as a feature. I can't recall the specifics, but IIRC there might have been other issues with the tests. Here's the [old E2E test branch](https://github.com/gflarity/grove/tree/e2e_gang_termination) I put together. Here's the [scenario document](https://docs.google.com/document/d/11dGjQh--s7QXggTyBxrcAttl2JHt1tbUKEV_UxvxAp8/edit?tab=t.0#heading=h.s62odkdoge8e) it was based on.

gflarity (approved):

shayasoolin (approved):

Ronkahn21 (approved):

PR #566, opened 2026-04-29
fix: stabilize PodCliqueSet hashes against map slice reorder

### LINKED ISSUE #565: Rapid PodCliqueSet spec updates trigger full gang roll instead of in-place replica scale
### What happened?

On a Dynamo `DynamoGraphDeployment` running with `optimization_target=latency`, the Dynamo Planner emits multiple back-to-back replica scale-up patches against the DGD (typical: 3–6 patches over 20–30 s while load is sustained). Each patch is translated by the Dynamo operator into a `PodCliqueSet` spec update. Grove appears to treat this rapid sequence of spec updates as a full rollout rather than an in-place replica bump, which has these consequences:

1. Every pod in the gang (Frontend, Planner, PrefillWorker, DecodeWorker) is killed and recreated inside a ~5 s window.
2. The race between the operator applying `replicas=N` and Grove applying a new spec hash produces intermediate prefill pods that are **created and then re-deleted mid-reconcile** (e.g. `prefill-5ckwr`, `prefill-zkxzw` in the run below).
3. The final `replicas` value on the impacted PodClique is 1, while `updatedReplicas` reaches 2 and is never reset — the DGD is left in `Ready=False` / `state: pending` indefinitely:

   ```yaml
   conditions:
   - message: 'Resources not ready: tc-2-2-3-late-da27: podclique/tc-2-2-3-late-da27-0-vllmprefillworker:
       desired=1, updated=2'
     reason: some_resources_are_not_ready
     status: "False"
     type: Ready
   services:
     VllmPrefillWorker:
       readyReplicas: 1
       replicas: 1
       updatedReplicas: 2
   ```

The same workload with `optimization_target=throughput` (which produces ~1 patch per scale-up cycle instead of a burst) takes the in-place `PodCreateSuccessful` path and does **not** trigger a gang roll until minutes later.

#### Observed pod lifecycle during the burst (run 2, ~2 s window)

\`\`\`
UpdatePodCliqueSet fires →
  CREATE prefill-8gbqv     (scale-up attempt #1)
  CREATE decode-wprtt      (Grove roll)
  CREATE planner-bxn7v     (Grove roll)
  CREATE frontend-58rql    (Grove roll)
  CREATE prefill-5ckwr     (scale-up attempt #2)
  DELETE planner-r9mcf     (Gen 0 gone)
  DELETE prefill-8stjg
  DELETE frontend-cbflf
  DELETE decode-fdqgs
  DELETE prefill-5ckwr     ← killed ~instantly
  CREATE prefill-zkxzw     (scale-up attempt #3)
... ~100 s later ...
  DELETE prefill-zkxzw     ← killed again
  Final: 1 prefill, 1 decode, 1 planner, 1 frontend
\`\`\`

Two independent runs reproduced the same failure mode (88% and 96% client-request failure during the roll window).

A series of replica-only spec updates on a `PodCliqueSet` should be reconciled as in-place replica changes (scale up/down only the affected `PodClique`), not as a full gang rollout. New replicas requested by an in-flight scale-up patch should not be created and then immediately deleted by a competing reconcile pass. `updatedReplicas` should not be left greater than `replicas` after reconciliation settles.

### Environment

- Kubernetes version: (Dynamo rc5 test cluster — version not captured in report)
- Grove version: shipped with `nvcr.io/nvstaging/ai-dynamo/dynamo-operator` 1.1.0rc5 — please advise which Grove SHA this corresponds to
- Scheduler: Grove (`grove-operator`, `PodCliqueSet`) + run:ai PodGroup sidecar
- Workload: Dynamo `DynamoGraphDeployment` (vLLM disagg, Qwen3-8B), Planner driving replica patches via the Dynamo operator
- Reproduces consistently on `optimization_target=latency`; does **not** reproduce on `optimization_target=throughput` (same cluster, same fixture, only that field differs)

Original NVIDIA bug report: https://nvbugspro.nvidia.com/bug/6109874 (internal). Full attachments — event timelines, planner logs, DGD snapshot — are linked from there; happy to mirror excerpts into this issue on request.

gflarity:
## Summary

Fixes two related bugs that together caused Grove to gang-roll a PodCliqueSet on every spec update where the upstream operator happened to emit `+listType=map` slices in a different order. See #565 for the full repro and pod-lifecycle trace.

Root cause: `dump.ForHash` (used by both `computeGenerationHash` and `ComputeHash`) sorts map keys but preserves slice order. Several Kubernetes API slices that are declared `+listType=map` (and `PodCliqueSet.Spec.Template.Cliques` itself) are name-keyed sets with no inherent order, but they were being treated as ordered by the hash. Non-deterministic Go map iteration in upstream code (e.g. the Dynamo operator) flipped the hashes on every reconcile, which scheduled real rolling updates and silently destroyed in-flight scale state.

**Fixes:**

1. `computeGenerationHash` (`operator/internal/controller/podcliqueset/reconcilespec.go`) sorts `Spec.Template.Cliques` by name before hashing. Slice order is mixed back into the hash only for `CliqueStartupTypeInOrder` (via a `startupOrderingMarker`), where the slice index does encode startup-chain semantics. `AnyOrder` and `Explicit` no longer leak slice order into the hash.

2. `ComputeHash` (`operator/internal/utils/kubernetes/pod.go`) canonicalizes order-independent `+listType=map` slices in PodSpec before hashing:
   - PodSpec: `Containers`, `Volumes`, `ImagePullSecrets`, `HostAliases`, `TopologySpreadConstraints`, `ResourceClaims`, `EphemeralContainers`
   - Container (regular, init, ephemeral): `Ports`, `VolumeMounts`, `VolumeDevices`, `ResizePolicy`

   `InitContainers` (sequential execution), `Container.Env` (`$(VAR)` substitution), `EnvFrom` / `Tolerations` (`+listType=atomic`), and `Args` / `Command` are intentionally NOT sorted — their order is part of the desired state.

## Test plan

- [x] `go vet ./...` clean on the affected packages
- [x] `go test ./internal/utils/kubernetes/... ./internal/controller/common/component/utils/... ./internal/controller/podcliqueset/... ./internal/controller/podclique/components/pod/...` all pass
- [x] Sort-invariance tests for every canonicalized slice (`pod_test.go::TestComputeHash_AdditionalListTypeMapSlices` + 8 sub-tests; mirrored at the per-PCLQ layer in `podclique_test.go::TestComputePCLQPodTemplateHash_AdditionalListTypeMapSlices`)
- [x] Reverse-direction regression tests (`TestComputeHash_RealSpecChangesStillFlipHash`, `TestComputePCLQPodTemplateHash_RealSpecChangesStillFlipHash`, `TestComputeGenerationHash_RealCliqueTemplateChangeFlipsHash`) — pin that real desired-state changes still flip the hash, so a future over-canonicalization can't silently break rolling updates
- [x] StartupType boundary cases (`TestComputeGenerationHash_InOrderToAnyOrderFlipsHash`, `TestComputeGenerationHash_StartupTypeChangeFlipsHash`, `TestComputeGenerationHash_InOrderStartupIsSensitiveToCliqueOrder`)
- [x] Realistic Dynamo-shaped PodSpec end-to-end test (`TestComputePCLQPodTemplateHash_RealisticDynamoLikePodSpec`) — same desired state, every `+listType=map` slice (including `VolumeMounts`) shuffled, hash must be stable
- [x] Nil-safety (`TestComputeHash_NilSafety`)
- [x] Combined replica-bump + clique-reorder no-op (`TestComputeGenerationHash_CombinedReplicaChangeAndCliqueReorderIsNoOp`) — the realistic scale-up patch shape from the bug report
- [x] E2E tests which upgrade Grove and confirms everything works as expected
Resolves #565

Original NVIDIA bug report: https://nvbugspro.nvidia.com/bug/6109874 (internal)

### DIFF
--- a/operator/internal/controller/common/component/utils/podcliquescalinggroup.go
+++ b/operator/internal/controller/common/component/utils/podcliquescalinggroup.go
@@ -124,6 +124,14 @@ func GetPCSGsByPCSReplicaIndex(ctx context.Context, cl client.Client, pcsObjKey 
 
 // GetPCLQTemplateHashes generates the Pod template hash for all PCLQs in a PCSG. Returns a map of [PCLQ Name : PodTemplateHas]
 func GetPCLQTemplateHashes(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev1alpha1.PodCliqueScalingGroup) map[string]string {
+	hashCandidates := GetPCLQTemplateHashCandidates(pcs, pcsg)
+	return lo.MapValues(hashCandidates, func(candidates HashCandidates, _ string) string {
+		return candidates.Canonical
+	})
+}
+
+// GetPCLQTemplateHashCandidates generates the canonical and legacy Pod template hash candidates for all PCLQs in a PCSG.
+func GetPCLQTemplateHashCandidates(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev1alpha1.PodCliqueScalingGroup) map[string]HashCandidates {
 	pclqTemplateSpecs := make([]*grovecorev1alpha1.PodCliqueTemplateSpec, 0, len(pcsg.Spec.CliqueNames))
 	for _, cliqueName := range pcsg.Spec.CliqueNames {
 		pclqTemplateSpec := FindPodCliqueTemplateSpecByName(pcs, cliqueName)
@@ -132,11 +140,11 @@ func GetPCLQTemplateHashes(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev
 		}
 		pclqTemplateSpecs = append(pclqTemplateSpecs, pclqTemplateSpec)
 	}
-	cliqueTemplateSpecHashes := make(map[string]string, len(pclqTemplateSpecs))
+	cliqueTemplateSpecHashes := make(map[string]HashCandidates, len(pclqTemplateSpecs))
 	for pcsgReplicaIndex := range int(pcsg.Spec.Replicas) {
 		for _, pclqTemplateSpec := range pclqTemplateSpecs {
 			pclqFQN := apicommon.GeneratePodCliqueName(apicommon.ResourceNameReplica{Name: pcsg.Name, Replica: pcsgReplicaIndex}, pclqTemplateSpec.Name)
-			cliqueTemplateSpecHashes[pclqFQN] = ComputePCLQPodTemplateHash(pclqTemplateSpec, pcs.Spec.Template.PriorityClassName)
+			cliqueTemplateSpecHashes[pclqFQN] = ComputePCLQPodTemplateHashCandidates(pclqTemplateSpec, pcs.Spec.Template.PriorityClassName)
 		}
 	}
 	return cliqueTemplateSpecHashes
@@ -147,12 +155,12 @@ func GetPCLQTemplateHashes(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev
 // a computed PodTemplateHash from the latest PodCliqueSet resource.
 func GetPCLQsInPCSGPendingUpdate(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev1alpha1.PodCliqueScalingGroup, existingPCLQs []grovecorev1alpha1.PodClique) []string {
 	pclqFQNsPendingUpdate := make([]string, 0, len(existingPCLQs))
-	expectedPCLQPodTemplateHashes := GetPCLQTemplateHashes(pcs, pcsg)
+	expectedPCLQPodTemplateHashes := GetPCLQTemplateHashCandidates(pcs, pcsg)
 	for _, existingPCLQ := range existingPCLQs {
 		existingPodTemplateHash := existingPCLQ.Labels[apicommon.LabelPodTemplateHash]
 		expectedPodTemplateHash := expectedPCLQPodTemplateHashes[existingPCLQ.Name]
-		if existingPodTemplateHash != expectedPodTemplateHash {
-			pclqFQNsPendingUpdate = append(pclqFQNsPendingUpdate, expectedPodTemplateHash)
+		if !expectedPodTemplateHash.Matches(existingPodTemplateHash) {
+			pclqFQNsPendingUpdate = append(pclqFQNsPendingUpdate, existingPCLQ.Name)
 		}
 	}
 	return pclqFQNsPendingUpdate
@@ -164,8 +172,14 @@ func IsPCSGUpdateInProgress(pcsg *grovecorev1alpha1.PodCliqueScalingGroup) bool 
 }
 
 // IsPCSGUpdateComplete returns whether the rolling update of the PodCliqueScalingGroup is complete.
-func IsPCSGUpdateComplete(pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pcsGenerationHash string) bool {
-	return pcsg.Status.CurrentPodCliqueSetGenerationHash != nil && *pcsg.Status.CurrentPodCliqueSetGenerationHash == pcsGenerationHash
+func IsPCSGUpdateComplete(pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pcsGenerationHash string, compatiblePCSGenerationHashes ...string) bool {
+	if pcsg.Status.CurrentPodCliqueSetGenerationHash == nil {
+		return false
+	}
+	if *pcsg.Status.CurrentPodCliqueSetGenerationHash == pcsGenerationHash {
+		return true
+	}
+	return slices.Contains(compatiblePCSGenerationHashes, *pcsg.Status.CurrentPodCliqueSetGenerationHash)
 }
 
 // GetPodCliqueFQNsForPCSG generates the PodClique FQNs for all PodCliques that are owned by a PodCliqueScalingGroup.
--- a/operator/internal/controller/podclique/components/pod/deletionsort.go
+++ b/operator/internal/controller/podclique/components/pod/deletionsort.go
@@ -18,6 +18,7 @@ package pod
 
 import (
 	apicommon "github.com/ai-dynamo/grove/operator/api/common"
+	componentutils "github.com/ai-dynamo/grove/operator/internal/controller/common/component/utils"
 
 	corev1 "k8s.io/api/core/v1"
 )
@@ -27,6 +28,8 @@ type DeletionSorter struct {
 	Pods []*corev1.Pod
 	// ExpectedPodTemplateHash is the hash that is expected as a label on the updated pods
 	ExpectedPodTemplateHash string
+	// ExpectedPodTemplateHashes are the canonical and legacy hashes accepted during hash migration.
+	ExpectedPodTemplateHashes componentutils.HashCandidates
 }
 
 // Len returns the length of the DeletionSorter
@@ -65,7 +68,7 @@ func (s DeletionSorter) Less(i, j int) bool {
 
 	// 4. Pods with older hashes < Pods with newer hashes
 	if s.Pods[i].Labels[apicommon.LabelPodTemplateHash] != s.Pods[j].Labels[apicommon.LabelPodTemplateHash] {
-		return s.Pods[i].Labels[apicommon.LabelPodTemplateHash] != s.ExpectedPodTemplateHash
+		return !s.expectedHashMatches(s.Pods[i].Labels[apicommon.LabelPodTemplateHash])
 	}
 
 	// 5. Empty creation time pods < newer pods < older pods
@@ -73,6 +76,13 @@ func (s DeletionSorter) Less(i, j int) bool {
 		return s.Pods[i].CreationTimestamp.IsZero()
 	}
 	return s.Pods[i].CreationTimestamp.After(s.Pods[j].CreationTimestamp.Time)
+}
+
+func (s DeletionSorter) expectedHashMatches(hash string) bool {
+	if s.ExpectedPodTemplateHashes.Canonical != "" {
+		return s.ExpectedPodTemplateHashes.Matches(hash)
+	}
+	return hash == s.ExpectedPodTemplateHash
 }
 
 // isPodReady checks if a pod is ready by looking for the PodReady condition with status True
--- a/operator/internal/controller/podclique/reconcilestatus.go
+++ b/operator/internal/controller/podclique/reconcilestatus.go
@@ -69,7 +69,10 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	}
 	// mutate PodClique Status Replicas, ReadyReplicas, ScheduleGatedReplicas and UpdatedReplicas.
 	mutateReplicas(pclq, podCategories, len(existingPods))
-	mutateUpdatedReplica(pcs, pclq, existingPods)
+	if err = mutateUpdatedReplica(pcs, pclq, existingPods); err != nil {
+		logger.Error(err, "failed to compute PodClique updated replicas")
+		return ctrlcommon.ReconcileWithErrors("failed to compute PodClique updated replicas", err)
+	}
 
 	// mutate the conditions only if the PodClique has been successfully reconciled at least once.
 	// This prevents prematurely setting incorrect conditions.
@@ -159,7 +162,7 @@ func mutateReplicas(pclq *grovecorev1alpha1.PodClique, podCategories map[corev1.
 }
 
 // mutateUpdatedReplica calculates and sets the number of pods with the expected template hash
-func mutateUpdatedReplica(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grovecorev1alpha1.PodClique, existingPods []*corev1.Pod) {
+func mutateUpdatedReplica(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grovecorev1alpha1.PodClique, existingPods []*corev1.Pod) error {
 	var expectedPodTemplateHashes componentutils.HashCandidates
 	// If UpdateProgress exists (update in progress or recently completed), use the target hash from it.
 	// This covers both the active update phase and the window after completion before CurrentPodTemplateHash is synced.
@@ -194,6 +197,7 @@ func mutateUpdatedReplica(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grovecorev1
 		}, 0)
 		pclq.Status.UpdatedReplicas = int32(updatedReplicas)
 	}
+	return nil
 }
 
 // mutateSelector creates and sets the label selector for autoscaler use when scaling is configured
--- a/operator/internal/controller/podclique/reconcilestatus.go
+++ b/operator/internal/controller/podclique/reconcilestatus.go
@@ -69,7 +69,10 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 	}
 	// mutate PodClique Status Replicas, ReadyReplicas, ScheduleGatedReplicas and UpdatedReplicas.
 	mutateReplicas(pclq, podCategories, len(existingPods))
-	mutateUpdatedReplica(pclq, existingPods)
+	if err = mutateUpdatedReplica(pcs, pclq, existingPods); err != nil {
+		logger.Error(err, "failed to compute PodClique updated replicas")
+		return ctrlcommon.ReconcileWithErrors("failed to compute PodClique updated replicas", err)
+	}
 
 	// mutate the conditions only if the PodClique has been successfully reconciled at least once.
 	// This prevents prematurely setting incorrect conditions.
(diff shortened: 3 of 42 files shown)


danbar2 (dismissed):

gflarity:
After discussing with the group I see two viable options. See below for more information. 

I've decided to implement the **two-hash compatibility window fix** because it keeps the upgrade self-contained in Grove: no external migration job, no need to coordinate with upstream writers such as Dynamo, and no requirement to block CR changes during the operator restart. 


## Option 1: Pre-Start Migration During Operator Restart

During upgrade, the old Grove pod stops. The new Grove pod starts but runs a migration before starting reconcilers. The migration computes canonical hashes and patches existing stored hashes/labels so normal reconciliation sees a consistent canonical state.

  It must patch PCS, PCLQ, PCSG, and Pod hash fields, and should fail if an update is actively in progress.

  Pros:

  - Avoids long-term controller complexity.
  - No legacy hash path in normal reconcile logic.
  - Can avoid the one-time roll if migration completes before reconcilers start.

  Cons:

  - Upgrade path becomes more operationally complex.
  - Migration must be complete and idempotent.
  - Upstream spec writers, including Dynamo, should be quiesced or at least not racing PCS spec changes during migration.
  - If any hash-bearing field is missed, the new operator may still roll or report stale status.


## Option 2: Two-Hash Compatibility
  A transition Grove release computes both hashes from the current desired spec:

  canonicalHash(current spec)
  legacyHash(current spec)

 If stored hash matches either, the object is treated as current. If it matches legacy, Grove patches it to canonical during reconciliation. If it matches neither, normal rolling-update behaviour applies.

  Pros:

  - No special upgrade choreography.
  - Avoids the one-time upgrade roll for steady-state workloads.
  - Safer than blind relabeling because it only accepts legacyHash(current spec).

  Cons:

  - Requires preserving the old hash algorithm temporarily.
  - Needs a follow-up release to remove legacy support.

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -108,22 +108,70 @@ func (r *Reconciler) isGenerationHashExpectationSatisfied(pcsObjectName string,
...
+// clique order and passes clique names as order keys to the shared hash helper.
 func computeGenerationHash(pcs *grovecorev1alpha1.PodCliqueSet) string {
-	podTemplateSpecs := lo.Map(pcs.Spec.Template.Cliques, func(pclqTemplateSpec *grovecorev1alpha1.PodCliqueTemplateSpec, _ int) *corev1.PodTemplateSpec {
-		podTemplateSpec := &corev1.PodTemplateSpec{
-			ObjectMeta: metav1.ObjectMeta{
-				Labels:      pclqTemplateSpec.Labels,
-				Annotations: pclqTemplateSpec.Annotations,
-			},
-			Spec: pclqTemplateSpec.Spec.PodSpec,
+	preserveCliqueOrder := isInOrderStartup(pcs)
```
sanjaychatterjee: Is this necessary? Startup ordering is a runtime requirement. Hash generation can be agnostic and canonical.

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -108,22 +108,70 @@ func (r *Reconciler) isGenerationHashExpectationSatisfied(pcsObjectName string,
...
+	if preserveCliqueOrder {
+		return k8sutils.ComputeHashWithOrderKeys(orderKeys, podTemplateSpecs...)
+	}
+
 	return k8sutils.ComputeHash(podTemplateSpecs...)
 }
 
+func podTemplateSpecForGenerationHash(pclqTemplateSpec *grovecorev1alpha1.PodCliqueTemplateSpec, priorityClassName string) *corev1.PodTemplateSpec {
+	podTemplateSpec := &corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
```
sanjaychatterjee: I am not sure if updating the PodCliqueTemplateSpec labels and annotations should necessarily trigger rolling updates. They need to propagated to all the resources not trigger updates. This is ok for now since it preserves current behavior.
cc @renormalize who is also looking into the fix.

operator/internal/utils/kubernetes/pod.go
```diff
@@ -135,15 +136,209 @@ func HasAnyContainerNotStarted(pod *corev1.Pod) bool {
...
 	}
 	return rand.SafeEncodeString(fmt.Sprint(podTemplateSpecHasher.Sum64()))
 }
 
+// canonicalizePodTemplateSpecForHashing returns a deep-copied PodTemplateSpec
+// with every order-independent +listType=map slice sorted by its API key,
+// so that two specs representing the same desired state always produce the
+// same byte-for-byte serialization. See the doc on ComputeHash for the full
+// list of slices canonicalized and the rationale for the slices left alone.
+func canonicalizePodTemplateSpecForHashing(in *corev1.PodTemplateSpec) *corev1.PodTemplateSpec {
```
sanjaychatterjee, line 196: I agree the canonicalization is necessary to avoid triggering unnecessary rolling updates but I have a couple of questions here:
- I am concerned how much performance hit this code will now introduce.
- Is there a standard utility that is used by Kubernetes operators that we can reuse here?

cc @unmarshall @kangclzjc

operator/internal/controller/podcliqueset/reconcilespec.go
gflarity: Good catch. After looking at the update validation again, cliqueStartupType is immutable and clique reordering is already rejected where order matters. The generation hash does not need to encode startup ordering. I’ll remove the InOrder special case and keep clique hashing canonical by name.

operator/internal/controller/podcliqueset/reconcilespec.go
gflarity: Ya, I wanted to preserve the current behvaiour as much as is practical right since it's a pretty impactful change.

operator/internal/utils/kubernetes/pod.go
gflarity: > - I am concerned how much performance hit this code will now introduce.

The existing hash path already reflects over the full template via dump.ForHash, so I expect the extra deep copy and bounded slice sorts to be small.  I'm not sure what the state of the performance testings suite is, but a before and after run might make sense when it's ready. 
> - Is there a standard utility that is used by Kubernetes operators that we can reuse here?

The  thought occurred to me,  but a quick search didn't find much.  I also wanted to keep the canonical hashing logic close to current hashing logic to make reviewing easy.

danbar2: the infra for scale test is ready, I think its a good use case to add a scale test with a scenario that highlight this issue. Run it on main as baseline and on your branch to compare the results - I did a similar thing [here](https://github.com/ai-dynamo/grove/pull/545) you can use it as an example

danbar2:
> After discussing with the group I see two viable options. See below for more information.
> 
> I've decided to implement the two-hash compatibility window fix because it keeps the upgrade self-contained in Grove: no external migration job, no need to coordinate with upstream writers such as Dynamo, and no requirement to block CR changes during the operator restart.
> 
> ## Option 1: Pre-Start Migration During Operator Restart
> During upgrade, the old Grove pod stops. The new Grove pod starts but runs a migration before starting reconcilers. The migration computes canonical hashes and patches existing stored hashes/labels so normal reconciliation sees a consistent canonical state.
> 
> It must patch PCS, PCLQ, PCSG, and Pod hash fields, and should fail if an update is actively in progress.
> 
> Pros:
> 
> * Avoids long-term controller complexity.
> * No legacy hash path in normal reconcile logic.
> * Can avoid the one-time roll if migration completes before reconcilers start.
> 
> Cons:
> 
> * Upgrade path becomes more operationally complex.
> * Migration must be complete and idempotent.
> * Upstream spec writers, including Dynamo, should be quiesced or at least not racing PCS spec changes during migration.
> * If any hash-bearing field is missed, the new operator may still roll or report stale status.
> 
> ## Option 2: Two-Hash Compatibility
> A transition Grove release computes both hashes from the current desired spec:
> 
> canonicalHash(current spec) legacyHash(current spec)
> 
> If stored hash matches either, the object is treated as current. If it matches legacy, Grove patches it to canonical during reconciliation. If it matches neither, normal rolling-update behaviour applies.
> 
> Pros:
> 
> * No special upgrade choreography.
> * Avoids the one-time upgrade roll for steady-state workloads.
> * Safer than blind relabeling because it only accepts legacyHash(current spec).
> 
> Cons:
> 
> * Requires preserving the old hash algorithm temporarily.
> * Needs a follow-up release to remove legacy support.

Im in favor of the first option. 
Note that we can never know if the user upgraded Grove on every released version, if he will jump from current behavior to a version where the old annotation is removed completely he will get all its PCS's to be updated

gflarity:
Scale-test follow-up for the performance concern: I ran both 1k-pod scale tests against a fair local main baseline (09da63b, with only the new scale-test harness/YAML applied) and this branch (ce2b190) on the same 100-KWOK node scale profile. The existing ScaleTest_1000 showed no measurable regression, with total time at 547.8s on baseline vs 535.0s on this branch (-2.3%). The "Dynamo-like" variant, which better exercises the hash-canonicalization concern, was also not worse overall: total time was 688.7s baseline vs 605.7s branch (-12.1%); early deploy markers were slightly slower (pods-created +2.1s, pcs-available +3.3s), but pods-ready was 86.1s faster and the steady-state availability check was not slower.

[2026-05-09-pr566-performance.tar.gz](https://github.com/user-attachments/files/27560255/2026-05-09-pr566-performance.tar.gz)

operator/internal/utils/kubernetes/pod.go
gflarity: Done,  no noticeable degradation in performance during the scale test. I even made one that looked more "Dynamo-like" with more volumes,  containers, mounts etc.

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -140,23 +159,35 @@ func mutateReplicas(pclq *grovecorev1alpha1.PodClique, podCategories map[corev1.
...
-	var expectedPodTemplateHash string
+func mutateUpdatedReplica(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grovecorev1alpha1.PodClique, existingPods []*corev1.Pod) {
+	var expectedPodTemplateHashes componentutils.HashCandidates
 	// If UpdateProgress exists (update in progress or recently completed), use the target hash from it.
 	// This covers both the active update phase and the window after completion before CurrentPodTemplateHash is synced.
 	if pclq.Status.UpdateProgress != nil {
-		expectedPodTemplateHash = pclq.Status.UpdateProgress.PodTemplateHash
+		expectedPodTemplateHashes = componentutils.HashCandidates{
+			Canonical: pclq.Status.UpdateProgress.PodTemplateHash,
+			Legacy:    pclq.Status.UpdateProgress.PodTemplateHash,
```
danbar2, line 169: why do we still need the legacy if an update is in progress?

operator/internal/utils/kubernetes/pod.go
```diff
@@ -134,16 +135,199 @@ func HasAnyContainerNotStarted(pod *corev1.Pod) bool {
...
 func ComputeHash(podTemplateSpecs ...*corev1.PodTemplateSpec) string {
+	return computeHash(canonicalizePodTemplateSpecForHashing, nil, podTemplateSpecs...)
+}
+
+// ComputeHashWithOrderKeys computes a hash for pod templates whose caller-level
+// identity/order matters in addition to the canonicalized PodTemplateSpec
+// content. orderKeys must be aligned with podTemplateSpecs by index.
+func ComputeHashWithOrderKeys(orderKeys []string, podTemplateSpecs ...*corev1.PodTemplateSpec) string {
+	if len(orderKeys) != len(podTemplateSpecs) {
+		panic("ComputeHashWithOrderKeys: orderKeys length must match podTemplateSpecs length")
```
oleg-kushniriov, line 155: Is this intentional? Consider returning an error

operator/internal/controller/common/component/utils/podcliquescalinggroup.go
```diff
@@ -124,6 +124,14 @@ func GetPCSGsByPCSReplicaIndex(ctx context.Context, cl client.Client, pcsObjKey
 
 // GetPCLQTemplateHashes generates the Pod template hash for all PCLQs in a PCSG. Returns a map of [PCLQ Name : PodTemplateHas]
 func GetPCLQTemplateHashes(pcs *grovecorev1alpha1.PodCliqueSet, pcsg *grovecorev1alpha1.PodCliqueScalingGroup) map[string]string {
```
oleg-kushniriov, line 126: Seems to be unused func

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -108,20 +114,23 @@ func (r *Reconciler) isGenerationHashExpectationSatisfied(pcsObjectName string,
...
+//
+// pcs.Spec.Template.Cliques is +listType=map +listMapKey=name, so it is an
+// unordered set; the hash must be stable across reorderings to avoid falsely
+// triggering rolling recreates when an upstream operator emits the same
+// cliques in a different sequence.
+//
+// Order is semantically meaningful when StartupType is CliqueStartupTypeInOrder,
+// so in that mode the original clique order is preserved and clique names are
+// passed as order keys to the shared hash helper.
 func computeGenerationHash(pcs *grovecorev1alpha1.PodCliqueSet) string {
```
oleg-kushniriov, line 128:
```
  func computeGenerationHash(pcs ...) string { return componentutils.ComputePCSGenerationHash(pcs) }
  func computeGenerationHashLegacy(pcs ...) string { return componentutils.ComputePCSGenerationHashLegacy(pcs) }
```
  Only used by tests. Tests can call `componentutils.ComputePCSGenerationHash` directly.

operator/internal/controller/podclique/components/pod/deletionsort.go
```diff
@@ -27,6 +28,8 @@ type DeletionSorter struct {
 	Pods []*corev1.Pod
 	// ExpectedPodTemplateHash is the hash that is expected as a label on the updated pods
 	ExpectedPodTemplateHash string
```
oleg-kushniriov, line 30: Consider dropping `ExpectedPodTemplateHash` in favor of `ExpectedPodTemplateHashes` — every constructor in this PR already sets both, and `HashCandidates.Matches(hash) `covers the canonical value that the single-string field used to hold, so the fallback branch in expectedHashMatches is unreachable.

operator/internal/utils/kubernetes/pod.go
gflarity: Resolving, feel free to reach out if there are still concerns.

operator/internal/controller/podcliqueset/reconcilespec.go
gflarity: Resolving, feel free to reach out if there are still concerns.

operator/internal/utils/kubernetes/pod.go
gflarity: This is intentional. `ComputeHashWithOrderKeys` is an internal helper, and the only production caller builds `orderKeys` and pod templates from the same clique slice, so a length mismatch is a programming invariant violation rather than recoverable user input. I kept the panic to avoid pushing an impossible error path through the hash/reconcile stack, but documented that behavior in `0ddddc1`.

operator/internal/controller/common/component/utils/podcliquescalinggroup.go
gflarity: Agreed. This helper became unused after switching callers to `HashCandidates`. Removed it in `0ddddc1`.

operator/internal/controller/podclique/reconcilestatus.go
gflarity: Good question. We still need this during the one-release hash migration window: `UpdateProgress.PodTemplateHash` may already contain either the canonical or legacy value, depending on which controller version wrote it. The local candidate starts from the stored progress hash, then expands back to the current desired canonical/legacy candidates when it matches the current spec. I added a clarifying comment and a regression test for the legacy `UpdateProgress` case in `0ddddc1`.

operator/internal/controller/podcliqueset/reconcilespec.go
gflarity: Agreed. These wrappers were only test convenience after the implementation moved into `componentutils`. Removed them and updated the tests to call `componentutils.ComputePCSGenerationHash*` directly in `0ddddc1`.

operator/internal/controller/podclique/components/pod/deletionsort.go
gflarity: Agreed. The single-hash fallback was no longer reachable in production now that constructors pass `ExpectedPodTemplateHashes`. Removed `ExpectedPodTemplateHash`, the sync-context single-hash field, and the fallback branches in `0ddddc1`.

oleg-kushniriov (approved):

danbar2 (approved):

PR #574, opened 2026-05-03
refactor(mnnvl): replace two-annotation model with single mnnvl-group annotation

shmuel-runai:
/kind feature

Ref #417

## Summary

- Remove `grove.io/auto-mnnvl` annotation — use `grove.io/mnnvl-group` as the sole annotation controlling MNNVL enrollment and group assignment
- Not backward compatible with Phase 1 (acceptable since the feature is still in alpha)
- Simplify resolution logic, webhook validation, propagation, and CD naming

## Changes

- **Constants:** Remove `AnnotationAutoMNNVL`, `AnnotationAutoMNNVLEnabled`, `AnnotationAutoMNNVLDisabled`; add `AnnotationMNNVLGroupOptOut` (`"none"`)
- **Helpers:** Introduce `groupStatus` enum (`groupEnrolled`/`groupWithdrawn`/`groupAbsent`) replacing the two-boolean return from `resolveGroupName`; simplify `ResolveGroupNameHierarchically`; use case-sensitive matching per K8s convention
- **Webhook:** Remove `auto-mnnvl` validation and conflict detection; immutability check now only covers `mnnvl-group`
- **Propagation:** `propagateMNNVLAnnotations` only propagates `mnnvl-group` (PCS → PCSG)
- **ComputeDomain:** `generateComputeDomainName` always uses `{pcs}-{replica}-{group}` format (single scheme)
- **Shared utility:** Extract generic `ValidateAnnotationsImmutability` to `utils/kubernetes/annotations.go` with full test coverage
- **Tests:** All unit tests updated for single-annotation semantics

## Test plan

- [x] `go build ./...` passes
- [x] All unit tests pass (`go test ./...`)
- [ ] E2E tests (deferred to follow-up PR)

Made with Cursor


athreesh:
worth fixing: can we reject the legacy `grove.io/auto-mnnvl` annotation instead of silently ignoring it?

i buy the alpha-breaking-change argument, but silent acceptance is a bad failure mode: a user can submit an old manifest, think MNNVL is enabled, and just get worse performance with no obvious signal. even if we do not support backward compatibility, the webhook should probably fail fast with “use grove.io/mnnvl-group: default instead” so the migration is explicit.

shmuel-runai:
@athreesh 
> worth fixing: can we reject the legacy `grove.io/auto-mnnvl` annotation instead of silently ignoring it?
> 
> i buy the alpha-breaking-change argument, but silent acceptance is a bad failure mode: a user can submit an old manifest, think MNNVL is enabled, and just get worse performance with no obvious signal. even if we do not support backward compatibility, the webhook should probably fail fast with “use grove.io/mnnvl-group: default instead” so the migration is explicit.

I thought about this, however, in phase 1, the legacy `grove.io/auto-mnnvl` annotation wasn't added by the user; it was added automatically by grove using a webhook. 
and the last thing I want for grove it to maintaine a "zombie" code to supprot for a rare case of a API that was depreacted after a month in alpha.

operator/internal/mnnvl/computedomain/computedomain_test.go
shayasoolin, line 82: There are 5 different tests, but they all share the same use case. I don't see how they differ from each other.

operator/internal/mnnvl/computedomain/computedomain_test.go
```diff
@@ -708,10 +710,10 @@ func createClientThatFailsOnList() client.Client {
 }
 
 // createTestCDs creates a slice of test ComputeDomains with proper labels and finalizers.
-func createTestCDs(pcsName, namespace string, count int) []*unstructured.Unstructured {
+func createTestCDs(pcsName, namespace, groupName string, count int) []*unstructured.Unstructured {
```
shayasoolin: I might be missing it - but do you have tests in this file for cases with multiple CDs of different groups (not just different PCS replicas)? For example, two PCLQs each with its own group.

operator/internal/mnnvl/constants.go
```diff
@@ -42,28 +42,17 @@ var ComputeDomainGVK = schema.GroupVersionKind{
...
-	// not required when mnnvl-group is set.
 	// The value must be a valid Kubernetes name component (lowercase alphanumeric or dashes,
-	// starting and ending with alphanumeric, max 63 characters).
+	// starting and ending with alphanumeric, max 63 characters), or the reserved value "none"
+	// to explicitly opt out of MNNVL (overriding a parent layer's setting).
 	AnnotationMNNVLGroup = "grove.io/mnnvl-group"
 
+	// AnnotationMNNVLGroupOptOut is the reserved value for AnnotationMNNVLGroup that
+	// explicitly opts out of MNNVL, overriding any parent layer's group assignment.
+	AnnotationMNNVLGroupOptOut = "none"
```
shayasoolin, line 54: Are `NONE` and `None` accepted as well? (I think they should)

operator/internal/mnnvl/helpers.go
```diff
@@ -21,103 +21,73 @@ import (
...
-		return fmt.Errorf("mnnvl-group value %q is not a valid DNS-1123 label: %s", name, strings.Join(errs, "; "))
-	}
-	return nil
-}
-
-// DetectMNNVLConflict returns an error if the annotations contain contradictory
-// MNNVL settings: auto-mnnvl: disabled combined with a mnnvl-group value.
-func DetectMNNVLConflict(annotations map[string]string) error {
-	if annotations == nil {
+	if name == AnnotationMNNVLGroupOptOut {
```
shayasoolin, line 47:
```suggestion
	if strings.ToLower(name) == AnnotationMNNVLGroupOptOut {
```

operator/internal/mnnvl/helpers.go
```diff
@@ -21,103 +21,73 @@ import (
...
 
 // ResolveGroupNameHierarchically resolves the MNNVL group from multiple
 // annotation layers, ordered from most specific to least specific
-// (e.g., PCLQ → PCSG/PCS). The first layer that requests MNNVL wins —
-// this lets a child layer intentionally override its parent's group,
-// including escaping a named group back to the default group by setting
-// only auto-mnnvl: enabled without mnnvl-group.
+// (e.g., PCLQ → PCSG → PCS). The first layer where the annotation is
+// present wins — even if it's "none" (withdrawn), which stops the walk.
 func ResolveGroupNameHierarchically(annotationLayers ...map[string]string) (string, bool) {
```
shayasoolin, line 75: Nit, I think a clearer implementation for this function could be more straighforward without the auxiliary function of `resolveGroupName`, which adds new terms of _enrolled_, _withdrawn_ and _absent_. 
You can simply stop the loop at 'none' or some group name and return the name / status.

operator/internal/mnnvl/webhook_test.go
```diff
@@ -33,27 +33,21 @@ func TestValidatePCSOnCreate_Metadata(t *testing.T) {
...
 		{
-			description:      "annotation disabled + feature disabled -> no error",
-			pcs:              createPCSWithGPU(map[string]string{AnnotationAutoMNNVL: AnnotationAutoMNNVLDisabled}),
-			autoMNNVLEnabled: false,
-			expectError:      false,
-		},
-		{
-			description:      "annotation disabled + feature enabled -> no error",
-			pcs:              createPCSWithGPU(map[string]string{AnnotationAutoMNNVL: AnnotationAutoMNNVLDisabled}),
+			description:      "mnnvl-group 'none' + feature enabled -> no error",
```
shayasoolin, line 49: What about "mnnvl-group 'none' + feature disabled -> no error"?

operator/internal/mnnvl/computedomain/computedomain_test.go
shmuel-runai: Good catch — reduced to 2 test cases: one for the default group and one for a named group with a higher replica index. The remaining cases were effectively testing string interpolation with different inputs.

operator/internal/mnnvl/computedomain/computedomain_test.go
shmuel-runai: Good point — added `TestSyncMultipleGroups` (creates CDs for two distinct groups across replicas) and `TestSyncMultipleGroupsScaleDown` (verifies excess CDs are removed per-group when scaling down). Both tests exercise the multi-group reconciliation path that was previously only covered by `TestGetRequiredCDNames`.

operator/internal/mnnvl/constants.go
shmuel-runai: We deliberately keep annotation values case-sensitive here. Kubernetes annotations, labels, and field selectors are case-sensitive by convention — for example, label selectors like `app=MyApp` vs `app=myapp` are distinct, and tools like `kubectl` treat annotation values as opaque case-sensitive strings. Since the group name becomes part of the ComputeDomain resource name (which must be a DNS-1123 label and therefore lowercase), the DNS validation layer already rejects uppercase values like `NONE` or `None`. Keeping the opt-out check case-sensitive (`== "none"`) is consistent with this and avoids surprises where `None` would silently opt out instead of being rejected by DNS validation.

operator/internal/mnnvl/helpers.go
shmuel-runai: See reply above on the case-sensitivity rationale — since the group name must pass DNS-1123 validation (lowercase only), keeping the opt-out check as an exact match with \`"none"\` is consistent. Using \`strings.ToLower\` would create an inconsistency where \`None\` or \`NONE\` silently opts out while \`Workers\` or \`DEFAULT\` would be rejected by DNS validation.

operator/internal/mnnvl/helpers.go
shmuel-runai: We considered inlining but prefer the current split for a couple of reasons:

1. **Separation of concerns**: `resolveGroupName` handles a single annotation map, `ResolveGroupNameHierarchically` handles the walk. If we need to change the resolution logic (e.g. new annotation values), only `resolveGroupName` needs to change.
2. **The `groupStatus` enum makes the three states explicit**: absent (inherit from parent), withdrawn (opt-out), and enrolled. Without it, `ResolveGroupNameHierarchically` would need to distinguish "annotation not present" from "annotation is 'none'" using string comparisons inline, which we found less readable during development.

That said, it's a style preference — happy to revisit if you feel strongly about it.

operator/internal/mnnvl/webhook_test.go
shmuel-runai: Great catch — added the test case and also fixed the webhook logic. Previously, `validateMNNVLAnnotationsOnCreate` rejected *any* annotation value when the feature was disabled, including "none". Now it skips the feature-enabled check for the opt-out value, since opting out when the feature is disabled is a harmless no-op.

operator/internal/mnnvl/constants.go
shayasoolin: If uppercase names are rejected earlier then it's fine.

operator/internal/mnnvl/computedomain/computedomain_test.go
```diff
@@ -88,39 +88,18 @@ func Test_generateComputeDomainName(t *testing.T) {
 		expected     string
 	}{
 		{
-			description:  "default group replica 0",
+			description:  "default group",
```
shayasoolin, line 91: Just to make sure I'm not missing something: what makes the name `default` special? What's the difference between `default` and `training` or `lalala`?
Asking as the second test that you left mentions 'named group'.

shayasoolin (approved):

danbar2 (approved):

PR #575, opened 2026-05-03
Scale Test CI

### LINKED ISSUE #550: Scheduled CI: scale-test trend tracking and comparison

A scheduled CI job that runs the scale test, compares the new result against the previously stored run, and persists the result to shared storage for trend tracking.

Concrete scope:

1. **Scheduled trigger.** A GitHub Actions workflow (e.g. `.github/workflows/scale-test.yaml`) that runs all existing scale tests on a cron (plus `workflow_dispatch` for manual runs). Each run is tagged with commit SHA, and timestamp.

2. **Shared result storage.** Each completed run uploads its `scale-test-results.json` (and pprof artifacts when available) to a shared backing store, indexed by commit SHA + timestamp. Open question for discussion: which backend — a dedicated branch in this repo (`scale-test-history`), an S3/GCS bucket, or a gh-pages site. Whichever we pick should be cheap, append-only, and queryable from CI without extra credentials beyond what GHA already has.

3. **Compare against latest.** After a run finishes, the job pulls the most recent stored result for the same test, and emits a side-by-side report:
   - Per-test, per-milestone wall-clock (e.g. `pods-created`, `pods-ready`, `pcs-available`, `pcs-deleted`, total) with absolute + percent delta vs. previous.
   - Per-test, per-phase totals (e.g. deploy / steady-state / delete).
   - When pprof is captured, top-N CPU samples per controller `Reconcile` with deltas.
   - Output in markdown so it can be posted as a job summary and (optionally) commented on the triggering commit / PR.

4. **Trend tracking.** Because every run is appended to shared storage, generating a "last N runs" trend chart for each milestone becomes a separate, cheap step (a small static page or a script that emits CSV/SVG from the stored JSONs).

Suggested implementation outline:
- Extend `operator/e2e/measurement/exporter` with a `CompareExporter` that takes two `TrackerResult`s and emits the markdown report + a structured diff.
- Small CLI under `operator/hack/` (`scale-compare`) that loads two JSON files and runs the exporter — usable both in CI and by developers locally.
- New workflow under `.github/workflows/` that wires up: scale-cluster-up → run-scale-test → fetch previous result from storage → run `scale-compare` → upload new result → post job summary.

## Why is this needed?

Operator perf work (the recent reconcile-CPU optimizations on the 1k-pod scenario, and the upcoming 5k-pod delete QPS work) keeps running into the same problem: there is no continuous signal on scale-test performance. We only know whether a change moved the numbers when someone manually runs baseline-vs-optimized and hand-builds a comparison table from `scale-test-results.json`.

A scheduled CI job that runs the scale test, persists results, and compares against the previous run gives us:

- **Trend visibility.** We see drift over time across many small changes, not just the deltas around a single perf-focused PR. Quiet regressions that nobody attributed to any single PR become visible.
- **A canonical history.** Reviewers and design docs can link to the stored run for a given commit instead of relying on `/tmp/...` dirs on a developer's laptop.
- **Lower cost-per-comparison.** The current manual workflow (run twice, open two JSONs, transcribe into a table) is slow enough that we don't do it for most PRs. Automating it means perf comparison becomes the default, not a special effort.

shayasoolin:
#### What type of PR is this?

  /kind feature

  Adds the first version of Grove scale-test CI history tracking.

  This PR adds:

  - operator/hack/scale-history.py for storing scale-test results in a history branch.
  - A compact index/metrics.ndjson metric index plus raw scale-test-results.json storage.
  - A minimal static dashboard under operator/hack/scale-dashboard/.
  - A scheduled/manual GitHub Actions workflow that runs the scale test and stores successful results.
  - README docs for local ingestion and dashboard validation.

  The workflow disables routine profiling/pprof collection, compiles the scale test before creating the KWOK cluster, and uploads short-retention diagnostics on failure.

  Fixes #550

  The workflow is intentionally manual/scheduled only for v1. It does not run on PR labels.

  The history branch is expected to be public and served by GitHub Pages. Routine runs store only timing history and raw scale-test JSON, not profiling artifacts.

  Validation done:

  - Parsed workflow YAML.
  - Ran git diff --check.
  - Verified scale-history.py --help.
  - Ran scale-history.py local on a sample result and confirmed duplicate ingestion is skipped.

  NONE

  operator/hack/README.md documents scale-history.py local/branch usage and local dashboard validation.


operator/hack/scale-history.py
```diff
@@ -0,0 +1,459 @@
...
+            raise SystemExit(f"workdir exists and is not empty: {workdir}")
+        workdir.parent.mkdir(parents=True, exist_ok=True)
+        git(["clone", repo, str(workdir)])
+        git(["fetch", "origin"], cwd=workdir)
+
+    if remote_branch_exists(workdir, branch):
+        git(["checkout", "-B", branch, f"origin/{branch}"], cwd=workdir)
+    else:
+        git(["checkout", "--orphan", branch], cwd=workdir)
+        clear_worktree(workdir)
```
gflarity, line 337: After `git checkout --orphan`, Git keeps the start point in the index. Since `clear_worktree()` only removes files from disk, the first commit on a missing history branch can still include the cloned source tree along with the history files. We should clear the index before adding the history output, for example with `git rm -r --cached .` or `git read-tree --empty`.

.github/workflows/scale-test-ci.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: Create scale cluster
+        working-directory: operator
+        run: |
+          make scale-cluster-up E2E_CREATE_FLAGS='--dind-memory-mode --set pyroscope.enabled=false --set grove.profiling=false'
+
+      - name: Run scale test
+        working-directory: operator/e2e/tests/scale
+        env:
+          GROVE_E2E_DIAG_DIR: ${{ runner.temp }}/grove-scale
```
gflarity, line 50: With `GROVE_E2E_DIAG_MODE=file`, failed scale tests write diagnostics under `${{ runner.temp }}/grove-scale`, but this workflow never uploads that directory. The files disappear when the runner exits, so failed scheduled runs lose the detailed `testctx` diagnostics. Can we add a failure-only `actions/upload-artifact` step for this path before cleanup?

operator/hack/README.md
```diff
@@ -62,6 +62,37 @@ pip3 install -r hack/requirements.txt
 ./hack/infra-manager.py setup --config scale.yaml --set kwok.nodes=1000
 ```
 
+### scale-history.py
+
+Stores a scale-test result in Grove's benchmark history format and copies the
+static dashboard files into the history root.
+
+```bash
+# Local testing; writes history under /tmp/grove-scale-history by default.
```
danbar2, line 71: do we need to add that to .gitignore ?

operator/hack/scale-dashboard/app.js
```diff
@@ -0,0 +1,622 @@
...
+  rows: [],
+  testName: "",
+};
+
+const els = {
+  status: document.getElementById("status"),
+  testSelect: document.getElementById("test-select"),
+  totalChart: document.getElementById("total-chart"),
+  phaseChart: document.getElementById("phase-chart"),
+  deployChart: document.getElementById("deploy-chart"),
```
danbar2: We need to be more generic than that, to support new tests/phases/milestones out of the box. it is possible to make some special adjustments of needed, but if none are defined the results still needs to show up

shayasoolin: Done, the implementation is now resilient to additions of new phases and milestones. It doesn't hard-code any names. Instead, it uses the original order from the result json files (which uses arrays and not maps, so order is preserved).
The `metrics.ndjson` file that held each metric as a separate line, was replaced with `runs.ndjson` that has a line for every run, not metric. It makes it easier to maintain the order of phases/milestones. (I think it also simplifies some other parts of the code as well)

danbar2 (dismissed):

gflarity (commented):
Nice!

operator/hack/scale-history.py
```diff
@@ -0,0 +1,465 @@
...
+    result_path = find_result_file(Path(args.input))
+    result = read_json(result_path)
+    commit = args.commit or best_effort_git(["rev-parse", "HEAD"])
+    run_url = args.run_url or ""
+
+    timestamp = run_timestamp(result)
+    test_name = required_str(result, "testName")
+    run_id = required_str(result, "runID")
+
+    if args.command == "local":
```
shmuel-runai: the flow here should be
```
history_dir = calcHistoryDir <-- new function the calc what the history dir should be
changed = append_history(...
copy_dashboard_files(Path(args.history_dir))
if args.command == "remote":
   commit_and_push(history_dir, args.branch, result, changed)
 return 0 if changed is not None else 1
```

operator/hack/scale-history.py
```diff
@@ -0,0 +1,465 @@
...
+def required_str(obj: dict[str, Any], key: str) -> str:
+    value = obj.get(key)
+    if not isinstance(value, str) or not value:
+        raise ValueError(f"missing required string field: {key}")
+    return value
+
+
+def safe_path_part(value: str) -> str:
+    value = value.strip().replace("/", "-")
+    return re.sub(r"[^A-Za-z0-9._-]+", "-", value) or "unknown"
```
oleg-kushniriov: Consider this:
```suggestion
  def safe_path_part(value: str) -> str:
      value = value.strip().replace("/", "-")
      value = re.sub(r"[^A-Za-z0-9._-]+", "-", value)
      value = value.lstrip(".")
      return value or "unknown"
```

result file containing testName: ".." or runID: ".." would let result_history_path write outside results

operator/hack/scale-history.py
```diff
@@ -0,0 +1,465 @@
...
+    with path.open() as f:
+        return json.load(f)
+
+
+def append_history(
+    history_dir: Path,
+    result_path: Path,
+    result: dict[str, Any],
+    run_record: dict[str, Any],
+) -> bool | None:
```
shmuel-runai: When does this return None? Also, the caller expects None

oleg-kushniriov (approved):

danbar2 (approved):

PR #576, opened 2026-05-03
bound UpdateProgress status payload

### LINKED ISSUE #567: Status field UpdatedPodCliques []string in PCS/PCSG does not scale —  unbounded status payload at thousands of PodCliques and O(N²) reconciler CPU
### What happened?

                                                                                                                                                                                                                                                                                            
  The UpdateProgress.UpdatedPodCliques field on PodCliqueSetStatus (https://github.com/ai-dynamo/grove/blob/main/operator/api/core/v1alpha1/podcliqueset.go#L159-L160) and PodCliqueScalingGroupStatus                                                                                      
  (https://github.com/ai-dynamo/grove/blob/main/operator/api/core/v1alpha1/scalinggroup.go#L153-L157) is an unbounded []string carrying the fully-qualified name of every PodClique that has been updated to the desired generation hash. The PCS additionally carries
  UpdatedPodCliqueScalingGroups []string with the same shape, and the deprecated RollingUpdateProgress.UpdatedPodCliques is kept in sync — so each FQN is effectively persisted twice on the PCS, and a PCSG-owned PCLQ can be recorded up to four times across the cluster (PCS + PCS      
  mirror + PCSG + PCSG mirror).

  Cost is driven by the total number of PodCliques in the PCS. Two failure modes appear at the scale Grove is targeting, well before any hard etcd limit:                                                                                                                                   
   
  1. Storage / wire cost grows linearly. Realistic FQN is ~50–60 chars (~60 bytes JSON-encoded, ~120 bytes per PCLQ once the deprecated mirror is included):                                                                                                                                
                  
  - 1 000 PCLQs → ~120 KiB of status payload.                                                                                                                                                                                                                                               
  - 2 000 PCLQs → ~240 KiB. kubectl get pcs -o yaml visibly slow.
  - 4 000 PCLQs → ~480 KiB. API-server / watch-cache pressure.                                                                                                                                                                                                                              
  - ~13 000 PCLQs → ~1.5 MiB, exceeding etcd's default per-object limit. Status writes are rejected, the controller cannot persist progress.                                                                                                                                                
                                                                                                                                                                                                                                                                                            
  Past ~100 KiB of status payload (reached at fewer than 1 000 PodCliques), serialization overhead and list/watch latency become noticeable. Every status write broadcasts the full object to every watcher (operator caches, dashboards, kubectl watches) because Kubernetes watches do not
   do field-level patches.                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                            
  2. Reconciler CPU is O(N²) per reconcile step.

  Status fields that summarize update progress should have bounded size and bounded per-reconcile cost, regardless of how many PodCliques the PCS contains.                                                                                                                                 
                  
  Concretely, I would expect:                                                                                                                                                                                                                                                               
                  
  - The PCS and PCSG status to carry UpdatedPodCliquesCount / TotalPodCliquesCount (and the equivalent pair for PCSGs on the PCS) as fixed-width int32 fields, plus a small derived display string (e.g. PodCliquesUpdateProgress = "344/600") for kubectl printer columns — instead of the 
  unbounded []string of FQNs.
  - Counts to be derived each reconcile from live child generation-hash labels (which the controller already loads into the informer cache for computeUpdateProgress), not accumulated. This is idempotent by construction — no drift on scale-in, re-queue, status-write retry, or         
  controller restart — and replaces today's O(N²) cleanup block with an O(N) in-memory walk over data already in memory, with zero new API-server reads.                                                                                                                                    
  - The deprecated RollingUpdateProgress.UpdatedPodCliques / UpdatedPodCliqueScalingGroups mirrors to be retired rather than kept in sync.
  - kubectl get pcs and kubectl get pcsg to surface progress as a single PCLQS-UPDATED 344/600 printer column instead of a list of FQNs that is unreadable at scale.                                                                                                                        
                                                                                                                                                                                                                                                                                            
  End result: status payload drops from O(N · 60 B) ×2 to a small constant, the reconciler hot path drops from O(N²) to O(N), and the operator works correctly at the fleet sizes Grove is designed for.                                                                                    
            
[issue-updated-podcliques-scale.md](https://github.com/user-attachments/files/27234806/issue-updated-podcliques-scale.md)

### Environment

The bug is design-level and reproducible on any environment that runs a PCS with enough PodCliques (the painful threshold begins at <1 000 PCLQs); the specific Kubernetes version, scheduler, or hardware do not gate the symptoms. Fill in your local details below for the report.     

- Grove version - v0.1.0-alpha.8

oleg-kushniriov:
#### What type of PR is this?

  /kind bug
  /kind api

  The `UpdateProgress.UpdatedPodCliques` and `UpdateProgress.UpdatedPodCliqueScalingGroups` fields on `PodCliqueSet`/`PodCliqueScalingGroup` status were unbounded `[]string` slices that grew linearly with the total number of PodCliques/PCSGs in the cluster. At the scale Grove
  targets, this produced two failure modes well before the etcd 1.5 MiB per-object limit:

  1. **Status payload growth.** Each reconcile broadcast the full FQN list to every watcher (operator caches, dashboards, `kubectl watch`). At ~1 000 PodCliques the payload was ~120 KiB; at ~13 000 it would exceed etcd's per-object limit. The deprecated `RollingUpdateProgress` mirror
   doubled the cost.
  2. **Reconciler O(N²) hot path.** The cleanup block in `rollingupdate.go` did `lo.Uniq(append(...))` followed by `slices.DeleteFunc` over the expected-FQN list every reconcile. A second O(M·E) pattern in `reconcileStatus` re-flattened `lo.Values(map)` per filtered element.

  This PR replaces the unbounded slices with bounded `int32` count + total fields:

  - `PodCliqueSetUpdateProgress`: `UpdatedPodCliquesCount`, `TotalPodCliquesCount`, `UpdatedPodCliqueScalingGroupsCount`, `TotalPodCliqueScalingGroupsCount`.
  - `PodCliqueScalingGroupUpdateProgress`: `UpdatedPodCliquesCount`, `TotalPodCliquesCount`.

  Counts are recomputed each reconcile from child `CurrentPodCliqueSetGenerationHash` labels already in the informer cache — idempotent by construction (no drift on scale-in, requeue, status-write retry, or controller restart) and O(N) instead of O(N²). The stray-resource filter is
  hoisted to a `map[string]struct{}` lookup with `slices.DeleteFunc`, making it O(M+E) instead of O(M·E).

  The deprecated `RollingUpdateProgress` mirror is preserved with the new bounded shape so existing consumers keep working at the field-presence level. Slice sub-fields are gone from the mirror as well.                                                                                  
   
  Adds four `kubectl get pcs` printer columns and two `kubectl get pcsg` printer columns surfacing the new counts.                                                                                                                                                                          
  2. **Reconciler O(N²) hot path.** The cleanup block in `rollingupdate.go` did `lo.Uniq(append(...))` followed by `slices.DeleteFunc` over the expected-FQN list every reconcile. A second O(M·E) pattern in `reconcileStatus` re-flattened `lo.Values(map)` per filtered element.

  This PR replaces the unbounded slices with bounded `int32` count + total fields:

  - `PodCliqueSetUpdateProgress`: `UpdatedPodCliquesCount`, `TotalPodCliquesCount`, `UpdatedPodCliqueScalingGroupsCount`, `TotalPodCliqueScalingGroupsCount`.
  - `PodCliqueScalingGroupUpdateProgress`: `UpdatedPodCliquesCount`, `TotalPodCliquesCount`.

  Counts are recomputed each reconcile from child `CurrentPodCliqueSetGenerationHash` labels already in the informer cache — idempotent by construction (no drift on scale-in, requeue, status-write retry, or controller restart) and O(N) instead of O(N²). The stray-resource filter is
  hoisted to a `map[string]struct{}` lookup with `slices.DeleteFunc`, making it O(M+E) instead of O(M·E).

  The deprecated `RollingUpdateProgress` mirror is preserved with the new bounded shape so existing consumers keep working at the field-presence level. Slice sub-fields are gone from the mirror as well.

  Adds four `kubectl get pcs` printer columns and two `kubectl get pcsg` printer columns surfacing the new counts.

  Fixes #567

  - **Breaking change to status fields.** Any consumer that read `pcs.Status.UpdateProgress.UpdatedPodCliques` (or the deprecated mirror's slice copy) will now find the field absent. The replacement is `UpdatedPodCliquesCount` / `TotalPodCliquesCount`. Justified at `v0.1.0-alpha.8`;
  called out in release notes.
  - **Semantic preserved from the original.** PCS-level counts cover *standalone* PCLQs only (matching the old slice's semantics — `pri.pclqs` was always loaded with `KindPodCliqueSet` owner). PCSG-owned PCLQs are tracked on their owning PCSG via
  `PodCliqueScalingGroupUpdateProgress.UpdatedPodCliquesCount`. We discussed during review whether to aggregate at the PCS level; current implementation keeps existing semantics.
  - **Mirror change.** `RollingUpdateProgress` is not removed (already deprecated; deprecation contract honored). Its inner slice fields are gone but timestamps + `CurrentlyUpdating` + the new counts are mirrored. Consumers iterating the old slices will see `nil`/empty and should
  migrate.
  - **Tests.** New unit tests cover `computeAvailableAndUpdatedReplicas` count outputs, `mutateReplicas` write paths (PCS + PCSG), and the count helpers (`countUpdatedPCLQs`, `countUpdatedPCSGs`, `countPCSGReplicaUpdatedPCLQs`, `flattenNamesToSet`). All target the new canonical
  fields, not the deprecated mirror.
  - **Generated artifacts regenerated.** `zz_generated.deepcopy.go`, both CRD YAMLs (`api/core/v1alpha1/crds/` + `charts/crds/`), and `docs/api-reference/operator-api.md` are regenerated via `make generate` and `hack/generate-api-docs.sh`.
  - **`make lint` and `go vet` (default + `-tags=e2e`) clean.** Full unit-test suite green except for one pre-existing `TestDecodeOperatorConfig` failure in `api/config/v1alpha1` that is unrelated to this change and reproduces on `main`.

  ```release-note
  action required: PodCliqueSet.Status.UpdateProgress.UpdatedPodCliques and UpdatedPodCliqueScalingGroups (and PodCliqueScalingGroup.Status.UpdateProgress.UpdatedPodCliques) are removed. Use the new bounded int32 counts instead: UpdatedPodCliquesCount/TotalPodCliquesCount and
  UpdatedPodCliqueScalingGroupsCount/TotalPodCliqueScalingGroupsCount on the PodCliqueSet, and UpdatedPodCliquesCount/TotalPodCliquesCount on the PodCliqueScalingGroup. The deprecated RollingUpdateProgress mirror is preserved with the new bounded shape; its inner slice fields are
  removed.

  docs/api-reference/operator-api.md regenerated to reflect the new count fields and the trimmed RollingUpdateProgress mirror. Four new printer columns on PodCliqueSet (`Replicas`, `Available`, `Updated`, `PCLQS-Updated`, `PCLQS-Total`, `PCSGS-Updated`, `PCSGS-Total`) and two on
  PodCliqueScalingGroup (`PCLQS-Updated`, `PCLQS-Total`) surface progress at scale where the previous FQN slices were unreadable.


operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -110,13 +114,40 @@ func mutateReplicas(logger logr.Logger, currentPCSGenerationHash *string, pcsg *
 		if isUpdated {
 			updatedReplicas++
 		}
+		updatedPCLQs += countPCSGReplicaUpdatedPCLQs(currentPCSGenerationHash, pclqs)
```
gflarity, line 125: nit: `mutateReplicas` still derives status from every PCLQ group returned by the cache, while the totals come from the current PCSG spec. During scale-down, stale non-terminating children for old replica indexes can briefly make the PCSG publish inflated replica/progress counts, including `UpdatedPodCliquesCount > TotalPodCliquesCount`. This should be incredibly brief (hence the nit), however we can handle this the same way PCS status handles child status: iterate only expected replica indexes and filter each replica's PCLQs to the expected names before computing status.

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -115,17 +123,32 @@ type PodCliqueSetUpdateStrategy struct {
 }
 
 // PodCliqueSetRollingUpdateProgress captures the progress of a rolling update of the PodCliqueSet.
-// Deprecated: Use PodCliqueSetUpdateProgress instead. This struct is maintained for backward compatibility.
+// Deprecated: Use PodCliqueSetUpdateProgress instead. Maintained for backward compatibility.
```
danbar2: if its already marked as deprecated - can we remove it instead of breaking the API ?

operator/api/core/v1alpha1/crds/grove.io_podcliquescalinggroups.yaml
```diff
@@ -16,7 +16,23 @@ spec:
...
+      name: Replicas
+      type: integer
+    - jsonPath: .status.availableReplicas
+      name: Available
+      type: integer
+    - jsonPath: .status.updatedReplicas
+      name: Updated
+      type: integer
+    - jsonPath: .status.updateProgress.updatedPodCliquesCount
+      name: PCLQS-Updated
```
shayasoolin:
```suggestion
      name: PCLQs-Updated
```

shayasoolin: See multiple places in multiple yamls

docs/api-reference/operator-api.md
```diff
@@ -545,8 +553,10 @@ _Appears in:_
 | --- | --- | --- | --- |
 | `updateStartedAt` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#time-v1-meta)_ | UpdateStartedAt is the time at which the rolling update started for the PodCliqueSet. |  |  |
 | `updateEndedAt` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#time-v1-meta)_ | UpdateEndedAt is the time at which the rolling update ended for the PodCliqueSet. |  |  |
-| `updatedPodCliqueScalingGroups` _string array_ | UpdatedPodCliqueScalingGroups is a list of PodCliqueScalingGroup names that have been updated to the desired PodCliqueSet generation hash. |  |  |
-| `updatedPodCliques` _string array_ | UpdatedPodCliques is a list of PodClique names that have been updated to the desired PodCliqueSet generation hash. |  |  |
+| `updatedPodCliquesCount` _integer_ | UpdatedPodCliquesCount mirrors PodCliqueSetUpdateProgress.UpdatedPodCliquesCount. | 0 |  |
```
shayasoolin: In the PCS level, when PCLQ is count, is it only the standalone PCLQs or both standalone and PCSG-owned ones? I assume you keep the same behavior as before just with counts instead of names.

operator/e2e/tests/update/utils.go
```diff
@@ -518,17 +518,17 @@ func waitForOrdinalUpdating(tc *testctx.TestContext, ordinal int32) error {
 		// Log status every few polls for debugging
 		if pollCount%3 == 1 {
 			currentOrdinal := int32(-1)
-			if pcs.Status.RollingUpdateProgress != nil && pcs.Status.RollingUpdateProgress.CurrentlyUpdating != nil {
-				currentOrdinal = pcs.Status.RollingUpdateProgress.CurrentlyUpdating.ReplicaIndex
+			if pcs.Status.UpdateProgress != nil && len(pcs.Status.UpdateProgress.CurrentlyUpdating) > 0 {
+				currentOrdinal = pcs.Status.UpdateProgress.CurrentlyUpdating[0].ReplicaIndex
```
shayasoolin, line 522: `CurrentlyUpdating.ReplicaIndex` is replaced with `CurrentlyUpdating[0].ReplicaIndex`, why? It was a slice and remained one, right? Trying to understand the original code that didn't refer a specific index in the slice.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -110,13 +114,40 @@ func mutateReplicas(logger logr.Logger, currentPCSGenerationHash *string, pcsg *
...
+	totalPCLQs = pcsg.Spec.Replicas * cliqueNamesPerReplica
 	logger.Info("Mutating PodCliqueScalingGroup replicas",
 		"pcsg", client.ObjectKeyFromObject(pcsg),
-		"scheduledReplicas", scheduledReplicas, "availableReplicas", availableReplicas, "updatedReplicas", updatedReplicas)
+		"scheduledReplicas", scheduledReplicas, "availableReplicas", availableReplicas, "updatedReplicas", updatedReplicas,
+		"updatedPCLQs", updatedPCLQs, "totalPCLQs", totalPCLQs)
 	pcsg.Status.ScheduledReplicas = scheduledReplicas
 	pcsg.Status.AvailableReplicas = availableReplicas
 	pcsg.Status.UpdatedReplicas = updatedReplicas
+	if pcsg.Status.UpdateProgress != nil {
```
shayasoolin, line 135: Where is UpdateProgress initialized? Make sure it is.. otherwise this is dead code

operator/internal/controller/podcliqueset/reconcilestatus.go
```diff
@@ -73,52 +76,77 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
-	availableReplicas, updatedReplicas, err := r.computeAvailableAndUpdatedReplicas(ctx, logger, pcs)
+	stats, err := r.computeAvailableAndUpdatedReplicas(ctx, logger, pcs)
 	if err != nil {
 		return fmt.Errorf("could not compute available replicas: %w", err)
 	}
-	pcs.Status.AvailableReplicas = availableReplicas
-	pcs.Status.UpdatedReplicas = updatedReplicas
+	pcs.Status.AvailableReplicas = stats.availableReplicas
+	pcs.Status.UpdatedReplicas = stats.updatedReplicas
+	if pcs.Status.UpdateProgress != nil {
```
shayasoolin, line 85: Same here (see if you need to init UpdateProgress here inside the if or it's ok)

operator/internal/controller/podcliqueset/reconcilestatus.go
```diff
@@ -73,52 +76,77 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
 	pcsgs, err := componentutils.GetPCSGsForPCS(ctx, r.client, pcsObjectKey)
 	if err != nil {
-		return availableReplicas, updatedReplicas, err
+		return stats, err
 	}
-	// Filter the PCSGs that belong to the expected set of PCSGs for PCS, this ensures that we do not
-	// consider any stray PCSGs that might have been created externally.
-	pcsgs = lo.Filter(pcsgs, func(pcsg grovecorev1alpha1.PodCliqueScalingGroup, _ int) bool {
-		return lo.Contains(lo.Flatten(lo.Values(expectedPCSGFQNsPerPCSReplica)), pcsg.Name)
+	pcsgs = slices.DeleteFunc(pcsgs, func(pcsg grovecorev1alpha1.PodCliqueScalingGroup) bool {
```
shayasoolin, line 133: Why is `lo.Filter` replaced with `slices.DeleteFunc`?

gflarity (commented):
One nit, otherwise LGTM.

docs/api-reference/operator-api.md
oleg-kushniriov: I kept the same behaviour as before

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
oleg-kushniriov: All three controllers initialize their UpdateProgress symmetrically — none of these nil-guards are dead code:                                                                                                                                                                             
                                                     
  PCS (podcliqueset/reconcilestatus.go:89 — if pcs.Status.UpdateProgress != nil):                                                                                                                                                                                                           
  - Initialized at podcliqueset/reconcilespec.go:139 in initUpdateProgress, called from processGenerationHashChange when newGenerationHash != *pcs.Status.CurrentGenerationHash. Triggered the moment a spec change produces a different hash.                                              
                                                                                                                                                                                                                                                                                            
  PCSG (podcliquescalinggroup/reconcilestatus.go:127 — already covered):                                                                                                                                                                                                                    
  - Initialized at podcliquescalinggroup/reconcilespec.go:128 in initOrResetUpdate, called from processPendingUpdates once the parent PCS picks this PCSG's PCS-replica index for rollout.                                                                                                  
                                                                                                                                                                                                                                                                                            
  PCLQ (podclique/reconcilestatus.go:116, :148 — pclq.Status.UpdateProgress == nil):                                                                                                                                                                                                        
  - Initialized at podclique/reconcilespec.go:172 in initOrResetUpdate, called when either firstEverUpdateRequired (line 141: PCS hash differs from the PCLQ's recorded hash) or the in-progress/completed update is stale relative to the current PCS hash.                                
                                                                                                                                                                                                                                                                                            
  So the chain is PCS → PCSG → PCLQ: each level's UpdateProgress is set when its parent's hash advances and the controller decides this object should roll. The nil-guards on the status side are the canonical "is an update in flight for this object?" check.

operator/internal/controller/podcliqueset/reconcilestatus.go
oleg-kushniriov: lo.Filter always allocates a fresh slice and copies retained elements into it. slices.DeleteFunc compacts in place

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
oleg-kushniriov: good catch. fixed

operator/internal/controller/podcliqueset/reconcilestatus.go
oleg-kushniriov: Not dead code. pcsg.Status.UpdateProgress is initialized in reconcilespec.go:128 inside initOrResetUpdate

operator/e2e/tests/update/utils.go
shayasoolin: ?

oleg-kushniriov: previously it was RollingUpdateProgress ( a deprecated field) now it is UpdateProgress

shayasoolin (dismissed):

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -115,17 +123,32 @@ type PodCliqueSetUpdateStrategy struct {
...
 	// +optional
 	UpdateEndedAt *metav1.Time `json:"updateEndedAt,omitempty"`
-	// UpdatedPodCliqueScalingGroups is a list of PodCliqueScalingGroup names that have been updated to the desired PodCliqueSet generation hash.
-	UpdatedPodCliqueScalingGroups []string `json:"updatedPodCliqueScalingGroups,omitempty"`
-	// UpdatedPodCliques is a list of PodClique names that have been updated to the desired PodCliqueSet generation hash.
-	UpdatedPodCliques []string `json:"updatedPodCliques,omitempty"`
+	// UpdatedPodCliquesCount mirrors PodCliqueSetUpdateProgress.UpdatedPodCliquesCount.
+	// +optional
+	// +kubebuilder:default=0
+	UpdatedPodCliquesCount int32 `json:"updatedPodCliquesCount,omitempty"`
```
gflarity: Given that PodCliqueSetRollingUpdateProgress is deprecated,  I don't see the point of adding the counts into this struct?  If you were relying on the slices before, they're gone now. This is a breaking changed (as required by the unbounded slices).  If you have to update something as a result of the breaking change, you might as well start using PodCliqueSetUpdateProgress.

operator/api/core/v1alpha1/scalinggroup.go
```diff
@@ -100,29 +105,41 @@ type PodCliqueScalingGroupStatus struct {
...
+	// +optional
 	UpdateEndedAt *metav1.Time `json:"updateEndedAt,omitempty"`
 	// PodCliqueSetGenerationHash is the PodCliqueSet generation hash corresponding to the PodCliqueSet spec that is
 	// being rolled out. While the update is in progress PodCliqueScalingGroupStatus.CurrentPodCliqueSetGenerationHash will
 	// not match this hash. Once the update is complete the value of this field will be copied to
 	// PodCliqueScalingGroupStatus.CurrentPodCliqueSetGenerationHash.
 	PodCliqueSetGenerationHash string `json:"podCliqueSetGenerationHash"`
-	// UpdatedPodCliques is the list of PodClique names that have been updated to the latest PodCliqueSet spec.
-	UpdatedPodCliques []string `json:"updatedPodCliques,omitempty"`
+	// UpdatedPodCliquesCount mirrors PodCliqueScalingGroupUpdateProgress.UpdatedPodCliquesCount.
```
gflarity: Same thing here. Just don't see the use case for adding in the counts. If you needed the slices, they are now gone. If you have to update our code anyways, then just start using PodCliqueScalingGroupUpdateProgress since PodCliqueScalingGroupRollingUpdateProgress is deprecated anyways.

gflarity (commented):
Missed the counts being  added to deprecated status fields (via structs). Let's chat about this briefly during the standup or design sync. Maybe I'm missing something.

operator/api/core/v1alpha1/scalinggroup.go
oleg-kushniriov: RollingUpdateStatus totally removed

operator/api/core/v1alpha1/podcliqueset.go
oleg-kushniriov: RollingUpdateProgress removed

operator/api/core/v1alpha1/podcliqueset.go
oleg-kushniriov: removed

danbar2 (dismissed):

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -63,6 +65,11 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 		logger.Error(err, "failed to list PodCliques for PodCliqueScalingGroup")
 		return ctrlcommon.ReconcileWithErrors(fmt.Sprintf("failed to list PodCliques for PodCliqueScalingGroup: %q", client.ObjectKeyFromObject(pcsg)), err)
 	}
+	// Prune children that no longer belong to the spec (replica index >= Spec.Replicas during
```
shayasoolin: Are we sure about this behavior? Another way to look at it, is that the status is derived by the existing pod cliques, and will be updated as cliques are deleted or created. Otherwise the status may be "more up-to-date" than the existing resources -- pod cliques were not yet deleted but the counters already filter them out.

oleg-kushniriov: We want UpdatedPodCliquesCount ≤ TotalPodCliquesCount to always be true. TotalPodCliquesCount is computed from the new spec (Spec.Replicas * len(Spec.CliqueNames)). If we rename a
   clique (say b → c), the old b PCLQs still exist briefly while the cascade delete runs. Without pruning, those lingering b PCLQs would bump UpdatedPodCliquesCount above
  TotalPodCliquesCount until the delete finishes — which is exactly the transient overshoot we're trying to remove.
  
  You're right that the status then looks "ahead" of the live resources. I think that's actually the right meaning here: these counters report how far we are from spec, not a live count
  of objects in the cache
  
  I updated the comment a little.

shayasoolin (approved):

danbar2 (approved):

PR #583, opened 2026-05-06
test(mnnvl): rewrite E2E tests for single-annotation model

shmuel-runai:
/kind feature

## Summary

- Rewrites the MNNVL E2E test suite for the single-annotation (`grove.io/mnnvl-group`) model introduced in #574
- Replaces Go builder functions with pre-written YAML fixtures under `operator/e2e/yaml/auto-mnnvl/`, aligning with the repository's existing E2E patterns
- Adds `applyMNNVLYAML` helper to dynamically patch PCS resource names for test isolation
- Adds comprehensive `testMNNVLEndToEnd` covering all annotation propagation layers (PCS → PCSG → PCLQ) with inherit/override/opt-out combinations, plus scale-out, scale-in, and deletion
- Removes redundant tests consolidated into the end-to-end test
- Updates `supported_but_disabled_test.go` and `shared_cases.go` to use `AnnotationMNNVLGroup`

## Test plan

- [x] `make check` passes (lint, format, codegen)
- [x] `go vet -tags=e2e ./e2e/tests/auto-mnnvl/...` compiles cleanly
- [x] All 4 E2E configurations pass locally:
  - Config1_UnsupportedAndDisabled: PASS
  - Config2_UnsupportedButEnabled: PASS
  - Config3_SupportedAndEnabled: PASS (12 test cases, all green)
  - Config4_SupportedButDisabled: PASS

Ref #417


operator/e2e/tests/auto-mnnvl/shared_cases.go
```diff
@@ -33,17 +32,15 @@ import (
...
 )
 
-// testNoMNNVLArtifactsWhenDisabled verifies that no ComputeDomains, auto-mnnvl
-// annotations, or resourceClaims are produced for a GPU-capable PCS. It is shared
-// across test suites because the expected behavior is identical regardless of why
-// the feature is inactive.
+// testNoMNNVLArtifactsWhenDisabled verifies that no ComputeDomains or
+// resourceClaims are produced for a GPU-capable PCS when the feature is
+// disabled. It is shared across test suites because the expected behavior
+// is identical regardless of why the feature is inactive.
```
shayasoolin: It's not clear what you mean by "feature is disabled" vs "feature is inactive". I understand that you use the same yaml (which has NO mnnvl annotations) when the feature is both enabled and disabled in the operator level. Please clarify the comment here and in the yaml itself.

operator/e2e/yaml/auto-mnnvl/mnnvl-comprehensive-bare.yaml
```diff
@@ -0,0 +1,85 @@
+# MNNVL E2E: Comprehensive PCS with mixed GPU/CPU cliques and PCSGs, NO annotations
```
shayasoolin: Since there are no annotations, I don't think you should have that complex stucture. I'd be confident with a much simpler yaml, of a PCS with two PCLQs - one with GPUs the other CPU only. I suggest to simplify.

operator/e2e/yaml/auto-mnnvl/mnnvl-comprehensive-mix.yaml
```diff
@@ -0,0 +1,209 @@
+# MNNVL E2E: Comprehensive mix — all inherit/override/none combinations
```
shayasoolin, line 1: Nice!

operator/e2e/yaml/auto-mnnvl/mnnvl-comprehensive.yaml
```diff
@@ -0,0 +1,87 @@
+# MNNVL E2E: Comprehensive PCS with mixed GPU/CPU cliques and PCSGs, mnnvl-group: "default"
```
shayasoolin: Can you reuse the comprehensive-mix yaml for this test as well?
In general I feel there are too many tests and yamls, while the e2e one is very broad. 
I appreciate testing step-by-step before reaching the e2e one but here it feels like too much. It may also have implication on the total e2e run time. Not sure how significant it is now, but the more we add it will accumulate.
I suggest to try and combine some tests together, as well as see for each and every test if it adds something unique or perhaps already covered by another test case.

operator/e2e/tests/auto-mnnvl/shared_cases.go
shmuel-runai: Clarified the comment — it now explicitly distinguishes operator-disabled (`autoMNNVLEnabled=false`) from CRD-unsupported (ComputeDomain CRD not installed), and explains why the test is shared: both scenarios produce the same outcome (no CDs, no claims).

operator/e2e/yaml/auto-mnnvl/mnnvl-comprehensive-bare.yaml
shmuel-runai: Good call — simplified to just 2 PCLQs (one GPU, one CPU). The complex structure was overkill for a test that only checks "no annotation → no MNNVL artifacts".

operator/e2e/yaml/auto-mnnvl/mnnvl-comprehensive.yaml
shmuel-runai: Agreed — removed 4 redundant tests (`testMNNVLComputeDomainAndClaimInjection`, `testMultipleGroupsInOnePCS`, `testPCLQOptOutOverridesParent`, `testThreeLayerPCSPCSGPCLQAllDifferent`) and their 4 YAML files. All their coverage is already handled by `testMNNVLEndToEnd` using the comprehensive-mix YAML. Also moved the finalizer-blocks-deletion check into `testMNNVLEndToEnd`. Down from 12 to 8 tests in Config3.

shayasoolin (approved):

danbar2 (approved):

PR #587, opened 2026-05-07
docs: update MNNVL user guide for single-annotation model

shmuel-runai:
/kind documentation
Ref #417

## Summary

- Rewrites `docs/user-guide/auto-mnnvl.md` to reflect the new single-annotation model (`grove.io/mnnvl-group`) introduced in #574, replacing the previous two-annotation system (`grove.io/auto-mnnvl` + `grove.io/mnnvl-group`).
- Adds usage examples covering simple opt-in, multiple groups, PCS-level opt-out, and a full three-layer hierarchy (PCS → PCSG → PCLQ) with overrides.
- Documents annotation propagation, ComputeDomain naming, group name validation, backward compatibility (breaking change from old annotation), and updated limitations.

## Test plan

- [ ] Review rendered markdown for correctness and readability
- [ ] Verify YAML examples match current API and annotation semantics
- [ ] Cross-check with GREP-417 and merged implementation (#574)


danbar2 (approved):

docs/user-guide/auto-mnnvl.md
```diff
@@ -1,32 +1,34 @@
 # Auto MNNVL (Multi-Node NVLink)
 
-Grove can automatically enable Multi-Node NVLink (MNNVL) acceleration without requiring any manual configuration. This guide explains how to enable, use, and manage the auto MNNVL feature.
+Grove can automatically manage Multi-Node NVLink (MNNVL) acceleration for your GPU workloads. This guide explains how to enable, configure, and manage the auto MNNVL feature using the `grove.io/mnnvl-group` annotation.
```
nvrohanv:
```suggestion
Grove can automatically manage Multi-Node NVLink (MNNVL) setup for your GPU workloads. This guide explains how to enable, configure, and manage the auto MNNVL feature using the `grove.io/mnnvl-group` annotation. It is Grove's intention to keep the core API vendor agnostic which is why MNNVL support is handled via an annotation.
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -1,32 +1,34 @@
...
+| A group name (e.g., `"my-group"`, `"workers"`) | Opt in to MNNVL. PodCliques with the same group name share a ComputeDomain per replica. |
+| `"none"` | Explicit opt-out. Overrides a parent layer's group assignment. |
+| Absent | Inherit from the parent layer. If no parent sets it, no MNNVL. |
 
 ## Prerequisites and Constraints
 
 Before enabling auto MNNVL, ensure your cluster meets the following requirements:
 
-1. **NVIDIA GPUs with MNNVL support** on all nodes that will run MNNVL workloads
+1. **NVIDIA GPUs with MNNVL support** on the nodes that will run MNNVL workloads.
```
nvrohanv:
```suggestion
1. **NVIDIA GPUs with MNNVL support** on the nodes that will run MNNVL workloads e.g. GB200 and GB300 NVL72 systems.
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
                     nvidia.com/gpu: "8"
 ```
 
-When opting out, the operator will **not** create ComputeDomains or any other MNNVL-related artifacts for that PCS.
+This creates ComputeDomains `my-workload-0-my-group` and `my-workload-1-my-group`, for all PodCliques.
```
nvrohanv:
```suggestion
In this example, spec.replicas: 2 produces two PCS replicas, my-workload-0 and my-workload-1. The operator creates one ComputeDomain per PCS replica by appending the MNNVL group name, resulting in my-workload-0-my-group and my-workload-1-my-group. Because no per-PodClique MNNVL group overrides are specified, all GPU PodCliques within each PCS replica use the PCS-level group and share that replica’s ComputeDomain.
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+
+### Multiple MNNVL Groups
+
+Assign different PodCliques to separate groups. Each group gets its own ComputeDomain per replica:
```
nvrohanv:
```suggestion
Assign different PodCliques to separate groups. Each group gets its own ComputeDomain per PodCliqueSet replica:
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+
+```yaml
+apiVersion: grove.io/v1alpha1
+kind: PodCliqueSet
+metadata:
+  name: training-job
```
nvrohanv: Since we dont really have training job support yet lets just call this inference-deployment, i guess you could imagine a situation where workers are the main model replicas, encoders are the vision encoders, and then lets change param-servers to frontend, will add some suggestions to make these changes.

```suggestion
  name: inference-deployment
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+        spec:
+          replicas: 1
+          podSpec:
+            containers:
+              - name: encoder
+                image: my-encoder:latest
+                resources:
+                  limits:
+                    nvidia.com/gpu: "4"
+      - name: param-servers
```
nvrohanv:
```suggestion
      - name: frontends
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+        spec:
+          replicas: 1
+          podSpec:
+            containers:
+              - name: ps
```
nvrohanv:
```suggestion
              - name: frontend
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+                image: my-ps:latest
```
nvrohanv:
```suggestion
                image: my-frontend:latest
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+                resources:
+                  limits:
+                    cpu: "4"
+```
+
+Result per replica:
+- `workers` → group `"workers"` → ComputeDomain `training-job-0-workers` / `training-job-1-workers`
+- `encoders` → group `"encoders"` → ComputeDomain `training-job-0-encoders` / `training-job-1-encoders`
+- `param-servers` → no annotation → no MNNVL
```
nvrohanv:
```suggestion
- `frontends` → non GPU PodClique → no MNNVL
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+          podSpec:
+            containers:
+              - name: ps
+                image: my-ps:latest
+                resources:
+                  limits:
+                    cpu: "4"
+```
+
+Result per replica:
```
nvrohanv:
```suggestion
Result per PCS replica:
```

docs/user-guide/auto-mnnvl.md
```diff
@@ -47,38 +49,51 @@ helm upgrade -i grove oci://ghcr.io/ai-dynamo/grove/grove-charts --version <vers
...
+
+| PCS annotation | PCLQ annotation | Effective PCLQ group |
+|---|---|---|
+| `"my-group"` | Absent | `"my-group"` (inherited) |
+| `"my-group"` | `"other-group"` | `"other-group"` (overridden) |
+| `"my-group"` | `"none"` | No MNNVL (opted out) |
+| Absent | `"my-group"` | `"my-group"` (PCLQ-level only) |
 
-For a PCS named `my-inference` with `replicas: 2`, Grove creates:
+### Non-GPU PodCliques
```
nvrohanv: When we say “non-GPU PodClique,” do we specifically mean any PodClique that does not request `nvidia.com/gpu`? I want to confirm that PodCliques requesting non-NVIDIA GPUs would also be ignored by MNNVL rather than treated as eligible GPU PodCliques. If this is correct we should specify that its non-nvidia GPU PodClique.

It may also be worth adding guidance for mixed GPU clusters. For example, in a cluster with both H100s and NVL72-capable GPUs, PodCliques that should not participate in MNNVL should either avoid inheriting an MNNVL group or explicitly opt out with `grove.io/mnnvl-group: "none"`. My understanding is that if the DRA driver is installed across all NVIDIA GPUs, including the non-GB200/GB300 cases in a ComputeDomain is technically safe because the driver is a no-op there, but for readability and intent it would be clearer to explicitly exclude those PodCliques from the MNNVL group so I think would be good to explain this.

docs/user-guide/auto-mnnvl.md
```diff
@@ -88,42 +103,232 @@ spec:
...
+Resolution for each PodClique:
+
+| PodClique | PCLQ annotation | PCSG annotation | PCS annotation | Effective group |
+|---|---|---|---|---|
+| `gpu-a` | Absent | `"none"` | `"my-group"` | No MNNVL — PCSG opts out |
+| `gpu-b` | `"my-group"` | `"none"` | `"my-group"` | `"my-group"` — PCLQ overrides PCSG back into the PCS group |
+| `gpu-c` | Absent | Absent | `"my-group"` | `"my-group"` — inherited through PCSG from PCS |
+| `gpu-d` | `"none"` | Absent | `"my-group"` | No MNNVL — PCLQ opts out |
+
+### Group Name Requirements
```
nvrohanv: I'm thinking Group Name Requirements and ComputeDomain Naming should go above the usage examples? The remaining stuff below seems more tied to the example above so i think fine to keep it as is, what do you think?

nvrohanv (commented):
Overall looks good, left suggestions for phrasing fixes for clarity, also I think we should provide a bit more guidance about how it works with mixed nvidia gpu clusters. Once those are addressed, will approve. Great work!

docs/user-guide/auto-mnnvl.md
shmuel-runai: fix

docs/user-guide/auto-mnnvl.md
shmuel-runai: fix

docs/user-guide/auto-mnnvl.md
shmuel-runai: fix

docs/user-guide/auto-mnnvl.md
shmuel-runai: Done

docs/user-guide/auto-mnnvl.md
shmuel-runai: Done — renamed to `inference-deployment` with `frontends`

docs/user-guide/auto-mnnvl.md
shmuel-runai: Good catch — updated to 'non-NVIDIA-GPU' and added mixed GPU cluster guidance

docs/user-guide/auto-mnnvl.md
shmuel-runai: Agreed — moved both sections above Usage Examples

danbar2 (approved):

nvrohanv (approved):

PR #560, opened 2026-04-29
feat: Add GREP-0376 for Volcano scheduler backend

xianlubird:
## Summary

- Proposes Volcano scheduler backend support for Grove through native Volcano `PodGroup`
- Uses Volcano 1.14+ `subGroupPolicy` to map each Grove PodGang role into a Volcano subgroup
- Documents the end-to-end scheduling flow from `PodCliqueSet` to Grove `PodGang` to Volcano `PodGroup`
- Adds the Volcano 1.14+ version requirement and backend initialization capability check for `PodGroup.spec.subGroupPolicy`

## Key design decisions

- Keep Grove `PodGang` as the portable scheduler API and translate it to Volcano-native resources in the backend
- Map each Grove `PodGroup.minReplicas` to the corresponding Volcano `subGroupPolicy[].subGroupSize`
- Preserve the overall workload gang requirement with Volcano `PodGroup.spec.minMember`
- Require Volcano 1.14+ because `subGroupPolicy` is needed now, and future topology-aware scheduling will also depend on Volcano 1.14+ capabilities
- Follow the direction already taken by KubeRay and Volcano Kthena for subgroup-based gang scheduling

Fixes #571

PR #590, opened 2026-05-08
Add GREP-377 for Volcano topology-aware scheduling

### LINKED ISSUE #571: Add volcano scheduler support as a backend in Grove

[Volcano](https://github.com/volcano-sh/volcano) is a popular scheduler. As many customer already use Volcano, we should support this as on backend

### Why is this needed?

[Volcano](https://volcano.sh/) is a Kubernetes-native batch scheduling system, extending and enhancing the capabilities of the standard kube-scheduler. It provides a comprehensive set of features specifically designed to manage and optimize various batch and elastic workloads, including Artificial Intelligence (AI) / machine learning (ML) / deep learning (DL), bioinformatics / genomics, and other "Big Data" applications.

As Grove is designed especially for AI/Inference/Train, Volcano and Grove should work great.

### LINKED ISSUE #637: Track GREP-377 Volcano topology-aware scheduling backend
This issue tracks GREP-377 for adding Volcano topology-aware scheduling backend support to Grove.

The proposal covers:

- Implementing TopologyAwareSchedBackend for the Volcano backend
- Translating Grove ClusterTopology into Volcano HyperNode resources
- Translating Grove PodCliqueSet and PodGang topology constraints into Volcano PodGroup networkTopology and subGroupPolicy networkTopology
- Supporting externally managed HyperNodes and topology drift detection

Related PR: #590

xianlubird:
## Summary

- Proposes Volcano topology-aware scheduling support for Grove through Volcano HyperNode and PodGroup networkTopology
- Documents how Grove ClusterTopology is materialized as a Volcano HyperNode tree
- Defines PodCliqueSet to PodGang to Volcano PodGroup translation with direct subGroupPolicy usage
- Adds future test expectations for HyperNode sync, drift detection, and topology translation

## Key design decisions

- Implement TopologyAwareSchedBackend in the Volcano backend
- Use Volcano HyperNode as the scheduler backend topology resource
- Translate Grove topology constraints to Volcano highestTierAllowed
- Generate a value-level HyperNode tree from Node topology labels
- Keep Grove PodCliqueSet as the user-facing workload API and translate through Grove PodGang to Volcano PodGroup

Fixes #637


docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+  - [PodCliqueSet to PodGang to Volcano PodGroup Translation](#podcliqueset-to-podgang-to-volcano-podgroup-translation)
+  - [Operator Configuration and Dependencies](#operator-configuration-and-dependencies)
+  - [Monitoring](#monitoring)
+  - [Test Plan](#test-plan)
+- [Appendix](#appendix)
+<!-- /toc -->
+
+## Summary
+
+Grove already defines a portable topology-aware scheduling API through `ClusterTopology`, `TopologyConstraint`, and `PodGang`. This GREP proposes implementing Volcano support for that API by extending the Volcano scheduler backend to implement `TopologyAwareSchedBackend`. The backend will translate Grove `ClusterTopology` resources into Volcano 1.14+ `HyperNode` resources and translate Grove topology constraints into Volcano `PodGroup.spec.networkTopology` and `PodGroup.spec.subGroupPolicy[*].networkTopology` so Grove workloads can use Volcano for topology-aware gang scheduling.
```
unmarshall:
```suggestion
Grove already defines a portable topology-aware scheduling API through `ClusterTopologyBinding`, `TopologyConstraint`, and `PodGang`. This GREP proposes implementing Volcano support for that API by extending the Volcano scheduler backend to implement `TopologyAwareSchedBackend`. The backend will translate Grove `ClusterTopologyBinding` resources into Volcano 1.14+ `HyperNode` resources and translate Grove topology constraints into Volcano `PodGroup.spec.networkTopology` and `PodGroup.spec.subGroupPolicy[*].networkTopology` so Grove workloads can use Volcano for topology-aware gang scheduling.
```

unmarshall: GREP-244 is a bit outdated. Recently we changed the name of the CR to `ClusterTopologyBinding` resource. Please see https://github.com/ai-dynamo/grove/issues/615 for more information.

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+
+## Motivation
+
+Grove workloads should be portable across scheduler backends. GREP-244 defines the Grove topology model, and GREP-375 defines the scheduler backend framework that lets a scheduler implement topology management without changing the user-facing workload API. PR #560 / GREP-376 introduces the base Volcano backend for Grove gang scheduling with Volcano `PodGroup` and `subGroupPolicy`. The next step is to make the Volcano backend understand the same topology-aware scheduling intent that KAI already supports.
```
unmarshall: Make `GREP-*` into links, same with PR #560

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+
+## Proposal
+
+The Volcano backend will become a topology-aware scheduler backend. When topology-aware scheduling is enabled and Volcano is configured as an active scheduler profile, the ClusterTopology controller will invoke the Volcano backend through `TopologyAwareSchedBackend` in the same way it invokes the KAI backend.
+
+For auto-managed topologies, Grove will create and maintain Volcano `HyperNode` resources from each `ClusterTopology`. The generated HyperNodes will form a tree: leaf HyperNodes select real Nodes, and non-leaf HyperNodes reference child HyperNodes. Workload constraints will target those tiers through `networkTopology.highestTierAllowed`, using the numeric tier that corresponds to the requested Grove pack domain.
+
+For externally managed topologies, the administrator can add a Volcano entry to `ClusterTopology.spec.schedulerTopologyReferences`. Grove will not create or update the referenced HyperNode, but it will check whether it is compatible with the Grove `ClusterTopology` and report drift through the standard `SchedulerTopologyDrift` condition and `schedulerTopologyStatuses`.
```
unmarshall:
```suggestion
For externally managed topologies, the administrator can add a Volcano entry to `ClusterTopology.spec.schedulerTopologyReferences`. Grove will not create or update the referenced HyperNodes, but it will check whether it is compatible with the Grove `ClusterTopology` and report drift through the standard `SchedulerTopologyDrift` condition and `schedulerTopologyStatuses`.
```

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+
+In this section, `networkTopology` refers to Volcano's workload-side API field on Volcano resources. Volcano `networkTopology` is the constraint that tells Volcano which HyperNode tier a job or subgroup may cross. Therefore this GREP uses both Volcano primitives:
+
+* `HyperNode` as the backend topology resource managed by `TopologyAwareSchedBackend`.
+* `PodGroup.spec.networkTopology` and `PodGroup.spec.subGroupPolicy[*].networkTopology` as the translated workload constraints created by `SyncPodGang`.
+
+### Limitations/Risks & Mitigations
```
unmarshall, line 73: Both `pack.required` and `pack.preferred` topology constraints can be set at multiple levels (PCS, PCLQ and PCSG). Can this be realised via `mode: hard | soft`  at `PodGroup.spec.networkTopology` and `PodGroup.spec.subGroupPolicy[*].networkTopology`? In the API there is only provision to have just one `mode` today.

If this is not possible then should this be a limitation that must be mentioned?

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+
+Volcano expresses the scheduling boundary with `highestTierAllowed`. In Volcano's HyperNode model, lower tier numbers represent narrower, higher-locality domains, while higher tier numbers represent broader domains.
+
+**Mitigation:** The Volcano backend will derive tier numbers deterministically from the `ClusterTopology` order. Grove `ClusterTopology.spec.levels` are ordered broadest to narrowest, so for `N` Grove levels, level index `i` maps to Volcano tier `N - i`. For example, `[block, rack, host]` maps to `block=3`, `rack=2`, and `host=1`. Workload constraints will use `highestTierAllowed`.
+
+## Design Details
+
+### Control Flow
+
+```mermaid
```
unmarshall, line 103: In this diagram - it is not clear what is the difference in the control flow for externally managed volcano topology (via HyperNode tree) and auto-created HyperNode tree. AFAIK drift detection is for externally managed network topology. For auto-managed, a reconcile should sync it, right?

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+```
+
+The control flow has two independent reconciliation paths. The ClusterTopology path creates or validates Volcano `HyperNode` topology resources. The workload path translates Grove `PodGang` scheduling intent into a Volcano `PodGroup`. Volcano only has enough information to enforce topology-aware scheduling when both paths are healthy: the HyperNodes describe the topology tiers, and the PodGroup `networkTopology` fields describe the workload's requested boundary.
+
+### Volcano Scheduler Backend Topology
+
+The Volcano backend will implement the existing optional topology interface:
+
+```go
+type TopologyAwareSchedBackend interface {
```
unmarshall: The name of the interface has changed to `TopologyAwareBackend`. Please reflect this change here.
I think it is best to just provide a link to the interface instead so that any changes are automatically reflected in the doc.

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+	) (bool, string, int64, error)
+}
+```
+
+For Volcano:
+
+* `TopologyGVR()` returns `topology.volcano.sh/v1alpha1, Resource=hypernodes`.
+* `TopologyResourceName(ct)` returns the synthetic root HyperNode name for the `ClusterTopology`. The default name is the `ClusterTopology` name.
+* `SyncTopology()` lists Nodes, derives the desired HyperNode tree from the `ClusterTopology` levels and Node label values, and creates or updates all auto-managed HyperNodes for the `ClusterTopology`.
+* `OnTopologyDelete()` is a no-op because every auto-managed HyperNode will have the `ClusterTopology` as its controller owner, allowing Kubernetes garbage collection to cascade-delete them.
```
unmarshall: I am just wondering that for all auto-managed HyperNode resources cascade-delete would do its job. This is true for all auto-managed resources irrespective of the backend scheduler. For unmanaged scheduler backend custom resources deletion should never be done by Grove as it does not manage them.

@enoodle This warrants the question on why do we even need this method in the TopologyAwareBackend interface?

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+* `CheckTopologyDrift()` compares the externally managed HyperNode tree rooted at `ref.topologyReference` with the desired topology derived from `ClusterTopology` and Node labels.
+
+The backend must add Volcano topology API types to the operator scheme and ensure the `hypernodes.topology.volcano.sh` CRD is present when Volcano TAS is enabled.
```
unmarshall: When you say the `backend must ensure that the `hypernodes.topology.volcano.sh` CRD is present when Volcano TAS is enabled.` - what does this mean precisely? Do you expect Grove to install the scheduler specific CRDs or did you mean something else?

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+| Node   | `topology.grove.io/block` | `topology.grove.io/rack` | `kubernetes.io/hostname` |
+| ------ | ------------------------- | ------------------------ | ------------------------ |
+| node-1 | block-a                   | rack-1                   | node-1                   |
+| node-2 | block-a                   | rack-1                   | node-2                   |
+| node-3 | block-a                   | rack-2                   | node-3                   |
+| node-4 | block-b                   | rack-3                   | node-4                   |
+
+> NOTE: Each generated HyperNode will include Grove management labels and a `ClusterTopology` controller owner reference. The example below omits repeated metadata fields for readability.
+
+```yaml
```
unmarshall, line 188: A simple example where you have 3 levels {block, rack and host} with minimal number of nodes, the number of HyperNode resources that are created are: 14. This model of representing network topology is unfortunately not scalable by any means. We are now seeing massive sized clusters and there HyperNode resources will itself create extremely large number of resources.

This would mean very expensive LIST calls and large number of CREATE calls from Grove operator.

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+    host3 --> node3
+    host4 --> node4
+
+    classDef hyperNode fill:#d8ebff,stroke:#333,stroke-width:1px,color:#111;
+    classDef node fill:#eef6ff,stroke:#333,stroke-width:1px,color:#111;
+    class root,blockA,blockB,rack1,rack2,rack3,host1,host2,host3,host4 hyperNode;
+    class node1,node2,node3,node4 node;
+```
+
+Generated names must be deterministic and valid Kubernetes object names. Value-level HyperNode names include the topology path so equal rack names under different blocks do not collide. If raw label values do not fit object-name constraints, the implementation will sanitize them and add a stable hash suffix. Generated HyperNodes must set the `ClusterTopology` as their controller owner so Kubernetes garbage collection deletes them with the owning topology. They should also carry Grove labels such as the owning `ClusterTopology` name and topology domain so they can be listed, updated, and repaired as a set.
```
unmarshall: >If raw label values do not fit object-name constraints, the implementation will sanitize them and add a stable hash suffix.

This is a direct consequence of generating too many resources to represent network switch topology.

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+
+For Volcano 1.14+, only leaf HyperNodes point directly to Nodes. Non-leaf HyperNodes point to child HyperNodes using `exactMatch`. `regexMatch` must not be used for members whose type is `HyperNode`.
+
+If a `ClusterTopology` includes a `host` domain backed by `kubernetes.io/hostname`, each host value is unique, so a faithful Volcano topology needs one leaf HyperNode per Node. This can create many HyperNodes in large clusters. Administrators who do not need Grove `pack.required: host` semantics for Volcano can reduce the object count by omitting the `host` level from the Volcano-targeted `ClusterTopology`; in that case the narrowest generated HyperNodes, such as rack HyperNodes, can select real Nodes directly with `labelMatch` or `regexMatch`. Grove must not collapse all hosts into a single host HyperNode, because that would make `pack.required: host` behave like a broader domain and weaken the user's requested constraint.
```
unmarshall: >Administrators who do not need Grove `pack.required: host` semantics for Volcano can reduce the object count by omitting the `host` level from the Volcano-targeted `ClusterTopology`

Is this also followed in the community where Volcano is used? Does Volcano support `numa` level, because that will create even larger number of HyperNodes than `host`, right?

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+  schedulerTopologyReferences:
+    - schedulerName: volcano
+      topologyReference: grove-volcano-root
+```
+
+When this reference is present, Grove will not create or mutate the referenced HyperNode. The Volcano backend will verify that the referenced HyperNode exists and that its tier and Node selector match the expected domain from `ClusterTopology.spec.levels`. It will report mismatches through `schedulerTopologyStatuses[*].inSync=false` and the aggregate `SchedulerTopologyDrift` condition.
+
+### PodCliqueSet to PodGang to Volcano PodGroup Translation
+
+Users do not create Volcano `PodGroup` resources directly, and they usually do not create Grove `PodGang` resources directly either. The user-facing workload API is `PodCliqueSet`. The translation path for Volcano topology-aware scheduling is:
```
unmarshall: >they usually do not create Grove `PodGang` resources directly either

When you use the word `usually` then it means that sometimes they do. But that is not true. The users never create `PodGang` resources directly and that is the clear statement.

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+
+If a topology constraint only has `pack.preferred`, the `PodCliqueSet` controller resolves it into `PodGang.spec.topologyConstraint.packConstraint.preferred`, and the Volcano backend translates it with `mode: soft`:
+
+```yaml
+networkTopology:
+  mode: soft
+  highestTierAllowed: 2
+```
+
+If both `pack.required` and `pack.preferred` are present on the same Grove topology constraint, the Volcano backend must preserve the hard scheduling boundary first. Volcano `networkTopology` has one mode and one tier boundary, so the backend will translate the required constraint with `mode: hard` and will not encode the preferred hint in the generated `PodGroup`. The implementation should emit a Kubernetes event or status message explaining that the preferred hint was not translated for Volcano.
```
unmarshall, line 533: What is the impact of not honoring soft pack constraint?
Can you also check https://github.com/ai-dynamo/grove/issues/368?

unmarshall (changes requested):

docs/proposals/377-volcano-topology-aware/README.md
```diff
@@ -0,0 +1,598 @@
...
+      minReplicas: 8
+      podReferences:
+        - namespace: default
+          name: decode-0
+        - namespace: default
+          name: decode-1
+        # ... decode-2 through decode-7 omitted for brevity
+```
+
+The Volcano backend translates this into a single Volcano `PodGroup`. The gang-level topology constraint becomes `PodGroup.spec.networkTopology`, and the subset constraint becomes an entry in `PodGroup.spec.subGroupPolicy`:
```
unmarshall, line 492: can you check if Volcano also has the same issue that is described in https://github.com/ai-dynamo/grove/issues/648?

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Done

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Done

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Done

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Thanks, good point. Volcano `networkTopology` only has one `mode` and one tier boundary per scope, so it cannot encode both `pack.required` and `pack.preferred` in the same `networkTopology`.

I updated the proposal to document this limitation and the translation rule:

```text
The translation is scope-local:
- whole-gang constraints map to PodGroup.spec.networkTopology
- subgroup constraints map to PodGroup.spec.subGroupPolicy[*].networkTopology

If a scope only has pack.preferred:
- translate it as mode: soft

If a scope has pack.required:
- preserve the hard boundary with mode: hard
- do not encode the preferred hint for that same scope
- report this through an event or status message

File: docs/proposals/377-volcano-topology-aware/README.md (hunk 5 above)
Comment (xianlubird) [reply docs/proposals/377-volcano-topology-aware/README.md:103]: Thanks, right. I updated the control-flow diagram to make the two topology paths explicit.

```text
Auto-managed topology:
- SyncTopology lists Nodes and Grove-owned HyperNodes
- derives the desired HyperNode tree
- creates, updates, or deletes owned HyperNodes during reconcile

Externally managed topology:
- CheckTopologyDrift reads the referenced HyperNode tree
- compares it with the expected Grove topology
- reports in-sync or SchedulerTopologyDrift
- does not mutate the referenced HyperNodes
```

I also updated the text below the diagram to clarify that drift detection is only for externally managed topology, while auto-managed topology is repaired by reconciliation.

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Thanks, updated.

I re-checked the current local code and aligned the proposal with it:

```text
TopologyAwareSchedBackend -> TopologyAwareBackend
schedulerTopologyReferences -> schedulerTopologyBindings
```

I also removed the copied Go interface definition from the proposal and replaced it with a link to the source interface, so future interface changes are reflected from the code instead of duplicated in the document.

docs/proposals/377-volcano-topology-aware/README.md
enoodle: Lets remove it - seems like it is no op currently for all scheduling backends. I think that we can bring it back in the future if we need any cleanup logic.

xianlubird: Yes, I agree. For auto-managed backend resources, Kubernetes GC should handle deletion as long as Grove sets the owning `ClusterTopologyBinding` as the controller owner.

For example, for Volcano auto-managed HyperNodes, Grove will create each generated HyperNode with an `ownerReference` pointing to the `ClusterTopologyBinding`. When the `ClusterTopologyBinding` is deleted, Kubernetes garbage collection will cascade-delete those HyperNodes. Grove does not need to explicitly delete them.

For externally managed scheduler resources, Grove should not delete them at all. It should only check compatibility / drift because those resources are owned outside Grove.

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Good point. I did not mean that Grove should install Volcano scheduler-specific CRDs.

The assumption here is that Volcano is already installed correctly, including the required Volcano CRDs such as `hypernodes.topology.volcano.sh`. If Volcano is installed incorrectly or the CRD is missing, Grove may fail when reconciling auto-managed HyperNodes, but that should be treated as an installation/configuration issue outside this GREP.

To avoid confusion, I will remove this sentence and keep the proposal based on the assumption that the Volcano 1.14+ installation already provides the required CRDs.

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: This is also my biggest concern. 
For auto-managed resources, Volcano's `HyperNode` management is extremely complex. 
It has a node label-based selection method, but if node labels change, new nodes are added, or old nodes are deleted, the Grove controller needs to track these events and handle them accordingly. 
I think this exceeds the scope of Grove's management capabilities. 
I suggest that Grove's topology-aware scheduling for Volcano should only support externally managed topologies, allowing external users to create the topologies, and Grove simply using them. 
I'd like to hear your opinions; this is a very important decision. @unmarshall @sanjaychatterjee @enoodle

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Just like mentioned above

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Yes, the community also advises against enabling host-level scheduling unless absolutely necessary; numa-level scheduling is not currently supported.

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: Already change to never

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: OK, I checked #368.

My understanding is that `pack.preferred` is a best-effort placement optimization. If Volcano cannot encode a soft pack constraint for a specific scope, the impact is not a correctness issue: the workload can still be scheduled, and any `pack.required` constraint is still honored. The loss is that Volcano may not try the preferred/local placement first, so placement may be less optimal for latency or performance.

This limitation only applies when the same Grove topology constraint scope has both `pack.required` and `pack.preferred`, because one Volcano `networkTopology` object has only one `mode`.

Different scopes can still be represented separately. For example:
- PCS-level preferred constraint -> `PodGroup.spec.networkTopology` with `mode: soft`
- subgroup-level required constraint -> `PodGroup.spec.subGroupPolicy[*].networkTopology` with `mode: hard`

So this still supports the kind of mixed-scope use case discussed in #368, such as preferring whole-workload locality while requiring subgroup locality. The unsupported case is only required + preferred on the same exact scope, where we preserve the hard constraint and report that the preferred hint was not translated for Volcano.

docs/proposals/377-volcano-topology-aware/README.md
xianlubird: OK, I checked #648.

My understanding is that Volcano has the same class of limitation if one PCS replica is split into multiple Grove `PodGang`s / multiple Volcano `PodGroup`s.

Volcano `networkTopology` is enforced at the Volcano `PodGroup` scope. So:
- if the whole PCS replica is represented by one Volcano `PodGroup`, then the PCS-level topology constraint can be enforced across that whole `PodGroup`;
- subgroup constraints can still be represented inside the same `PodGroup` through `subGroupPolicy[*].networkTopology`;
- but if Grove splits one PCS replica into multiple `PodGang`s, and therefore multiple Volcano `PodGroup`s, Volcano does not have cross-`PodGroup` coordination to guarantee that all those `PodGroup`s choose the same topology domain.

So the issue from #648 is not KAI-specific in principle. It is a scheduler contract limitation whenever the backend only sees independent gang objects. For this GREP, the safe statement is that Volcano can enforce the topology constraint within each generated Volcano `PodGroup`, but it cannot guarantee a PCS-level topology invariant across multiple Volcano `PodGroup`s for the same PCS replica.

xianlubird:
@unmarshall @sanjaychatterjee 
I checked Volcano's HyperNode auto discovery mechanism. Volcano can discover the topology and create / maintain the corresponding HyperNode resources by itself, so Grove may not need to own HyperNode generation in that mode.

Reference:
- Volcano docs: https://volcano.sh/en/docs/user-guide/how_to_use_hypernode_auto_discovery/
- Local implementation: https://github.com/volcano-sh/volcano/blob/master/pkg/controllers/hypernode/discovery/label/label.go

However, there is one important limitation: the label-based auto discovery does not generate HyperNodes for `kubernetes.io/hostname`. It treats Kubernetes Nodes as leaf members under the lowest HyperNode tier. This means the generated topology can represent domains like rack / TOR  / hypercluster, but it does not provide a real host-level HyperNode tier.

Because of that, this mechanism cannot express node-level topology semantics through `networkTopology`. For example, it cannot make a group of pods spread across different Kubernetes Nodes, and it also cannot enforce host-level topology through the final `kubernetes.io/hostname` level.

Given this, maybe the simpler contract is that Grove does not generate Volcano HyperNodes. Instead, users or Volcano auto discovery prepare the HyperNode tree, and Grove only references it when translating workload topology constraints.

xianlubird:
image

docs/proposals/377-volcano-topology-aware/README.md
unmarshall: Ok, i think its a design mistake of choosing a single resource to represent two concerns - `gang-scheduling` and `topology-packing`.  Gang scheduling is always going to be specific to a Gang whereas topology packing can span across pod gangs. Since a PodGroup maps 1:1 to PodGang therefore it becomes a misfit to represent topology pack constraints.

This i see as a common mistake across KAI and Volcano. Can you raise an issue on Volcano github repo, giving a link to the issues created on Grove?

xianlubird: Thanks, agreed. I created an upstream Volcano issue for this:

https://github.com/volcano-sh/volcano/issues/5430

I linked Grove #648 and this PR discussion there. The issue frames the gap as: `PodGroup` is the gang scheduling unit, but topology packing may need to span multiple `PodGroup`s that belong to the same higher-level workload replica.

xianlubird: https://github.com/volcano-sh/volcano/issues/5430#issuecomment-4688349212
some link here

shivansh-gohem:
Hi @xianlubird! Adding to what @JesseStutler mentioned, mapping your entire logical workload replica to a single `PodGroup` and utilizing `Task` topologies (subgroups) is currently the idiomatic Volcano approach to guarantee co-location of distinct units under the same application.

However, if your higher-level controller (like Grove) strictly *must* generate entirely separate `PodGroup` CRDs—perhaps because they have vastly different lifecycles, require independent queuing, or due to controller limitations—this represents a very interesting use case. It suggests the need for a generalized `topologyGroup` annotation or label across multiple `PodGroup`s to coordinate placement, which aligns conceptually with the inter-group affinity discussions in #5347.

Could you elaborate a bit on *why* the external controller splits the workload into multiple `PodGroup`s instead of submitting one unified `PodGroup` with multiple subgroups? Understanding whether this is a hard architectural limitation of the controller versus a specific lifecycle/queuing requirement would really help clarify the design direction here!

xianlubird:
Hello @unmarshall  @sanjaychatterjee 
Any update on this ?

PR #595, opened 2026-05-09
test(e2e): support multiple scheduler backends

### LINKED ISSUE #594: [E2E] Multi-backend E2E test framework
## What

Restructure the Grove E2E suite from KAI-only to support multiple scheduler backends.

## Why

[GREP-375](https://github.com/ai-dynamo/grove/blob/main/docs/proposals/375-scheduler-backend-framework/README.md) introduced a pluggable `Backend` interface and shipped two implementations (`kai-scheduler`, `default-scheduler`). The E2E suite still hardcodes `schedulerName: kai-scheduler` in 22 workload YAMLs and runs only against KAI. As more backends arrive (default-scheduler today, third-party schedulers later per GREP-375 Story 1), this single-backend assumption blocks per-backend regression coverage.

### First PR (#595): Framework + KAI + default-scheduler

- Remove hardcoded `schedulerName` from 22 workload YAMLs; operator's `PreparePod()` injects from `defaultProfileName`.
- Add per-backend Skaffold profile + infra-manager preset.
- `RequireCapability(t, ...)` gate + `DiscoverCapabilities()` reading live `OperatorConfiguration` + a hardcoded backend→capability table cross-checked against actual Go interface implementations.
- CI matrix grows three default-scheduler rows for sensitive suites (RU/OD/SO).
- Three-tier test classification lands (Agnostic / Sensitive / Capability-gated).

### Follow-ups

- **Path-aware PR matrix**: refine the existing `operator/**` filter so backend-specific changes select only the relevant backend's rows.
- **Nightly workflow**: exhaustive (suite × capable backend) matrix on a daily schedule.
- **Scheduler version coverage**: capability table currently assumes a single pinned version per backend; multi-version coverage is a known gap to address later.

## Out of scope

- Modifying GREP-375's `Backend` interface.
- Rewriting existing E2E test code.
- Any change to Dynamo.

## Credits

Originally prototyped by @kangclzjc in #584 (closed and superseded).

brluobt:
#### What type of PR is this?

/kind feature

Restructures the E2E suite from KAI-only to support multiple scheduler backends — KAI as primary, default-scheduler as the first additional backend. Continuation of @kangclzjc's prototype in #584 (closed) with a small polish commit on top.

Three-tier test classification (per `RequireCapability` runtime gating):

- **Agnostic** (CM, CRD): primary backend only — no scheduler-specific code path.
- **Sensitive** (RU, OD, SO): every enabled backend — behavior may diverge subtly across implementations.
- **Capability-gated** (GS, TAS, AutoMNNVL): backends that declare the capability — `RequireCapability(t, ...)` auto-skips otherwise.

Six configuration layers touched (purely additive — no Makefile/test-runner/Helm-rendering changes):

1. **Workload YAMLs** (×22): drop `schedulerName: kai-scheduler`; operator's `PreparePod()` injects from `defaultProfileName`. Same workload YAML now runs unmodified on every backend.
2. **Skaffold profiles**: rename `topology-test` → `e2e-kai`; add `e2e-default-scheduler`.
3. **infra-manager presets**: new `hack/e2e-default-scheduler.yaml` overlay; KAI image prepull gated on `cfg.scheduler.kai.enabled` so default-scheduler rows skip the unused pull.
4. **Capability discovery**: `operator/e2e/tests/{capabilities.go, capability_discovery.go}`; cross-check unit test in `capabilities_test.go` fails the build if the hardcoded backend→capability table drifts from actual Go interface assertions.
5. **Test gates**: `RequireCapability(t, GangScheduling)` etc. added to GS and TAS suites.
6. **CI matrix**: new `create_flags` field threads `-f hack/<preset>.yaml` through `E2E_CREATE_FLAGS`. default-scheduler rows: `rolling_updates_default-scheduler`, `ondelete_updates_default-scheduler`, `startup_ordering_default-scheduler`. e2e-skip mirrors for branch protection.

Refs #594

(This PR does not resolve #594. The issue tracks ongoing multi-backend E2E work including path-filtered PR matrix and nightly runs as follow-ups.)

**Two commits in this PR**:

1. `test(e2e): support multiple scheduler backends` — cherry-picked from @kangclzjc's branch, message prefix updated from `GREP:` to `test(e2e):`. Author preserved as Kang.
2. `test(e2e): polish multi-backend matrix naming and infra preset` — two cosmetic items:
   - CI matrix: rename `_default` → `_default-scheduler` (full backend name in the GHA UI; matches `configv1alpha1.SchedulerNameKube` value).
   - `hack/e2e-default-scheduler.yaml`: replace dangling reference to a non-existent design doc with a concrete activation pointer.

**Merge strategy**: please prefer **rebase-merge** over squash to preserve @kangclzjc's authorship on commit 1. If squash is required, the trailer block should retain `Co-authored-by: Kang Zhang <kangz@nvidia.com>` and both `Signed-off-by` lines.

**Out of scope** for this PR (deferred to follow-ups, tracked by #594):

- **Path-aware PR matrix**: today the matrix runs all configured rows when `operator/**` or `.github/**` changes. Refining to backend-aware filters (`scheduler/kai/**` → KAI rows only, etc.) is a separate PR.
- **Nightly workflow**: exhaustive (suite × capable backend) matrix on schedule.
- **Scheduler version coverage**: the capability table currently assumes a single pinned version per backend; multi-version coverage is a known gap to address later.

**L20 validation** ✅ — lightweight validation on a single-server k3d (l20-6, 30 KWOK workers) completed:
- Run 1 — KAI baseline (`cert_management` + `crd_installer`): **PASS**
- Run 2 — default-scheduler new path (`rolling_updates`): **PASS**

Full evidence (per-test results, what each run proves, environment notes) is in the [validation summary comment](https://github.com/ai-dynamo/grove/pull/595#issuecomment-4411825923).

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/.github/workflows/build-check-test.yaml
+++ b/.github/workflows/build-check-test.yaml
@@ -151,13 +151,13 @@ jobs:
 
           # --- default-scheduler (sensitive tests only; agnostic skipped by policy,
           #     capability-gated skipped by RequireCapability at runtime) ---
-          - test_name: rolling_updates_default
+          - test_name: rolling_updates_default-scheduler
             test_pattern: "^Test_RU"
             create_flags: "-f hack/e2e-default-scheduler.yaml"
-          - test_name: ondelete_updates_default
+          - test_name: ondelete_updates_default-scheduler
             test_pattern: "^Test_OD"
             create_flags: "-f hack/e2e-default-scheduler.yaml"
-          - test_name: startup_ordering_default
+          - test_name: startup_ordering_default-scheduler
             test_pattern: "^Test_SO"
             create_flags: "-f hack/e2e-default-scheduler.yaml"
             make_target: "run-e2e-real-full"
@@ -232,9 +232,9 @@ jobs:
           - test_name: auto_mnnvl
           - test_name: crd_installer
           - test_name: resource_sharing
-          - test_name: rolling_updates_default
-          - test_name: ondelete_updates_default
-          - test_name: startup_ordering_default
+          - test_name: rolling_updates_default-scheduler
+          - test_name: ondelete_updates_default-scheduler
+          - test_name: startup_ordering_default-scheduler
     name: E2E - ${{ matrix.test_name }}
     steps:
       - name: Skip E2E (no relevant changes)
--- a/.github/workflows/build-check-test.yaml
+++ b/.github/workflows/build-check-test.yaml
@@ -97,6 +97,8 @@ jobs:
   # Matrix entries can set:
   #   test_name     (required) - name shown in the GitHub Actions UI
   #   test_pattern  (optional) - Go test -run pattern (standard e2e tests)
+  #   create_flags  (optional) - extra flags appended to E2E_CREATE_FLAGS;
+  #                              empty string means "use the base e2e.yaml preset (KAI)"
   #   make_target   (optional) - Makefile target, defaults to run-e2e-full
   e2e:
     needs: [test, build, check, changes]
@@ -113,28 +115,52 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # --- kai-scheduler (primary backend: agnostic + sensitive + KAI-supported capabilities) ---
           - test_name: gang_scheduling
             test_pattern: "^Test_GS"
+            create_flags: ""
           - test_name: gang_termination
             test_pattern: "^Test_GT"
+            create_flags: ""
           - test_name: rolling_updates
             test_pattern: "^Test_RU"
+            create_flags: ""
           - test_name: ondelete_updates
             test_pattern: "^Test_OD"
+            create_flags: ""
           - test_name: startup_ordering
             test_pattern: "^Test_SO"
+            create_flags: ""
             make_target: "run-e2e-real-full"
           - test_name: Topology_Aware_Scheduling
             test_pattern: "^Test_TAS"
+            create_flags: ""
           - test_name: cert_management
             test_pattern: "^Test_CM"
+            create_flags: ""
           - test_name: auto_mnnvl
             test_pattern: "^Test_AutoMNNVL"
+            create_flags: ""
             make_target: "run-e2e-mnnvl-full"
           - test_name: crd_installer
             test_pattern: "^Test_CRD_Installer"
+            create_flags: ""
           - test_name: resource_sharing
             test_pattern: "^Test_RS"
+            create_flags: ""
+
+          # --- default-scheduler (sensitive tests only; agnostic skipped by policy,
+          #     capability-gated skipped by RequireCapability at runtime) ---
+          - test_name: rolling_updates_default
+            test_pattern: "^Test_RU"
+            create_flags: "-f hack/e2e-default-scheduler.yaml"
+          - test_name: ondelete_updates_default
+            test_pattern: "^Test_OD"
+            create_flags: "-f hack/e2e-default-scheduler.yaml"
+          - test_name: startup_ordering_default
+            test_pattern: "^Test_SO"
+            create_flags: "-f hack/e2e-default-scheduler.yaml"
+            make_target: "run-e2e-real-full"
     name: E2E - ${{ matrix.test_name }}
     steps:
       # print runner specs so we have a record in case of failures
@@ -156,7 +182,7 @@ jobs:
 
       - name: Run e2e tests - ${{ matrix.test_name }}
         run: |
-          make ${{ matrix.make_target || 'run-e2e-full' }} TEST_PATTERN='${{ matrix.test_pattern }}' E2E_CREATE_FLAGS='--dind-memory-mode'
+          make ${{ matrix.make_target || 'run-e2e-full' }} TEST_PATTERN='${{ matrix.test_pattern }}' E2E_CREATE_FLAGS='${{ matrix.create_flags }} --dind-memory-mode'
         working-directory: operator
 
       # The test code handles cleanup via Teardown(), but this step provides
@@ -193,15 +219,22 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # Must mirror the e2e job's matrix.test_name list exactly so required
+          # branch-protection checks resolve the same set of names regardless
+          # of whether E2E ran or was skipped.
           - test_name: gang_scheduling
           - test_name: gang_termination
           - test_name: rolling_updates
+          - test_name: ondelete_updates
           - test_name: startup_ordering
           - test_name: Topology_Aware_Scheduling
           - test_name: cert_management
           - test_name: auto_mnnvl
           - test_name: crd_installer
           - test_name: resource_sharing
+          - test_name: rolling_updates_default
+          - test_name: ondelete_updates_default
+          - test_name: startup_ordering_default
     name: E2E - ${{ matrix.test_name }}
     steps:
       - name: Skip E2E (no relevant changes)
--- a/operator/e2e/k8s/k8sclient/client.go
+++ b/operator/e2e/k8s/k8sclient/client.go
@@ -74,11 +74,32 @@ func init() {
 	crlog.SetLogger(klog.Background())
 }
 
+// Default client-side rate limits for e2e tests.
+//
+// client-go's defaults (5 QPS / 10 burst) are far too low for parallel e2e
+// scenarios like RU-11 (30 nodes + PCS scale-out + rolling update), where the
+// client rate limiter can saturate and cause tests to fail with
+// "client rate limiter Wait returned an error: context deadline exceeded".
+// Bumping to 50/100 matches the ballpark used by most k8s test suites while
+// still being conservative enough for a test kind cluster.
+const (
+	e2eClientQPS   = 50
+	e2eClientBurst = 100
+)
+
 // New creates a Client from a rest.Config.
 func New(restConfig *rest.Config) (*Client, error) {
 	scheme, err := newScheme()
 	if err != nil {
 		return nil, fmt.Errorf("build scheme: %w", err)
+	}
+
+	// Bump client-side rate limits unless the caller already customized them.
+	if restConfig.QPS == 0 {
+		restConfig.QPS = e2eClientQPS
+	}
+	if restConfig.Burst == 0 {
+		restConfig.Burst = e2eClientBurst
 	}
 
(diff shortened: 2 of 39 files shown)


.github/workflows/build-check-test.yaml
```diff
@@ -109,26 +111,49 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # --- kai-scheduler (primary backend: agnostic + sensitive + KAI-supported capabilities) ---
           - test_name: gang_scheduling
             test_pattern: "^Test_GS"
+            create_flags: ""
```
kangclzjc, line 121: if this is empty, can we just don't set this parameters?

.github/workflows/build-check-test.yaml
```diff
@@ -187,14 +212,21 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # Must mirror the e2e job's matrix.test_name list exactly so required
+          # branch-protection checks resolve the same set of names regardless
+          # of whether E2E ran or was skipped.
           - test_name: gang_scheduling
           - test_name: rolling_updates
+          - test_name: ondelete_updates
```
kangclzjc, line 228: Is this because we don't do ondelete_updates test before?

.github/workflows/build-check-test.yaml
brluobt: Good catch — technically yes, GitHub Actions silently expands an undeclared matrix field to an empty string, so we could drop these 8 lines and the workflow would still run.

I left them in for two reasons:

1. **Symmetry with the default-scheduler block** right below, which uses this same field with a non-empty value (`-f hack/e2e-default-scheduler.yaml`). Keeping `create_flags` explicit on every row makes the matrix read as "two backends along the same axis" rather than making default-scheduler look like a special case.

2. **Forward compatibility**: when a third backend lands (e.g. the Volcano backend from GREP-0376 / #560), it'll just be another `create_flags` value, no schema change needed for the include block.

Happy to remove them if you'd rather optimize for line count over the explicit-axis read — let me know which way you prefer.

(Also: pushed a rebase onto current main just now to clear the conflict from #601 / #616.)

.github/workflows/build-check-test.yaml
brluobt: Not new — the `Test_OD*` (ondelete updates) tests have existed in `operator/e2e/tests/` for a while and are already part of the `e2e` job's matrix. The reason `ondelete_updates` is *appearing* in the `e2e-skip` matrix in this PR is that one of the goals here is to make the `e2e-skip` job's `test_name` list a **1:1 mirror** of the `e2e` job's, so the same set of branch-protection check names resolve regardless of whether E2E ran or was skipped.

I just verified the two matrices are aligned now — both list the same 12 entries (including `ondelete_updates` and the three `*_default-scheduler` variants). The comment block at the top of `e2e-skip` flags this contract so future contributors know to update both blocks together when adding a backend or test.

kangclzjc (dismissed):

brluobt:
### CI status: 3 default-scheduler jobs failing, root cause analysis

The 3 `*_default-scheduler` E2E jobs in this PR fail consistently (incl. retry) with the same error:
```
failed calling webhook "pcs.defaulting.webhooks.grove.io":
no endpoints available for service "grove-operator"
```
All 14 other channels (kai-scheduler, Topology, cert/crd/etc.) pass.

**Believe this is a pre-existing race in the e2e setup that this PR exposes, not a defect introduced by this PR.**

#### Evidence

1. The webhook readiness check in `hack/infra_manager/grove.py` does a single `kubectl create --dry-run=server` against `workload1.yaml`. Once that passes, setup declares webhook ready.
2. From a failing job's log:
   - `19:02:00` — "Grove webhook is ready" (dry-run passed)
   - `19:02:46` — first real `apply` from test → no endpoints available
   46-second gap (go test compile) shouldn't matter — but endpoints disappeared.
3. The CI matrix runs jobs against a `shared-e2e-test-cluster`, and each job's setup does `helm uninstall grove-operator` followed by skaffold deploy. When 3 default-scheduler jobs (+ kai jobs) run in parallel on the same cluster, any later `helm uninstall` deletes the deployment, briefly removing webhook endpoints — and a sibling job that just finished its own ready check will hit a webhook call with zero endpoints.
4. This is consistent with: 100% reproducible, retry also red, only the new default-scheduler channels affected (3-way mutual interference).

#### Suggested fix options

- **A. Channel isolation (preferred):** give the default-scheduler matrix entries their own k3d cluster (or stagger setup). Structural fix, costs ~1min CI per job.
- **B. Stronger readiness check:** poll `kubectl get endpoints grove-operator` after the dry-run, ensuring non-empty subsets. Defensive, but won't help when a sibling job's uninstall later wipes endpoints again.
- **C. Helm chart-side guard** (e.g. `maxUnavailable: 0`): scope creep, and `helm uninstall` removes the deployment outright so it doesn't actually help.

Happy to take a stab at A if that's the preferred direction — wanted to flag the root cause first since it touches the e2e setup framework, not the capability table I'm adding here.

shayasoolin:
@brluobt I'm afraid that the root cause analysis in point 3 isn't quite right. Each matrix job runs on its own dedicated runner from the `prod-grove-e2e-v1` pool with its own isolated k3d cluster — jobs don't share a cluster, so a sibling job's `helm uninstall` can't be responsible.
See `github/workflows/build-check-test.yaml` - each job in the matrix sets its own runner and installations.

brluobt:
@shayasoolin thanks for the correction — you're right, my point 3 was wrong; runners and clusters are isolated per matrix entry. The failures were actually fixed on the same day by [758f8d1](https://github.com/ai-dynamo/grove/pull/595/commits/758f8d1): the `e2e-default-scheduler` skaffold profile was inheriting the chart's 4-backend default in the rendered ConfigMap, which made the operator's `/readyz` flap long enough for the endpoints to disappear. Scoping the profile to a single backend (symmetric with `e2e-kai`) resolved it, and all 3 default-scheduler jobs have been green since ([12/12 pass](https://github.com/ai-dynamo/grove/actions/runs/28160910602)).

brluobt:
**RU-11 flake analysis (rolling_updates on both KAI & default-scheduler)**

TL;DR: this is a flake, not a regression. Both `rolling_updates` and `rolling_updates_default-scheduler` fail only on `Test_RU11_RollingUpdateWithPCSScaleOut` with:

```
client rate limiter Wait returned an error: context deadline exceeded
```

The other 20 RU_* tests pass on both backends.

**Why flake, not regression**
- RU-11 is the tightest scenario: 30 nodes + PCS 2→3 scale-out (10→30 pods) + parallel rolling update, all under a hard 2-minute test timeout.
- The failure is client-go's built-in rate limiter saturating (default 5 QPS / 10 burst), not a controller bug.
- Timing lines up: 08:11:31 → 08:13:51 = exactly 2 min (rate-limited API waits) + 20 s diag.
- `rolling_recreate_test.go` last touched 2026-04-19; test code hasn't changed.
- Last CI on this branch (commit 758f8d1, 06-25) was 12/12 green with identical test code.
- `main` branch's last ~15 E2E runs are green — no systemic regression.

**Fix plan (pushing shortly)**
1. Bump `Test_RU11`'s timeout from 2 min → 3 min (small headroom for the tightest scenario).
2. Raise the e2e k8s client's default QPS/burst (5/10 → 50/100) in `k8sclient.New` so parallel operations aren't throttled by client-side limiter.
3. Rebase onto latest `main` (currently ~4 weeks behind).

Will follow up with commits + a re-run request. Thanks!

brluobt:
**Update on the follow-up run: RU-11 is stable, RU-12 now flakes under the same rate-limiter symptom**

After [771efc27](https://github.com/ai-dynamo/grove/pull/595/commits/771efc27) (bump QPS/burst to 50/100 + RU-11 timeout to 3 min), CI [run 29978288967](https://github.com/ai-dynamo/grove/actions/runs/29978288967) is 18/19 green — RU-11 now passes on both backends. However, `rolling_updates_default-scheduler` still fails on a different test: `Test_RU12_RollingUpdateAfterPCSScaleOut`.

**Analysis**

- The `client rate limiter Wait ... context deadline exceeded` string is the messenger, not the cause: it's what client-go's throttled `Wait()` reports when the surrounding `ctx.Done()` fires. The real cause is the surrounding poll timeout expiring, not QPS/burst saturation.
- RU-12 (`rolling_recreate_test.go:371`) sets `tcOrdinalTimeout.Timeout = 60 * time.Second` and calls `waitForOrdinalUpdating` to wait for the final ordinal to enter `CurrentlyUpdating`. Timeline confirms exact 60 s expiry:
  - `04:08:34.296` — start of step 4 (`waitForOrdinalUpdating`)
  - `04:09:34.297` — timeout at exactly +60 s (`context deadline exceeded`)
- Under default-scheduler, pod pending → scheduling → running is noticeably slower than under KAI (KAI/volcano variants of rolling_updates completed in ~43 s for RU-11 while default-scheduler needed 87 s for RU-12), so the same 60 s wait crosses the ceiling on this backend even after 771efc27.
- RU-11 (which the previous commit fixed) used a test-level 2 min → 3 min bump; RU-12 has an internal `tcOrdinalTimeout` of 60 s that the previous commit did not touch.

**Suggested next step** (open to reviewer preference before pushing):

- Bump `Test_RU12`'s `tcOrdinalTimeout.Timeout` from 60 s to 120 s, aligning with the `tcLongTimeout = 2 * time.Minute` used immediately after it in the same test.
- No global change needed — QPS/burst bump in 771efc27 is still doing its job; this is a per-test headroom miss.

Will push the timeout bump on ack, or defer if you'd prefer a different scope (e.g. sweep all `tcOrdinalTimeout` sites to a shared constant).

brluobt:
Follow-up investigation supersedes my previous RU-12 timeout analysis.

The failure is not caused by the default scheduler taking longer. RU-12 uses a 5-second poll to observe a transient `CurrentlyUpdating` transition that can disappear within 0.3–0.4 seconds. This race is scheduler-agnostic; #595 only increased its exposure by adding the default-scheduler matrix row.

The issue is tracked in #733 and fixed independently by #734. I am keeping this PR unchanged until #734 is merged, then I will rebase it onto main and rerun the multi-backend matrix.

sanjaychatterjee:
@brluobt let's prioritize this PR once you are back. We are rapidly adding support for multiple scheduler backends for our customers.

Issue #596, opened 2026-05-10
GREP: PodClique Disruption Policy


andrewpaprotsky:
## Summary

Track a GREP for an opt-in PodClique disruption policy for taint-manager-driven replacement. When Kubernetes starts deleting a managed Pod because it does not tolerate a NoExecute taint, Grove deletes the owning PodClique so the owner recreates the PodClique and normal PodClique pod sync creates the desired Pods.

## Motivation

Grove currently repairs missing Pods one-for-one. That is correct for ordinary replicated Pods, but not for workloads where disruption of one member means the whole PodClique should be recreated. One-for-one repair can leave the remaining Pods bound to the old placement and deny the scheduler a clean chance to place the replacement unit.

## Proposed scope

- Preserve current behavior by default.
- Add a simple opt-in PodClique policy for taint-manager-driven replacement.
- Match Kubernetes Pod DisruptionTarget=True with reason DeletionByTaintManager.
- Support action Recreate only.
- Keep replacement scoped to the configured PodClique.
- Keep placement decisions in the scheduler.

## Non-goals

- Supporting every DisruptionTarget reason.
- Replacing PodCliqueSet or PodCliqueScalingGroup replicas as a larger unit.
- Changing scheduler placement logic.
- Enforcing PodDisruptionBudget semantics for Pods removed by PodClique deletion.

## Proposal draft

A GREP draft will be added under docs/proposals using this issue number.

oleg-kushniriov:
Hi @andrewpaprotsky — thanks for opening this. Before the GREP draft, would it be worth evaluating whether existing gang termination already covers (or could be lightly extended to cover) what you need here?

 The current gang-termination action — createPCSReplicaDeleteTask in gangterminate.go (and its PCSG counterpart) — already does exactly what you propose: delete all PodCliques for a replica so the owner recreates them. The action layer looks reusable as-is.

 Where the gap sits is the trigger, not the action. Gang termination today fires on aggregate MinAvailableBreached after TerminationDelay. Your proposal needs a more specific, threshold-free signal: per-pod DisruptionTarget=True, reason=DeletionByTaintManager. That means today's path:
 - won't fire when a single pod is taint-evicted but minAvailable isn't breached,
 - waits out TerminationDelay even when the taint signal already tells us the placement is bad,
 - is cause-agnostic, so it can't be limited to taint-manager only.

andrewpaprotsky:
[...]

PR #597, opened 2026-05-10
GREP-596: PodClique Disruption Policy

andrewpaprotsky:
Adds GREP-596 for an opt-in `PodClique` disruption policy that replaces a managed `PodClique` when Grove observes a taint-manager-driven Pod deletion signal.

Related to #596

### DIFF
--- a/docs/proposals/596-podclique-disruption-policy/README.md
+++ b/docs/proposals/596-podclique-disruption-policy/README.md
@@ -1,0 +1,222 @@
+# GREP-596: PodClique Disruption Policy
+
+<!-- toc -->
+- [Summary](#summary)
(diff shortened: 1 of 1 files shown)


andrewpaprotsky:
Closing this.

After mapping the recovery case back to the actual object boundaries, the behavior we need is not PodClique-scoped. The affected recovery unit can be only a subset of Pods inside a generated Grove object, so deleting/recreating the whole PodClique would restart too much.

The better implementation is in the higher-level controller that understands that recovery unit: watch terminal/evicted Pods, identify the related sibling Pods, delete that subset, and let Grove recreate the missing Pods through its existing reconciliation loop.

Issue #393, opened 2026-02-04
☂️ Grove Workload Update Strategies


sanjaychatterjee:
### What you would like to be added?

This is an umbrella issue to track all work on different update strategies supported by Grove.

### Why is this needed?

Grove workloads have various update requirements which cannot be met by only one strategy. Some of the considerations used are:
1. backward compatible vs incompatible updates
2. training vs inference updates
3. And more.

Issue #292, opened 2025-12-18
Rolling Update support inplace update pod image


xulinfei1996:
### What you would like to be added?

Rolling update supports update image while not delete pod just update pod(InPlaceIfPossible), so that kubelet will download image and restart container.

### Why is this needed?

Currently, kubelet supports downloading the image and restarting the container without the deletion of the entire Pod to update the image.
When the cluster is busy, deleting a Pod and rescheduling it (may be PodGroup), incurs significant overhead (e.g., resource reallocation, scheduling delays, etc.). Therefore, avoiding Pod deletion and directly updating the image can significantly reduce this overhead.

gflarity:
Hi @xulinfei1996, this is an interesting idea, but this approach is generally discouraged even though it works.  Can you give a more concrete example of why for AI workloads in particular this overhead is bad and something like this would be needed?

danbar2:
We will start an investigation on the [kubernetes feature](https://kubernetes.io/blog/2026/01/02/kubernetes-v1-35-restart-all-containers/). Based on the Research, we will provide a GREP

AllenXu93:
> We will start an investigation on the [kubernetes feature](https://kubernetes.io/blog/2026/01/02/kubernetes-v1-35-restart-all-containers/). Based on the Research, we will provide a GREP

May I take this issue? 
I can provide a KEP for supporting pod inplace update in grove.

PR #598, opened 2026-05-11
Implement in-place Pod image update

AllenXu93:
## Summary
- Add GREP-292 for in-place Pod image updates
- Add InPlaceIfPossible and InPlaceOnly update strategies
- Patch eligible Pod container images in place and advance the Grove template hash after kubelet reports completion
- Add readiness gate handling, fallback/block behavior, CRD schema updates, and focused unit tests

## Validation
- make --directory=operator/api generate
- go test ./internal/controller/common/component/utils ./internal/controller/podclique/components/pod
- go test ./...
- make verify-toc


danbar2:
Thanks for the PR and the detailed work here, @AllenXu93 — and for writing up the accompanying GREP in #599 as well.

We did consider this in-place update approach (it’s been on our radar via #292, under the broader update-strategies umbrella #393), but we’ve decided to hold off on it for now. The team is currently focused on a more advanced update mechanism: Coherent updates (GREP-393, merged in #640, with implementation tracked in #638 and #650). That work substantially reworks how updates are orchestrated across PodGangs — introducing per-component rolling-update configuration, a new PodGangMap CRD, and multi-version update step planning — so landing a new update strategy in parallel would create significant overlap and likely require rework once it merges.

Once the coherent update machinery stabilizes, in-place image updates become a natural extension on top of it, and we’d be glad to revisit this then. We’ll keep #292 open to track it — your input there on use cases and requirements would be very welcome.

Thanks again for the contribution!

PR #599, opened 2026-05-11
Add GREP for in-place Pod image updates

AllenXu93:
## Summary
- Add GREP-292 for opt-in in-place Pod image updates
- Define InPlaceIfPossible and InPlaceOnly update strategies
- Document eligibility checks, readiness gate behavior, completion detection, status/events, and test plan

## Key design decisions
- Keep RollingRecreate as the default behavior; in-place update is explicitly opt-in.
- Support only regular container image changes for the initial implementation.
- Use a Grove-managed readiness gate so Pods are marked not ready while kubelet restarts updated containers.
- Advance the Grove Pod template hash only after kubelet reports the updated container image ID and readiness.
- Let InPlaceIfPossible fall back to RollingRecreate for unsupported changes, while InPlaceOnly blocks without deleting Pods.

Related to  https://github.com/ai-dynamo/grove/issues/292

### DIFF
--- a/docs/proposals/292-in-place-pod-image-update/README.md
+++ b/docs/proposals/292-in-place-pod-image-update/README.md
@@ -6,16 +6,11 @@
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
-  - [User Stories](#user-stories)
-    - [Story 1: Image-Only Rollout for GPU Inference](#story-1-image-only-rollout-for-gpu-inference)
-    - [Story 2: Strict No-Delete Upgrade](#story-2-strict-no-delete-upgrade)
-    - [Story 3: Fallback for Mixed Changes](#story-3-fallback-for-mixed-changes)
   - [Limitations/Risks &amp; Mitigations](#limitationsrisks--mitigations)
     - [Unsupported Pod Template Changes](#unsupported-pod-template-changes)
     - [Premature Update Completion](#premature-update-completion)
     - [Readiness and Traffic Impact](#readiness-and-traffic-impact)
     - [Image Pull or Startup Failures](#image-pull-or-startup-failures)
-    - [Older Kubernetes Versions](#older-kubernetes-versions)
 - [Design Details](#design-details)
   - [API Changes](#api-changes)
   - [High-Level Architecture](#high-level-architecture)
@@ -26,81 +21,56 @@
   - [PodCliqueScalingGroup Flow](#podcliquescalinggroup-flow)
   - [Status and Conditions](#status-and-conditions)
   - [Monitoring](#monitoring)
-  - [Dependencies](#dependencies)
   - [Test Plan](#test-plan)
     - [Unit Tests](#unit-tests)
     - [Integration Tests](#integration-tests)
     - [E2E Tests](#e2e-tests)
   - [Graduation Criteria](#graduation-criteria)
-    - [Alpha](#alpha)
-    - [Beta](#beta)
-    - [GA](#ga)
-- [Implementation History](#implementation-history)
-- [Alternatives](#alternatives)
-  - [Add an <code>inPlace</code> Flag Under <code>RollingRecreate</code>](#add-an-inplace-flag-under-rollingrecreate)
-  - [Always Try In-Place Update Under <code>RollingRecreate</code>](#always-try-in-place-update-under-rollingrecreate)
-  - [Reuse <code>OnDelete</code>](#reuse-ondelete)
-  - [Introduce ControllerRevision History](#introduce-controllerrevision-history)
-- [Appendix](#appendix)
 <!-- /toc -->
 
 ## Summary
 
-This GREP extends `PodCliqueSet` updates with an opt-in in-place pod image update strategy. When a workload update only changes regular container images, Grove can update existing Pods by patching `spec.containers[*].image` and waiting for kubelet to restart the affected containers, instead of deleting the Pods and forcing rescheduling. This reduces update cost for large AI workloads, especially workloads that rely on gang scheduling, scarce accelerators, or expensive scheduler placement decisions. Existing `RollingRecreate` and `OnDelete` behavior remain unchanged unless users explicitly select the new strategy.
+This GREP extends `PodCliqueSet` updates with opt-in in-place Pod image updates. When a workload update only changes regular container images, Grove can patch existing Pods by updating `spec.containers[*].image` and wait for kubelet to restart the affected containers, instead of deleting Pods and forcing rescheduling. Existing `RollingRecreate` and `OnDelete` behavior remain unchanged unless users explicitly select one of the new in-place strategies.
 
 ## Motivation
 
-Grove currently updates PodCliqueSet templates by either automatically recreating resources with `RollingRecreate` or waiting for user-driven deletion with `OnDelete`. Both strategies create a new Pod whenever the updated specification must be applied. For AI inference and training workloads, Pod recreation can be expensive even when the only meaningful change is a container image tag:
+Grove currently applies PodCliqueSet template changes through either automatic recreate with `RollingRecreate` or user-driven deletion with `OnDelete`. Recreating Pods is expensive for AI workloads even when the only change is an image tag:
 
-- The Pod has to pass through scheduler queues again, and for PodCliqueScalingGroups the replacement may involve gang scheduling.
-- Scarce GPU resources can be reallocated to another workload between deletion and replacement.
-- Existing node-local placement, warm caches, mounted resources, IP address continuity, and scheduler backend state can be lost.
-- Large clusters under load can spend more time rescheduling a Pod than kubelet spends pulling and restarting a container image.
+- Pods enter scheduler queues again, and PodCliqueScalingGroup replacements may require gang scheduling.
+- Scarce accelerator placement can be lost between deletion and replacement.
+- Node-local placement, warm caches, mounted resources, IP address continuity, and scheduler backend state can be disrupted.
+- Under cluster pressure, rescheduling can take longer than pulling and restarting a new image on the same node.
 
-Kubernetes permits updating Pod container images in place. When `spec.containers[*].image` changes, kubelet pulls the new image and restarts the affected container without deleting the Pod object. Grove should expose this capability in its update orchestration while preserving the current recreate-based strategy for changes that cannot be safely applied in place.
+Kubernetes allows updating regular container images on an existing Pod. Grove should use this capability for image-only changes while preserving recreate semantics for unsupported changes.
 
 ### Goals
 
 - Add explicit `PodCliqueSet` update strategy types for in-place image updates.
-- Support in-place updates when the effective Pod template change is limited to regular container image changes and Grove-managed metadata.
-- Preserve `RollingRecreate` as the default update strategy for backward compatibility.
-- Reuse Grove's existing update progress and template hash model so users can track convergence with `updatedReplicas` and `updateProgress`.
-- Keep PodCliqueScalingGroup updates gang-aware by updating one selected PodCliqueSet replica at a time, while avoiding deletion when in-place update is possible.
-- Provide events, status fields, and conditions that show whether an update is progressing in place, completed, blocked, or falling back to recreate.
+- Apply in-place updates when the effective Pod template change is limited to regular container image changes.
+- Preserve `RollingRecreate` as the default update strategy.
+- Reuse Grove's existing update progress and template hash model.
+- Keep PodCliqueScalingGroup updates gang-aware by updating one selected replica at a time.
+- Surface in-place update progress, completion, blocked state, and fallback behavior through status and events.
 
 ### Non-Goals
 
-- Supporting in-place updates for fields other than regular container images in the initial implementation.
+- Supporting in-place updates for fields other than regular container images.
 - Supporting in-place updates for `initContainers`, ephemeral containers, resources, commands, args, env, probes, volumes, scheduling fields, resource claims, topology constraints, startup order, or service discovery fields.
 - Changing the default `RollingRecreate` behavior.
 - Providing application-level compatibility checks between old and new images.
-- Guaranteeing zero traffic impact for workloads that do not opt into the proposed readiness gate.
-- Introducing `ControllerRevision` resources for PodCliqueSet history. Grove already has generation hashes and template hashes that are sufficient for this scoped feature.
+- Guaranteeing zero traffic impact while kubelet restarts updated containers.
+- Introducing a new history resource for PodCliqueSet revisions.
 
 ## Proposal
 
 Introduce two new `PodCliqueSet` update strategy types:
 
-- `InPlaceIfPossible`: Grove first attempts to apply eligible image-only updates in place. If the change is not eligible, Grove falls back to the existing `RollingRecreate` behavior.
-- `InPlaceOnly`: Grove attempts only in-place updates. If the change is not eligible, Grove does not delete Pods and instead exposes a blocked update state through status and events.
+- `InPlaceIfPossible`: Grove attempts eligible image-only updates in place. If the change is not eligible, Grove falls back to the existing `RollingRecreate` behavior.
+- `InPlaceOnly`: Grove only attempts in-place updates. If the change is not eligible, Grove blocks the update and does not delete Pods.
 
 The new strategies are opt-in. Existing users continue to get the current `RollingRecreate` behavior when `spec.updateStrategy` is omitted or explicitly set to `RollingRecreate`. The existing `OnDelete` strategy remains manual and does not automatically patch or delete existing Pods.
 
-In-place update is evaluated per Pod. For each Pod that has an outdated template hash, Grove compares the existing Pod against the desired Pod that would be created from the current PodCliqueSet template. If the only pod spec differences are regular container image changes, Grove patches the existing Pod. After kubelet restarts the updated containers and reports new container status, Grove marks the Pod as updated by changing its Grove template hash label to the target hash.
-
-### User Stories
-
-#### Story 1: Image-Only Rollout for GPU Inference
-
-As an inference platform operator, I want to deploy a new model server image without deleting Pods so that existing GPU placement and scheduling decisions are preserved. I configure `spec.updateStrategy.type: InPlaceIfPossible`, update the image in the PodCliqueSet template, and Grove rolls the image across Pods without rescheduling eligible Pods.
-
-#### Story 2: Strict No-Delete Upgrade
-
-As a platform operator running a highly constrained GPU cluster, I want Grove to avoid Pod deletion during a rollout even if the update cannot be applied in place. I configure `spec.updateStrategy.type: InPlaceOnly`. If my change includes unsupported fields, Grove blocks the update and emits a clear event instead of falling back to Pod deletion.
-
-#### Story 3: Fallback for Mixed Changes
-
-As an application developer, I want the same update strategy to optimize image-only changes while still allowing broader template changes. I use `InPlaceIfPossible`. When I update only images, Grove patches Pods in place. When I later change resources or environment variables, Grove automatically falls back to `RollingRecreate`.
+In-place update is evaluated per Pod. For each Pod with an outdated template hash, Grove builds the desired Pod from the current PodCliqueSet template and compares it with the existing Pod. If the only Pod spec differences are regular container image changes, Grove patches the existing Pod. After kubelet restarts the updated containers and reports new container status, Grove marks the Pod as updated by changing its Grove template hash label to the target hash.
 
 ### Limitations/Risks & Mitigations
 
@@ -108,31 +78,25 @@ As an application developer, I want the same update strategy to optimize image-o
 
 Only regular container image changes are eligible for in-place update. Any other Pod spec change requires recreate semantics.
 
-*Mitigation*: `InPlaceIfPossible` falls back to `RollingRecreate`; `InPlaceOnly` blocks the update with a status condition and event that lists the first unsupported field path.
+*Mitigation*: `InPlaceIfPossible` falls back to `RollingRecreate`; `InPlaceOnly` blocks the update with status and events that describe why the Pod is not eligible.
 
 #### Premature Update Completion
 
 If Grove updates the Pod template hash label before kubelet actually restarts containers, `status.updatedReplicas` could incorrectly report success.
 
-*Mitigation*: Grove records in-place update state separately and updates the Pod template hash label only after kubelet reports completion through changed `status.containerStatuses[*].imageID` or equivalent runtime metadata.
+*Mitigation*: Grove records in-place update state separately and updates the Pod template hash label only after kubelet reports changed `status.containerStatuses[*].imageID` for updated containers and those containers are ready.
 
 #### Readiness and Traffic Impact
 
-Patching an image causes kubelet to restart the container. Without readiness gate integration, the Pod may continue to be considered ready until kubelet transitions its conditions.
+Patching an image causes kubelet to restart the container. The Pod should not stay ready from Grove's perspective during the update.
 
-*Mitigation*: Grove injects a Grove-managed readiness gate into newly created Pods when in-place update is enabled. During an in-place update, Grove sets the condition to `False` before patching the Pod and restores it to `True` after completion. For existing Pods that do not carry the readiness gate, the alpha behavior is conservative: `InPlaceIfPossible` falls back to recreate and `InPlaceOnly` blocks the update.
+*Mitigation*: Grove injects a Grove-managed readiness gate into newly created Pods when in-place update is enabled. Before patching images, Grove sets the condition to `False`; after completion, Grove sets it back to `True`. Existing Pods without the readiness gate are not patched in place.
 
 #### Image Pull or Startup Failures
 
 If the new image cannot be pulled or fails to become ready, the Pod remains in an update-in-progress state.
 
-*Mitigation*: Grove surfaces the failure through Pod and PodClique status, Kubernetes Events, and unchanged `updatedReplicas`. Grove does not automatically delete the Pod as a recovery action in `InPlaceOnly`. In `InPlaceIfPossible`, fallback is decided before the in-place patch; after the Pod has been patched, Grove waits for kubelet or user intervention rather than switching strategies mid-update.
-
-#### Older Kubernetes Versions
-
-Pod image mutation behavior depends on Kubernetes support for updating `spec.containers[*].image` on existing Pods.
-
-*Mitigation*: The feature depends on Kubernetes versions that allow Pod image updates. The controller should treat API validation failures during Pod patch as update failure and surface a clear event. Documentation will call out the required Kubernetes support level.
+*Mitigation*: Grove keeps the old template hash label until completion and surfaces the failure through status and events. After a Pod has been patched, Grove waits for kubelet or user intervention rather than switching strategies mid-update.
 
 ## Design Details
 
@@ -165,29 +129,6 @@ const (
 )
 ```
 
-Add optional in-place settings to `PodCliqueSetUpdateStrategy`:
-
-```go
-type PodCliqueSetUpdateStrategy struct {
-    // Type indicates the type of update strategy.
-    // Default is RollingRecreate.
-    // +kubebuilder:default=RollingRecreate
-    Type UpdateStrategyType `json:"type,omitempty"`
-
-    // InPlaceUpdate configures behavior for InPlaceIfPossible and InPlaceOnly.
-    // +optional
-    InPlaceUpdate *InPlaceUpdateStrategy `json:"inPlaceUpdate,omitempty"`
-}
-
-type InPlaceUpdateStrategy struct {
-    // GracePeriodSeconds is the delay between marking a Pod not-ready through
-    // the Grove in-place readiness gate and patching container images.
-    // Defaults to 0.
-    // +optional
-    GracePeriodSeconds *int32 `json:"gracePeriodSeconds,omitempty"`
-}
-```
-
 Example usage:
 
 ```yaml
@@ -199,8 +140,6 @@ spec:
   replicas: 2
   updateStrategy:
     type: InPlaceIfPossible
-    inPlaceUpdate:
-      gracePeriodSeconds: 30
   template:
     cliques:
       - name: decode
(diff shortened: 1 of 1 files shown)


oleg-kushniriov:
Hi @AllenXu93, thanks for putting this together.

Heads-up:there's a parallel effort in progress for another update strategy called MVU (Multi-Version Update) which also avoids rescheduling — it restarts all Pods in place on their existing nodes rather than recreating them. The GREP for it isn't published yet, but there is meaningful overlap with this proposal

FYI: @sanjaychatterjee @renormalize @danbar2 @unmarshall

AllenXu93:
> Hi @AllenXu93, thanks for putting this together.
> 
> Heads-up:there's a parallel effort in progress for another update strategy called MVU (Multi-Version Update) which also avoids rescheduling — it restarts all Pods in place on their existing nodes rather than recreating them. The GREP for it isn't published yet, but there is meaningful overlap with this proposal
> 
> FYI: @sanjaychatterjee @renormalize @danbar2 @unmarshall

Thanks for the heads-up and for pointing me to the MVU effort.

To clarify, this GREP is intentionally narrow: it focuses on opt-in, image-only in-place updates for Grove-managed Pods, while keeping RollingRecreate as the default and fallback path.

I’d like to better understand whether MVU is expected to cover this use case as well, and what primitive it plans to use for the in-place restart behavior, especially whether it depends on Kubernetes v1.35 RestartAllContainers.

Would it make sense to merge these discussions, or should I participate directly in the MVU proposal instead?

renormalize:
Hi @AllenXu93 @oleg-kushniriov - apologies for the delayed response, I wanted to gain some more clarity on a few things before I responded here.

> Heads-up:there's a parallel effort in progress for another update strategy called MVU (Multi-Version Update) which also avoids rescheduling

This is not entirely accurate. I would say that the efforts in these update strategies are orthogonal. Sure, it does touch the same code paths but I do not see any reason to merge these efforts. I would suggest not to work too hard on the implementation for in-place until we have the other strategy merged.

> I’d like to better understand whether MVU is expected to cover this use case as well, and what primitive it plans to use for the in-place restart behavior, especially whether it depends on Kubernetes v1.35 RestartAllContainers.

The efforts in that strategy do not depend on `RestartAllContainers` from 1.35, as per my understanding.

From my first pass at the GREP, I have a few questions/comments, I will post them soon.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+- Keep PodCliqueScalingGroup updates gang-aware by updating one selected replica at a time.
+- Surface in-place update progress, completion, blocked state, and fallback behavior through status and events.
+
+### Non-Goals
+
+- Supporting in-place updates for fields other than regular container images.
+- Supporting in-place updates for `initContainers`, ephemeral containers, resources, commands, args, env, probes, volumes, scheduling fields, resource claims, topology constraints, startup order, or service discovery fields.
+- Changing the default `RollingRecreate` behavior.
+- Providing application-level compatibility checks between old and new images.
+- Guaranteeing zero traffic impact while kubelet restarts updated containers.
```
renormalize: I do not think this can be guaranteed even if we wanted to. The responsibility will be on the user to ensure they have enough replicas above their `minAvailable` during their updates to ensure service availability. Feel free to correct me if I am wrong.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+- Introducing a new history resource for PodCliqueSet revisions.
+
+## Proposal
+
+Introduce two new `PodCliqueSet` update strategy types:
+
+- `InPlaceIfPossible`: Grove attempts eligible image-only updates in place. If the change is not eligible, Grove falls back to the existing `RollingRecreate` behavior.
```
renormalize: Could you perhaps point me to the scenarios in which we would need a `InPlaceIfPossible`? What entails `If the change is not eligible`?

AllenXu93:
> Hi @AllenXu93 @oleg-kushniriov - apologies for the delayed response, I wanted to gain some more clarity on a few things before I responded here.
> 
> > Heads-up:there's a parallel effort in progress for another update strategy called MVU (Multi-Version Update) which also avoids rescheduling
> 
> This is not entirely accurate. I would say that the efforts in these update strategies are orthogonal. Sure, it does touch the same code paths but I do not see any reason to merge these efforts. I would suggest not to work too hard on the implementation for in-place until we have the other strategy merged.
> 
> > I’d like to better understand whether MVU is expected to cover this use case as well, and what primitive it plans to use for the in-place restart behavior, especially whether it depends on Kubernetes v1.35 RestartAllContainers.
> 
> The efforts in that strategy do not depend on `RestartAllContainers` from 1.35, as per my understanding.
> 
> From my first pass at the GREP, I have a few questions/comments, I will post them soon.

Thanks for the clarification, that makes sense.
I’ll wait for your detailed comments before making further changes.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+In-place update is evaluated per Pod. For each Pod with an outdated template hash, Grove builds the desired Pod from the current PodCliqueSet template and compares it with the existing Pod. If the only Pod spec differences are regular container image changes, Grove patches the existing Pod. After kubelet restarts the updated containers and reports new container status, Grove marks the Pod as updated by changing its Grove template hash label to the target hash.
+
+### Limitations/Risks & Mitigations
+
+#### Unsupported Pod Template Changes
+
+Only regular container image changes are eligible for in-place update. Any other Pod spec change requires recreate semantics.
+
+*Mitigation*: `InPlaceIfPossible` falls back to `RollingRecreate`; `InPlaceOnly` blocks the update with status and events that describe why the Pod is not eligible.
+
```
renormalize, line 93: I would rather depend on the validating webhook to evaluate if an update is allowed to be triggered with `InPlace`, by checking the mentioned fields; or deny it altogether.

Having a `IfPossible` strategy makes it confusing for consumers, and a strict acceptance or denial of an in-place update is a better API, in my opinion.

If you have an particular use case in mind where `IfPossible` is absolutely necessary, please add it to the document.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+If Grove updates the Pod template hash label before kubelet actually restarts containers, `status.updatedReplicas` could incorrectly report success.
+
+*Mitigation*: Grove records in-place update state separately and updates the Pod template hash label only after kubelet reports changed `status.containerStatuses[*].imageID` for updated containers and those containers are ready.
+
+#### Readiness and Traffic Impact
+
+Patching an image causes kubelet to restart the container. The Pod should not stay ready from Grove's perspective during the update.
+
+*Mitigation*: Grove injects a Grove-managed readiness gate into newly created Pods when in-place update is enabled. Before patching images, Grove sets the condition to `False`; after completion, Grove sets it back to `True`. Existing Pods without the readiness gate are not patched in place.
```
renormalize: My understanding is that when the container is restarted (after the in-place update patch call is made), the Pod has its `Ready` condition move from true to false. When this happens, the pod is removed from the endpointslices, causing no traffic to go to it. So the pod is never really 'ready' from grove's perspective or a kubernetes perspective.

So I ask, do we really need a  readiness gate?

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+- The Pod carries the Grove in-place readiness gate.
+- The existing Pod and desired Pod have the same regular container set, matched by container name.
+- The only Pod spec changes are `spec.containers[*].image`.
+- `initContainers`, ephemeral containers, volumes, resource claims, scheduling fields, resources, probes, env, command, args, security context, restart policy, and DNS settings are unchanged.
+
+When eligibility fails, Grove records a structured reason for status and events.
+
+### Pod In-Place Update State
+
+Grove stores in-place update state on the Pod:
```
renormalize: Could you also describe that this is stored in the PodCliqueStatus, as a slice?

However, I see some potential issues with this approach. If a PodClique has multiple thousands of pods, storing all of these fields for every single pod will not be scalable. If the in-place update will always be done one pod at a time, then we can store this for the currently updating pod only.

renormalize (commented):
@AllenXu93 these are a few comments I have so far. Let's discuss more on my current comments.

Also, the PR does not really mention restarting all containers of the pod when the pod image is patched. For certain workloads, we would *need* to establish startup order (through grove-initc) after image updates. So how are you planning on approaching on restarting all containers of the pod (`RestartAllContainers`) with this update strategy?

I experimented a bit yesterday, and unfortunately we can not use `RestartAllContainers` under `restartPolicyRules` to restart all containers when a container image is updated, by relying on the exit code of the container being updated when it is stopped. I think we would need a sidecar to run that helps trigger the restart of all containers, but I am not sure what the approach would be. Feel free to correct me if I am wrong.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+
+```go
+type InPlaceUpdateState struct {
```
oleg-kushniriov: Could InPlaceUpdateState lean on the existing progress model instead of being its own type?
  
  Looking at this against PodCliqueUpdateProgress in operator/api/core/v1alpha1/podclique.go:142-166, it seems like most of these fields might already be covered by what Grove tracks today. That progress type follows a nice pattern — one shape per resource, with optional sub-fields for strategy-specific bits (e.g. ReadyPodsSelectedToUpdate only set for auto strategies). It might be
  worth seeing how much of InPlaceUpdateState fits in there.

  Walking through the fields:
  
  - PodTemplateHash — looks like this overlaps with PodCliqueUpdateProgress.PodTemplateHash and the grove.io/pod-template-hash Pod label.
  - PodCliqueSetGenerationHash — also seems to overlap with PodCliqueUpdateProgress.PodCliqueSetGenerationHash.
  - UpdateStartedAt — PodCliqueUpdateProgress.UpdateStartedAt already carries this.
  - ContainerImages — could this be derived from the desired Pod spec at reconcile time? Grove already builds the desired Pod for eligibility detection (lines 180–189), so it should be available without persisting it.
  - LastContainerStatuses — this feels like the only piece that's genuinely new. And even here, it might be avoidable: comparing containerStatuses[*].image against the desired image directly would skip the snapshot step (raised in a separate comment on completion detection).

My main worry isn't this specific type — it's the precedent. If every new update strategy gets its own <Strategy>UpdateState struct, we'll end up with parallel state shapes that duplicate hash/timestamp fields and drift over time. Grove already chose the opposite pattern in PodCliqueUpdateProgress (one shape, optional sub-fields per strategy). It would be good to stay on that path here.

docs/proposals/292-in-place-pod-image-update/README.md
oleg-kushniriov: Reconsider modeling this as a new strategy — it's a rolling update, just by restart.
  
  At the orchestration level, InPlaceIfPossible and RollingRecreate are the same loop: pick the next out-of-hash Pod, make it converge to the target spec, wait for ready, repeat. They differ only in the per-Pod mechanism (patch + kubelet restart vs delete + recreate). OnDelete is genuinely different because Grove does nothing automatic — that's the natural cleavage line for the
  top-level enum.

  Treating InPlaceIfPossible as a sibling of RollingRecreate mis-classifies it. It's a knob on RollingRecreate, not a peer.

  The K8s precedent fits cleanly. StatefulSet.spec.updateStrategy is { type: RollingUpdate|OnDelete, rollingUpdate: {...} }. We can mirror it directly, or use a single field on the existing struct:

  type PodCliqueSetUpdateStrategy struct {
      Type            UpdateStrategyType  `json:"type,omitempty"`
      // PodUpdatePolicy controls how Grove updates each Pod when Type=RollingRecreate.
      // Ignored for OnDelete.
      PodUpdatePolicy PodUpdatePolicy     `json:"podUpdatePolicy,omitempty"` // Recreate (default) | InPlaceIfPossible
  }

  spec:
    updateStrategy:
      type: RollingRecreate
      podUpdatePolicy: InPlaceIfPossible

  Benefits:
  
  - The "rolling, by restart" framing is explicit in the API.
  - The top-level enum stays tight (RollingRecreate, OnDelete) — no proliferation as we add more per-Pod mechanisms later (e.g. in-place resource resize when K8s graduates it; we'd add another PodUpdatePolicy value, not another strategy).
  - Combined with dropping InPlaceOnly (separate comment), the API collapses from 4 strategies in the current proposal to 2 strategies + 1 policy field — same expressive power, no enum bloat, no per-strategy state types.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+- Pod condition `grove.io/InPlaceUpdateReady`
+
+The condition is injected as a Pod readiness gate for new Pods created when the selected update strategy is `InPlaceIfPossible` or `InPlaceOnly`.
+
+### Completion Detection
+
+An in-place Pod update is complete when:
+
+- The Pod still has the target in-place update state.
+- For each updated container, the current `status.containerStatuses[name].imageID` differs from the image ID recorded before the patch.
```
oleg-kushniriov: - Mutable tags / re-pushed digests. If a user bumps a tag that resolves to the same digest (or no digest change), containerStatuses[*].imageID never changes and Grove waits forever. Compare against the desired image string on the containerStatus rather than against a snapshotted prior imageID — or supplement with a timeout/condition.
  - Snapshot race. LastContainerStatuses is taken before patching. If kubelet restarts the container between snapshot and patch for an unrelated reason (OOMKilled, liveness probe failure), the post-patch imageID may differ even though the new image never actually ran. Consider comparing the current image spec on containerStatus to the desired image.

docs/proposals/292-in-place-pod-image-update/README.md
oleg-kushniriov: "Existing Pods without the readiness gate are not patched in place." That means a user who flips an existing PCS from RollingRecreate to InPlaceIfPossible gets a full RollingRecreate on the next image bump (no Pods carry the gate yet). Worse, with InPlaceOnly they get a permanently blocked update with zero progress.

Does Grove auto-inject the gate on next reconcile, or only at Pod creation? If the latter, the strategy effectively only takes effect on the second update after opt-in.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+1. `PodCliqueSet` initializes `status.updateProgress` with the target generation hash.
+2. `PodClique` initializes or resets `status.updateProgress` with the target Pod template hash.
+3. The Pod component lists existing Pods and identifies Pods whose `grove.io/pod-template-hash` does not match the target hash.
+4. The Pod component updates one eligible ready Pod at a time.
+5. Old non-ready Pods are not patched in place. `InPlaceIfPossible` recreates them; `InPlaceOnly` blocks until they become suitable or are manually handled.
+6. The PodClique update completes when all Pods have the target template hash and the current minAvailable condition is satisfied.
+
+### PodCliqueScalingGroup Flow
+
+For PodCliqueScalingGroups, Grove continues to select one PodCliqueSet replica for update at a time.
```
oleg-kushniriov: PCSG one-at-a-time ordering — is the motivating invariant still load-bearing for in-place?
                                                                                                                                                                                                                                                                                                                                                                           
  Today's serial pacing exists to scope gang-termination suspension to a single PCSG during recreate. While a PCSG is updating, Grove forces its MinAvailableBreached to Unknown (podcliquescalinggroup/reconcilestatus.go:219-235) so the inevitable scheduledReplicas dip from deleting Pods doesn't trigger gang termination of the whole PodGang. Updating one replica at a time
  (rollingupdate.go:159-164, gated in podcliquescalinggroup/reconcilespec.go:91-98) keeps that suspension narrowly scoped.

  In-place patching doesn't trip the invariant: the Pod stays scheduled, only the readiness gate flips False. MinAvailableBreached is derived from scheduledReplicas, not Ready count, so the signal that motivated serialization is never at risk.

  If that's right, in-place inherits recreate's pacing without recreate's reason. Worth either:

  1. Stating an in-place-specific reason for keeping it serial (happy to be corrected), or
  2. Allowing parallel in-place updates across PCSG replicas, ideally behind a concurrency knob.

  Removing the artificial serialization could make in-place rollouts meaningfully faster on top of the per-Pod savings.

docs/proposals/292-in-place-pod-image-update/README.md
```diff
@@ -0,0 +1,327 @@
...
+  - [PodCliqueScalingGroup Flow](#podcliquescalinggroup-flow)
+  - [Status and Conditions](#status-and-conditions)
+  - [Monitoring](#monitoring)
+  - [Test Plan](#test-plan)
+    - [Unit Tests](#unit-tests)
+    - [Integration Tests](#integration-tests)
+    - [E2E Tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+<!-- /toc -->
+
```
oleg-kushniriov, line 31: Not covered: user patches updateStrategy.type while an in-place update is in progress, or while InPlaceOnly is in Blocked state. What happens to Pods already carrying InPlaceUpdateState annotation, the InPlaceUpdateReady=False gate, and an outdated template hash label? Add this to the test plan and design details.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Thanks, this makes sense.

The reason I originally proposed `IfPossible` is based on a platform/CD use case, similar to the way OpenKruise exposes in-place update behavior.

In many environments, the update policy is configured at Day 0 by the platform or CD system owner. For Day 1 updates, the CD pipeline or application owner usually only changes workload fields, such as the image, and should not need to decide the low-level update mechanism on every rollout.

For that use case, `IfPossible` is useful as a platform-level optimization policy: prefer in-place updates whenever the change is eligible, to avoid resource preemption, rescheduling, and gang scheduling overhead; fall back to recreate when the change is not eligible.

That said, I agree the current API shape can be improved. I’ll clarify this Day0/Day1 CD-platform use case in the GREP, and I’ll also revisit whether this should be modeled as a top-level strategy or as a per-Pod update policy under the rolling update strategy.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: I think native Pod readiness will eventually remove the Pod from EndpointSlices, but that happens after kubelet has already started processing the image change and the container restart is underway.

The reason for a Grove-managed readiness gate is pre-update draining: mark the Pod not ready before patching the image, wait for that readiness change to propagate, and then patch `spec.containers[*].image`. This reduces the window where traffic can still reach a container that is about to restart.

That said, the GREP should make this clearer. I’ll update it to describe the readiness gate as a pre-drain mechanism, and clarify that it reduces disruption but does not guarantee zero impact by itself.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: You are right that Grove cannot guarantee zero traffic impact by itself, and users still need to configure enough replicas above `minAvailable` to maintain service availability.

What I meant here is narrower: with a Grove-managed readiness gate, Grove can try to drain the Pod before patching the image. The flow would be: mark the Pod not ready, wait for it to be removed as a ready Service endpoint, and only then patch `spec.containers[*].image`. This is intended to prevent new traffic from being routed to a Pod whose container is about to restart.

I’ll update the wording to avoid implying a full zero-impact guarantee, and describe this as best-effort pre-update draining that depends on readiness propagation and the user’s availability settings.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: I added more context in the discussion below, but the short version is that the main use case is a platform/CD workflow.

The update policy can be configured at Day 0 by the platform or CD owner. For Day 1 changes, the application owner or CD pipeline usually only changes workload fields such as the image, and should not need to choose the low-level update mechanism for each rollout.

By “not eligible”, I mean changes that cannot be safely handled by patching `spec.containers[*].image` only, such as container set changes, init container changes, resources, env/command/args, volumes, scheduling fields, or workloads that require all containers to restart and re-establish startup order.

I’ll make this clearer in the GREP.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Agreed. A separate full `InPlaceUpdateState` is not the right shape.

Most of the fields are already covered by `PodCliqueUpdateProgress` or can be derived during reconciliation. I’ll revise the GREP to reuse the existing progress model and only add minimal in-place-specific state if needed, such as the currently updating Pod and its phase.

I’ll also remove the dependency on snapshotted previous container statuses, since completion detection should compare against the desired image instead.

docs/proposals/292-in-place-pod-image-update/README.md
renormalize: Thanks for the reply @AllenXu93.

Okay, I think I understand what your intention was. However, I want to add a few more observations based on my current understanding:

1. The gate will prevent new traffic from reaching the pod that is gated, and will also enable the pod to respond to inflight requests. However, what would the behavior be when the application does not terminate existing connections, which were made before the gate was added? The connection was opened -> the gate was added -> pod is `NotReady` -> pod updated -> connection never closed. We would have abruptly terminated connections, and we can't really control this from the operator either. Should we then allow users to configure this pod-update-delay (pre-drain as you called it) time based on how long they expect the application to keep connections open?
2. The readiness gate will only work when all clients use the service abstraction only (which relies on endpointslices). This is not the case in Grove, since we publish IP addresses of pods through `hostname`s (headless service per PCS replica), specifically for pod-pod communication. This is used for KV cache transfer.

I obviously see validity in the gate after your explanation, but I'm just curious what the approach to solve point 1 above would be, perhaps there's no real way to solve it and we'll have to live with it?

Pod readiness gates would not work for point 2. Please feel free to correct me.

docs/proposals/292-in-place-pod-image-update/README.md
renormalize: Thanks for the explanation. I wonder if this is something we should directly support at the level of grove, or perhaps at an inference framework level like dynamo. I don't have a strong opinion on this for now, let's discuss this with the maintainers.

docs/proposals/292-in-place-pod-image-update/README.md
renormalize: Any ideas on this @AllenXu93?

docs/proposals/292-in-place-pod-image-update/README.md
renormalize: Any opinions, @AllenXu93?

AllenXu93: > Thanks for the reply @AllenXu93.
> 
> Okay, I think I understand what your intention was. However, I want to add a few more observations based on my current understanding:
> 
> 1. The gate will prevent new traffic from reaching the pod that is gated, and will also enable the pod to respond to inflight requests. However, what would the behavior be when the application does not terminate existing connections, which were made before the gate was added? The connection was opened -> the gate was added -> pod is `NotReady` -> pod updated -> connection never closed. We would have abruptly terminated connections, and we can't really control this from the operator either. Should we then allow users to configure this pod-update-delay (pre-drain as you called it) time based on how long they expect the application to keep connections open?
> 2. The readiness gate will only work when all clients use the service abstraction only (which relies on endpointslices). This is not the case in Grove, since we publish IP addresses of pods through `hostname`s (headless service per PCS replica), specifically for pod-pod communication. This is used for KV cache transfer.
> 
> I obviously see validity in the gate after your explanation, but I'm just curious what the approach to solve point 1 above would be, perhaps there's no real way to solve it and we'll have to live with it?
> 
> Pod readiness gates would not work for point 2. Please feel free to correct me.

Thanks, this makes sense.

After looking into Dynamo's routing/discovery path more carefully, I agree that a Grove-managed readiness gate should not be a required part of this proposal.

Dynamo may use Kubernetes readiness-backed discovery in some paths, but routing is not purely Kubernetes Service endpoint selection. Also, readiness gates do not solve already-established or long-lived connections. Those should be handled by the framework/application graceful shutdown behavior after SIGTERM, within terminationGracePeriodSeconds.

I’ll update the GREP to remove the readiness gate / pre-drain mechanism and keep the proposal focused on orchestrating eligible image-only in-place updates.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: I removed the readiness gate / pre-drain mechanism from the GREP. 
If we want Grove-level readiness/drain integration later, I think it should be discussed as a separate topic instead of being coupled to this in-place image update proposal.

AllenXu93: Good point. This migration problem is another reason not to make a Grove-managed readiness gate a required part of this proposal.

In the latest GREP update, I removed the readiness gate / pre-drain requirement entirely. Existing Pods no longer need a gate to become eligible for in-place image updates; eligibility is now based on whether the effective Pod change is limited to regular container image changes and whether the Pod is otherwise safe to patch.

If we later want Grove-managed readiness/drain integration, I think that should be discussed as a separate topic with its own migration behavior.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: I agree with this framing.

After considering the feedback here and the related comments, I think the API can be simplified further. In-place image update should not be modeled as a new top-level update strategy. It is still part of the automatic rolling update flow; only the per-Pod update mechanism changes.

I’ll update the GREP to remove both `InPlaceIfPossible` and `InPlaceOnly` as strategy types. Instead, the proposal will keep `RollingRecreate` and `OnDelete` as the top-level strategies, and add an opt-in setting under rolling update, for example:

```yaml
updateStrategy:
  type: RollingRecreate
  rollingUpdate:
    inPlaceUpdate: true
```

When this is enabled, Grove will decide per Pod whether the effective change is eligible for an in-place image patch. If it is eligible, Grove patches spec.containers[*].image; otherwise it falls back to the existing recreate path.

A stricter “in-place only / deny otherwise” mode can be discussed separately if there is a concrete need for it. Does this API direction sound reasonable to you?

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: I think this should be addressed by simplifying the API.

I’ll remove `InPlaceIfPossible` and `InPlaceOnly` as strategy types. Instead, in-place image update will be an opt-in setting under `RollingRecreate`, e.g. `rollingUpdate.inPlaceUpdate: true`. Grove will then decide per Pod whether the change is eligible for an in-place image patch; otherwise it falls back to the existing recreate path.

```yaml
updateStrategy:
  type: RollingRecreate
  rollingUpdate:
    inPlaceUpdate: true
```

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: I agree with this. The original gang-scheduling rationale for serial PCSG updates does not apply in the same way to in-place image updates, because the Pods remain scheduled.

So I think the proposal should not hard-code one-at-a-time ordering for in-place updates. Instead, it should follow a Deployment-style rolling update control and let users bound concurrency with `maxUnavailable`.

I’ll update the GREP to add a `rollingUpdate` block under `RollingRecreate`, with `inPlaceUpdate: true` and `maxUnavailable` controlling how many Pods may be patched and not yet converged at the same time.

For example:

```yaml
updateStrategy:
  type: RollingRecreate
  rollingUpdate:
    inPlaceUpdate: true
    maxUnavailable: 1
```
This keeps the default conservative while allowing users to opt into parallel in-place updates by increasing maxUnavailable.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Updated.

Completion detection now compares each updated container status against the desired image from the target Pod spec and requires the container to be ready before advancing the Grove template hash label. `imageID` is no longer the primary convergence signal and is only useful for diagnostics.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Updated the GREP to follow this direction.

The proposal now reuses the existing `PodCliqueStatus.UpdateProgress` model instead of adding a separate `InPlaceUpdateState` shape. With the latest API update, in-place-specific progress tracks only the bounded set of Pods currently being patched in place, capped by `rollingUpdate.maxUnavailable`.

Target images are derived from the desired Pod during reconciliation, and previous container status snapshots are no longer persisted.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Updated the GREP to cover strategy/policy changes during an active update.

With the latest API shape, `InPlaceOnly` has been removed, so there is no separate blocked state for an in-place-only strategy. The remaining case is a user changing `updateStrategy` or `rollingUpdate` settings while an in-place update is active.

The proposed behavior is: Pods already patched in place are not rolled back; Grove waits for them to converge before advancing their template hash. Pods not yet patched are reconciled using the latest `updateStrategy` and `rollingUpdate` settings. If in-place update is disabled mid-update, Grove stops starting new in-place patches and remaining outdated Pods use the existing recreate path.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Updated the GREP to describe this.

In-place-specific progress is stored under `PodCliqueStatus.UpdateProgress`, but it is not an unbounded per-Pod list. With the latest API update, it tracks only the bounded set of Pods currently being patched in place, capped by `rollingUpdate.maxUnavailable`.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Follow-up after the API update: the proposal no longer exposes `InPlaceIfPossible` as a strategy type. In-place image update is now an opt-in setting under `RollingRecreate`, and Grove falls back to the existing recreate path when the effective Pod change is not eligible.

docs/proposals/292-in-place-pod-image-update/README.md
AllenXu93: Follow-up clarification: this completion check assumes the desired Pod spec contains a changed image reference for the updated container.

If the same tag is re-pushed to a different digest without changing `spec.containers[name].image`, Grove cannot reliably detect or trigger that as an in-place image update because Kubernetes does not expose it as a Pod spec change. For deterministic rollouts, users should use unique tags or digest-pinned image references.

So the convergence check is: after Grove patches the Pod image field, wait until `status.containerStatuses[name].image` matches the desired image reference from the target Pod spec and the container is running and ready. `imageID` remains diagnostic only.

danbar2:
Hi @AllenXu93,

Thank you again for putting this proposal together and for the thoughtful discussion in this thread - the design work here is genuinely appreciated.

We’ve discussed this internally and, at this point, in-place Pod image updates aren’t a required capability on the Grove roadmap. Given that, I’m going to close this PR for now. This isn’t a reflection on the quality of the proposal - it’s purely a matter of current priorities and scope.

If new requirements or user stories emerge that call for this capability, the right path forward would be to bring it to the Arch & Design meeting so we can align on direction (including how it relates to the ongoing MVU work) before investing further in the design and implementation.

Thanks again for your contribution and for engaging so constructively - we’d welcome your participation in that discussion if and when it happens.

PR #608, opened 2026-05-12
add soak-churn scale test

### LINKED ISSUE #607: Add long-running soak / churn benchmark to e2e scale tests

Add a long-running benchmark under `operator/e2e/tests/scale/` that
boots a small `PodCliqueSet` (~100 pods) and then drives repeated
scale-up / scale-down cycles against it over an extended duration.
The goal is to surface bugs that only appear after many incremental
reconciles — leaks, monotonically growing slices, gradually drifting
counters, finalizer pile-ups — that single-shot benchmarks cannot see.

**Scope (single test, parameterized):**

| Parameter | Default | Env override | Rationale |
|---|---|---|---|
| Base PCS replicas | 25 (50 pods) | `SOAK_BASE` | Small enough to fit dev clusters; large enough to exercise multi-child reconciles. |
| Peak PCS replicas | 50 (100 pods) | `SOAK_PEAK` | 2x amplitude per cycle. |
| Cycles | 10 | `SOAK_CYCLES` | Enough per-cycle iterations to surface slow drift; bounded total runtime. |
| Per-cycle hold | ~30 s | — | Lets watch events flush and pprof capture overlap consecutive no-op reconciles. |
| Worker nodes | 30 kwok nodes | — | Same fixture as scale-up/down tests. |
| Timeout | 45 min | — | Soft cap; expected runtime ~25–30 min. |

**Deliverables:**

- `operator/e2e/tests/scale/soak_test.go`.
- `operator/e2e/yaml/soak-churn.yaml` (PCS sized at base; ~50 pods).
- Reuse `runScaleTest`, `ScalePCS`, `PodsCreatedCondition`,
  `TimerCondition`. The same scale-down milestone gap noted in the
  scale-up-down issue applies here (`PodsAtCountCondition` is
  prerequisite).
- Final-check phase asserting no leaks/orphans/unbounded counters at
  the end of the run.

**Cycle shape (executed N times):**

1. **scale-up** — patch `spec.replicas` base → peak. Wait until live
   pod count reaches `peak * podsPerReplica`.
2. **hold-peak** — 30 s window.
3. **scale-down** — patch `spec.replicas` peak → base. Wait until live
   pod count reaches `base * podsPerReplica`.
4. **hold-base** — 30 s window.

### Why is this needed?

Today's scale benchmarks run for minutes and measure a single event
(deploy, resize, delete). They are good at catching gross regressions
in per-event throughput, but they pass cleanly on a controller that is
slowly leaking heap, accumulating stale entries in an informer index,
or letting one status field drift by an item per reconcile.

oleg-kushniriov:
#### What type of PR is this?

  /kind feature
  
  Adds a long-running soak / churn benchmark under `operator/e2e/tests/scale/` that boots a small `PodCliqueSet` (~50 pods) and drives repeated scale-up / scale-down cycles against it over an extended duration. The goal is to surface bugs that single-shot benchmarks cannot see — slow
   leaks, monotonically growing status fields, gradually drifting counters, finalizer pile-ups — by accumulating many incremental reconciles against the same controller instance.

  The test is **opt-in only**. It's gated by a separate build tag (`//go:build e2e && soak`) and is invisible to `make run-e2e` and `make run-scale-test`. Default runtime is ~30 minutes, too long for routine CI; it runs on demand when investigating a regression or capturing a
  baseline.

  **Single test, parameterized by env vars**

  | Parameter | Default | Env override |
  |---|---|---|
  | Base PCS replicas | 25 (50 pods) | `SOAK_BASE` |
  | Peak PCS replicas | 50 (100 pods) | `SOAK_PEAK` |
  | Cycles | 10 | `SOAK_CYCLES` |
  | Per-cycle hold | 30 s | — |
  | Worker nodes | 30 kwok | — |
  | Timeout | 60 min | — |

  **Cycle shape (executed N times)**

  1. `scale-up-cN` — patch `spec.replicas` base → peak, wait for `peak-pods-ready`.
  2. `hold-peak-cN` — 30 s window for watch events to flush.
  | Cycles | 10 | `SOAK_CYCLES` |
  | Per-cycle hold | 30 s | — |
  | Worker nodes | 30 kwok | — |
  | Timeout | 60 min | — |

  **Cycle shape (executed N times)**

  1. `scale-up-cN` — patch `spec.replicas` base → peak, wait for `peak-pods-ready`.
  2. `hold-peak-cN` — 30 s window for watch events to flush.
  3. `scale-down-cN` — patch `spec.replicas` peak → base, wait for `base-pods-restored` (live count drops to ≤ base).
  4. `hold-base-cN` — 30 s window.

  **What's in the PR**

  - `operator/e2e/tests/scale/soak_test.go` — `Test_SoakChurn`, gated `//go:build e2e && soak`. Loads env config, drives N cycles, runs final-check as the last timeline phase.
  - `operator/e2e/yaml/soak-churn.yaml` — PCS at base size (25 replicas × 2 pods/clique = 50 pods).
  - `operator/e2e/measurement/condition/pod.go` — adds `PodsAtCountCondition` (fires when live pod count drops to ≤ ExpectedCount). Required for the scale-down leg of each cycle since the existing `PodsCreatedCondition` is `≥`-only and would fire immediately.
  - `operator/Makefile` — `make run-soak-test` target with `SOAK_BASE` / `SOAK_PEAK` / `SOAK_CYCLES` / `DIAG_DIR` env-var forwarding.
  - `operator/e2e/tests/scale/soak_churn.md` — design doc with motivation, decisions, and the placement reasoning for the final-check phase.

  **How to run**

  ```sh
  cd operator
  make scale-cluster-up E2E_CREATE_FLAGS="--set kwok.nodes=30"

  # Quick smoke (~5 min) — 3 cycles
  make run-soak-test SOAK_CYCLES=3

  # Full default run (~30 min)
  make run-soak-test

  # Baseline capture for cross-revision comparison
  make run-soak-test DIAG_DIR=/tmp/soak-rev-A

  make scale-cluster-down
  ```

  Results land in `<DIAG_DIR>/SoakChurn/<runID>/scale-test-results.json` with per-phase timings — diff with `jq` across revisions, diff pprof bundles (when Pyroscope is on) with `go tool pprof -base A.pb.gz B.pb.gz`. No comparison logic in the test itself — keeps it deterministic, no
   baselines committed to the repo.

  Fixes #607


operator/e2e/tests/scale/soak_churn.md
danbar2, line 1: I dont think this file is needed

operator/Makefile
```diff
@@ -222,6 +222,24 @@ run-scale-test:
 	@echo "> Running scale tests..."
 	@cd e2e && go test -count=1 -tags=e2e ./tests/scale/... -v -timeout 40m
 
+# Run the soak / churn benchmark against an existing cluster.
+# Gated behind the `soak` build tag so it does NOT run as part of
+# run-e2e or run-scale-test. Long-running (~30 min default).
```
danbar2, line 227: why not to run it in scale tests?

operator/e2e/tests/scale/soak_churn.md
oleg-kushniriov: removed

operator/Makefile
oleg-kushniriov: The initial reason this test was separated was its long runtime — it's designed to surface issues that only appear after many cycles. But you're right, adding it to CI is probably the correct path. We'd need to bump the workflow timeout (current job is 60 min, scale test alone uses 30 min) to fit soak alongside it, since scale tests run serially.

danbar2 (dismissed):

operator/e2e/measurement/condition/pod.go
```diff
@@ -135,3 +135,38 @@ func (c *PodsReadyCondition) Met(ctx context.Context) (bool, error) {
 func (c *PodsReadyCondition) Progress(_ context.Context) string {
 	return fmt.Sprintf("%d/%d pods ready", c.lastReady, c.ExpectedCount)
 }
+
+// PodsAtCountCondition fires when the live pod count drops to ExpectedCount or
+// below. Intended for scale-down milestones where PodsCreatedCondition (≥-only)
+// would fire immediately because the starting count already exceeds the target.
+// Using ≤ rather than == makes the condition robust to a transient overshoot
+// during cascade-delete where two consecutive polls might skip the exact target.
+type PodsAtCountCondition struct {
```
shayasoolin: The name of this struct better reflect the fact that the condition only serves the case of scale-down, but not scale-up.

operator/e2e/tests/scale/soak_test.go
```diff
@@ -0,0 +1,257 @@
...
+		ActionFn: func(ctx context.Context) error {
+			return runSoakFinalChecks(ctx, tc, cfg)
+		},
+	})
+}
+
+// runSoakFinalChecks asserts end-state invariants. Returns an error rather
+// than calling t.Fatalf so it can be invoked from a phase ActionFn; the
+// tracker propagates the error into a test failure.
+func runSoakFinalChecks(ctx context.Context, tc *testctx.TestContext, cfg soakConfig) error {
```
shayasoolin: I feel this function should check more things. The time spent on the test is long, and we better maximize the insights we can have from the test results.

Not sure everything below will make sense, we can discuss if you want, but here's an ai-generated list of potential checks to have here:

  1. Expected child object graph
      - List PodCliques and PodCliqueScalingGroups with app.kubernetes.io/part-of=soak-churn.
      - Assert exact expected counts after final scale-down.
      - Assert no object has grove.io/podcliqueset-replica-index >= SOAK_BASE.
      - For this fixture, likely expect cfg.base PodCliques and 0 PodCliqueScalingGroups.
  2. Owner reference consistency
      - Every final PodClique should be owned by the PCS.
      - Every final pod should be owned by an existing PodClique.
      - No pod should reference a missing/deleting PodClique.
  3. Pod identity uniqueness
      - For each PodClique, assert pod labels form exactly the expected pod index set.
      - No duplicate (podCliqueName, podIndex).
      - No pod with stale scale-out replica index.
  4. Pod health, not just count
      - Final pods should be Running.
      - Final pods should be Ready.
      - No final pod should have DeletionTimestamp.
      - No final pod should still have scheduling gates.
  5. PCS status convergence
      - pcs.Spec.Replicas == cfg.base.
      - pcs.Status.Replicas == cfg.base.
      - pcs.Status.AvailableReplicas == cfg.base.
      - pcs.Status.UpdatedReplicas == cfg.base.
      - ObservedGeneration == Generation.
      - UpdateProgress == nil or UpdateEndedAt != nil and CurrentlyUpdating empty.
  6. Child status convergence
      - For every PodClique: LastErrors empty, ObservedGeneration == Generation, ReadyReplicas == Spec.Replicas, UpdatedReplicas == Spec.Replicas.
      - If PCSGs exist in future soak fixtures: same idea for PCSG.Status.Replicas, AvailableReplicas, UpdatedReplicas, LastErrors, and update progress bounds.
  7. PodGang cleanup
      - List PodGangs for the PCS and assert the expected final set.
      - This is especially relevant because churn can leave stale scheduling artifacts even when pod count looks correct.
  8. No stuck deletions/finalizers
      - Assert no managed Pod, PodClique, PodCliqueScalingGroup, or PodGang is in deletion.
      - Existing live objects may legitimately have finalizers, but DeletionTimestamp != nil at final check is suspicious.
  9. Operator health
      - Assert Grove operator pod did not restart during the test.
      - Optionally fail if it has OOMKilled/restarted containers.

operator/Makefile
shayasoolin: If this is added to the CI (which I'm not against!), I suggest to have it run with relatively low numbers so run time won't get too long.
You wrote "scale test alone uses 30 min" but that's not what I see - the scale tests that run as part of the new CI flow take about 6 minutes, including the cluster setup and teardown.
Not sure we're talking about the same flows though. There's the scale-test CI, there's `make run-e2e` and make `run-scale-test`.

oleg-kushniriov: the runtime very much dependent on the value of SOAK_CYCLES. For 20 cycles - runtime will be 20+ minutes

operator/e2e/tests/scale/soak_test.go
oleg-kushniriov: added more validations

operator/e2e/tests/scale/soak_test.go
```diff
@@ -207,35 +264,186 @@ func addCyclePhases(tracker *measurement.TimelineTracker, tc *testctx.TestContex
...
-	// 1. Live pod count is back at base.
+func runSoakFinalChecks(ctx context.Context, tc *testctx.TestContext, cfg soakConfig, baseline *operatorBaseline) error {
 	pods, err := tc.ListPods()
 	if err != nil {
 		return fmt.Errorf("list pods: %w", err)
 	}
-	if got, want := len(pods.Items), cfg.basePods(); got != want {
-		return fmt.Errorf("live pod count = %d, want %d (potential leak)", got, want)
+
+	pcqList := &grovecorev1alpha1.PodCliqueList{}
```
shayasoolin: we use PCLQ not PCQ for pod cliques.

shayasoolin (approved):

danbar2 (approved):

PR #613, opened 2026-05-14
fix: PodCliqueSet webhook crashes on unknown scheduler name

### LINKED ISSUE #619: webhook crashes on unknown scheduler name
### What happened?

In a PodCliqueSet YAML, users set spec.template.cliques[].spec.podSpec.schedulerName to choose the scheduler backend. Valid values are "default-scheduler" and "kai-scheduler".
This field is plain corev1.PodSpec.SchedulerName (a string) — it has no CRD-level enum constraint, so the API server does not reject unknown values like "volcano" at the schema layer.

If a user mistakenly sets this to an unsupported value or typo , the validating webhook runs validateSchedulerNames, which detects the mismatch and records a validation error. However, validateSchedulerNames only accumulates errors into a field.ErrorList; it does not return early or prevent the handler from continuing. validatePodCliqueSetWithBackend is called
unconditionally immediately after.

Inside that function, GetOrDefault("kaikai-scheduler") or GetOrDefault("volcano") returns nil because the name is non-empty but not in the registry, and the subsequent backend.ValidatePodCliqueSet() dereferences nil, panicking the webhook process before it can return the recorded validation error to the user.
In short: the webhook does validate the name, but that validation does not stop the code from reaching the nil-deref site.

suggest a PR https://github.com/ai-dynamo/grove/pull/613

_No response_

### Environment

- Kubernetes version
- Grove version
- Scheduler details

weizhoublue:
#### What type of PR is this?

/kind bug

  In a PodCliqueSet YAML, users set spec.template.cliques[].spec.podSpec.schedulerName to choose the scheduler backend. Valid values are "default-scheduler" and "kai-scheduler". 
This field  is plain corev1.PodSpec.SchedulerName (a string) — it has no CRD-level enum constraint, so the API server does not reject unknown values like "volcano" at the schema layer.

 If a user mistakenly sets this to an unsupported value or typo , the validating webhook runs validateSchedulerNames, which detects the mismatch and records a validation error. However,  validateSchedulerNames only accumulates errors into a field.ErrorList; it does not return early or prevent the handler from continuing. validatePodCliqueSetWithBackend is called
  unconditionally immediately after. 

Inside that function, GetOrDefault("kaikai-scheduler") or GetOrDefault("volcano")  returns nil because the name is non-empty but not in the registry, and the subsequent  backend.ValidatePodCliqueSet() dereferences nil, panicking the webhook process before it can return the recorded validation error to the user. 

In short: the webhook does validate the  name, but that validation does not stop the code from reaching the nil-deref site.

Fixes #619 

```release-note

```

```docs

```

### DIFF
--- a/operator/internal/webhook/admission/pcs/validation/handler.go
+++ b/operator/internal/webhook/admission/pcs/validation/handler.go
@@ -134,6 +134,9 @@ func (h *Handler) validatePodCliqueSetWithBackend(ctx context.Context, pcs *v1al
 	}
 
 	backend := h.schedRegistry.GetOrDefault(schedulerName)
+	if backend == nil {
+		return fmt.Errorf("schedulerName %q is not a supported backend", schedulerName)
+	}
 	return backend.ValidatePodCliqueSet(ctx, pcs)
 }
 
--- a/operator/internal/webhook/admission/pcs/validation/handler_test.go
+++ b/operator/internal/webhook/admission/pcs/validation/handler_test.go
@@ -116,6 +116,41 @@ func TestValidateCreate(t *testing.T) {
 			expectError:   true,
 			errorContains: "failed to cast object to PodCliqueSet",
 		},
+		{
+			name: "unknown scheduler name returns error without panic",
+			obj: &grovecorev1alpha1.PodCliqueSet{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-pcs",
+					Namespace: "default",
+				},
+				Spec: grovecorev1alpha1.PodCliqueSetSpec{
+					Replicas: 1,
+					Template: grovecorev1alpha1.PodCliqueSetTemplateSpec{
+						TerminationDelay: ptr.To(metav1.Duration{Duration: 4 * time.Hour}),
+						StartupType:      ptr.To(grovecorev1alpha1.CliqueStartupTypeAnyOrder),
+						Cliques: []*grovecorev1alpha1.PodCliqueTemplateSpec{
+							{
+								Name: "test-pclq",
+								Spec: grovecorev1alpha1.PodCliqueSpec{
+									RoleName:     "test-role",
+									Replicas:     1,
+									MinAvailable: ptr.To(int32(1)),
+									PodSpec: corev1.PodSpec{
+										SchedulerName: "unknown-scheduler",
+										Containers: []corev1.Container{{
+											Name:  "test",
+											Image: "test",
+										}},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expectError:   true,
+			errorContains: "schedulerName must be an enabled scheduler backend",
+		},
 	}
 
 	for _, tt := range tests {


operator/internal/controller/podcliqueset/components/podgang/podgang_test.go
```diff
@@ -295,3 +296,25 @@ func TestBuildResource(t *testing.T) {
 func TestMirrorPCSMetadataNeverReturnsNil(t *testing.T) {
 	assert.NotNil(t, mirrorPCSMetadata(nil, nil, nil))
 }
+
+// TestGetSchedulerNameForPCSNilDefault verifies that getSchedulerNameForPCS
+// returns "" instead of panicking when the default backend is nil.
```
shayasoolin: Nit, the default backend in this test case is not nil, but an invalid (nonexistent) string.

weizhoublue: thanks, modified

shayasoolin (dismissed):
Thanks for the fix!

operator/internal/webhook/admission/pcs/validation/handler.go
```diff
@@ -134,6 +134,9 @@ func (h *Handler) validatePodCliqueSetWithBackend(ctx context.Context, pcs *v1al
 	}
 
 	backend := h.schedRegistry.GetOrDefault(schedulerName)
+	if backend == nil {
+		return nil
```
sanjaychatterjee: If backend is nil, why is this returning `nil` error?

sanjaychatterjee (commented):
@kangclzjc can you please review this PR? Please ensure the validations are correct.

operator/internal/controller/podcliqueset/components/podgang/podgang.go
```diff
@@ -260,7 +260,10 @@ func (r _resource) getSchedulerNameForPCS(pcs *grovecorev1alpha1.PodCliqueSet) s
 			return c.Spec.PodSpec.SchedulerName
 		}
 	}
-	return r.schedRegistry.GetDefault().Name()
+	if def := r.schedRegistry.GetDefault(); def != nil {
+		return def.Name()
```
kangclzjc: GetDefault can never return nil since we always enable default-scheduler backend. It doesn't need to validate whether it's nil.

operator/internal/webhook/admission/pcs/validation/handler.go
kangclzjc: Yes, we should return an error like backend nil error or schedulerName not valid. And this is a good catch here if use schedulerName is typo like `volllcano`. We should check it here.

kangclzjc:
@weizhoublue it's a good catch, I left some comments here. And could you also create an issue, and link it in `Which issue(s) this PR fixes:` here.

operator/internal/controller/podcliqueset/components/podgang/podgang.go
weizhoublue: You're right that in production runtime, GetDefault() is guaranteed non-nil by the config validation in validateSchedulerConfiguration (which requires DefaultProfileName to match a configured profile, and default-scheduler is always required)

operator/internal/webhook/admission/pcs/validation/handler.go
weizhoublue: thanks for pointing out

weizhoublue:
> @weizhoublue it's a good catch, I left some comments here. And could you also create an issue, and link it in `Which issue(s) this PR fixes:` here.

sure,  https://github.com/ai-dynamo/grove/issues/619 was created

kangclzjc (approved):

sanjaychatterjee (approved):
Approved. Thanks for the contribution @weizhoublue!

PR #614, opened 2026-05-14
Gang-termination breach formula rework + E2E scenarios (GT-1 through GT-6)

danbar2:
Follow-up to #561 (which fixed the initial cordon-then-kill regression) — picks up the open
questions that PR left behind. Refs #277.

## Summary

Two related changes, one commit each:

1. **`c8820c9` Rework MinAvailableBreached semantics.** Drops the `scheduledReplicas == 0`
   suppression that #561 introduced as a churn-loop guard, and rewrites the PCSG-level
   breach formula to match the spec-implied definition.
2. **`7f151af` + `294882b` Add gang-termination E2E scenarios GT-1 through GT-6.**

### Drop the scheduled==0 suppression
The previous suppression made it impossible to recover when a PodClique lost every
scheduled pod — gang termination never fired even after the workload had once been
healthy. The new rule: `scheduledReplicas < MinAvailable` always breaches.

The natural debounce is `TerminationDelay` (default 4h). During a normal startup the
breach flickers True briefly and resolves; a workload still below MinAvailable past
`TerminationDelay` is genuinely stuck and recycling it gives the scheduler a fresh
PodGang against the current cluster state.

### Rework PCSG breach formula
Replaces \`availableReplicas = scheduledReplicas - breachedReplicas\` with
\`notInBreachReplicas = existingReplicas - breachedReplicas\`. The previous formula
double-counted: a PCSG replica that failed to schedule was excluded from
\`scheduledReplicas\` AND counted again in \`breachedReplicas\`, flipping the PCSG into
breach unnecessarily and causing PCS-level gang termination to fire when only the
PCSG-replica-scoped restart was warranted.

Matches the canonical definition:
- A PCSG **replica** is in breach if at least one of its PodCliques is in breach.
- A PCSG **replica** in breach > TerminationDelay → PCSG-scoped restart (just that
  replica's PCLQs).
- The **PCSG itself** is in breach when (replicas not in breach) < MinAvailable.
- Standalone PCLQ in breach → PCS-level restart.
- PCSG-level in breach → PCS-level restart.

Unit tests rewritten around the new formula; sanity-pin cases for "1 of 2 replicas
breached with MinAvailable=1 — not in breach (PCSG-scoped restart only)" and "both
replicas breached — escalates to PCS-level" capture the GT-4 / GT-5 expectations.

## E2E scenarios (commits `7f151af` + `294882b`)

| Test | Workload | What it exercises |
|---|---|---|
| GT-1 | WL1 (full replicas) | Cordon + kill a PCS-owned pod → full-workload gang term |
| GT-2 | WL1 (full replicas) | Cordon + kill a PCSG-owned pod → full-workload gang term |
| GT-3 | WL2 (min replicas) | Incremental kill of pc-a; first kill no-op, second triggers term |
| GT-4 | WL2 (min replicas) | Multi-stage PCSG-owned: PCSG-scoped restart then full term |
| GT-5 | WL5 (pc-c min=3) | Kill all of one PCSG replica's pc-c → that PCSG replica only |
| GT-6 | WL2 | Kill a scaled-PodGang pod → no PCS-level term (pc-a survives) |

All six pass locally against an operator running with this PR's changes.

## Test plan

- [x] \`go test ./...\` clean (operator + scheduler/api + api modules)
- [x] \`go vet ./...\` clean
- [x] gofmt clean
- [x] E2E: \`make run-e2e TEST_PATTERN=Test_GT\` — 6 PASS in ~9 minutes against k3d/Kai
- [x] Recommend a CI run on \`prod-grove-e2e-v1\` before merge

## Out of scope (for a follow-up)

The original test plan also lists GT-7/9/10/11 (crash-loop variants), GT-12 (readiness
probe failure), GT-13 (no gang term during rolling update), and GT-14 (recovery before
TerminationDelay). These are not in this PR — each needs new test infrastructure that
doesn't exist in \`operator/e2e\` today:

- GT-7/9/10/11 + GT-12: pod-exec from Go (\`client-go/tools/remotecommand\` + SPDY) plus
  workload variants with a toggleable liveness/readiness probe keyed on a file in an
  emptyDir volume.
- GT-13: a \`PatchPCSTemplate\` helper in \`testctx\` so a test can trigger a rolling
  update mid-run.
- GT-14 actually works against this PR's operator changes, but its assertion shape is
  different enough from GT-1..6 that I'd rather pull it in once the infra above lands.

### DIFF
--- a/operator/api/common/constants/constants.go
+++ b/operator/api/common/constants/constants.go
@@ -145,8 +145,6 @@ const (
 	ConditionReasonInsufficientScheduledPods = "InsufficientScheduledPods"
 	// ConditionReasonSufficientScheduledPods indicates that the number of scheduled pods in the PodClique greater or equal to PodCliqueSpec.MinAvailable.
 	ConditionReasonSufficientScheduledPods = "SufficientScheduledPods"
-	// ConditionReasonInsufficientScheduledPCSGReplicas indicates that the number of scheduled replicas in the PodCliqueScalingGroup is below the PodCliqueScalingGroupSpec.MinAvailable.
-	ConditionReasonInsufficientScheduledPCSGReplicas = "InsufficientScheduledPodCliqueScalingGroupReplicas"
 	// ConditionReasonScheduledReplicasBelowMinAvailable indicates that scheduledReplicas is below MinAvailable but greater than zero.
 	ConditionReasonScheduledReplicasBelowMinAvailable = "ScheduledReplicasBelowMinAvailable"
 	// ConditionReasonInsufficientAvailablePCSGReplicas indicates that the number of ready replicas in the PodCliqueScalingGroup is below the PodCliqueScalingGroupSpec.MinAvailable.
--- a/operator/api/common/constants/constants.go
+++ b/operator/api/common/constants/constants.go
@@ -120,11 +120,14 @@ const (
 	// ConditionTypePodCliqueScheduled indicates that the PodClique has been successfully scheduled.
 	// This condition is set to true when number of scheduled pods in the PodClique is greater than or equal to PodCliqueSpec.MinAvailable.
 	ConditionTypePodCliqueScheduled = "PodCliqueScheduled"
-	// ConditionTypeGangTerminationInProgress indicates that gang termination has fired for this resource and the
-	// gang termination is still in flight. The condition is set when gang termination deletes the PodCliques
-	// (either PCSG-replica scoped or PCS-level) and is cleared when MinAvailableBreached transitions back to
-	// False (workload recovered). While the condition is True, further gang-termination action is suppressed —
-	// at most one fire per breach episode, regardless of how long the workload stays below MinAvailable.
+	// ConditionTypeGangTerminationInProgress indicates that PCS-level gang termination has fired for this
+	// PodCliqueScalingGroup and is still in flight. It is set on the PCSG when the PCS-level handler deletes
+	// the PodCliques of the whole PCS replica, and is cleared when the PCSG's MinAvailableBreached transitions
+	// back to False (recovered). While it is True, further PCS-level gang termination for the PCS replica is
+	// suppressed — at most one fire per breach episode, regardless of how long the workload stays below
+	// MinAvailable. This is a PCS-level-only mechanism: the PCSG-replica-scoped recycle path does not use this
+	// flag and instead breaks its own re-fire loop via WasPCLQEverScheduled, since a freshly recreated
+	// PodClique has never been scheduled and is therefore excluded from the breached set.
 	// Its only Reason is ConditionReasonGangTerminationActive.
 	ConditionTypeGangTerminationInProgress = "GangTerminationInProgress"
 	// ConditionTopologyLevelsUnavailable indicates that the required topology levels defined on a PodCliqueSet for topology-aware scheduling are no longer available.
--- a/operator/api/common/constants/constants.go
+++ b/operator/api/common/constants/constants.go
@@ -120,6 +120,13 @@ const (
 	// ConditionTypePodCliqueScheduled indicates that the PodClique has been successfully scheduled.
 	// This condition is set to true when number of scheduled pods in the PodClique is greater than or equal to PodCliqueSpec.MinAvailable.
 	ConditionTypePodCliqueScheduled = "PodCliqueScheduled"
+	// ConditionTypeGangTerminationInProgress indicates that gang termination has fired for this resource and the
+	// gang termination is still in flight. The condition is set when gang termination deletes the PodCliques
+	// (either PCSG-replica scoped or PCS-level) and is cleared when MinAvailableBreached transitions back to
+	// False (workload recovered). While the condition is True, further gang-termination action is suppressed —
+	// at most one fire per breach episode, regardless of how long the workload stays below MinAvailable.
+	// Its only Reason is ConditionReasonGangTerminationActive.
+	ConditionTypeGangTerminationInProgress = "GangTerminationInProgress"
 	// ConditionTopologyLevelsUnavailable indicates that the required topology levels defined on a PodCliqueSet for topology-aware scheduling are no longer available.
 	// This can happen when the ClusterTopologyBinding resource is modified which removes one or more levels required by the PodCliqueSet.
 	ConditionTopologyLevelsUnavailable = "TopologyLevelsUnavailable"
@@ -145,6 +152,12 @@ const (
 	ConditionReasonSufficientAvailablePCSGReplicas = "SufficientAvailablePodCliqueScalingGroupReplicas"
 	// ConditionReasonUpdateInProgress indicates that the resource is undergoing rolling update.
 	ConditionReasonUpdateInProgress = "UpdateInProgress"
+	// ConditionReasonGangTerminationActive is the (only) Reason paired with
+	// ConditionTypeGangTerminationInProgress=True. The Kubernetes condition API requires a
+	// non-empty Reason, and the flag has exactly one cause, so this Reason simply restates the
+	// Type in the CamelCase token form callers branch on. If a second cause for the flag is ever
+	// introduced, add a distinct Reason then.
+	ConditionReasonGangTerminationActive = "GangTerminationActive"
 	// ConditionReasonClusterTopologyNotFound indicates that the ClusterTopologyBinding resource required for topology-aware scheduling was not found.
 	ConditionReasonClusterTopologyNotFound = "ClusterTopologyNotFound"
 	// ConditionReasonTopologyLevelsUnavailable indicates that the one or more required topology levels defined on a
--- a/operator/e2e/tests/gang_termination_test.go
+++ b/operator/e2e/tests/gang_termination_test.go
@@ -220,7 +220,9 @@ func Test_GT4_GangTerminationMinReplicasPCSGOwned(t *testing.T) {
 	pcAUIDsBeforeStep5 := capturePodUIDsForClique(pods, "workload2-gt-0-pc-a")
 	cordonAndKillPodsFromClique(ctx, t, tc, pcsg1Target, 1)
 	time.Sleep(terminationDelayInWorkloadYAML + gangTerminationGrace)
-	verifyNoGangTermination(t, tc, pcAUIDsBeforeStep5, 0)
+	// minRunning=totalWLPods-1 also proves the killed pod's replacement (or the rest of the
+	// workload) wasn't silently lost — with 0 the check devolved to UID-survival only.
+	verifyNoGangTermination(t, tc, pcAUIDsBeforeStep5, totalWLPods-1)
 
 	Logger.Info("6. Kill the remaining 2 ready pods from sg-x-1-pc-c — test plan expects all PCS pods terminated")
 	pods, err = tc.ListPods()
@@ -334,7 +336,12 @@ func Test_GT6_ScaledPodGangPodDeletion(t *testing.T) {
 	time.Sleep(terminationDelayInWorkloadYAML + gangTerminationGrace)
 
 	Logger.Info("6. Verify pc-a (standalone PCLQ) kept its UIDs — only PCS-level gang term would delete pc-a")
-	verifyNoGangTermination(t, tc, pcAUIDs, 0)
+	verifyNoGangTermination(t, tc, pcAUIDs, totalWLPods-1)
+
+	Logger.Info("7. Verify the deleted scaled pod was actually recreated and the workload is fully ready again")
+	if err := tc.WaitForReadyPods(totalWLPods); err != nil {
+		t.Fatalf("scaled pod was not recreated/ready after deletion: %v", err)
+	}
 }
 
 // ---------- helpers ----------
@@ -490,7 +497,10 @@ func dumpPodsByClique(t *testing.T, pods *corev1.PodList) {
 
 // verifyAllPodsRecreated polls until every UID in originalUIDs is absent from
 // the workload's current pods AND the workload has exactly expectedPods
-// non-terminating pods. Strong signal that PCS-level gang termination fired.
+// non-terminating pods, all of them Ready. Strong signal that PCS-level gang
+// termination fired AND the replacement gang actually recovered — without the
+// readiness requirement a replacement gang stuck Pending (e.g. unschedulable)
+// would be reported as a successful recycle.
 func verifyAllPodsRecreated(t *testing.T, tc *testctx.TestContext, originalUIDs map[types.UID]struct{}, expectedPods int) {
 	t.Helper()
 	deadline := time.Now().Add(tc.Timeout)
@@ -500,20 +510,24 @@ func verifyAllPodsRecreated(t *testing.T, tc *testctx.TestContext, originalUIDs 
 		if err != nil {
 			t.Fatalf("Failed to list pods during recreate check: %v", err)
 		}
-		survivors, nonTerminating := 0, 0
-		for _, p := range pods.Items {
+		survivors, nonTerminating, ready := 0, 0, 0
+		for i := range pods.Items {
+			p := &pods.Items[i]
 			if _, ok := originalUIDs[p.UID]; ok {
 				survivors++
 			}
 			if p.DeletionTimestamp == nil {
 				nonTerminating++
+				if isPodReady(p) {
+					ready++
+				}
 			}
 		}
-		if survivors == 0 && nonTerminating == expectedPods {
-			Logger.Infof("Gang termination confirmed: 0/%d original UIDs survive, %d non-terminating pods", len(originalUIDs), nonTerminating)
+		if survivors == 0 && nonTerminating == expectedPods && ready == expectedPods {
+			Logger.Infof("Gang termination confirmed: 0/%d original UIDs survive, %d non-terminating pods, all ready", len(originalUIDs), nonTerminating)
 			return
 		}
-		lastErr = fmt.Sprintf("survivors=%d non-terminating=%d (want survivors=0 non-terminating=%d)", survivors, nonTerminating, expectedPods)
+		lastErr = fmt.Sprintf("survivors=%d non-terminating=%d ready=%d (want survivors=0 non-terminating=%d ready=%d)", survivors, nonTerminating, ready, expectedPods, expectedPods)
 		time.Sleep(tc.Interval)
 	}
 	t.Fatalf("Gang termination did not occur within %s — final state: %s", tc.Timeout, lastErr)
@@ -579,7 +593,12 @@ func verifyPCSGReplicaRecreatedOnly(t *testing.T, tc *testctx.TestContext, pcsgR
 			currentPCA[p.UID] = struct{}{}
 		}
 
-		targetRecreated := overlap(currentByPCSGIdx[pcsgReplicaIndex], originalPCSG0) == 0
+		// Requiring the full replacement pod count (not just zero UID overlap) prevents a
+		// false-positive during the transient deletion window: right after the delete the
+		// target index has no non-terminating pods at all, so overlap(nil, original)==0
+		// would hold before any replacement pod exists.
+		targetCurrent := currentByPCSGIdx[pcsgReplicaIndex]
+		targetRecreated := len(targetCurrent) == len(originalPCSG0) && overlap(targetCurrent, originalPCSG0) == 0
 		otherIntact := overlap(currentByPCSGIdx[otherIndex(pcsgReplicaIndex)], originalPCSG1) == len(originalPCSG1)
 		pcAIntact := overlap(currentPCA, originalPCA) == len(originalPCA)
 
--- a/operator/e2e/tests/gang_termination_test.go
+++ b/operator/e2e/tests/gang_termination_test.go
@@ -207,11 +207,12 @@ func Test_GT4_GangTerminationMinReplicasPCSGOwned(t *testing.T) {
 	verifyPCSGReplicaRecreatedOnly(t, tc, "0", pcsg0OriginalUIDs, pcsg1OriginalUIDs, pcAOriginalUIDs)
 
 	Logger.Info("5. Kill 1 ready pod from sg-x-1-pc-c — expect NO PCS-level gang termination (pc-a must survive)")
-	// PCSG-0 is still cycling (PCSG-replica-scoped restart fires every TerminationDelay
-	// because its replacement pods stay Pending on cordoned nodes), so we cannot assert
-	// on PCSG-0's UIDs. The signal that the workload is NOT PCS-gang-terminated is that
-	// pc-a (the standalone PCLQ) keeps its UIDs — only PCS-level termination would
-	// delete pc-a.
+	// We assert on pc-a (the standalone PCLQ) rather than on PCSG-0's UIDs: PCSG-0 was already
+	// recycled once by the PCSG-replica-scoped restart in step 4, so its pods no longer carry
+	// their original UIDs. That path does not keep re-firing — WasPCLQEverScheduled excludes the
+	// freshly recreated, never-scheduled PodCliques from the breached set — but the UIDs still
+	// reflect the step-4 recycle, so they are not a useful signal here. pc-a's UIDs change only
+	// under PCS-level gang termination, so pc-a surviving proves no PCS-level termination fired.
 	pods, err = tc.ListPods()
 	if err != nil {
 		t.Fatalf("Failed to list pods: %v", err)
--- a/operator/e2e/tests/gang_termination_test.go
+++ b/operator/e2e/tests/gang_termination_test.go
@@ -232,6 +232,110 @@ func Test_GT4_GangTerminationMinReplicasPCSGOwned(t *testing.T) {
 	verifyAllPodsRecreated(t, tc, finalOriginalUIDs, totalWLPods)
 }
 
+// Test_GT5_IndividualPCSGReplicaTermination: WL5 (PCS pc-a min=1 r=2;
+// sg-x min=1 r=2 of pc-b min=1 r=1, pc-c min=3 r=3). Kill all 3 pods of one
+// PCSG replica's pc-c, breaching that replica's MinAvailable. Only that PCSG
+// replica's PCLQs should be recreated; the other PCSG replica + pc-a survive.
+//
+// Differs from GT-4 step 4 in that pc-c here has min=3 (no tolerance), so
+// killing even one pod breaches pc-c immediately — we kill all 3 at once.
+func Test_GT5_IndividualPCSGReplicaTermination(t *testing.T) {
+	ctx := context.Background()
+	workloadName := "workload5-gt"
+	pcsg0Target := "workload5-gt-0-sg-x-0-pc-c"
+
+	Logger.Infof("1. Initialize a %d-node Grove cluster for WL5 scenario", totalWLPods)
+	tc, cleanup := testctx.PrepareTest(ctx, t, totalWLPods,
+		testctx.WithWorkload(&testctx.WorkloadConfig{
+			Name:         workloadName,
+			YAMLPath:     "../yaml/workload5-gt.yaml",
+			Namespace:    "default",
+			ExpectedPods: totalWLPods,
+		}),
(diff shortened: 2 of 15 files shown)


shayasoolin:
@danbar2 I'd like to know if the new decision of breaching even when `scheduled == 0` only means setting the condition, or actually recreating resources like `Pods` and `PogGangs`.
Considering the simplest case where gang-scheduling doesn't take place due to insufficient resources, I'm okay with having the breach condition set, but reluctant to recreating resources indefinitely every `terminationDelay` seconds. Grove should be silent in such cases, only the scheduler should do this kind of polling after its pending resources.

operator/e2e/yaml/workload5-gt.yaml
```diff
@@ -0,0 +1,116 @@
+# Workload 2: Flexible Startup
```
shayasoolin: This file is a duplicate of `operator/e2e/yaml/workload2-gt.yaml`. I assume you don't need it, but if you do, please change the contents and the comment header accordingly.

operator/api/common/constants/constants.go
```diff
@@ -145,6 +151,9 @@ const (
 	ConditionReasonSufficientAvailablePCSGReplicas = "SufficientAvailablePodCliqueScalingGroupReplicas"
 	// ConditionReasonUpdateInProgress indicates that the resource is undergoing rolling update.
 	ConditionReasonUpdateInProgress = "UpdateInProgress"
+	// ConditionReasonRecycleInFlight indicates that gang termination has fired for this resource and the
+	// recycle is in flight. Paired with ConditionTypeGangTerminationInProgress.
+	ConditionReasonRecycleInFlight = "RecycleInFlight"
```
shayasoolin: I didn't understand the difference between this and `ConditionTypeGangTerminationInProgress`. If they are both required, please clarify in the inline comments, and I'd also suggest using the term of `GangTermination` in this condition as well, rather than `Recycle` which may be kept more code-internal than user-facing.

operator/e2e/yaml/workload5-gt.yaml
```diff
@@ -0,0 +1,121 @@
+# Workload 5: GT-5 PCSG-replica gang-termination scenario.
```
shayasoolin: consider moving most of this description to the test case itself. the yaml comment better just describe the technical details of what's inside. after all, the same workload yaml can serve us for multiple tests later on.

operator/e2e/yaml/workload1-gt.yaml
```diff
@@ -0,0 +1,115 @@
+# Workload 1: Strict Startup (AnyOrder)
```
shayasoolin: here we have 'strict startup' and in workload2 we have 'flexible startup'. I'm not sure what each of these terms mean (how is strict == AnyOrder? it doesn't make sense to me), and I don't see any reference in the yamls for startup order.

operator/internal/controller/common/component/utils/podclique.go
```diff
@@ -86,8 +86,53 @@ func GroupPCLQsByPCSReplicaIndex(pclqs []grovecorev1alpha1.PodClique) map[string
...
+// has never been healthy is left alone — only regressions get recycled.
+func WasPCLQEverScheduled(pclq *grovecorev1alpha1.PodClique) bool {
+	sched := meta.FindStatusCondition(pclq.Status.Conditions, constants.ConditionTypePodCliqueScheduled)
+	if sched == nil {
+		return false
+	}
+	if sched.Status == metav1.ConditionTrue {
+		return true
+	}
+	return sched.LastTransitionTime.After(pclq.CreationTimestamp.Add(InitialScheduleGrace))
```
shayasoolin, line 117: Concern: this timestamp test is not a reliable `ever scheduled` signal.

The code treats `PodCliqueScheduled=False` with `LastTransitionTime > CreationTimestamp + InitialScheduleGrace` as proof that the PCLQ must have been scheduled before and then regressed. That inference is not safe. Example:

1. PCLQ is created at 12:00:00.
2. The controller is delayed by cache sync, restart, leader election, API latency, or backlog.
3. The first status reconcile happens at 12:00:10.
4. It writes `PodCliqueScheduled=False` and `MinAvailableBreached=True`.
5. This function returns true because 12:00:10 is after creation + 5s.

That workload was never scheduled, but this now arms gang termination after `TerminationDelay`, which can recreate the same Pending resources against the same insufficient cluster state. The same issue exists in `WasPCSGEverHealthy` below: a late first `MinAvailableBreached=True` write can be misread as prior health.

I think this should be an explicit durable signal set only when the good state is actually observed (`PodCliqueScheduled=True` for PCLQ, `MinAvailableBreached=False` for PCSG), rather than inferred from condition timing.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -254,38 +267,22 @@ func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1a
...
-			Status:  metav1.ConditionTrue,
-			Reason:  constants.ConditionReasonScheduledReplicasBelowMinAvailable,
-			Message: fmt.Sprintf("Scheduled replicas (%d) below MinAvailable (%d)", scheduledReplicas, minAvailable),
-		}
-	}
-	minAvailableBreachedReplicas := computeMinAvailableBreachedReplicas(logger, pcsg, pclqsPerPCSGReplica)
-	availableReplicas := scheduledReplicas - minAvailableBreachedReplicas
-	if availableReplicas < minAvailable {
+	breachedReplicas := computeMinAvailableBreachedReplicas(logger, pcsg, pclqsPerPCSGReplica)
+	existingReplicas := len(pclqsPerPCSGReplica)
```
shayasoolin: Concern: `len(pclqsPerPCSGReplica)` counts replica-index buckets, not complete PCSG replicas.

That means a partially-created replica can be counted as “not in breach” as long as none of the PCLQs currently present in that bucket has `MinAvailableBreached=True`. But elsewhere, `computeReplicaStatus` correctly treats a replica as unscheduled/unavailable unless all expected non-terminating PCLQs exist.

Example:

- PCSG has `replicas=2`, `minAvailable=1`.
- Each PCSG replica should contain `pc-b` and `pc-c`.
- Replica 0 has only `pc-b`; `pc-c` does not exist yet, or has been deleted and not recreated.
- Replica 1 has a breached PCLQ.

Current formula:
`existingReplicas = len(map) = 2`
`breachedReplicas = 1`
`notInBreachReplicas = 1`
`1 >= minAvailable`, so PCSG condition becomes `MinAvailableBreached=False`.

But replica 0 is incomplete, so it should not count as a valid not-in-breach replica. This can incorrectly mark the PCSG healthy and also poison the new was-healthy gate, because a false `MinAvailableBreached=False` later becomes evidence that the PCSG was once healthy.

Please count a PCSG replica as not-in-breach only when the expected non-terminating PCLQs for that replica all exist and none of them is breached.

operator/api/common/constants/constants.go
```diff
@@ -120,6 +120,13 @@ const (
 	// ConditionTypePodCliqueScheduled indicates that the PodClique has been successfully scheduled.
 	// This condition is set to true when number of scheduled pods in the PodClique is greater than or equal to PodCliqueSpec.MinAvailable.
 	ConditionTypePodCliqueScheduled = "PodCliqueScheduled"
+	// ConditionTypeGangTerminationInProgress indicates that gang termination has fired for this resource and the
+	// gang termination is still in flight. The condition is set when gang termination deletes the PodCliques
+	// (either PCSG-replica scoped or PCS-level) and is cleared when MinAvailableBreached transitions back to
```
shayasoolin: This comment says `GangTerminationInProgress` covers “either PCSG-replica scoped or PCS-level” gang termination, but the implementation only wires it into the PCS-level path.

The PCS-level path:
- filters candidates in `getMinAvailableBreachedPCSGInfo`
- sets the condition in `createPCSReplicaDeleteTask`
- clears it from PCSG status when the PCSG recovers

The PCSG-replica scoped path does not use this condition:
- `processMinAvailableBreachedPCSGReplicas` calls `createDeleteTasks`
- `createDeleteTasks` only `DeleteAllOf`s the PCLQs for that PCSG replica
- `getMinAvailableBreachedPCSGIndices` only uses `GetMinAvailableBreachedPCLQInfo`; it does not check `GangTerminationInProgress`

So this flag does not currently suppress PCSG-replica scoped recycle loops. The E2E comment in `gang_termination_test.go` saying PCSG-0 is “still cycling” also points at this mismatch. Either this condition should be wired into the PCSG-replica scoped path too, or the constant comment and test comments should be narrowed to say this is PCS-level behavior only.

shayasoolin (commented):
Review comments on gang-termination edge cases and comment/code consistency.

operator/internal/controller/common/component/utils/podclique.go
danbar2: its a known issue, I think its reasonable.

operator/internal/controller/common/component/utils/podclique.go
```diff
@@ -86,8 +86,53 @@ func GroupPCLQsByPCSReplicaIndex(pclqs []grovecorev1alpha1.PodClique) map[string
...
+}
+
+// WasPCSGEverHealthy reports whether the PodCliqueScalingGroup has ever reached the
+// MinAvailableBreached=False state since creation. Mirrors WasPCLQEverScheduled but reads the
+// PCSG's own MinAvailableBreached condition (PCSGs have no PodCliqueScheduled equivalent).
+// Used to gate gang-termination so an initial-startup PCSG that has not yet stabilized is left
+// alone — only regressions from a previously-healthy state get recycled.
+func WasPCSGEverHealthy(pcsg *grovecorev1alpha1.PodCliqueScalingGroup) bool {
+	cond := meta.FindStatusCondition(pcsg.Status.Conditions, constants.ConditionTypeMinAvailableBreached)
```
Ronkahn21, line 127: This logic only holds true if the operator actually observes the transition. If the operator was offline and missed the "healthy" state, it doesn't break the logic. This limitation applies to all status checks, and we should either document it or avoid using this method entirely.

operator/internal/controller/podcliqueset/components/podcliquesetreplica/gangterminate.go
```diff
@@ -187,26 +204,61 @@ func getMinAvailableBreachedPCSGInfo(pcsgs []grovecorev1alpha1.PodCliqueScalingG
...
+			pcsgList := &grovecorev1alpha1.PodCliqueScalingGroupList{}
+			if err := r.client.List(ctx, pcsgList,
+				client.InNamespace(pcs.Namespace),
+				client.MatchingLabels(pcsReplicaLabels)); err != nil {
+				logger.Error(err, "failed to list PCSGs to mark GangTerminationInProgress", "pcsReplicaIndex", pcsReplicaIndex)
+				return err
+			}
+			for i := range pcsgList.Items {
+				pcsg := &pcsgList.Items[i]
+				patch := client.MergeFrom(pcsg.DeepCopy())
```
Ronkahn21: **[Finding 1] Missing optimistic locking on status patch**

`client.MergeFrom(pcsg.DeepCopy())` produces a JSON merge-patch with no `resourceVersion` precondition. The PCSG status reconciler runs concurrently and mutates `Status.Conditions` (e.g. clearing `GangTerminationInProgress` on recovery, updating `MinAvailableBreached`). A write from that reconciler between this function's `List()` and `Patch()` calls will be silently overwritten.

Use `MergeFromWithOptimisticLock` and retry on conflict:
```go
if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
    latest := &grovecorev1alpha1.PodCliqueScalingGroup{}
    if err := r.client.Get(ctx, client.ObjectKeyFromObject(pcsg), latest); err != nil {
        return err
    }
    patch := client.MergeFromWithOptions(latest.DeepCopy(), client.MergeFromWithOptimisticLock{})
    meta.SetStatusCondition(&latest.Status.Conditions, metav1.Condition{...})
    return r.client.Status().Patch(ctx, latest, patch)
}); err != nil {
    logger.Error(err, "failed to mark GangTerminationInProgress on PCSG", ...)
    return err
}
```

operator/e2e/tests/gang_termination_test.go
```diff
@@ -0,0 +1,612 @@
...
+	}
+	if err := tc.Client.Delete(ctx, scaled); err != nil {
+		t.Fatalf("Failed to delete pod: %v", err)
+	}
+
+	Logger.Infof("5. Wait %s — base PodGang min still met, no PCS-level gang term should fire", terminationDelayInWorkloadYAML+gangTerminationGrace)
+	time.Sleep(terminationDelayInWorkloadYAML + gangTerminationGrace)
+
+	Logger.Info("6. Verify pc-a (standalone PCLQ) kept its UIDs — only PCS-level gang term would delete pc-a")
+	verifyNoGangTermination(t, tc, pcAUIDs, 0)
```
Ronkahn21: **[Finding 3] Assertion is weaker than the doc comment claims (GT-6)**

`verifyNoGangTermination(t, tc, pcAUIDs, 0)` — `minRunning=0` disables the pod-count floor entirely. The test description says "the controller recreates the pod and the workload keeps running" but nothing verifies the deleted scaled pc-c pod was actually recreated. A regression where the pod is never recreated (with pc-a untouched) would pass this test.

Suggested fix:
```go
verifyNoGangTermination(t, tc, pcAUIDs, totalWLPods-1)
if err := tc.WaitForReadyPods(totalWLPods); err != nil {
    t.Fatalf("scaled pod was not recreated/ready after deletion: %v", err)
}
```

Same issue applies to GT-4 step 5 (line ~284, also `minRunning=0`).

operator/e2e/tests/gang_termination_test.go
```diff
@@ -0,0 +1,612 @@
...
+				if currentByPCSGIdx[idx] == nil {
+					currentByPCSGIdx[idx] = map[types.UID]struct{}{}
+				}
+				currentByPCSGIdx[idx][p.UID] = struct{}{}
+				continue
+			}
+			currentPCA[p.UID] = struct{}{}
+		}
+
+		targetRecreated := overlap(currentByPCSGIdx[pcsgReplicaIndex], originalPCSG0) == 0
```
Ronkahn21: ** False-positive during the transient deletion window**

`currentByPCSGIdx[pcsgReplicaIndex]` is `nil` when the target PCSG's pods have all been deleted but none yet recreated (terminating pods are filtered at line 569). `overlap(nil, originalPCSG0)` iterates zero keys → returns 0, so `0 == 0` is true mid-deletion, before any replacement pod exists. Combined with `otherIntact` and `pcAIntact` already holding, the poll loop can return success prematurely.

Require that the new pods actually exist:
```go
targetCurrent := currentByPCSGIdx[pcsgReplicaIndex]
targetRecreated := len(targetCurrent) == len(originalPCSG0) && overlap(targetCurrent, originalPCSG0) == 0
```

operator/e2e/tests/gang_termination_test.go
```diff
@@ -0,0 +1,612 @@
...
+		survivors, nonTerminating := 0, 0
+		for _, p := range pods.Items {
+			if _, ok := originalUIDs[p.UID]; ok {
+				survivors++
+			}
+			if p.DeletionTimestamp == nil {
+				nonTerminating++
+			}
+		}
+		if survivors == 0 && nonTerminating == expectedPods {
```
Ronkahn21, line 547: ** Readiness not checked — Pending pods satisfy this condition**

`nonTerminating == expectedPods` passes when pods are stuck `Pending` (e.g. if the new PodGang can't schedule due to cordoned nodes). This reports "gang termination confirmed" on a workload that never actually recovered.

Adding a `ready` counter would catch this:
```go
survivors, nonTerminating, ready := 0, 0, 0
for _, p := range pods.Items {
    if _, ok := originalUIDs[p.UID]; ok { survivors++ }
    if p.DeletionTimestamp == nil {
        nonTerminating++
        if isPodReady(&p) { ready++ }
    }
}
if survivors == 0 && nonTerminating == expectedPods && ready == expectedPods {
```

Note: the e2e cluster uses 80Mi-per-pod busybox containers so the new PodGang should fit across the remaining uncordoned nodes — but worth confirming before applying this.

operator/internal/controller/podcliqueset/components/podcliquesetreplica/gangterminate.go
```diff
@@ -187,26 +204,61 @@ func getMinAvailableBreachedPCSGInfo(pcsgs []grovecorev1alpha1.PodCliqueScalingG
...
+				meta.SetStatusCondition(&pcsg.Status.Conditions, metav1.Condition{
+					Type:    apiconstants.ConditionTypeGangTerminationInProgress,
+					Status:  metav1.ConditionTrue,
+					Reason:  apiconstants.ConditionReasonGangTerminationActive,
+					Message: fmt.Sprintf("Gang termination fired at PCS-replica scope for PCS replica %d; this PCSG's PodCliques were deleted as part of the recycle", pcsReplicaIndex),
+				})
+				if err := r.client.Status().Patch(ctx, pcsg, patch); err != nil {
+					logger.Error(err, "failed to mark GangTerminationInProgress on PCSG", "pcsg", client.ObjectKeyFromObject(pcsg))
+					return err
```
oleg-kushniriov: After the delete, the code lists PCSGs and patches each in a loop; the first `Status().Patch` error returns from the whole task. If PCSG A is flagged and PCSG B's patch fails:
- The task errors and is retried. `DeleteAllOf` runs again and deletes the freshly-recreated PodCliques.
- Some PCSGs carry the flag and some don't, so the gate result is inconsistent until convergence.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -232,17 +237,25 @@ func mutateMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1al
 			"reason", newCondition.Reason)
 		meta.SetStatusCondition(&pcsg.Status.Conditions, newCondition)
 	}
+	if newCondition.Status == metav1.ConditionFalse &&
```
oleg-kushniriov, line 240: doc says "True→False transition" but the code clears on any `False`

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -254,58 +267,61 @@ func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1a
 	}
 
 	minAvailable := int(*pcsg.Spec.MinAvailable)
-	scheduledReplicas := int(pcsg.Status.ScheduledReplicas)
-	if scheduledReplicas < minAvailable {
-		if scheduledReplicas == 0 {
-			return metav1.Condition{
-				Type:    constants.ConditionTypeMinAvailableBreached,
-				Status:  metav1.ConditionFalse,
-				Reason:  constants.ConditionReasonInsufficientScheduledPCSGReplicas,
```
oleg-kushniriov, line 263: ConditionReasonInsufficientScheduledPCSGReplicas - is dead constant now - consider removing

operator/internal/controller/common/component/utils/podclique.go
```diff
@@ -86,8 +86,53 @@ func GroupPCLQsByPCSReplicaIndex(pclqs []grovecorev1alpha1.PodClique) map[string
...
+	if cond == nil {
+		return false
+	}
+	if cond.Status == metav1.ConditionFalse {
+		return true
+	}
+	return cond.LastTransitionTime.After(pcsg.CreationTimestamp.Add(InitialScheduleGrace))
```
oleg-kushniriov, line 134: The value of InitialScheduleGrace (5s) may be too small to correctly compute the "was-ever-healthy/scheduled" signal. if the first reconcile takes longer than that,
a workload that was never scheduled gets its initial PodCliqueScheduled=False stamped with a LastTransitionTime past the grace window. It's then misclassified as previously-healthy, marked eligible for gang  termination, and recycled - producing the Pending-pod churn loop.
                                                                                                                                                                                                                                                                                          Consider recording it explicitly instead of inferring from time - 
set a status field or annotation the first time PodCliqueScheduled goes True, and read that. That's immune to reconcile timing.

operator/internal/controller/common/component/utils/podclique.go
danbar2: Thats correct, its a design approach we took a while ago. It will still be eventually consistent. Since there is a grace time window, we need to mark from when the window starts

operator/internal/controller/podcliqueset/components/podcliquesetreplica/gangterminate.go
```diff
@@ -187,26 +204,61 @@ func getMinAvailableBreachedPCSGInfo(pcsgs []grovecorev1alpha1.PodCliqueScalingG
...
+			for i := range pcsgList.Items {
+				pcsg := &pcsgList.Items[i]
+				patch := client.MergeFrom(pcsg.DeepCopy())
+				meta.SetStatusCondition(&pcsg.Status.Conditions, metav1.Condition{
+					Type:    apiconstants.ConditionTypeGangTerminationInProgress,
+					Status:  metav1.ConditionTrue,
+					Reason:  apiconstants.ConditionReasonGangTerminationActive,
+					Message: fmt.Sprintf("Gang termination fired at PCS-replica scope for PCS replica %d; this PCSG's PodCliques were deleted as part of the recycle", pcsReplicaIndex),
+				})
+				if err := r.client.Status().Patch(ctx, pcsg, patch); err != nil {
```
oleg-kushniriov: Two controllers now write PCSG.Status.Conditions with plain client.MergeFrom : the PCSG status reconciler (here and in reconcilestatus.go:57). A merge patch replaces the conditions array wholesale, so concurrent writes clobber each other: either GangTerminationInProgress is dropped (→ re-fire) or the recovery MinAvailableBreached=False is reverted (→ flag never clears, gang termination stuck off). 

a simple fix if considered can be: add optimistic locking to both writers and retry on conflict


```
  retry.RetryOnConflict(retry.DefaultRetry, func() error {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
      patch := client.MergeFromWithOptions(pcsg.DeepCopy(), client.MergeFromWithOptimisticLock{})                                                                                                                                                                                             
      meta.SetStatusCondition(&pcsg.Status.Conditions, ...)                                                                                                                                                                                                                                   
      return r.client.Status().Patch(ctx, pcsg, patch)                                                                                                                                                                                                                                        
  })    
```

operator/internal/controller/common/component/utils/podclique.go
danbar2: Already discussed, its an edge case that we can accept

operator/internal/controller/podcliqueset/components/podcliquesetreplica/gangterminate.go
danbar2: fixed already in previous comment

operator/internal/controller/podcliqueset/components/podcliquesetreplica/gangterminate.go
```diff
@@ -187,31 +206,97 @@ func getMinAvailableBreachedPCSGInfo(pcsgs []grovecorev1alpha1.PodCliqueScalingG
...
-				logger.Error(err, "failed to delete PodCliques for PCS Replica index", "pcsReplicaIndex", pcsReplicaIndex, "reason", reason)
-				r.eventRecorder.Eventf(pcs, corev1.EventTypeWarning, constants.ReasonPodCliqueSetReplicaDeleteFailed, "Error deleting PodCliqueSet replica %d: %v", pcsReplicaIndex, err)
+				client.MatchingLabels(pcsReplicaLabels)); err != nil {
+				logger.Error(err, "failed to list PCSGs for PCS replica gang termination", "pcsReplicaIndex", pcsReplicaIndex)
 				return err
 			}
-			logger.Info("Deleted PCS replica PodCliques", "pcsReplicaIndex", pcsReplicaIndex, "reason", reason)
-			r.eventRecorder.Eventf(pcs, corev1.EventTypeNormal, constants.ReasonPodCliqueSetReplicaDeleteSuccessful, "PodCliqueSet replica %d deleted", pcsReplicaIndex)
-			return nil
+			alreadyDeleted := lo.SomeBy(pcsgList.Items, func(pcsg grovecorev1alpha1.PodCliqueScalingGroup) bool {
```
oleg-kushniriov: **Failure scenario (PCS replica with two PCSGs `sg-a`, `sg-b`):**
1. Both regress and cross `TerminationDelay` → fire. DeleteAllOf runs; `sg-a` and `sg-b` both flagged.
2. `sg-a` recovers → its flag is cleared. `sg-b` is slower to recover → `sg-b` keeps the flag.
3. `sg-a` regresses **again**, crosses `TerminationDelay` → it becomes a valid candidate (breached, was-healthy, no flag) and a fire is scheduled.
4. In `createPCSReplicaDeleteTask`: `List` returns `[sg-a (no flag), sg-b (flag=True)]` → `alreadyDeleted == true` (because of `sg-b`) → **`DeleteAllOf` is skipped**. Only the flag writes run, stamping `sg-a`.
5. `sg-a`'s PodCliques were never deleted; `sg-a` stays breached and is now flagged → suppressed indefinitely.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -254,58 +267,61 @@ func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1a
 	}
 
 	minAvailable := int(*pcsg.Spec.MinAvailable)
```
oleg-kushniriov: consider sanity check: pcsg.Spec.MinAvailable != nil

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -254,58 +267,61 @@ func computeMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1a
...
+// Completeness is required because a partially-created replica (e.g. one whose pc-c was deleted and
+// not yet recreated) has no breached PodClique yet is still not a valid healthy replica. Counting it
+// as not-in-breach would spuriously report the PCSG healthy and would also poison the was-healthy gate,
+// which reads a MinAvailableBreached=False as evidence that the PCSG was once healthy. This mirrors
+// computeReplicaStatus, which likewise treats a replica as unscheduled/unavailable unless all expected
+// PodCliques exist.
+//
+// Iteration is bounded to expected replica indexes [0, Spec.Replicas) so stale-index children left
+// behind during scale-down neither inflate nor deflate the count.
+func computeNotInBreachReplicas(logger logr.Logger, pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pclqsPerPCSGReplica map[string][]grovecorev1alpha1.PodClique) int {
```
oleg-kushniriov, line 309: New replica indexes are "incomplete" (no PCLQs yet). Raising `Replicas + MinAvailable` together (`3/3 -> 5/5`) makes `notInBreachReplicas=3 < 5` -> `MinAvailableBreached=True`; if the new capacity takes >`TerminationDelay`, the whole replica (including healthy members) is recycled.

Fix suggestion: don't count not-yet-created replicas against availability - clamp the denominator to observed replicas, e.g. gate on `pcsg.Status.Replicas` / observedGeneration rather than `Spec.Replicas` for the breach decision, or exclude never-created indexes from the `< MinAvailable` comparison.

danbar2: 1. MinAvailable is immutable
2. Still, its a valid termination and its a good reason for termination delay

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
danbar2: Spec.MinAvailable defaults to 1. cant be nil

oleg-kushniriov (dismissed):

Ronkahn21 (dismissed):

oleg-kushniriov (approved):

Ronkahn21 (approved):

PR #616, opened 2026-05-15
Rename ClusterTopology CRD to ClusterTopologyBinding

### LINKED ISSUE #615: Rename ClusterTopology to ClusterTopologyBinding and handle upgrade/migration
**What you would like to be added?**

Rename the Grove topology API type from `ClusterTopology` to `ClusterTopologyBinding` everywhere it still appears in API types, comments, docs, controller/status wording, tests, and related helper names.

This rename should not be treated as cosmetic only. The updated naming and documentation should make the intended semantics explicit:

- this resource is Grove’s source-of-truth topology hierarchy
- it is also the binding point between that Grove topology model and each scheduler backend’s topology representation
- it is not itself the scheduler backend topology resource

Because this changes the CRD, the issue should also cover the upgrade path. Existing `ClusterTopology` resources must be handled during upgrade, and the `crd-upgrader` should be updated to support the migration.

**Why is this needed?**

The old name `ClusterTopology` reads like “this CR is the topology object itself,” which is misleading.

In practice, the resource serves two related purposes:

- it defines Grove’s topology hierarchy in Grove terms
- it binds that hierarchy to one or more backend-specific topology resources, which may be Grove-managed or externally managed

That distinction matters for understanding why `schedulerTopologyBindings` exists, why the controller may create backend topology resources or only check them for drift, and where this API sits relative to scheduler-specific topology objects.

The CRD rename also creates an upgrade concern for existing clusters, so Grove needs a defined migration path either implemented it in the `crd-upgrader` or clear documentation.

That said, this is expected to be a relatively easy migration because there has not been a Grove release since the change that allowed administrators to create `ClusterTopology` resources freely. So while migration support is still required, the practical upgrade surface should remain limited.

enoodle:
#### What type of PR is this?
/kind api
/kind cleanup
/kind documentation

Rename ClusterTopology CRD to ClusterTopologyBinding

Fixes #615 

```release-note
Rename ClusterTopology CRD to ClusterTopologyBinding
```

Updated docs/installation.md with upgrade instructions

```docs

```


danbar2 (dismissed):

docs/installation.md
```diff
@@ -116,6 +116,43 @@ On clusters with NVIDIA MNNVL support, you can enable automatic Multi-Node NVLin
...
+2. Delete the old `ClusterTopology` instances.
+3. Delete the old `ClusterTopology` CRD.
+
+Example:
+
+```bash
+# 1. Verify any old ClusterTopology instances that still exist
+kubectl get clustertopologies.grove.io
+
+# 2. Delete the old ClusterTopology instances
```
shayasoolin: Add a note between 1 and 2 that this is where any `ClusterTopology` resources that are found (by `get`) and are relevant to keep, should be recreated.

docs/installation.md
```diff
@@ -116,6 +116,43 @@ On clusters with NVIDIA MNNVL support, you can enable automatic Multi-Node NVLin
...
+kubectl delete clustertopologies.grove.io --all
+
+# 3. Delete the old ClusterTopology CRD
+kubectl delete crd clustertopologies.grove.io
+```
+
+If no `ClusterTopology` instances exist in the cluster, delete the old CRD
```
shayasoolin: I think this addition is redundant, the full example above is clear enough.

shayasoolin (commented):
Added some minor comment, LGTM, thanks

docs/installation.md
enoodle: I will remove this

shayasoolin (approved):

danbar2 (approved):

PR #621, opened 2026-05-18
Refine scale test delete measurements

shayasoolin:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
Removes the 5000-pod deletion scale test and changes `ScaleTest_1000` to measure only the user-facing PCS delete request latency. After cascade-delete, waiting for all child resources and pods to disappear mostly measures Kubernetes API server and garbage collector behavior rather than Grove controller overhead.

Adds a small regular e2e test that verifies PCS deletion cascades through the owned Grove resource tree and managed child resources.

#### Which issue(s) this PR fixes:
Related to #423

#### Special notes for your reviewer:
The scale result intentionally excludes Kubernetes cascade-cleanup latency. Cascade-delete correctness is covered by the new small e2e test instead.

#### Does this PR introduce a API change?
```release-note
NONE
```

#### Additional documentation e.g., enhancement proposals, usage docs, etc.:
```docs
NONE
```

PR #626, opened 2026-05-20
Fix: Prevent potential test environment pollution in shared scale clusters

weizhoublue:
…ster pollution

#### What type of PR is this?

/kind bug

Issue Summary

 Running scale tests (e.g., make run-scale-test) against a persistent or reused Kubernetes cluster.

In  PR #621, WithSkipCleanupWait() was intentionally introduced in operator/e2e/tests/scale/scale_test.go to measure only the PodCliqueSet deletion latency without waiting for the slow Kubernetes API server and GC to delete all sub-resources (Pods) .  
However, an oversight occurred: no thorough final cleanup was enforced at the suite level. Consequently, the test suite exits without ever waiting for the cascading deletion to  complete.
A large number of Terminating resources may  remain in the cluster for seconds after the scale tests finish. 

When the same cluster is reused for subsequent test runs, these leftover resources pollute the environment, leading to scheduling conflicts, test misjudgments, and resource exhaustion.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/operator/e2e/tests/scale/main_test.go
+++ b/operator/e2e/tests/scale/main_test.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/ai-dynamo/grove/operator/e2e/k8s/k8sclient"
 	"github.com/ai-dynamo/grove/operator/e2e/measurement"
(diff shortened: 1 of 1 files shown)


renormalize (commented):
Thanks for the PR @weizhoublue!

Each contribution in grove requires a GitHub issue describing the bug/feature, and the corresponding issue is linked with an implementation PR.

So could you please specify the bug as you've described in this PR's description in a GitHub issue, and then link to that issue in this PR? We have an action that verifies this, and that is currently failing.

PR #627, opened 2026-05-21
fix: suppress cascade-delete reconcile noise across PCS/PCSG/PCLQ

### LINKED ISSUE #622: Reduce noisy PodClique NotFound logs during normal cascade-delete flow
Follow-up to https://github.com/ai-dynamo/grove/pull/556.

After the cascade-delete change, deleting a PodCliqueSet removes the PCS, then its owned PodCliques, and then Kubernetes garbage-collects the pods. During this normal delete flow, the PodClique controller can receive reconcile requests for PodCliques that were already deleted and logs many entries like this at `info` level:

```json
{"level":"info","ts":"2026-05-18T08:36:45.836Z","logger":"podclique-controller","msg":"PodClique not found","controller":"podclique-controller","controllerGroup":"grove.io","controllerKind":"PodClique","PodClique":{"name":"scale-test-5000-deletion-0-decode-465-dworker","namespace":"default"},"namespace":"default","name":"scale-test-5000-deletion-0-decode-465-dworker","reconcileID":"384bdb1d-6169-4040-af7a-8f2c9f7f4a0e","objectKey":{"name":"scale-test-5000-deletion-0-decode-465-dworker","namespace":"default"}}
```

At scale this produces a large amount of expected log noise during normal deletion.

The tricky part is that `PodClique not found` is not always noise. If a PodClique is unexpectedly missing while its parent PCS/PCSG still expects it to exist, that is useful information and should remain visible, probably as a warning or as part of parent reconciliation/status/debugging.

We should improve this logging so expected NotFound reconciles during cascade-delete do not spam `info`, while still preserving visibility for unexpected missing PodCliques.

Possible directions:
- Lower the generic `PodClique not found` log in the PodClique reconciler/get helper to a verbose/debug level.
- Add a more contextual warning in the owning PCS/PCSG reconciliation path when a desired PodClique is missing while the parent is not being deleted.
- Consider whether the NotFound helper needs a caller-controlled log policy instead of always logging at `info`.

Acceptance criteria:
- Normal PCS cascade-delete does not emit one `info` log per already-deleted PodClique.
- Unexpected missing PodCliques while the parent still expects them are still observable in logs and/or status.
- The behavior is covered by unit tests around the relevant reconciliation/get helper or owner reconciliation path.

yankay:
## What this PR does

Fixes #622. When a `PodCliqueSet` is deleted, Kubernetes garbage collection cascades the deletion through PCSG → PodClique → Pod. Each child reconciler could observe the missing parent and emit info-level `not found` logs, producing tens to thousands of noisy lines during a normal delete flow.

This PR suppresses the expected cascade-delete noise across PCS / PCSG / PCLQ.

## Changes

1. **Demote `NotFound` logs to `V(1)`** in the shared reconciler helpers (`GetPodCliqueSet`, `GetPodClique`, `GetPodCliqueScalingGroup`).

2. **Filter cascade-delete events at the watch predicate source.** New `IsOwnerBeingDeleted` helper checks whether the child's owner is being deleted, already gone, or replaced by a same-name object with a different UID.

3. **Bound predicate owner lookups with a 5s timeout** so a slow API/cache lookup cannot stall controller-runtime event processing indefinitely.

## Coverage of review feedback

Per the feedback in https://github.com/ai-dynamo/grove/issues/622#issuecomment-4497899390, Delete predicates now filter cascade-delete events at the watch source, and the fix covers the PCS / PCSG / PCLQ cascade path.

## Test plan

- [x] Unit tests for `IsOwnerBeingDeleted`:
  - owner alive
  - owner deleting
  - owner already gone (`NotFound`)
  - same-name owner recreated with a different UID
- [x] Unit tests for PodClique Delete predicates:
  - manual Pod delete still observes expectations and enqueues reconcile
  - cascade Pod delete is filtered without consuming expectations
  - PodClique delete is filtered when the owning PCSG is deleting
- [x] Unit test for `GetPodClique` `NotFound` log verbosity (`V(0)` suppressed, `V(1)` emitted)
- [x] Local `go test ./...`
- [x] E2E on kind: head-to-head comparison of operator logs during `kubectl delete pcs simple1`

### E2E results — before vs. after

Same kind cluster, same `simple1` sample (1 PCS + 1 PCSG + 4 PCLQs + 9 Pods). Operator restarted before each `kubectl delete pcs simple1`; logs scoped via `--since-time`.

| Metric | Before (main) | After (this PR) |
| --- | ---: | ---: |
| Total log lines | 162 | **53** |
| `error`-level | 2 | **0** |
| `"not found"` mentions | 54 | **0** |
| &nbsp;&nbsp;PodClique / PCSG / PCS | 40 / 5 / 7 | 0 / 0 / 0 |

Raw logs: [gist](https://gist.github.com/yankay/cb2bbdf1b1cc76d3fe7197bf9fb967f7) (`logs-parent.json` / `logs-pr.json`).

The 2 `error` lines on `main` are reconcile failures from the cascade race: the PCSG controller is enqueued for `simple1-0-sga` after `PodCliqueSet/simple1` is already gone, `GetPodCliqueSet` returns `NotFound`, and controller-runtime logs `Reconciler error`. The new predicate filter drops these events at the watch source, eliminating both error-class and info-class noise.


operator/internal/controller/utils/predicate_helpers.go
```diff
@@ -0,0 +1,42 @@
...
+// &grovecorev1alpha1.PodCliqueSet{}. On transient errors it returns false to avoid
+// swallowing real events.
+func IsOwnerBeingDeleted(ctx context.Context, cl client.Client, child client.Object, ownerKind string, ownerObj client.Object) bool {
+	ownerRef := k8sutils.FindOwnerRefByKind(child.GetOwnerReferences(), ownerKind)
+	if ownerRef == nil {
+		return false
+	}
+	key := client.ObjectKey{Namespace: child.GetNamespace(), Name: ownerRef.Name}
+	if err := cl.Get(ctx, key, ownerObj); err != nil {
+		return apierrors.IsNotFound(err)
```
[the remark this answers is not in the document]
yankay: Good catch. Fixed in 7d05fd2 by also comparing `ownerObj.GetUID()` against `ownerRef.UID`, so a same-name recreate is treated as deleted.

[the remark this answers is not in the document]
shayasoolin: I don't see the fix for that.

shayasoolin (commented):
LGTM but please fix the comment about using UID. Thanks

operator/internal/controller/utils/predicate_helpers.go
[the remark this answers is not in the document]
yankay: Thanks @shayasoolin — re-pushed in 4c03c39. Added the `ownerObj.GetUID() != ownerRef.UID` check below:

shayasoolin (dismissed):

danbar2 (dismissed):

renormalize:
I'm mid-review, can the merge of this wait for a bit longer? Thanks.

operator/internal/controller/podclique/register.go
```diff
@@ -83,15 +83,22 @@ func (r *Reconciler) RegisterWithManager(mgr ctrl.Manager) error {
...
 	return predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
 			return grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...)
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			return grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...)
+			if !grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...) {
+				return false
+			}
+			ctx := context.Background()
```
renormalize: We should introduce a timeout for this context. Pehaps we could use the default timeout that's used from similar calls in the codebase? The predicates already diverge from controller-runtime conventions by invoking the client. 

Not having a timeout here could completely block the predicate goroutine, stalling reconciliation for all resources of that kind.

operator/internal/controller/podcliqueset/register.go
```diff
@@ -137,12 +137,16 @@ func mapClusterTopologyToPodCliqueSets(cl client.Client) handler.MapFunc {
...
+// podCliquePredicate keeps events for PCS-owned PodCliques and drops cascade-deletes (issue #622).
+func (r *Reconciler) podCliquePredicate() predicate.Predicate {
 	return predicate.Funcs{
 		CreateFunc: func(_ event.CreateEvent) bool { return false },
 		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
-			return grovectrlutils.IsManagedPodClique(deleteEvent.Object, constants.KindPodCliqueSet)
+			if !grovectrlutils.IsManagedPodClique(deleteEvent.Object, constants.KindPodCliqueSet) {
+				return false
+			}
+			return !grovectrlutils.IsOwnerBeingDeleted(context.Background(), r.client, deleteEvent.Object,
```
renormalize: https://github.com/ai-dynamo/grove/pull/627#discussion_r3309208409

operator/internal/controller/podcliquescalinggroup/register.go
```diff
@@ -176,12 +176,16 @@ func mapPCLQToPCSG() handler.MapFunc {
...
+// podCliquePredicate keeps events for PCSG-owned PodCliques and drops cascade-deletes (issue #622).
+func (r *Reconciler) podCliquePredicate() predicate.Predicate {
 	return predicate.Funcs{
 		CreateFunc: func(_ event.CreateEvent) bool { return false },
 		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
-			return ctrlutils.IsManagedPodClique(deleteEvent.Object, constants.KindPodCliqueScalingGroup)
+			if !ctrlutils.IsManagedPodClique(deleteEvent.Object, constants.KindPodCliqueScalingGroup) {
+				return false
+			}
+			return !ctrlutils.IsOwnerBeingDeleted(context.Background(), r.client, deleteEvent.Object,
```
renormalize: https://github.com/ai-dynamo/grove/pull/627#discussion_r3309208409

operator/internal/controller/podclique/register.go
```diff
@@ -83,15 +83,22 @@ func (r *Reconciler) RegisterWithManager(mgr ctrl.Manager) error {
...
+			if grovectrlutils.IsOwnerBeingDeleted(ctx, r.client, e.Object, constants.KindPodCliqueScalingGroup, &grovecorev1alpha1.PodCliqueScalingGroup{}) {
+				return false
+			}
+			return !grovectrlutils.IsOwnerBeingDeleted(ctx, r.client, e.Object, constants.KindPodCliqueSet, &grovecorev1alpha1.PodCliqueSet{})
```
renormalize: nit: simplify these statements:

```suggestion
			return !(grovectrlutils.IsOwnerBeingDeleted(ctx, r.client, e.Object, constants.KindPodCliqueScalingGroup, &grovecorev1alpha1.PodCliqueScalingGroup{}) ||
				grovectrlutils.IsOwnerBeingDeleted(ctx, r.client, e.Object, constants.KindPodCliqueSet, &grovecorev1alpha1.PodCliqueSet{}))
```

renormalize (changes requested):
Thanks for the PR @yankay. Feel free to ignore the nit, but please introduce a timeout in the client call. I'm unsure if such timeout has been introduced yet; so please suggest an appropriate timeout given the context, and we can take it forward from there. It shouldn't exceed a 5 second duration in my opinion.

yankay:
Thanks @renormalize! Added 5s timeout to all four predicate `client.Get` calls and collapsed the owner check per your suggestion.

renormalize (commented):
@yankay something must have gone wrong in the rebase, I see files that were introduced in the PR that was merged last to main being removed.

Can you perform the rebase again and make sure only the relevant files are modified?

yankay:
> @yankay something must have gone wrong in the rebase, I see files that were introduced in the PR that was merged last to main being removed.
> 
> Can you perform the rebase again and make sure only the relevant files are modified?

  Sorry @renormalize, bad rebase on my side. Redoing against latest main and force-pushing. Thanks for
  catching.

renormalize (approved):

danbar2 (dismissed):

renormalize:
We should get this in by getting another reviewer's approval.

sanjaychatterjee:
> We should get this in by getting another reviewer's approval.

@gflarity can you please review as well?

shayasoolin (dismissed):

renormalize (dismissed):

operator/internal/controller/podclique/register.go
```diff
@@ -83,15 +84,21 @@ func (r *Reconciler) RegisterWithManager(mgr ctrl.Manager) error {
...
 			return grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...)
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			return grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...)
+			if !grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...) {
+				return false
+			}
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			return !grovectrlutils.IsOwnerBeingDeleted(ctx, r.client, e.Object, constants.KindPodCliqueScalingGroup, &grovecorev1alpha1.PodCliqueScalingGroup{}) &&
```
unmarshall: Is there a real need to make a client call from within the predicate? This is not a recommended pattern for any controller-runtime based operator for the following reasons:

* Predicates are called far more frequently than reconciles. Every Add/Update/Delete from the informer hits the predicate, including resync events (every ~10 min by default for each watched kind).
* The manager's cached client is fully populated only after mgr.Start() finishes its initial sync. Predicates can fire before the referenced kind's cache is in sync, especially right after restart — you can get spurious NotFounds and make the wrong filtering decision.
* Predicates run synchronously on the informer callback before enqueueing. Anything blocking in the predicate blocks delivery of every subsequent event from that informer to every registered handler.
* If `client.Get` returns an error other than `NotFoundError` then you still have to return a bool from within the predicate.
* `For/Watches` is a declarative specification of what the controller observes. If a predicate calls `client.Get` to another resource then it creates a coupling to that resource KIND to be cached and in sync.

operator/internal/controller/podcliquescalinggroup/register.go
```diff
@@ -176,12 +177,18 @@ func mapPCLQToPCSG() handler.MapFunc {
...
+// podCliquePredicate keeps events for PCSG-owned PodCliques and drops cascade-deletes (issue #622).
+func (r *Reconciler) podCliquePredicate() predicate.Predicate {
 	return predicate.Funcs{
 		CreateFunc: func(_ event.CreateEvent) bool { return false },
 		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
-			return ctrlutils.IsManagedPodClique(deleteEvent.Object, constants.KindPodCliqueScalingGroup)
+			if !ctrlutils.IsManagedPodClique(deleteEvent.Object, constants.KindPodCliqueScalingGroup) {
+				return false
+			}
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
oleg-kushniriov: The predicates are called synchronously in the controller-runtime watch loop. Using `context.Background()` means the 5s timeout is the only cancellation mechanism. If the manager is shutting down, the in-flight predicate won't be interrupted until the timeout fires. Passing the manager's root context would be more precise.

Consider using manager's context

same usage of `context.Background()` inside predicate — non-cancellable parent (5 occurrences)

operator/internal/controller/podclique/register.go
```diff
@@ -83,15 +84,21 @@ func (r *Reconciler) RegisterWithManager(mgr ctrl.Manager) error {
...
 	return predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
 			return grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...)
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			return grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...)
+			if !grovectrlutils.IsManagedPodClique(e.Object, expectedOwnerKinds...) {
+				return false
+			}
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
unmarshall: I am not sure why is there a timeout of 5s. Currently it is used to make a `client.Get` call. This call always is from the shared informer's cache (local store) and is not a remote call to the kube-apiserver. So there is no network I/O and therefore having this timeout is cosmetic at best.

operator/internal/controller/utils/predicate_helpers.go
```diff
@@ -0,0 +1,46 @@
...
+// deleted or already gone. ownerObj is an empty typed object of the owner's kind, e.g.
+// &grovecorev1alpha1.PodCliqueSet{}. On transient errors it returns false to avoid
+// swallowing real events.
+func IsOwnerBeingDeleted(ctx context.Context, cl client.Client, child client.Object, ownerKind string, ownerObj client.Object) bool {
+	ownerRef := k8sutils.FindOwnerRefByKind(child.GetOwnerReferences(), ownerKind)
+	if ownerRef == nil {
+		return false
+	}
+	key := client.ObjectKey{Namespace: child.GetNamespace(), Name: ownerRef.Name}
+	if err := cl.Get(ctx, key, ownerObj); err != nil {
```
unmarshall: you should change this to:
```go
if err := cl.Get(ctx, key, ownerObj); err != nil {
    if apierrors.IsNotFound(err) {
         return true
    }
    return false  // any other error: don't filter, let reconcile sort it out
}
```

unmarshall (changes requested):

yankay:
Closing this PR as superseded by #641, which addresses #622. Thanks everyone for the reviews.

PR #630, opened 2026-05-25
Add preferred topology packing API

shayasoolin:
## Summary
- add topologyConstraint.pack.preferred as the soft topology packing hint
- move the existing required pack domain API to topologyConstraint.pack.required while keeping legacy packDomain for existing workloads
- update admission validation, warnings, generated CRDs/docs, and TAS e2e fixtures for the new API shape

## Validation
- go test ./... in operator/api
- go test ./... in operator
- go test -c -tags=e2e for e2e tests, update, scale, and auto-mnnvl packages
- make run-e2e TEST_PATTERN=Test_TAS
- git diff --check
- make verify-toc

References #368


docs/api-reference/operator-api.md
```diff
@@ -825,7 +825,8 @@ _Appears in:_
...
+| `pack` _[TopologyPackConstraint](#topologypackconstraint)_ | Pack specifies topology packing constraints for each replica of the resource. |  |  |
+| `packDomain` _[TopologyDomain](#topologydomain)_ | PackDomain specifies the required topology domain using the legacy field name.<br />Controls placement constraint for EACH individual replica instance.<br />Must reference a domain in the topology levels defined in the ClusterTopologyBinding named by TopologyName.<br />Example: "rack" means each replica independently placed within one rack.<br />Note: Does NOT constrain all replicas to the same rack together.<br />Different replicas can be in different topology domains.<br />Deprecated: use Pack.RequiredDomain. |  | MaxLength: 63 <br />MinLength: 1 <br />Pattern: `^[a-z][a-z0-9-]*$` <br /> |
```
danbar2, line 829: mark as deprecated

shayasoolin: there's "Deprecated: use Pack.RequiredDomain." in this line of code, what do you mean? what kind of mark is missing?

danbar2 (dismissed):

docs/api-reference/operator-api.md
danbar2: missed that sorry, approved

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -273,14 +273,74 @@ type TopologyConstraint struct {
...
 	// Controls placement constraint for EACH individual replica instance.
 	// Must reference a domain in the topology levels defined in the ClusterTopologyBinding named by TopologyName.
 	// Example: "rack" means each replica independently placed within one rack.
 	// Note: Does NOT constrain all replicas to the same rack together.
 	// Different replicas can be in different topology domains.
-	// +required
-	PackDomain TopologyDomain `json:"packDomain"`
+	// Deprecated: use Pack.RequiredDomain.
+	// +optional
+	PackDomain TopologyDomain `json:"packDomain,omitempty"`
```
Ronkahn21, line 294: add here cel validation for create to reject new workloads with old

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -714,6 +715,17 @@ func (v *pcsValidator) validateTopologyConstraintsUpdate(oldPCS *grovecorev1alph
 	return immutabilityValidator.validateUpdate(oldPCS)
 }
 
+func (v *pcsValidator) validateTopologyConstraintsForLegacyRepair(ctx context.Context) field.ErrorList {
```
Ronkahn21, line 718: almost all of the  validation can be done only using CEL expression

operator/api/core/v1alpha1/podcliqueset.go
Ronkahn21: example:
// +kubebuilder:validation:XValidation:rule="!has(oldSelf) && has(self.pack) ...

Ronkahn21 (commented):
looks good, I would use CEL where it possible
I would add a mutation webhook rule, mostly for UX  where if you set the old pack domain it will set the new field instead

shayasoolin:
> looks good, I would use CEL where it possible I would add a mutation webhook rule, mostly for UX where if you set the old pack domain it will set the new field instead

Thanks. I'm not sure about the mutation idea, it will practically make us accept the old field instead of forcing the user to fix the original yamls. Then, when the deprecated field is removed from the API, those yamls that kept working so far, will break.

operator/api/core/v1alpha1/podcliqueset.go
shayasoolin: done, thanks

Ronkahn21 (approved):

danbar2 (approved):

PR #499, opened 2026-03-24
GREP-498: Controller Reconciliation Prometheus Metrics

brluobt:
## Summary

Add Prometheus metrics to all three Grove controllers (PodCliqueSet, PodClique, PodCliqueScalingGroup).

Grove currently has **zero custom Prometheus metrics**. This GREP proposes a metrics framework providing:
- Reconcile duration, count, error rate, and in-flight tracking via `ObservedReconciler` wrapper
- Sub-operation timing within each reconcile loop via `MetricsContext`
- 409 Conflict tracking with target-resource attribution for cross-controller contention diagnosis

**Tracking issue:** Fixes #498

## Motivation

During large-scale inference graph deployments (e.g., via NVIDIA Dynamo), reconciliation performance degrades significantly due to high event rates and cross-controller resource contention on PCS/PC objects. Without metrics, diagnosing these issues requires ad-hoc log analysis.

## Key Design Decisions

- **ObservedReconciler wrapper**: Generic wrapper for all 3 controllers, minimal code change
- **MetricsContext via context.Context**: Nil-safe, zero overhead when not present, no interface changes
- **Conflict tracking**: `grove_operator_status_update_conflict_total` with `target_kind` label for precise contention attribution
- **No new dependencies**: Uses existing prometheus/client_golang via controller-runtime
- **No logic changes**: Purely additive instrumentation

## Test Plan

- Unit tests for ObservedReconciler, MetricsContext, and error categorization
- Integration test verifying metrics at /metrics endpoint
- E2E test at scale verifying sub-operation and conflict metrics

PR #633, opened 2026-05-28
feat: add Prometheus metrics to Grove operator controllers

### LINKED ISSUE #498: GREP: Controller Reconciliation Prometheus Metrics
## Summary

Add Prometheus metrics to all three Grove controllers (PodCliqueSet, PodClique, PodCliqueScalingGroup). Grove currently has **zero custom Prometheus metrics**, providing no visibility into reconcile duration, sub-operation latency, error categorization, or cross-controller conflict rates.

## Motivation

During large-scale inference graph deployments using Grove (e.g., via NVIDIA Dynamo), we observed significant reconciliation performance degradation. Deploying an inference graph with 5 services x 10 replicas produces 50+ PodClique Pods, each transitioning through multiple states, generating a high rate of status changes cascading through the controller hierarchy.

**Key observability gaps:**
- No reconcile duration tracking for any controller
- No sub-operation latency breakdown (e.g., sync_pods vs reconcile_status)
- No visibility into 409 Conflict rates when external controllers (e.g., Dynamo) concurrently modify PCS/PC objects
- No work queue depth or controller saturation metrics

## Proposal

A GREP proposal introduces:

1. **ObservedReconciler wrapper** - Automatically records reconcile duration, count, errors, in-flight count, and requeue reasons for all three controllers
2. **MetricsContext** - Enables sub-operation timing within each reconcile loop
3. **Conflict tracking** - Records 409 Conflict errors with target-resource attribution on Status Update/Patch calls

All metrics use the grove_operator namespace prefix. No reconciliation logic is changed; this is purely additive instrumentation.

AsadShahid04:
## Summary
- Adds a new `internal/metrics` package with five `grove_operator_*` Prometheus metrics covering reconcile count, latency, in-flight gauge, sub-operation latency, and 409 Conflict rate
- Wraps all three controllers (PodCliqueSet, PodClique, PodCliqueScalingGroup) with `ObservedReconciler` via `Complete()` in each `RegisterWithManager` — zero changes to reconcile logic
- Times `reconcile_spec` and `reconcile_status` sub-phases in each controller via `StartOperation` / `done()` pairs

## Problem
Grove currently has zero custom Prometheus metrics. During large-scale Dynamo inference graph deployments (e.g. 5 services × 10 replicas = 50+ PodClique Pods), operators have no visibility into reconcile duration, sub-operation latency, 409 Conflict rates, or controller saturation — making it impossible to diagnose performance degradation.

## Solution
**`operator/internal/metrics/metrics.go`** — new package:
- `ObservedReconciler` wraps any `reconcile.Reconciler`; records duration, in-flight count, result label (success/error/requeue/requeue_after), and conflict count automatically
- `StartOperation(controller, operation)` returns a `done()` closure for timing named sub-operations
- All five metrics are registered on the controller-runtime global registry (`sigs.k8s.io/controller-runtime/pkg/metrics`) so they appear on the existing `/metrics` endpoint without any additional configuration

**Controller wiring (no logic changes):**
- `podcliqueset/register.go`: `Complete(grovemetrics.NewObservedReconciler(controllerName, r))`
- `podclique/register.go`: same
- `podcliquescalinggroup/register.go`: same
- Each `reconciler.go`: `doneSpec := grovemetrics.StartOperation(..., "reconcile_spec"); ...; doneSpec()`

**Metrics exposed:**
| Metric | Type | Labels |
|--------|------|--------|
| `grove_operator_reconcile_total` | Counter | `controller`, `result` |
| `grove_operator_reconcile_duration_seconds` | Histogram | `controller` |
| `grove_operator_in_flight_reconciles` | Gauge | `controller` |
| `grove_operator_operation_duration_seconds` | Histogram | `controller`, `operation` |
| `grove_operator_conflict_total` | Counter | `controller` |

## Testing
- `go build ./...`: passed
- Unit tests (`go test ./internal/metrics/... ./internal/controller/...`): passed — 6 new tests covering success, error, conflict, requeue, requeue_after, and sub-operation histogram population

Closes #498

### DIFF
--- a/operator/go.mod
+++ b/operator/go.mod
@@ -4,13 +4,13 @@ go 1.26.1
 
 require (
 	github.com/ai-dynamo/grove/operator/api v0.0.0
-	github.com/ai-dynamo/grove/operator/client v0.0.0-00010101000000-000000000000
 	github.com/ai-dynamo/grove/scheduler/api v0.0.0
 	github.com/docker/docker v28.3.3+incompatible
 	github.com/go-logr/logr v1.4.3
 	github.com/go-logr/zapr v1.3.0
 	github.com/kai-scheduler/KAI-scheduler v0.14.0
 	github.com/open-policy-agent/cert-controller v0.14.0
+	github.com/prometheus/client_golang v1.23.2
 	github.com/samber/lo v1.52.0
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.11.1
@@ -87,6 +87,7 @@ require (
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/compress v1.18.0 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/lann/builder v0.0.0-20180802200727-47ae307949d0 // indirect
 	github.com/lann/ps v0.0.0-20150810152359-62de8c46ede0 // indirect
 	github.com/lib/pq v1.10.9 // indirect
@@ -113,7 +114,6 @@ require (
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/prometheus/client_golang v1.23.2 // indirect
 	github.com/prometheus/client_model v0.6.2 // indirect
 	github.com/prometheus/common v0.67.5 // indirect
 	github.com/prometheus/procfs v0.20.1 // indirect
--- a/operator/internal/controller/podclique/reconciler.go
+++ b/operator/internal/controller/podclique/reconciler.go
@@ -84,17 +84,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return r.triggerDeletionFlow(ctx, logger, pclq).Result()
 	}
 
-	doneSpec := grovemetrics.StartOperation(controllerName, "reconcile_spec")
 	reconcileSpecFlowResult := r.reconcileSpec(ctx, logger, pclq)
-	doneSpec()
-
-	doneStatus := grovemetrics.StartOperation(controllerName, "reconcile_status")
 	statusReconcileResult := r.reconcileStatus(ctx, logger, pclq)
-	doneStatus()
 
 	if ctrlcommon.ShortCircuitReconcileFlow(statusReconcileResult) {
-		return statusReconcileResult.Result()
+		result, err := statusReconcileResult.Result()
+		grovemetrics.RecordReconcileError(controllerName, err)
+		return result, err
 	}
 
-	return reconcileSpecFlowResult.Result()
+	result, err := reconcileSpecFlowResult.Result()
+	grovemetrics.RecordReconcileError(controllerName, err)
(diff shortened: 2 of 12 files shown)


shayasoolin:
Hi @AsadShahid04 thanks for the PR. I think there's some confusion as while you follow the original issue #498, there is a follow-up GREP issue #499 which is the one with the details and full discussion. There you can see the actual (two) metrics we agreed can be implemented.
Please check that and revisit the PR contents accordingly.

AsadShahid04:
Thanks for pointing me to PR #499, @shayasoolin — I had based the implementation on the original issue #498 and missed that the GREP proposal had narrowed scope significantly after review.

Pushed a revision (2fef9e21e89d1f412345b1120cdac62b75fe0ebc) that reduces to the two metrics agreed in PR #499:

- **`grove_operator_status_update_conflict_total{controller, target_kind}`** — recorded at each `Status().Update()` / `Status().Patch()` call site when a 409 Conflict occurs, with `target_kind` set to the Grove resource kind (PodCliqueSet / PodClique / PodCliqueScalingGroup)
- **`grove_operator_reconcile_errors_total{controller, error_type}`** — recorded at the `Reconcile()` exit path, categorized into: `conflict`, `not_found`, `invalid`, `transient`, `server_error`, `unknown`

Removed: `ObservedReconciler` wrapper, `StartOperation` / sub-operation histogram, and the three reconcile-level metrics that duplicate `controller_runtime_reconcile_total` / `controller_runtime_reconcile_time_seconds` / `controller_runtime_active_workers`.

shayasoolin (changes requested):
Thanks for iterating on this and for narrowing the PR toward the two-metric direction discussed in GREP-498. The overall shape is much closer now: no wrapper, no duplicate reconcile-level metrics, and no sub-operation histogram. I found a few implementation gaps that should be addressed before this lands. Process note: GREP-498 / PR #499 is not approved or merged yet, so there may still be requested changes later if that proposal changes before landing; the items below are based on the current proposal text and discussion, not on a finalized contract.

1. `grove_operator_status_update_conflict_total` is currently recorded only in the three `reconcileStatus` methods, but the controllers also write the same Grove status subresources from spec/update-progress/error-recording paths. For example, `podcliqueset/reconcilespec.go` updates the generation hash and observed generation, `podclique/reconcilespec.go` patches update progress and observed generation, `podcliquescalinggroup/reconcilespec.go` does the same for PCSG, and `controller/common/reconcileerrorrecorder.go` patches `LastErrors`. Conflicts from these status writes would not be counted, so the metric would under-report the contention signal this proposal is trying to expose. Please record the conflict metric on all PCS/PCLQ/PCSG `Status().Update()` / `Status().Patch()` error paths where `apierrors.IsConflict(err)` is true.

2. `grove_operator_reconcile_errors_total` is not recorded on all reconcile error exits. Some early returns bypass `RecordReconcileError`, including initial `Get*` failures and deletion-flow failures in the reconcilers. Since this is intended to be a reconcile-loop error metric, each returned reconcile error should increment exactly one `error_type` series. A small helper around the final `(ctrl.Result, error)` return values in each `Reconcile` method may make this easier to keep consistent.

3. The `error_type` label set does not match the current GREP text in PR #499. That proposal currently describes bounded Kubernetes classes: `not_found`, `conflict`, `already_exists`, `timeout`, `forbidden`, `invalid`, `rate_limited`, and `internal`. The implementation currently uses `transient`, `server_error`, and `unknown`, and does not classify `AlreadyExists` or `Forbidden`. Please align the classifier and tests with the current GREP direction, unless the GREP changes before this implementation lands.

PR #634, opened 2026-05-28
feat: add pod clique pod index to pod names

### LINKED ISSUE #635: Add pod clique pod index to pod name for better visibility

Pod clique set and pod clique scaling group indexes are added to each pod name in order to disambiguate their naming. I would like the pod clique index to also be part of the pod name.

### Why is this needed?

When investigating a pod failure that includes the hostname, such as `ubuntu-0-worker-0`, it isn't possible to easily pinpoint the actual pod name to debug:

```
$ kubectl get po
NAME                    READY   STATUS    RESTARTS   AGE
ubuntu-0-worker-2tnab   1/1     Running   0          7h41m
ubuntu-0-worker-5mfde   1/1     Running   0          7h41m
```

Custom columns can help, but require additional tools or aliases and don't work with standard options like `-o wide`:
```
$ kubectl get pods -o custom-columns=NAME:.metadata.name,INDEX:'.metadata.labels.grove\.io/podclique-pod-index',STATUS:.status.phase
NAME                    INDEX   STATUS
ubuntu-0-worker-2tnab   0       Running
ubuntu-0-worker-5mfde   1       Running
```

Ideally, adding the pod clique pod index would make it easier to pinpoint the exact pod:
```
$ kubectl get po
NAME                      READY   STATUS    RESTARTS   AGE
ubuntu-0-worker-0-2tnab   1/1     Running   0          7h41m
ubuntu-0-worker-1-5mfde   1/1     Running   0          7h41m
```

steved:
#### What type of PR is this?

/kind feature

For better at-a-glance visibility, we'd like to encode the PCLQ pod index in the pod name.

Fixes #635 

* This moves the pod name generation out of `GenerateName` and into Grove. ~Rather than appending a 5-char suffix, we can rely on the randomness of 3 characters since pod names can only conflict with their same index.~
* This tightens the validation to check names against replicas. Previously, if a 45 char name had many replicas it could exceed the hostname limit of 63 characters.
    * A PCS that was previously valid may now be invalid, but that should be theoretically impossible since the pod hostnames are invalid.
* Pod names _can_ now go up to 69 chars, which is allowed (though discouraged), since we're explicitly setting the hostname and subdomain.
* ~Note: there is currently no validation on `/scale` API calls~

```release-note
Pod names now contain the pod clique pod index and may exceed the 63-character DNS limit
```

_N/A_


operator/internal/controller/podclique/register.go
```diff
@@ -317,22 +316,14 @@ func mapPodGangToPCLQs() handler.MapFunc {
 			if len(podGroup.PodReferences) == 0 {
 				continue
 			}
-			podRefName := podGroup.PodReferences[0].Name
```
danbar2, line 320: why removed?

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -989,33 +1059,45 @@ func (v *pcsValidator) validatePodCliqueUpdate(oldCliques []*grovecorev1alpha1.P
...
-//
-// Constraints:
-// - Random string + hyphens: 10 chars for PCSG pods, 8 chars for non-PCSG pods
-// - Max sum of all resource name characters: 45 chars
-func validatePodNameConstraints(pcsName, pcsgName, pclqName string) error {
-	// Check resource name constraints
-	resourceNameLength := len(pcsName) + len(pclqName)
+// <pcs-name>-<pcs-index>-<pclq-name>-<pod-index>-<random>
+func validatePodNameConstraints(pcsName string, pcsReplicas int32, pcsgName string, pcsgReplicas int32, pclqName string, pclqReplicas int32) error {
+	pclqOwnerNameReplica := apicommon.ResourceNameReplica{Name: pcsName, Replica: int(max(0, pcsReplicas-1))}
```
danbar2: why are you using the max between 0 to max replicas instead of the exact pcs replica index? same question for all usages of max replicas in the ResourceNameReplica builder

operator/api/core/v1alpha1/podclique.go
```diff
@@ -80,6 +80,15 @@ type PodCliqueSpec struct {
 	ScaleConfig *AutoScalingConfig `json:"autoScalingConfig,omitempty"`
 }
 
+// MaxReplicas returns the maximum number of replicas for this PodClique.
+func (p PodCliqueSpec) MaxReplicas() int32 {
+	replicas := p.Replicas
```
yankay, line 85: Small question — this is a bit asymmetric with
  `PodCliqueScalingGroupConfig.MaxReplicas()` below, which floors at 1 via
  `max(1, ptr.Deref(p.Replicas, 0))`. Is that intentional (e.g. zero replicas is
  a valid scaled-to-zero state for a PodClique), or would it be worth flooring
  here too for consistency? Happy either way — just wanted to flag it. Thanks!

operator/internal/controller/podclique/register.go
steved: The podgroup name is already the fully qualified podclique name (since a09e37014a9f6eea4f35cb9c784fc9dc2affd72c) so I think it was just an oversight that this still parses the pod references. Happy to back it out.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
steved: Some of these could be removed. Podclique replicas must be `>= 1`, but PCS replicas can be 0. This was to ensure we validate the exact index (`replicas: 10` creates PCS 0 - 9) and we don't validate against `-1` if `replicas: 0` is set.

operator/api/core/v1alpha1/podclique.go
steved: Ah yeah, they do have slightly different behaviors. `0` isn't valid for replicas and the field is required, but the default for scaling configs is only set in the API so maybe we just trust that and deref it directly.

https://github.com/ai-dynamo/grove/blob/43987df7d550241a93fd20df19e28922ce6f3d81/operator/internal/webhook/admission/pcs/validation/podcliqueset.go#L512

https://github.com/ai-dynamo/grove/blob/43987df7d550241a93fd20df19e28922ce6f3d81/operator/internal/webhook/admission/pcs/validation/podcliqueset.go#L360

https://github.com/ai-dynamo/grove/blob/43987df7d550241a93fd20df19e28922ce6f3d81/operator/api/core/v1alpha1/podcliqueset.go#L371

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -494,7 +499,12 @@ func (v *pcsValidator) validateScalingGroupPodCliqueNames(pcsgName string, allPc
 
 	// validate scaling group  PodClique pods names are valid.
 	for i, pclqName := range pclqNameInScalingGrp {
-		if err := validatePodNameConstraints(v.pcs.Name, pcsgName, pclqName); err != nil {
+		maxPCLQReplicas := int32(1)
+		if templateSpec := componentutils.FindPodCliqueTemplateSpecByName(v.pcs, pclqName); templateSpec != nil {
+			maxPCLQReplicas = templateSpec.Spec.MaxReplicas()
```
unmarshall: If HPA is not used but lets say dynamo planner is used, then ScaleConfig is not set. This means that at best you will take template replicas. This is not an accurate representation of real time replicas since PCLQ and PCSG have their own `Scale` subresources and any scale-out will directly change the PCLQ and PCSG replicas and will never be reflected in PCS.  So at best this check is only valuable when you solely look at PCS spec (using templates) - but with scale-outs this no longer provides any guard.

Therefore `MaxReplicas` name for the function is highly misleading - as it gives an impression that it always returns an accurate value during the lifetime of the PCS but that is not really true.
On top of that the doc string:
```
// MaxReplicas returns the maximum number of replicas for this PodClique.
```
is also very misleading.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -494,7 +499,12 @@ func (v *pcsValidator) validateScalingGroupPodCliqueNames(pcsgName string, allPc
...
+		}
+
+		if err := validatePodNameConstraints(v.pcs.Name, v.pcs.Spec.Replicas, scalingGroupConfig.Name, scalingGroupConfig.MaxReplicas(), pclqName, maxPCLQReplicas); err != nil {
```
unmarshall: If you wish to keep this then clearly document that this is `best-effort` and `pre-flight` and does not reflect the current state of PCS and its constituent PCLQs and PCSGs.

operator/internal/controller/podclique/register.go
unmarshall: yes that is correct. The PodGroup name is the PCLQ FQN and can be directly used. Thanks for the fix, this was missed.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -504,6 +514,66 @@ func (v *pcsValidator) validateScalingGroupPodCliqueNames(pcsgName string, allPc
...
+	allErrs := field.ErrorList{}
+	for _, ownerName := range ownerNames {
+		allErrs = append(allErrs, validatePodResourceClaimReferenceNames(ownerName, cliqueTemplateSpec.Spec.MaxReplicas(), cliqueTemplateSpec.ResourceSharing, fldPath)...)
+	}
+	return allErrs
+}
+
+func validatePodResourceClaimReferenceNames(ownerName string, maxReplicas int32, refs []grovecorev1alpha1.ResourceSharingSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	maxReplicaIndex := int(max(0, maxReplicas-1))
```
unmarshall: maxReplicas is not the correct representation of what the actual max is. This value is solely derived from template spec.

operator/api/common/namegen.go
```diff
@@ -20,6 +20,14 @@ import (
...
 
 	"github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
+
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
+)
+
+const (
+	// PodNameRandomSuffixLength is the length of the random suffix Grove adds to
+	// generated Pod names.
+	PodNameRandomSuffixLength = 5
```
renormalize, line 30: The PR description says 3, but I see a 5 here?

If the length is already going to exceed 63 characters in the pod name, what reason do we have to generate the random suffix ourselves, instead of relying the API Server? We should just validate that the hostname is 63 characters, and let the API Server generate the actual pod name.

The API Server performs an internal retry when pod creation fails due to a conflict, instead of immediately returning a HTTP 409, when `GenerateName` is used. We should utilize native features instead of doing the suffix generation ourselves.

operator/api/common/namegen.go
```diff
@@ -77,6 +85,21 @@ func GeneratePodCliqueScalingGroupName(pcsNameReplica ResourceNameReplica, pclqS
 	return fmt.Sprintf("%s-%d-%s", pcsNameReplica.Name, pcsNameReplica.Replica, pclqScalingGroupName)
 }
 
+// GeneratePodName generates a Pod name based on the PodClique FQN, Pod index, and a random suffix.
+func GeneratePodName(pclqName string, podIndex int) string {
+	return GeneratePodNameWithSuffix(pclqName, podIndex, utilrand.String(PodNameRandomSuffixLength))
+}
+
+// GeneratePodNameWithSuffix generates a Pod name with a caller-provided suffix.
+func GeneratePodNameWithSuffix(pclqName string, podIndex int, suffix string) string {
```
renormalize, line 94: https://github.com/ai-dynamo/grove/pull/634#discussion_r3371560814

operator/api/common/namegen.go
```diff
@@ -77,6 +85,21 @@ func GeneratePodCliqueScalingGroupName(pcsNameReplica ResourceNameReplica, pclqS
 	return fmt.Sprintf("%s-%d-%s", pcsNameReplica.Name, pcsNameReplica.Replica, pclqScalingGroupName)
 }
 
+// GeneratePodName generates a Pod name based on the PodClique FQN, Pod index, and a random suffix.
+func GeneratePodName(pclqName string, podIndex int) string {
```
renormalize, line 89: https://github.com/ai-dynamo/grove/pull/634#discussion_r3371560814

operator/api/core/v1alpha1/podclique.go
```diff
@@ -80,6 +80,15 @@ type PodCliqueSpec struct {
 	ScaleConfig *AutoScalingConfig `json:"autoScalingConfig,omitempty"`
 }
 
+// MaxReplicas returns the maximum number of replicas for this PodClique.
+func (p PodCliqueSpec) MaxReplicas() int32 {
```
renormalize: Should have a pointer receiver, following the conventions in the rest of the API.

```suggestion
func (p *PodCliqueSpec) MaxReplicas() int32 {
```

I would also prefer if the method name is changed, since we have fields called `MaxReplicas`, and the method does not identically return that value.

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -334,6 +335,15 @@ type PodCliqueScalingGroupConfig struct {
 	TopologyConstraint *TopologyConstraint `json:"topologyConstraint,omitempty"`
 }
 
+// MaxReplicas returns the maximum number of replicas allowed with this scaling group config.
+func (p PodCliqueScalingGroupConfig) MaxReplicas() int32 {
```
renormalize: https://github.com/ai-dynamo/grove/pull/634#discussion_r3371645786

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -334,6 +335,15 @@ type PodCliqueScalingGroupConfig struct {
...
+	replicas := max(1, ptr.Deref(p.Replicas, 0))
```
renormalize: As mentioned in https://github.com/ai-dynamo/grove/pull/634#discussion_r3331748720, we can remove `max` and `Deref`.

operator/internal/controller/podclique/register.go
renormalize: @steved is correct. We did not need that extra code. Thanks for the clean-up.

operator/internal/controller/podclique/components/pod/pod.go
```diff
@@ -150,10 +150,10 @@ func (r _resource) buildResource(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grov
...
 	labels := getLabels(pclq.ObjectMeta, pcsName, podGangName, pcsReplicaIndex, podIndex)
 	pod.ObjectMeta = metav1.ObjectMeta{
-		GenerateName: fmt.Sprintf("%s-", pclq.Name),
-		Namespace:    pclq.Namespace,
-		Labels:       labels,
-		Annotations:  pclq.Annotations,
+		Name:        apicommon.GeneratePodName(pclq.Name, podIndex),
+		Namespace:   pclq.Namespace,
+		Labels:      labels,
+		Annotations: pclq.Annotations,
```
renormalize, line 156: https://github.com/ai-dynamo/grove/pull/634#discussion_r3371560814

```suggestion
		GenerateName: fmt.Sprintf("%s-", apicommon.GeneratePodHostname(pclq.Name, podIndex)),
```

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
renormalize: Can we rename the variable name to `pcsNameReplica`?
```suggestion
	pcsNameReplica := apicommon.ResourceNameReplica{Name: pcsName, Replica: int(max(0, pcsReplicas-1))}
```

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -989,33 +1059,45 @@ func (v *pcsValidator) validatePodCliqueUpdate(oldCliques []*grovecorev1alpha1.P
...
 
-	if resourceNameLength > maxCombinedResourceNameLength {
-		if pcsgName != "" {
-			return fmt.Errorf("combined resource name length %d exceeds 45-character limit required for pod naming. Consider shortening: PodCliqueSet '%s', PodCliqueScalingGroup '%s', or PodClique '%s'",
-				resourceNameLength, pcsName, pcsgName, pclqName)
-		}
-		return fmt.Errorf("combined resource name length %d exceeds 45-character limit required for pod naming. Consider shortening: PodCliqueSet '%s' or PodClique '%s'",
-			resourceNameLength, pcsName, pclqName)
+	pclqFQN := apicommon.GeneratePodCliqueName(pclqOwnerNameReplica, pclqName)
+	podIndex := int(max(0, pclqReplicas-1))
```
renormalize: As you mentioned, this `max` can be removed since `pclqReplicas` is always `>=1`.

renormalize (commented):
Thanks for the PR! A few comments from me.

operator/api/common/namegen.go
steved: The API server truncates to 63: https://github.com/kubernetes/kubernetes/blob/v1.36.2/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L51

The 3 char suffix was changed back to 5 after some internal discussion, my fault.

renormalize: Okay, that makes sense. I was completely unaware of the truncation. Thanks for pointing me to this!

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -506,6 +516,66 @@ func (v *pcsValidator) validateScalingGroupPodCliqueNames(pcsgName string, allPc
...
+	for i, ref := range refs {
+		var rcName string
+
+		switch ref.Scope {
+		case grovecorev1alpha1.ResourceSharingScopeAllReplicas:
+			rcName = resourceclaim.RCName(ownerName, &ref, nil)
+		case grovecorev1alpha1.ResourceSharingScopePerReplica:
+			rcName = resourceclaim.RCName(ownerName, &ref, &maxReplicaIndex)
+		default:
+			continue
```
renormalize, line 570: Nit: Since this is the validation flow, we can return an error here that the specified resource sharing scope is neither `AllReplicas`, nor `PerReplica`. The CRD validation markers already enforce that only one of these two declared types is used, but perhaps we can be safe and return an error here in case things are missed in the API if/when it is extended in the future. Feel free to dismiss this since I don't feel too strongly about this.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -999,33 +1069,45 @@ func (v *pcsValidator) validatePodCliqueUpdate(oldCliques []*grovecorev1alpha1.P
...
 	}
 
-	if resourceNameLength > maxCombinedResourceNameLength {
-		if pcsgName != "" {
-			return fmt.Errorf("combined resource name length %d exceeds 45-character limit required for pod naming. Consider shortening: PodCliqueSet '%s', PodCliqueScalingGroup '%s', or PodClique '%s'",
-				resourceNameLength, pcsName, pcsgName, pclqName)
-		}
-		return fmt.Errorf("combined resource name length %d exceeds 45-character limit required for pod naming. Consider shortening: PodCliqueSet '%s' or PodClique '%s'",
-			resourceNameLength, pcsName, pclqName)
+	pclqFQN := apicommon.GeneratePodCliqueName(pcsNameReplica, pclqName)
```
renormalize: Nit: rename the variable name for semantic correctness.

```suggestion
	parentNameReplica := apicommon.ResourceNameReplica{Name: pcsName, Replica: int(pcsReplicas - 1)}
	if pcsgName != "" {
		pcsgFQN := apicommon.GeneratePodCliqueScalingGroupName(parentNameReplica, pcsgName)
		parentNameReplica = apicommon.ResourceNameReplica{Name: pcsgFQN, Replica: int(pcsgReplicas - 1)}
	}

	pclqFQN := apicommon.GeneratePodCliqueName(parentNameReplica, pclqName)
```

renormalize (dismissed):
Thanks for addressing the comments @steved!

The PR looks good now - just a couple nits. We can push this in once we have more approvals from reviewers.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
steved: makes sense, added!

steved:
> Thanks for addressing the comments @steved!
> 
> The PR looks good now - just a couple nits. We can push this in once we have more approvals from reviewers.

Thanks for taking a look @renormalize!

operator/api/core/v1alpha1/podclique.go
```diff
@@ -88,6 +80,15 @@ type PodCliqueSpec struct {
 	ScaleConfig *AutoScalingConfig `json:"autoScalingConfig,omitempty"`
 }
 
+// MaxAllowedReplicas returns the maximum number of replicas that can be configured for this PodClique.
+func (p *PodCliqueSpec) MaxAllowedReplicas() int32 {
```
unmarshall, line 84: Grove allows external pluggable horizontal autoscalers. `ScaleConfig` is only applicable when you wish to use HPA, otherwise this is not set (is nil). Dynamo uses `planner` to horizontally scale replicas. So in this case the max allowed will be limited by spec.Replicas which will not be correct.

unmarshall (changes requested):

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -401,6 +400,15 @@ type PodCliqueScalingGroupConfig struct {
 	TopologyConstraint *TopologyConstraint `json:"topologyConstraint,omitempty"`
 }
 
+// MaxAllowedReplicas returns the maximum number of replicas that can be configured with this scaling group config.
+func (p *PodCliqueScalingGroupConfig) MaxAllowedReplicas() int32 {
```
unmarshall, line 404: Please see [comment](https://github.com/ai-dynamo/grove/pull/634/changes#r3485159534)

unmarshall (changes requested):

operator/api/core/v1alpha1/podclique.go
steved: Just so I understand - theoretically, this is still accurate, but because we're not validating the PCLQ and `/scale` admission we're missing changes that are applied directly? Would you recommend we add validation to those admission paths?

steved: Bump @unmarshall if you have any thoughts, thanks!

operator/internal/webhook/admission/scale/validation/validation.go
```diff
@@ -0,0 +1,292 @@
...
+			pcsNameReplica,
+			config,
+			nil,
+			i,
+			templatesByName,
+			templateIndexes,
+			templatePath,
+		)...)
+
+		for replica := range pcs.Spec.Replicas {
```
yankay: Thanks @steved 
Could we iterate over the existing PCSGs instead of every value up to spec.replicas? Since spec.replicas is unbounded, a large value could keep the webhook busy even after its 10-second timeout.

steved: yep - fair enough. pushed a fix.

PR #640, opened 2026-06-01
Introduce GREP-393: Coherent updates

### LINKED ISSUE #639: [GREP]: Coherent update
### How would you describe the priority of this documentation request

Critical (currently preventing usage)

### Describe the future/missing documentation

Introduce a proposal doc detailing how the approach to realise `coherent` updates.

### Where have you looked?

_No response_

### Code of Conduct

- [x] I agree to follow Grove's Code of Conduct
- [x] I have searched the project's documentation and have found no duplicates for this request

unmarshall:
Introduces the GREP-393 proposal document for the Coherent rolling update strategy for PodCliqueSet. The proposal covers motivation, goals, design decisions, API changes, and the implementation approach for coordinating multi-version rolling updates across PodCliques.

#### What type of PR is this?

/kind documentation

Fixes #639 

```release-note
NONE
```

```docs
Introduce GREP: 393 Coherent updates
```


docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+
+### Goals
+
+* Enable rolling updates at a user-chosen granularity: the user selects which components to update in one event, and the system replaces them in lockstep within each MVU so cross-version mixing within the boundary is impossible.
+* Each MVU is gang-scheduled as a single unit, so an MVU's pods either come up together at the new version or none of them do.
+* Preserve `PodCliqueSet` availability during rolling updates to serve incoming traffic with sets of compatible interdependent components.
+
+### Non-Goals
+
+* Re-use of topology optimized resources during rolling update using resource reservations. Will be handled in future as a separate feature.
```
Ronkahn21: What does it mean topology optimized resources?

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+* Explicit support for `maxSurge` and `maxUnavailable` API. However, similar concurrency controls and functionality will be supported in future.
+* User-configurable concurrency control during a coherent update — neither the number of `PodCliqueSet` replicas updated simultaneously nor the number of MVU iterations in flight per replica is configurable in the current iteration. Both default to one. Configurable knobs will be supported in future.
+* `scale-out` and `scale-in` of scale sub-resources (`PodClique`, `PodCliqueScalingGroup`, `PodCliqueSet`) during a coherent update. The current iteration blocks these operations for the duration of a coherent update. How scale operations should compose with an in-flight coherent update will be supported in future.
```
Ronkahn21: In what level, meaning any pcsg/pclq of as part of the update, any resource in any pcs replica or anything in between

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+    // +kubebuilder:default=Coherent
+    Type UpdateStrategyType `json:"type,omitempty"`
+}
+```
+
+#### `PodGangMap` — new CRD
+
+`PodGangMap` (PGM) is a new namespaced custom resource that captures the **desired-state mapping between PodGangs and their constituent PodClique pod counts and PodCliqueScalingGroup replica indices** for a single `PodCliqueSet` replica. One `PodGangMap` exists per PCS replica, named `<pcs-name>-<pcs-replica-index>`.
+
+`PodGangMap` has no `Status` subresource as it only captures the desired-state. Mappings captured in this resource are read by the `PodGang`, `PodClique` and `PodCliqueScalingGroup` reconcilers.
```
Ronkahn21: we have new podGang reconciler?, today the pcs is the one handle the podGang

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+
+```go
+type PodCliqueSetReplicaUpdateProgress struct {
+    // ... existing fields ...
+
+    // InFlightPodGangs are the names of PodGangs that are part of the current update iteration
+    // for this replica. The orchestrator waits for all of them to become Available before
+    // advancing to the next iteration.
+    // +optional
+    InFlightPodGangs []string `json:"inFlightPodGangs,omitempty"`
```
Ronkahn21, line 382: It used by any rolling update type

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+##### Capturing MVU scope in `PodCliqueSetUpdateProgress`
+
+Two new fields are added to `PodCliqueSetUpdateProgress` to capture the **frozen set of components** detected as out-of-date when a coherent update started. This is used to deterministically compute the MVU template.
+
+```go
+type PodCliqueSetUpdateProgress struct {
+    // ... existing fields ...
+
+    // UpdatedStandalonePodCliques captures the names of standalone PodCliques whose pod template
+    // was detected as out-of-date when this coherent update started. The set is frozen for the
```
Ronkahn21: I am not sure I understand the sentence, it will include all the standalone pclq that are park of this update
or standalone pclq  that are not part of this MVU update but the current pod template is not match desire ?

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+
+    // ErrorMessage captures the reason the update of this replica is stalled or failing, if any.
+    // +optional
+    ErrorMessage *string `json:"errorMessage,omitempty"`
+}
+```
+
+`InFlightPodGangs` is the orchestrator's hand-off to the PodGangMap component and back. In addition it also provide visibility into what PodGang(s) are currently getting updated. A coherent update for a PodCliqueSet replica proceeds in two phases: 
+
+* The **MPG phase**: In this phase MPGs are taken up one at a time — each MPG is rolled to the new hash, and only once it reports `PodGangConditionTypeAvailable=True` does the orchestrator move on to the next MPG
```
Ronkahn21: Where is the step we calc the gang three, what belong to each podGang ?

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+    // ErrorMessage captures the reason the update of this replica is stalled or failing, if any.
+    // +optional
+    ErrorMessage *string `json:"errorMessage,omitempty"`
+}
+```
+
+`InFlightPodGangs` is the orchestrator's hand-off to the PodGangMap component and back. In addition it also provide visibility into what PodGang(s) are currently getting updated. A coherent update for a PodCliqueSet replica proceeds in two phases: 
+
+* The **MPG phase**: In this phase MPGs are taken up one at a time — each MPG is rolled to the new hash, and only once it reports `PodGangConditionTypeAvailable=True` does the orchestrator move on to the next MPG 
+* The **TPG phase**: Once every MPG is at the new hash, all remaining TPGs are rolled together in a single iteration.
```
Ronkahn21: What does it mean `Once every MPG is at the new hash` ? is it means the podGang is ready ?

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+
+**New behavior under Coherent.** New PCS deployments no longer produce a BPG/SPG layout. Instead, the PodGangMap component creates the replica's `PodGangMap` at PCS deployment time and seeds it from the PCS template using the same MVU template that drives a coherent update. The MVU template captures the smallest set of components that must come up together (MinAvailable replicas of every standalone PCLQ and every PCSG), and the PGM entry builder partitions the spec across:
+
+- **One MPG (Minimum-Viable PodGang)** carrying MinAvailable replicas of every standalone PCLQ and every PCSG.
+- **Zero or more TPGs**, one per tranche of PCSG replicas above MinAvailable. TPGs depend on the MPG via `DependsOn`, so the scheduler places them only after the MPG is up. 
+- **Excess standalone-PCLQ replicas** (when a standalone PCLQ has more than MinAvailable) roll into the highest index MPG entry as additional pods — no separate PodGang is created for them.
```
Ronkahn21: at startup is act basically the same as BPG and SPG ?

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+
+#### Story 2
+
+As a platform engineer managing a large-scale disaggregated inference fleet with many `PodCliqueSet` replicas, I need to control the blast radius of a rolling update. If a new version turns out to be faulty, I want to limit the number of replicas upgraded simultaneously so that a bug is contained to a small fraction of live traffic. I also need the ability to pause an in-progress update and roll back all affected MVUs to the previously known-good version, restoring compatibility within each replica instantly without manual intervention.
+
+#### Story 3
+
+As an ML infrastructure team member deploying a disaggregated inference system where the prefill tier and decode tier are updated on different release cadences, I need to independently update only the decode `PodClique` (e.g., to pick up a memory-efficiency fix) without touching the prefill `PodClique`. The system should recognise that this is a backward compatible, single-component update and updates decode pods incrementally (up to a configurable concurrency limit), and leave prefill pods untouched — all without requiring a full MVU replacement.
+
+### Limitations/Risks & Mitigations
```
Ronkahn21, line 146: Wanted to flag something I don't think is covered in the limitations section yet.

Before a coherent update, excess PCSG replicas sit in isolated single-component SPGs — `{prefill[1]}` and `{decode[1]}` are separate gangs. So if the scheduler needs to preempt due to resource pressure, only decode[1] gets preempted. Prefill[1] keeps running. After the update those two replicas are gang-scheduled together in `MPG-NEW2: {prefill[1]v1, decode[1]v1}` — the scheduler will preempt the entire gang, taking down both prefill and decode together. This persists permanently in steady state, not just during the update window. Worth calling out in the limitations section.

Separate question — after the first MPG is up and available, why do we continue creating additional MPGs for the remaining replicas rather than keeping them as TPGs that depend on the new MPG, same as the initial setup? The first MPG already satisfies `minAvailable`, so the remaining replicas are tail capacity. Keeping them as TPGs would preserve the pre-update isolation and avoid the preemption blast radius problem entirely. If the right behavior differs across use cases, we might need to expose this as a user-controlled knob on the update strategy API.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+- **PodCliqueScalingGroup.** When `PodCliqueScalingGroup.Spec.Replicas` shrinks, the PodCliqueScalingGroup reconciler's PodClique component runs a deterministic tier walk over the existing PodGang names — legacy SPG entries first, then unified-naming entries — sorted by trailing PodGang-name suffix descending, and pops the highest replica index from each entry until the scale-in count is satisfied. So although the PCSG-side pick is fully deterministic, the work — selecting which entry to drain and which replica index to pop from it — still happens inside the PodCliqueScalingGroup reconciler. The PodCliqueScalingGroup reconciler records the resulting decrement in `PodCliqueScalingGroupStatus.PodGangMapping` so the PodGangMap component can rebuild PGM from it on the next reconcile.
+
```
Ronkahn21, line 435: What happens here during a post-update scale-in when a PCSG replica is part of an MPG with another component? After a coherent update, replicas that were previously isolated in single-component SPGs are now gang-scheduled together — e.g. `MPG-NEW3: {prefill[2]v1, decode[2]v1}`. If the user scales in decode from 3 → 2, the tier walk hits `MPG-NEW3` and removes `decode[2]`, leaving `prefill[2]` orphaned in a half-empty MPG with no decode counterpart. The two PCSG reconcilers make their scale-in decisions independently with no cross-component coordination — prefill has no visibility that its gang partner was removed. Is this expected? Does the doc need to address what happens to the orphaned component?

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: Clarified.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: ah that is a honest mistake. There is no `PodGang` reconciler yet - at least not in this PR. I have now reworded it so it does not create confusion.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: No this is only used by Coherent update. For RollingRecreate this will be nil.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: When a PCS coherent update starts - this will capture the names of all the standalone PCLQs whose PodSpec's have been updated.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: question is not clear. Can you rephrase it?

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: reworded to avoid confusion

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: not sure what the question is. But in case your question is that when a new PCS is deployed will it continue to create BPG and SPG - then the answer is NO. This is also mentioned in this section. But if that is not what you wanted to ask, then perhaps you can rephrase the question clearly.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+### Goals
+
+* Enable rolling updates at a user-chosen granularity: the user selects which components to update in one event, and the system replaces them in lockstep within each MVU so cross-version mixing within the boundary is impossible.
+* Each MVU is gang-scheduled as a single unit, so an MVU's pods either come up together at the new version or none of them do.
+* Preserve `PodCliqueSet` availability during rolling updates to serve incoming traffic with sets of compatible interdependent components.
+
+### Non-Goals
+
+* Re-use of topology optimized resources during rolling update using resource reservations. Will be handled in future as a separate feature.
+* Explicit support for `maxSurge` and `maxUnavailable` API. However, similar concurrency controls and functionality will be supported in future.
```
shayasoolin: Mention the defaults that are going to be used for now (maxSurge=0, maxUnavailable=1, right?)

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
...pport disaggregated LLM serving, where stages like prefill and decode run as separate, networked components. While this can improve throughput and resource efficiency, it complicates standard deployment practices. A standard Kubernetes rolling update inevitably creates a period where old and new version pods run at the same time and may communicate. In disaggregated systems, this cross-version communication is unsafe, so applications must prevent it. However, once cross-version communication is disabled, rolling updates introduce another issue: different components often update at different rates, which leads to mismatched pools of compatible instances. For example, you might still have many old-version prefill instances running while most old-version decode instances have already been replaced. Since old prefill can only talk to old decode, a portion of the prefill capacity becomes unusable due to the lack of matching decode capacity. This kind of mismatch reduces effective end-to-end serving capacity during the update. Our goal is to design a rolling update strategy that maintains balanced, compatible capacity across components, so version upgrades do not reduce serving capacity.
```
shayasoolin: Consider mentioning distributed inference servers, rather than the underlying runtime engines:
`LLM serving systems such as vLLM, llm-d, and NVIDIA Dynamo support disaggregated serving, where...`
I think it makes more sense as this is the application layer that the user is using directly under Grove. Plus, it mentions `NVIDIA Dynamo` which is a significant use-case for Grove.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+## Proposal
+
+The GREP introduces a new rolling update strategy, named **Coherent Rolling Updates**, based on the concept of a **Minimal Viable Unit** (a.k.a. MVU).
+
+An MVU is **the set of `MinAvailable` replicas of every component the user has changed in this update event**. The user implicitly defines the compatibility boundary: it is exactly the set of components touched in this update event. Changing only one component (e.g. only `decode`) produces an MVU that contains MinAvailable replicas of just that component; changing multiple components in a single update event (e.g. `prefill` and `decode` together for a non-backward-compatible upgrade) produces an MVU that contains MinAvailable replicas of each.
+
+Components the user has not changed in this update event are not part of the MVU. They continue to serve traffic in their existing PodGangs and are not co-rolled. This is what makes Coherent updates granular: the disruption per update event is bounded by what the user actually changed.
```
shayasoolin: _"They continue to serve traffic"_ might not be accurate, as those components that are not updated are not necessarily related to traffic serving (e.g. Dynamo's Planner component). Perhaps it's better to simply say they continue to run, rather than serve.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+The system must update prefill and decode pods together as a single atomic unit (MVU), ensuring that at no point does an old-version prefill hand off a KV-cache block to a new-version decode, or vice versa. While the update is in progress, replicas that have not yet been updated must continue serving traffic using only old-version components, and replicas that have already been updated must serve using only new-version components. The update should proceed replica-by-replica (or MVU-by-MVU within a replica) without requiring a full deployment restart, so that overall serving capacity is preserved throughout the rollout.
+
+#### Story 2
+
+As a platform engineer managing a large-scale disaggregated inference fleet with many `PodCliqueSet` replicas, I need to control the blast radius of a rolling update. If a new version turns out to be faulty, I want to limit the number of replicas upgraded simultaneously so that a bug is contained to a small fraction of live traffic. I also need the ability to pause an in-progress update and roll back all affected MVUs to the previously known-good version, restoring compatibility within each replica instantly without manual intervention.
```
shayasoolin: Is this user story part of what's covered by the GREP? I think not, see the non-goals above.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+### Limitations/Risks & Mitigations
+
+The current iteration of Coherent Rolling Updates carries the following known limitations. Each is a deliberate scope boundary, with planned follow-up where applicable.
+
+1. **Re-triggering an update while one is in progress is not yet well-supported.** If the user mutates the PCS spec a second time while a coherent update is mid-flight, the system has no formal cancel-and-restart semantics for the in-flight update. The current behavior — the in-flight iteration completes against the previous hash, then the next iteration mints under the newer hash — has not been hardened against corner cases (e.g., a second update that reverts the first while iterations are still running, or that changes the MVU template shape).
+
+   *Mitigation:* operators should let an in-flight coherent update reach completion before triggering another spec change.
```
shayasoolin: What is considered 'completion'? What if the update involves changing the number of replicas or requested resources such that the workload is stuck pending? Will it be possible to perform a second update to mitigate that (i.e. reducing the number of replicas/resources)?

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+```
+PodGang-1: {  # this is the base PodGang that must be scheduled
+  frontend: 5 Pod,
+  prefill: { prefill-leader: 1 Pod, prefill-worker: 3 Pods (minAvailable=2) },
+  decode: { decode-leader:  1 Pod, decode-worker: 4 Pods (minAvailable=2) }
+}
+# In short represented as {5F, 1P, 1D}
+```
+
+*At time T4 (T4 > T3)* - An update is triggered
```
shayasoolin:
```suggestion
*At time T4 (T4 > T3)* - An update is triggered.
```
(This and next lines are rendered in one line, missing period in the end of the first)

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+3. **Gate on prior iteration availability.** The PodGangMap component will not emit the next iteration until every previously-generated new-hash PodGang reports `PodGangConditionTypeAvailable=True`. This is how "exactly one MVU in flight at a time" is realized.
+
+4. **Materialize PodGangs.** The PodGang component observes the new PGM entry, creates the corresponding `PodGang` resource with `MinReplicas` set to the gang's MinAvailable, and stamps the labels described in [Labels on `PodGang` resources](#labels-on-podgang-resources).
+
+5. **Take down old, recreate new.** The PodClique reconciler (for standalone PCLQs) and the PodCliqueScalingGroup reconciler (for PCSG-owned PCLQs) consume the PGM. Old-hash pods on the take-down indices are deleted; replacements are created **schedule-gated** against the new PodGang. The schedule gate prevents the scheduler from placing any new-version pod until the entire MPG is present, preserving gang-scheduling semantics.
```
oleg-kushniriov: Batch pod deletion during update iterations

The PCLQ and PCSG reconcilers currently delete take-down pods individually, issuing one DELETE call per pod. For large take-down sets — high minAvailable values, many constituent PCLQs per PCSG, or high replica counts — this produces O(n) sequential API-server round-trips and proportional etcd write amplification per iteration.

 The pods already carry grove.io/podcliqueset-generation-hash and grove.io/podgang labels, which are precise enough to drive a DeleteCollection call scoped to exactly the old-hash pods being replaced in this iteration. Should the pod component switch to label-selector-based bulk deletion instead of per-pod deletion?

 Trade-offs to consider: DeleteCollection is asynchronous (pods enter Terminating, not gone immediately), it bypasses per-pod pre-delete hooks if any, and the reconciler must still gate on all targeted pods reaching Terminated before marking the iteration complete. But for fleets where a single MPG iteration touches dozens of pods, the round-trip reduction is significant.

docs/proposals/393-coherent-rolling-updates/README.md
Ronkahn21: The doc describes InFlightPodGangs as the hand-off signal between the orchestrator and the PodGangMap component, but I'm not finding where the actual composition of each PodGang is computed. Specifically, which step determines which pods/replicas belong to which MPG or TPG entry in the PGM? Is that the PodGangMap component's responsibility on every reconcile, and if so, where in the update flow does it happen relative to the orchestrator setting InFlightPodGangs

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+          - dworker
+```
+
+Prior to update, replicas of each of the child resources of the `disagg-serving` PCS is as shown in the resource YAML above. The initial set of `PodGang`s that are created have the following composition:
+
+*At time T1:*
+
+```
+PodGang-1: {  # this is the base PodGang that must be scheduled
+  frontend (F): 3 Pod,
```
shayasoolin: It's worth mentioning here that the minAvailable for frontend is 2, and is specified as part of the `PodGang` definition.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+Following are the steps demonstrating the creation and update of MPGs during the update:
+```
+Step-1:
+  Take-down set: {2F, 1P, 1D}
+  Recreate order: {2Fv1, 1Pv1, 1Dv1}
+  Expected state: PG: {3F}, 3 * {P}, 2 * {D}, MPG: {2Fv1, 1Pv1, 1Dv1}
+Step-2:
+  Take-down set: {3F}, 3 * {P}, 2 * {D}
+  Recreate order: [{3Fv1, 1Pv1, 1Dv1}] -> [{1Pv1}, {1Dv1}, {1Pv1}]
```
shayasoolin: This is unclear to me, what does the "->" stands for? Why is 3Fv1 replaced by 1Fv1?

docs/proposals/393-coherent-rolling-updates/README.md
Ronkahn21, line 435: Flagging a regression I noticed at initial deployment, no update involved.

Setup: `PCS {8F, 8P, 8D}`, minAvailable=4. Cluster has capacity for only 7D.

**Previous BPG/SPG:**
```
BPG: {4F, 4P, 4D}   schedules fine
SPG: 4*{1P}, 4*{1D}  8th {1D} fails, rest run
Result: 8F, 8P, 7D running
```

**New MPG:**
```
MPG-0: {4F, 4P, 4D}   schedules fine
MPG-1: {4F, 4P, 4D}   fails, only 3D left
Result: 4F, 4P, 4D running
```

The strict gang requirement above minAvailable ends up blocking F and P from running even though the cluster has plenty of capacity for them. And this gets worse the higher minAvailable is set — a larger floor means a bigger MPG and a higher bar to place it. The old layout handled partial capacity gracefully; this one doesn't.

Stepping back — is the intent for initial deployment to always use the new MPG layout, or should it stay as BPG/SPG and only switch to MPG/TPG when a coherent update actually kicks off? Feels like those are two different design positions worth calling out explicitly.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,893 @@
...
+
+**New behavior under Coherent.** New PCS deployments no longer produce a BPG/SPG layout. Instead, the PodGangMap component creates the replica's `PodGangMap` at PCS deployment time and seeds it from the PCS template using the same MVU template that drives a coherent update. The MVU template captures the smallest set of components that must come up together (MinAvailable replicas of every standalone PCLQ and every PCSG), and the PGM entry builder partitions the spec across:
+
+- **One MPG (Minimum-Viable PodGang)** carrying MinAvailable replicas of every standalone PCLQ and every PCSG.
```
Ronkahn21: Flagging a regression in the new MPG layout at initial deployment, no update involved.

Setup: `PCS {8F, 8P, 8D}`, minAvailable=4. Cluster capacity: 7D.

**Previous BPG/SPG:**
```
BPG: {4F, 4P, 4D}   schedules fine
SPG: 4*{1P}, 4*{1D}  8th {1D} fails, rest run
Result: 8F, 8P, 7D running
```

**New MPG:**
```
MPG-0: {4F, 4P, 4D}   schedules fine
MPG-1: {4F, 4P, 4D}   fails, only 3D left
Result: 4F, 4P, 4D running
```

The strict gang requirement above minAvailable blocks F and P from running even though the cluster has capacity for both. This gets worse as minAvailable grows — a larger floor means a larger MPG and a higher bar to place it. The BPG/SPG model tolerated partial capacity gracefully; this layout does not.

Broader question: should initial deployment stay as BPG/SPG and only switch to MPG/TPG when a coherent update is triggered, or is the new layout unified across both with this tradeoff accepted and documented?

@unmarshall @nvrohanv @sanjaychatterjee

docs/proposals/393-coherent-rolling-updates/README.md
sanjaychatterjee: Maybe rephrase this as: 
```
Ensure equal or better topology optimized placement of the workload after rolling update.
```

docs/proposals/393-coherent-rolling-updates/README.md
sanjaychatterjee: Regarding the preemption use-case, it should be possible for the scheduler to preempt a sub-group should it choose to do so, since the minReplicas post-scheduling will be set to 0. This may not be ideal right now but it avoids the the full gang preemption scenario. When the KAI scheduler backend comes online, it should be possible to mutate the PodGroup structure with additional sub-groups when new scaled PCSG replicas are created and that should provide KAI with enough information needed for sub-group preemption.
For the second question, the update process needs to progress MPG-by-MPG so that the application is rolled in a load-balanced manner without compromising availability.

docs/proposals/393-coherent-rolling-updates/README.md
sanjaychatterjee: True, but given the current philosophy on MPG being the functional unit, would it be useful to have a partial deployment of the second MPG and hog resources unnecessarily if the whole MPG cannot be gang-scheduled? But I guess it would also be dependent on the kind of application and so your suggestion of starting with BPG/SPG model is a valid one.

docs/proposals/393-coherent-rolling-updates/README.md
sanjaychatterjee: True, but given the current philosophy on MPG being the functional unit, would it be useful to have a partial deployment of the second MPG and hog resources unnecessarily if the whole MPG cannot be gang-scheduled? But I guess it would also be dependent on the kind of application and so your suggestion of starting with BPG/SPG model is a valid one.

docs/proposals/393-coherent-rolling-updates/README.md
sanjaychatterjee: Ordering. [{3Fv1, 1Pv1, 1Dv1}] -> [{1Pv1}, {1Dv1}, {1Pv1}] means that {3Fv1, 1Pv1, 1Dv1} needs to be scheduled and become ready before any of the TPGs [{1Pv1}, {1Dv1}, {1Pv1}] can be scheduled.

docs/proposals/393-coherent-rolling-updates/README.md
shayasoolin: MPG is the functional unit when it comes to the first unit. But obviously, next MPGs can be partial - just like there are TPGs each representing only one of the involved PCSGs.
Grove should assume that a partial MPG deployment is valid, as otherwise the user should have grouped the components together into a single PCSG.

docs/proposals/393-coherent-rolling-updates/README.md
shayasoolin: Yes got it, worth adding an explanation above that, for clarity.

docs/proposals/393-coherent-rolling-updates/README.md
shayasoolin: If it's unsupported, Grove should block updates during updates (e.g., using a webhook). Please clarify this in the requirements.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
```
shayasoolin: Isn't this supposed to be "One or more MPGs"? For example where the desired number of replicas of each component (PCLQ/PCSG) is `2 * MinAvailable`.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+Grove encodes the MVU's gang-scheduling intent by generating new `PodGang` resources, **MVU PodGangs (MPGs)**, that hold exactly the MVU's pods. PCSG replicas above MinAvailable go into separate **Tail PodGangs (TPGs)** — one per excess replica. Both MPGs and TPGs are gang-scheduled. Each TPG `DependsOn` every MPG of the same generation. The scheduler places TPG pods only after all those MPGs report `Available=True`.
+
+A coherent update runs in two phases. **MPG phase**: MPGs are rolled one at a time. Each MPG must reach `Available=True` before the next one starts. **TPG phase**: once every MPG is at the new hash, all remaining TPGs roll together in a single iteration. This bounds disruption — at most one MPG's worth of pods is unavailable at any moment.
```
shayasoolin: _"at most one MPG's worth of pods is unavailable at any moment"_ - the TPGs may potentially be a lot of pods, even that they do not form an MPG together. For example if the MPG has {1,1,1} and the TPGs are only of the 2 last components, and have excess of 100 replicas each...

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+| **PodCliqueSet reconciler** (orchestrator) | Detects out-of-date children, captures the `UpdatedStandalonePodCliques`/`UpdatedPodCliqueScalingGroups` set in `Status.UpdateProgress`, picks the next PCS replica to update, populates `InFlightPodGangs`, and waits for them to reach `PodGangConditionTypeAvailable=True` before advancing the iteration. Closes the update out by setting `UpdateEndedAt`. Currently only one PCS replica is updated at a time. |
+| **PodGangMap component** (in PodCliqueSet reconciler) | Computes the next-iteration PGM entries during an update from the frozen MVU template and the existing PGM state. Waits for the previous iteration to be `Available` before emitting the next, so only one is in flight at a time. In steady state it follows PCLQ/PCSG `Status.PodGangMapping`. |
+| **PodGang component** (in PodCliqueSet reconciler) | Reads the PGM as its single source of truth for desired PodGang composition. Materializes (creates / patches / deletes) `PodGang` resources from the PGM entries. Sets `MinReplicas` on creation and releases it (sets `MinReplicas=0`) once the PodGang reaches `MinReplicas` ready pods, then sets `PodGangConditionTypeAvailable=True`. |
```
shayasoolin: _"releases it (sets MinReplicas=0)"_ - at this stage in the doc it's not clear why this is done, add a short explanation that it's related to scale-in and gang-termination.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+These rules define what goes into a single MPG — how many pods of each updated standalone PCLQ and how many replicas of each updated PCSG are co-located in one MPG entry. TPG formation is covered in [Coherent update flow](#coherent-update-flow).
+
+**For standalone PodCliques**
+
+When one or more standalone PCLQs are updated, each MPG contains `minAvailable` replicas of every updated standalone PCLQ. Residual pods that cannot fill a complete MVU are absorbed into the last MPG of the iteration.
+
+| Case# | Description | MPG composition |
+| ----- | ----------- | --------------- |
+| 1     | Only one standalone PCLQ is updated | Each MPG contains `minAvailable` replicas of that PCLQ. The last MPG absorbs any remaining pods. |
```
shayasoolin: These two cases are basically the same one, aren't they? The first is just a specific case of the second.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+| 2     | More than one standalone PCLQ are updated | Each MPG contains `minAvailable` replicas of each updated standalone PCLQ. The last MPG absorbs any remaining pods of all updated PCLQs. |
+
+**For PodCliques belonging to PodCliqueScalingGroups**
+
+At the PCSG level, the MPG will contain `MinAvailable` replicas of the updated PCSGs. The entire set of PCLQs (containing all their replicas) of each PCSG are included in the MPG even if a subset of PCLQs are updated.
+
+| Case# | Description | Gang Scheduling behavior |
+| ----- | ----------- | ------------------------ |
+| 1     | One or more PCLQs of a PCSG updated | MPG will contain the `MinAvailable` replicas of this PCSG. All the constituent PCLQs (with all their replicas) are included as `PodGroups` in the MPG. |
+| 2     | One or more PCLQs from more than one PCSG are updated | One or more new MPG created with `minAvailable` replicas from each of the PCLQs of every PCSG and replicated over all the replicas of each PCSG |
```
shayasoolin: Same here. Consider simplifying this section. Case #3 here includes standalone PCLQs even that we're under the case of "PodCliques belonging to PodCliqueScalingGroups". Perhaps case #3 below can serve as a single description including all cases, followed by some examples. It will make it clearer for the reader.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+A coherent update progresses one PCS replica at a time, and within each replica one MVU iteration at a time. The five actors collaborate as follows:
+
+1. **Trigger.** The user mutates the `PCS.spec`. The PCS generation hash advances. The PodCliqueSet reconciler resets `Status.UpdateProgress` with `UpdateStartedAt` and freezes the set of out-of-date components in `UpdatedStandalonePodCliques` and `UpdatedPodCliqueScalingGroups`. This snapshot is the MVU template's input for the lifetime of the update.
+
+2. **Compute the next iteration's PGM entries.** The PodGangMap component reads the existing PGM, separates entries by PCS generation hash (old vs new), and computes the next iteration's MPG entry from the frozen MVU template. Old-hash entries are decremented by exactly what the new MPG consumes — this is the **take-down set** for this iteration. On the final iteration, when no further full MVU can be formed from the remaining old-hash entries, residual standalone PCLQ pods are absorbed into that last MPG, and any remaining PCSG replicas are each emitted as a **TPG** that depends on all MPGs created so far. The updated entries are written back to the PGM.
```
shayasoolin: _"Old-hash entries are decremented by exactly what the new MPG consumes"_ - by which order? Always starting from the first MPG (the formerly known as "base" one)? It makes sense to me I just couldn't find it written explicitly.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+7. **Advance.** Once all in-flight PodGangs are Available, the orchestrator clears `InFlightPodGangs` and explicitly requeues (the status-only patch does not fire a new reconcile on its own since the PCS For-watch uses `GenerationChangedPredicate`). On the next reconcile the PodGangMap component computes the next iteration's entries (step 2). When no old-hash entries remain for the replica, the orchestrator marks the replica's `UpdateEndedAt` and seeds `CurrentlyUpdating[0]` for the next pending replica. When all replicas are done, `UpdateProgress.UpdateEndedAt` is set and the strategy enters steady state.
```
shayasoolin: Somewhat repeating an earlier comment as it's important for me to understand it - what if the update can't be completed: it's stuck on one of the MPGs not getting available (either cannot be scheduled or pods fail to become ready). What can the user do in order to rescue the workload?

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+During a coherent update, every iteration deliberately drives `ReadyReplicas` for the in-flight components below `MinAvailable` for as long as it takes to take down the old-hash pods and bring up new-hash replacements. Without suppression, the evaluator would observe `MinAvailableBreached=True` on the in-flight PCLQs and PCSGs, and any iteration running longer than `TerminationDelay` would tear down the replica mid-roll.
+
+To prevent that, **gang termination is suppressed for the duration of any in-flight update**. While a PCLQ or PCSG is being updated, its `MinAvailableBreached` condition is held at `Unknown` rather than evaluated against live replica counts; the gang-termination evaluator skips `Unknown` children, so they cannot accumulate against `TerminationDelay`. Once the update completes for a given child, its condition resumes reflecting live state on the next reconcile and the evaluator returns to normal behavior.
```
shayasoolin, line 778: What if the update process is stuck mid-way? The MPGs that did get scheduled may lose some of their pods, and no gang-termination will take care of it, leaving precious resources allocated but not functioning.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,1029 @@
...
+|------|------------------------|-----------|---------------------|---------|----------------------|
+| 1 | 10, 11, 13 | 5 | 1, 3 | MPG{2F, 2P, 2D}                         | 8, 9, 11 |
+| 2 | 8, 9, 11   | 4 | 1, 3 | MPG{2F, 2P, 2D}                         | 6, 7, 9  |
+| 3 | 6, 7, 9    | 3 | 1, 3 | MPG{2F, 2P, 2D} + 1×TPG{1D}             | 4, 5, 6  |
+| 4 | 4, 5, 6    | 2 | 1, 2 | MPG{2F, 2P, 2D} + 1×TPG{1D}             | 2, 3, 3  |
+| 5 | 2, 3, 3    | 1 | 1, 1 | MPG{2F, 2P, 2D} + 1×TPG{1P} + 1×TPG{1D} | 0, 0, 0  |
+
+Total: 5 MPGs (10F, 10P, 10D) + 1 P-TPG + 3 D-TPGs = 10F, 11P, 13D.
+
+#### Example 2 — 4F, 100P, 150D, minAvailable = (2F, 2P, 3D)
```
oleg-kushniriov: The current algorithm emits all TPGs for a given component in a single batch per step: `48×TPG{1P} and 72×TPG{1D}` land as two independent groups. Because pods are scheduled asynchronously, these batches can become Ready in any order.

Consider the worst case mid-step 1: all 48 P-TPGs become Ready before any D-TPGs are scheduled. As each v2 P-TPG becomes Ready, the corresponding v1 P-pod is taken down. This leaves both versions simultaneously in an invalid state:

`v2 = {2F, 50P,   3D} ` ← almost all Ps, nearly no Ds → unusable
`v1 = {2F,  2P, 147D}`  ← almost all Ds, nearly no Ps → unusable

Neither version can serve traffic at this point: v2 has no decode capacity to match its prefill replicas, and v1 has almost no prefill capacity to feed its decode replicas.

Proposed fix: emit TPGs in ratio-preserving sub-steps, where the sub-step size is determined by the ratio of component replicas - ratio[P] : ratio[D] - reduced to lowest terms. In this example 100P : 150D = 2 : 3, so each sub-step emits 2 P-TPGs and 3 D-TPGs together:

Current: `MPG{2F, 2P, 3D} + 48×TPG{1P} + 72×TPG{1D}`

Proposed: `MPG{2F, 2P, 3D} + 24×(2×TPG{1P} + 3×TPG{1D})`

Each sub-step maintains the component ratio, bounding the worst-case mid-sub-step skew to a single sub-step's worth `({2P, 0D} or {0P, 3D})` rather than the full `{50P, 3D} / {2P, 147D}` drift possible today. Total TPG count is unchanged; only the emission granularity changes.

Ronkahn21:
Following up on what I talked through with @unmarshall on the single-MPG step approach. Wanted to put my position in writing.

To be clear up front: I really like this for **coherent updates**. It's the right tool there, and that's what this GREP is about. During an update the workload is already running and serving with a known, working ratio, and the step gating is exactly what you want to keep that ratio intact while you roll so you don't lose serving capacity. No notes on the update path, I'm fully on board.

Where I push back is **init**. My ask is about scope: initial deployment should stay out of scope for this GREP, and this suggestion should not change current init behavior at all. The coherent machinery is great for updates, but I don't think it should reshape init as a side effect. Leave init exactly as it behaves today and keep the GREP focused on updates.

A few reasons I don't want to fold init into this:

- At deploy time we have no real signal about user intent beyond `minAvailable` and `replicas`. The step model assumes every component is equal and that there's a ratio worth defending through the whole bring-up. For a classic Prefill/Decode setup with a Frontend, sure, keeping that ratio makes sense. But we're at the infra level and that assumption doesn't generalize. The same PCS could be a training job, or any workload where there's no ratio between components to defend at all.
- It doesn't clearly improve init, it just swaps one set of trade-offs for another. Yes, the current BPG/SPG has real flaws (standalone pods eating SPG capacity, all SPG gates lifting together, the knock-on TAS issues), but the step model introduces its own. Each step gates on the previous step being scheduled, so a single pod that can't be placed blocks everything behind it. Example: `PCS(10F, 11P, 13D)`, `minAvailable (2F, 2P, 2D)`. If one component in step 1 can't get its pods scheduled (a capacity or TAS limit on that one component), the gate to step 2 never lifts, so not a single P or D pod beyond `minAvailable` is released, even when there's plenty of capacity for them. One blocked pod throttles the whole deploy and you end up with fewer components scheduled than BPG/SPG would have brought up. So a user who never asked for coherent behavior and isn't running an update still pays for it on first deploy.
- Fixing the BPG/SPG flaws is a separate effort. It deserves its own discussion, not a ride-along on the update GREP.

If we're overruled and do decide to touch init here, I'd keep it to the absolute minimum, a two-step layout instead of N:

1. **Step 1:** the single MPG as proposed, `minAvailable` of every standalone PCLQ plus `minAvailable` of every PCSG.
2. **Step 2**, gated on step 1: emit all the TPGs at once and fold the remaining standalone PCLQ pods into the MPG.

That removes the standalone-pod-over-scheduling problem (step 1 holds them at `minAvailable`) without the N gated waves, since the per-wave gating only pays off when you're protecting an already-running workload's ratio, which isn't the case at first deploy.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: The reason inference frameworks have been mentioned is that, it is where the incompatibilities come from - be it KV cache transfer protocols, RPC protocols, block sizes for KV cache etc. It fits better than specifying inference servers IMO.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: addressed

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: We do not support any `maxSurge` and `maxUnavailable` therefore mentioning defaults can be very misleading. I would leave it at this. When the user reads the coherent update flow they will realise that it is delete first followed by create.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: Dynamo's planner component is not really part of a PCS. It is an external horizontal autoscaler which is not managed/deployed by Grove. Every component of a PCS is part of a serving unit. From that perspective the above statement as mentioned in the GREP is valid, right?

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: This has been removed. We can include this when we do provide concurrency controls.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: If there is an on-going update, irrespective of whether it is stuck, we should always support another update. We support this today for rolling-recreate. The reason for triggering another update can be that the previous update got stuck or it could be any other reason.

>What is considered 'completion'?

Completion is when there is nothing more to update.

unmarshall: We do not prevent it today. I have already taken it down as a point that we should test. RollingRecreate already supports it today and therefore coherent updates should also support this. We will test it explicitly.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: PGM only prescribes the quantities of each PodGang - it should contain x pods for a standalone PCLQ and y replicas of PCSG etc.
For standalone PCLQ pods - selection of pods is done in the Pod component of the PCLQ reconciler.
For PCSG replicas - selection is done in the PCLQ component in PCSG reconciler.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: We have now reverted this to follow BPG/SPG structure only for initial deployment of PCS. We lack a capability to specify scheduling constraints in the PCS. We will take that up as future work.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: Unavailability here means - those pods are taken down.
What is your point here precisely?

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: rewritten to now give it a reference to a section which details it further.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: Yes you are right. I think when i wrote it was only meant for clarity but maybe its not really required.
I have removed it now.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: yes its been collapsed now.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: yes it starts from the first MPG. I have changed the text to reflect that clearly.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: Yes that can happen and in that case a human operator needs to either increase the capacity (if the stuck updates are due to capacity issues) or perhaps fix the PCS issue and start another update. Since the issue due to which the updates are stuck can be aplenty its not possible to foresee all kinds of cases.

As I have mentioned before, we support triggering another update when there is already an on-going update for RollingRecreate strategy. We will test this also for coherent udpates.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: done.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: done.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: added

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: This is an implementation detail. Perhaps you can add this comment to the implementation PR.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -548,24 +559,30 @@ sequenceDiagram
...
+| `PodGangConditionTypeReady` | Set to `True` once `MinReplicas` pods of every `PodGroup` are `Ready` (passing readiness probes). |
 
-1. **Set on creation.** When a PodGang is first materialized from a PGM entry, each `PodGroup`'s `MinReplicas` is set to the `minAvailable` value defined in the PCS spec for that PCLQ or PCSG. This forces the scheduler to place the whole gang at once before any constituent pod can run.
-2. **Wait for minAvailable pods to become ready.** The PodGang component tracks each PodGroup against the live pod set and computes whether at least `minAvailable` pods of that group are currently `Ready`.
-3. **Release once ready, then mark `Available`.** Once the gang reaches minAvailable-ready, the PodGang component patches `MinReplicas` to 0 on every PodGroup of that PodGang, then sets `Status.Conditions[Type=Available]=True` with `Reason=PodGangAvailable`.
+These conditions are progressive: a PodGang reaches `Scheduled=True` first, then `Ready=True`. Both stay `True` for the rest of the PodGang's lifetime — they are not flipped back to `False` after, for example, a pod failure or scale-in.
```
shayasoolin: This is an important point, we rely on these conditions when filtering PGs by labels and conditions. So it makes perfect sense to me that the conditions are progressive. Wondering if their names should reflect that, to avoid user confusion, who might expect them to always be up-to-date rather than progressive. 
For example we can consider "WasReady" instead of "Ready". You know the standards better than me, I'm just raising here but trust your judgement.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -548,24 +559,30 @@ sequenceDiagram
...
 
-The same lifecycle applies in steady state to PodGangs created by PCSG scale-out — the new SPG follows set → wait → release → mark Available exactly as an MPG does during an update.
+`MinReplicas` is released at stage 2 — when the scheduler has done its job — rather than at stage 3. The reason: `MinReplicas` is a *scheduler* directive (gang-placement floor and, on some backends, gang-termination floor); pod readiness is the *workload's* concern, not the scheduler's. Holding `MinReplicas` until pods pass readiness probes would conflate the two and would risk the backend scheduler terminating the gang on any pod failure that occurs *before* the gang reaches readiness — a single pod crash on an otherwise-placed gang could trigger termination before the workload has even had a chance to start. Releasing `MinReplicas` at `Scheduled=True` decouples the two concerns: once placement is done, readiness is purely a workload signal that the orchestrator (and only the orchestrator) consults.
+
+The same lifecycle applies in steady state to PodGangs created by PCSG scale-out — the new SPG follows set → release + `Scheduled=True` → `Ready=True` exactly as an MPG does during an update.
```
shayasoolin: I'm still mid-way in the review of the last changes, but I expect to see a reference to the decision that non-first MPGs have `MinReplicas` of 1 (rather than `MinAvailable`).
Maybe I'll find it written later, but consider mentioning it here as well.
It's an important change we agree on, that only the first MPG acts by MinAvailable.

docs/proposals/393-coherent-rolling-updates/README.md
shayasoolin: It's nit but I still think that "serving traffic" tightly couples us not only to inference workloads but also to each component within the PCS. It can be a parameter server for example, which doesn't serve any traffic. Minor issue though.

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,929 @@
...
+
+Grove encodes the MVU's gang-scheduling intent by generating new `PodGang` resources, **MVU PodGangs (MPGs)**, that hold exactly the MVU's pods. PCSG replicas above MinAvailable go into separate **Tail PodGangs (TPGs)** — one per excess replica. Both MPGs and TPGs are gang-scheduled. Each TPG `DependsOn` every MPG of the same generation. The scheduler places TPG pods only after all those MPGs report `Ready=True`.
+
+A coherent update runs in two phases. **MPG phase**: MPGs are rolled one at a time. Each MPG must reach `Ready=True` before the next one starts. **TPG phase**: once every MPG is at the new hash, all remaining TPGs roll together in a single iteration. This bounds disruption — at most one MPG's worth of pods is unavailable at any moment.
```
oleg-kushniriov: The TPG phase rolls all PCSG TPGs in a single iteration, but each component's TPGs (e.g. P-TPGs and D-TPGs) are independent `PodGang` resources that the scheduler places asynchronously. This creates a window where the ratio between PCSG components can be severely violated in both the old and new versions simultaneously.

**Example 2 - 4F, 100P, 150D, minAvailable = (2F, 2P, 3D):**

F is the bottleneck - `floor(4 / 2) = 2` MPGs are emitted, each `MPG{2F, 2P, 3D}`. The remaining 96P and 144D go into TPGs.

**MPG phase** (2 iterations):

```
Start:          v1 = {4F, 100P, 150D},  v2 = {}
After MPG-1:    v1 = {2F,  98P, 147D},  v2 = {2F,   2P,   3D}
After MPG-2:    v1 = {0F,  96P, 144D},  v2 = {4F,   4P,   6D}
```

After the MPG phase, v1 already has **zero F** - it is entirely unusable before a single TPG rolls.

**TPG phase** - all 96×TPG{1P} and 144×TPG{1D} emitted together in one iteration:

Normal completion (all TPGs become Ready simultaneously):
```
v1 = {},                 v2 = {4F, 100P, 150D}  ✓
```

Worst-case skewed completion (all P-TPGs Ready before any D-TPGs are scheduled - each v2 P-TPG becoming Ready causes its corresponding v1 P-pod to be taken down):
```
v2 = {4F, 100P,   6D}   ← all Ps, almost no Ds → unusable
v1 = {0F,   0P, 144D}   ← no F, no Ps          → unusable
```

**Neither version can serve traffic**. The full 96P and 144D pool is at risk simultaneously because the TPG phase operates on the entire TPG pool at once rather than a bounded per-step batch.

**Proposed fixed behaviour:** 
**Initial State before the update: 4F, 20P, 30D, minAvailable = (1F, 2P, 3D):**
(The example for a smaller workload just to show all steps during the update )
**ratio-unit** : the per-bottleneck-unit share of each component; computed by dividing each component count by the smallest (F) component count.
Example: 4F, 20P, 30D → ratio-unit = {4/4 F, 20/4 P, 30/4 D} = {1F, 5P, 7.5D}

Steps of Update:
------------------------------------------------------------------
```
Boundary Step = 1
-- v1                  v2                    → v2 (moved this step)   step type        --
v1 = {4F, 20P, 30D}   v2 = {}               —                        initial state
v1 = {3F, 18P, 27D}   v2 = {1F,  2P,  3D}  → {1F, 2P, 3D}          full: F + P + D
v1 = {3F, 16P, 24D}   v2 = {1F,  4P,  6D}  → {2P, 3D}               partial: P + D only
v1 = {3F, 14P, 21D}   v2 = {1F,  6P,  9D}  → {2P, 3D}               partial: P + D only 
(9 >= 7.5 * 1 -> moving to increment the smallest)
```
------------------------------------------------------------------
```
Boundary Step = 2
-- F boundary: next F unit moves to v2 --------
v1 = {2F, 12P, 18D}   v2 = {2F,  8P, 12D}  → {1F, 2P, 3D}          full: F + P + D
v1 = {2F, 10P, 15D}   v2 = {2F, 10P, 15D}  → {2P, 3D}               partial: P + D only
(15 >= 7.5 * 2 -> moving to increment the smallest)
```
------------------------------------------------------------------
```
Boundary Step = 3
-- F boundary: next F unit moves to v2 --------
v1 = {1F,  8P, 12D}   v2 = {3F, 12P, 18D}  → {1F, 2P, 3D}          full: F + P + D
v1 = {1F,  6P,  9D}   v2 = {3F, 14P, 21D}  → {2P, 3D}               partial: P + D only
v1 = {1F,  4P,  6D}   v2 = {3F, 16P, 24D}  → {2P, 3D}               partial: P + D only
(24 >= 7.5 * 3 -> moving to increment the smallest)
```
------------------------------------------------------------------
```
Boundary Step = 4
-- F boundary: next F unit moves to v2 --------
v1 = {0F,  2P,  3D}   v2 = {4F, 18P, 27D}  → {1F,2P, 3D}               partial: P + D only
v1 = {}                       v2 = {4F, 20P, 30D}  → { 2P, 3D}          full: F + P + D
```

oleg-kushniriov (commented):
This comment was not addressed in previous version: https://github.com/ai-dynamo/grove/pull/640#discussion_r3409462398

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: ok changed.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: `PodGangConditionTypeScheduled` on the other hand is one-time. Once pods are placed then minReplica pods will remain placed unless there is PCS replica termination due to what we incorrectly call today gang termination. But when that happens all things in that PCS replica are gone.

changed `PodGangConditionTypeReady` to always reflect the current state. This also helps in ensuring better observability

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: addressed

docs/proposals/393-coherent-rolling-updates/README.md
shayasoolin: The wording of "at most one MPG's worth of pods is unavailable" sounds like the bounds are pretty limited, while in reality there's the "remove all TPGs together" stage where lots of pods may be taken down.
Not asking to change the way it works, just to phrase it a bit differently as in some cases the impact of the update is not as small as you put it.

shayasoolin (dismissed):

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: addressed

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: For initial deployment we continue with BPG/SPG model.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: This has now been changed. Also the same issue is also present even today for all initial deployments of PCS. We do not solve the issue with initial deployment as part of this GREP but it will be taken up later.

docs/proposals/393-coherent-rolling-updates/README.md
unmarshall: The algo has been changed. Please review again.

shayasoolin (dismissed):

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,1155 @@
...
+
+###### Worked example
+
+Consider a PCS with three updated components — `frontend` (standalone PCLQ), `prefill` (PCSG), and `decode` (PCSG):
+
+| Component | `replicas` | `minAvailable` | `maxUnavailable` |
+| --- | --- | --- | --- |
+| `frontend` (F) | 10 | 2 | 2 |
+| `prefill`  (P) | 10 | 3 | 3 |
+| `decode`   (D) | 20 | 3 | 4 |
```
sanjaychatterjee, line 552: Decode maxUnavailable should be 3?

| `decode`   (D) | 20 | 3 | 3 |

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,1155 @@
...
+- **Prefill (P)** — PCSG.
+- **Decode (D)** — PCSG.
+
+##### Case A — Clean divisibility
+
+| Component | `replicas` | `minAvailable` | `maxUnavailable` |
+| --- | --- | --- | --- |
+| FrontEnd (F) | 10 | 1 | 1 |
+| Prefill (P)  | 100 | 3 | 3 |
+| Decode (D)   | 80 | 3 | 4 |
```
sanjaychatterjee, line 969: Decode maxUnavailable should be 3?

| `decode`   (D) | 20 | 3 | 3 |

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,1155 @@
...
+- Sub-step N.3: emit 3 more P-TPGs and 1 more D-TPG.
+- Sub-step N.4: emit 1 more P-TPG. Step done.
+
+##### Case B — Residual and leftover
+
+| Component | `replicas` | `minAvailable` | `maxUnavailable` |
+| --- | --- | --- | --- |
+| FrontEnd (F) | 20 | 2 | 2 |
+| Prefill (P)  | 10 | 3 | 3 |
+| Decode (D)   | 20 | 3 | 4 |
```
sanjaychatterjee, line 991: Decode maxUnavailable should be 3?

| `decode`   (D) | 20 | 3 | 3 |

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -0,0 +1,1155 @@
...
+
+The MPG of each step carries only F (`{1 F}`); pods of P and D continue running in their existing PodGangs untouched. The MPGs depend on no anchor — each is itself an anchor.
+
+##### Case H — Invalid: MaxUnavailable < MinAvailable
+
+| Component | `replicas` | `minAvailable` | `maxUnavailable` |
+| --- | --- | --- | --- |
+| FrontEnd (F) | 20 | 2 | 1 |
+| Prefill (P)  | 10 | 3 | 3 |
+| Decode (D)   | 20 | 3 | 4 |
```
sanjaychatterjee, line 1029: Decode maxUnavailable should be 3?

| `decode`   (D) | 20 | 3 | 3 |

sanjaychatterjee (approved):

shayasoolin (approved):

PR #645, opened 2026-06-03
fix: remove legacy hash migration window

gflarity:
#### What type of PR is this?

/kind cleanup

Discussion-only draft: this PR removes the legacy hash support/migration window and makes the operator use only the canonical PodCliqueSet and PodCliqueScalingGroup hash behavior.

The implication is that, when released, existing PCS/PCSG objects carrying legacy hashes would be treated as out of date and would update/roll to the canonical hash behavior, even when their desired spec has not otherwise changed.

- Drops legacy hash candidate helpers and legacy pod template / generation hash computation paths.
- Updates reconcile, status, and rolling update logic to compare stored hashes against the canonical expected hash only.
- Removes the legacy hash migration E2E scenario, fixture, and obsolete unit coverage.

NONE

This should not be merged until we explicitly agree that causing all affected PCS/PCSG objects to update/roll on upgrade is acceptable. This draft is for discussion of that tradeoff. 

```release-note
NONE
```

```docs
NONE
```


danbar2:
Closing this one out. The canonical hash behavior this draft builds on (introduced in #566) has since been reverted in #661, and Grove is back on the legacy single-hash model — so there’s no canonical hash to standardize on and no migration window left to remove. The code paths this PR deletes no longer exist on main.

PR #646, opened 2026-06-03
feat(chart): Allow setting security context in Helm

### LINKED ISSUE #647: Allow setting security context with Helm installation

I would like to explicitly set the UID/GID for Helm deployments as well as utilize Prometheus operator to collect metrics.

### Why is this needed?

In our environment we have strict Kyverno policies to disallow any attempts to run privileged pods.  In our environment it's critical that all pods run as unprivileged users.

treydock:
#### What type of PR is this?

/kind feature

Allow setting security contexts so that it's possible to force the containers to run as non-root and specific UID/GID.

Fixes #647

```release-note
- Allow setting `securityContext` and `podSecurityContext` for the Helm chart.
- Add optional serviceMonitor and allow annotations for service
```

```docs

```

### DIFF
--- a/operator/charts/templates/servicemonitor.yaml
+++ b/operator/charts/templates/servicemonitor.yaml
@@ -1,9 +1,9 @@
-{{- if .Values.serviceMonitor.enable }}
+{{- if .Values.serviceMonitor.enabled }}
 ---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: {{ .Values.serviceMonitor.name }}
+  name: {{ required ".Values.serviceMonitor.name is required" .Values.serviceMonitor.name }}
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "operator.servicemonitor.labels" . | nindent 4 }}
--- a/operator/charts/templates/servicemonitor.yaml
+++ b/operator/charts/templates/servicemonitor.yaml
@@ -1,0 +1,18 @@
+{{- if .Values.serviceMonitor.enable }}
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: grove-operator-monitor
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{  include "operator.servicemonitor.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "operator.service.labels" . | nindent 6 }}
+  endpoints:
+  - port: metrics
+    path: /metrics
+    scheme: http
+{{- end }}
--- a/operator/charts/templates/servicemonitor.yaml
+++ b/operator/charts/templates/servicemonitor.yaml
@@ -3,14 +3,14 @@
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: grove-operator-monitor
+  name: {{ .Values.serviceMonitor.name }}
   namespace: {{ .Release.Namespace }}
   labels:
-    {{  include "operator.servicemonitor.labels" . | nindent 4 }}
+    {{- include "operator.servicemonitor.labels" . | nindent 4 }}
 spec:
   selector:
     matchLabels:
-      {{- include "operator.service.labels" . | nindent 6 }}
+      {{- include "operator.service.matchLabels" . | nindent 6 }}
   endpoints:
   - port: metrics
     path: /metrics
--- a/operator/charts/templates/servicemonitor.yaml
+++ b/operator/charts/templates/servicemonitor.yaml
@@ -1,9 +1,9 @@
-{{- if .Values.serviceMonitor.enabled }}
+{{- if .Values.metrics.serviceMonitor.enabled }}
 ---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: {{ required ".Values.serviceMonitor.name is required" .Values.serviceMonitor.name }}
+  name: {{ required ".Values.metrics.serviceMonitor.name is required" .Values.metrics.serviceMonitor.name }}
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "operator.servicemonitor.labels" . | nindent 4 }}
--- a/operator/charts/values.yaml
+++ b/operator/charts/values.yaml
@@ -20,6 +20,11 @@ resources:
     memory: 128Mi
     ephemeral-storage: 128Mi
 
+podSecurityContext:
+  seccompProfile:
+    type: RuntimeDefault
+securityContext: {}
+
 # Node selector to constrain the operator pod to nodes with matching labels.
 nodeSelector: {}
 
--- a/operator/charts/values.yaml
+++ b/operator/charts/values.yaml
@@ -149,11 +149,15 @@ service:
     app.kubernetes.io/component: operator-service
     app.kubernetes.io/name: grove-operator
     app.kubernetes.io/part-of: grove
+  # -- Annotations to add to the Grove Operator Service resource
   annotations: {}
 
 serviceMonitor:
+  # -- Enable the Prometheus ServiceMonitor resource
   enabled: false
+  # -- Name of the Prometheus ServiceMonitor resource
   name: grove-operator-monitor
+  # -- Labels for the Prometheus ServiceMonitor resource
   labels:
     app.kubernetes.io/component: operator-service-monitor
     app.kubernetes.io/name: grove-operator
--- a/operator/charts/values.yaml
+++ b/operator/charts/values.yaml
@@ -158,6 +158,8 @@ serviceMonitor:
     app.kubernetes.io/component: operator-service-monitor
     app.kubernetes.io/name: grove-operator
     app.kubernetes.io/part-of: grove
+    # If the Prometheus Operator requires the 'release' label, set this value:
+    # release: prometheus
 
 clusterRole:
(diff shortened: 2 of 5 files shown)


renormalize (approved):
@treydock thank you for the contribution. I think the PR looks good!

I'm just wondering if there would be ever a case where consumers would need/want to specify different `securityContext` values for the crd-upgrade init-container and the grove-operator container.

Either way, I don't want to block the PR for that.

treydock:
> I'm just wondering if there would be ever a case where consumers would need/want to specify different `securityContext` values for the crd-upgrade init-container and the grove-operator container.
> 

For an operator not using PVCs, probably not.  We have a policy that nothing can run as root without specific reasons so we have to ensure everything has a specific / predefined UID/GID.  Since Grove is not mounting storage I think it's fine to keep the security context uniform across the various containers.

treydock:
@renormalize Let me know if any changes are needed or if this is good to merge.

operator/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.serviceMonitor.enable }}
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: grove-operator-monitor
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{  include "operator.servicemonitor.labels" . | nindent 4 }}
```
renormalize: for the sake of consistency, include the leading dash:

```suggestion
    {{- include "operator.service.matchLabels" . | nindent 4 }}
```

I would also prefer it if you introduced a `service.matchLabels` the way we have defined the `deployment.matchLabels`. `service.matchLabels` will not include the common chart labels since the matchlabels must only contain labels relevant to the service.

Something like this:

```
{{- define "operator.service.matchLabels" -}}
{{- range $key, $val := .Values.service.labels }}
{{ $key }}: {{ $val }}
{{- end -}}
{{- end -}}
```

operator/charts/values.yaml
```diff
@@ -139,6 +144,16 @@ service:
...
+  annotations: {}
+
+serviceMonitor:
+  enable: false
+  labels:
+    app.kubernetes.io/component: operator-service-monitor
+    app.kubernetes.io/name: grove-operator
+    app.kubernetes.io/part-of: grove
+    release: prometheus-stack
+
```
renormalize, line 168:
```suggestion
```

operator/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.serviceMonitor.enable }}
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: grove-operator-monitor
```
renormalize: The name can be templatized, and passed from `values.yaml`, as it is being done for deployment, role, service, etc.

operator/charts/values.yaml
```diff
@@ -139,6 +144,16 @@ service:
     app.kubernetes.io/component: operator-service
     app.kubernetes.io/name: grove-operator
     app.kubernetes.io/part-of: grove
+  annotations: {}
+
+serviceMonitor:
+  enable: false
+  labels:
+    app.kubernetes.io/component: operator-service-monitor
+    app.kubernetes.io/name: grove-operator
```
renormalize: https://github.com/ai-dynamo/grove/pull/646#discussion_r3411788791

renormalize (changes requested):
@treydock I was testing this PR, and rendering with `helm template` works fine. However, while using skaffold, I get parse errors because the label key `release` is being specified in two different places:
- `common.chart.labels` as specified in `operator/charts/templates/_helpers.tpl`.
- `serviceMonitor.labels` in `operator/charts/values.yaml`.

`skaffold` does not handle the label override well and fails, which `helm` tolerates.

We use `skaffold` in this repository in the `deploy` make target as specified in `operator/Makefile` for development, and we aim to support users who use skaffold to deploy to their clusters.

So my recommendation would be to remove `release` from `serviceMonitor.labels`, or remove the `common.chart.labels` from the seviceMonitor labels. I would like to get opinions from other reviewers about this.

****

To reproduce:

```
$ cd {your path}/grove/operator
$ make kind-up deploy
```

operator/charts/templates/servicemonitor.yaml
treydock: This is implemented and used new service matchLabels to match the service with ServiceMonitor.

treydock:
@renormalize I think I resolved the requested changes.  Let me know if the scaffold errors are resolved or if any other changes are needed.

operator/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.serviceMonitor.enable }}
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ .Values.serviceMonitor.name }}
```
renormalize: All other resources which have templatized this field have the name as required.

```suggestion
  name: {{ required ".Values.serviceMonitor.name is required" .Values.serviceMonitor.name }}
```

operator/charts/values.yaml
```diff
@@ -139,6 +144,15 @@ service:
     app.kubernetes.io/component: operator-service
     app.kubernetes.io/name: grove-operator
     app.kubernetes.io/part-of: grove
+  annotations: {}
+
+serviceMonitor:
+  enable: false
```
renormalize: All fields use `enabled` instead of `enable`. I do see one existing field which uses `enable` in `values.yaml` but the corresponding Go type does not even have that field. I will remove that later.

```suggestion
  enabled: false
```

operator/charts/values.yaml
```diff
@@ -139,6 +144,15 @@ service:
...
+  name: grove-operator-monitor
+  labels:
+    app.kubernetes.io/component: operator-service-monitor
+    app.kubernetes.io/name: grove-operator
+    app.kubernetes.io/part-of: grove
```
renormalize: Nit: Maybe we can document that we are not adding the `prometheus-stack` label that is typically used required by the `kube-prometheus-stack` install? We can add that label as a comment here if you think it is necessary. I don't have a strong opinion on this, we can go ahead with whatever you think is the right way to go about this.

renormalize (commented):
thanks for addressing the review @treydock! The issue with skaffold has been resolved as well.


I found a bunch more minor nits unfortunately, which I should have spotted in my previous review, sorry about that. Other than these nits, it's an lgtm from me. I will reach out to the review team so we get another approval, and we can proceed with the merge.

operator/charts/templates/servicemonitor.yaml
treydock: Changes made in new commit

operator/charts/values.yaml
treydock: Changes made in new commit

operator/charts/values.yaml
treydock: @renormalize I personally don't require specific labels with my Prometheus operator deployment but I went ahead and added a comment to values.yaml in case others do require the label to be set.

treydock:
@renormalize I think I've addressed the new requested changes.

renormalize (dismissed):
lgtm! 

Let's wait for the second approval from another reviewer and get this in. thank you for your contribution @treydock!

shayasoolin (dismissed):

operator/charts/values.yaml
```diff
@@ -139,6 +144,17 @@ service:
     app.kubernetes.io/component: operator-service
     app.kubernetes.io/name: grove-operator
     app.kubernetes.io/part-of: grove
+  annotations: {}
+
+serviceMonitor:
```
unmarshall: Can you rename this to `prometheusServiceMonitor` as `serviceMonitor` is too generic?

unmarshall: Add a comment explaining the purpose of this configuration. In general we lack proper documentation in values.yaml but lets improve that from now onwards.

unmarshall (changes requested):

operator/charts/values.yaml
treydock: @unmarshall Most community charts use `serviceMonitor` and `podMonitor`, for example the main Dynamo operator uses those keys.  Sometimes what sites do is like this:

```
prometheus:
  serviceMonitor:
    ...
```

Let me know if you still want `prometheusServiceMonitor` or something different.

treydock: I added helm-docs style documentation in a new commit.

treydock: @unmarshall Please see my previous comment, thanks.

shayasoolin (dismissed):

renormalize (dismissed):
lgtm

operator/charts/values.yaml
unmarshall: @treydock perfect then lets do what dynamo does - have the hierarchy as that way it is very clear that `serviceMonitor` is for prometheus.

treydock: @unmarshall I moved `serviceMonitor` under `metrics` which is what Dynamo operator chart uses.

Dynamo uses `dynamo.metrics.podMonitors` and uses plural as there are numerous `PodMonitor` resources but since this chart uses a single `ServiceMonitor` I used `metrics.serviceMonitor`.

renormalize (approved):

unmarshall (approved):

shayasoolin (approved):

PR #653, opened 2026-06-07
Store all scale test results

shayasoolin:
#### What type of PR is this?

/kind bug

The scale-test workflow now runs more than one scale test in the same action. Each test writes its own `scale-test-results.json`, but `scale-history.py` only picked the latest result file under the output directory.

As a result, recent scheduled runs stored only the later `SoakChurn` result and skipped the earlier `ScaleTest_1000` result, even though both tests passed.

This updates the history script to ingest all nested `scale-test-results.json` files when the input is a parent output directory, while preserving single-file and direct-run-directory behavior.

Related to #550

Validated locally with a temp input directory containing one `ScaleTest_1000` result and one `SoakChurn` result:

- first run stored two rows in `index/runs.ndjson`
- both raw result files were copied under `results/`
- second run skipped both as duplicates
- direct single-file input still stored one row

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/hack/scale-history.py
+++ b/operator/hack/scale-history.py
@@ -54,31 +54,36 @@ RUN_ID_RE = re.compile(r"^run-(\d{8})-(\d{6})$")
 
 def main() -> int:
     args = parse_args()
-    result_path = find_result_file(Path(args.input))
-    result = read_json(result_path)
+    result_paths = find_result_files(Path(args.input))
     commit = args.commit or best_effort_git(["rev-parse", "HEAD"])
     run_url = args.run_url or ""
 
-    timestamp = run_timestamp(result)
-    test_name = required_str(result, "testName")
-    run_id = required_str(result, "runID")
     history_dir = history_dir_for_args(args)
-    changed = append_history(
-        history_dir=history_dir,
-        result_path=result_path,
-        result=result,
-        run_record=build_run_record(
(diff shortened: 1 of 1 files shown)


Ronkahn21 (approved):

shmuel-runai (approved):
LGTM

operator/hack/scale-history.py
```diff
@@ -339,19 +345,22 @@ def prepare_branch_workdir(workdir: Path, branch: str, repo: str | None) -> Path
...
+    if len(changed_results) == 1:
+        result = changed_results[0]
         message = (
             f"Add scale test result {result.get('testName')} "
             f"{result.get('runID')}"
         )
+    elif changed_results:
+        message = f"Add {len(changed_results)} scale test results"
     else:
         message = "Update scale test dashboard"
```
oleg-kushniriov, line 365: Do we need to commit in case msg is Update scale test dashboard?
if no changed_results registered?

shayasoolin: I don’t think the behavior needs a fix.

That else only happens after this check:

```
if not git_has_changes(history_dir):
    print("no git changes to commit")
    return

```
So if changed_results is empty and we still reach the commit, it means there are git changes from something else, practically the copied dashboard files

PR #657, opened 2026-06-08
perf(operator): list PodClique pods via owner-UID field index

### LINKED ISSUE #655: perf(operator): GetPCLQPods scans the whole namespace per call (label selector is not cache-indexed)
## Summary

`GetPCLQPods` lists a PodClique's pods with a **label selector** against the manager's cache. controller-runtime's cache does **not** index label selectors, so every call iterates **all grove-managed pods in the namespace** and matches the label per-pod. The function is called per PodClique reconcile, so a full reconcile cycle over `C` cliques sharing a namespace costs `O(C² · R)` (R = replicas per clique).

## Where

`operator/internal/controller/common/component/utils/pod.go` — `GetPCLQPods`:

```go
cl.List(ctx, podList,
    client.InNamespace(pclq.Namespace),
    client.MatchingLabels{apicommon.LabelPodClique: pclq.Name},
)
// then filtered in-memory with metav1.IsControlledBy
```

## Root cause (controller-runtime v0.22.4)

`pkg/cache/internal/cache_reader.go`, `CacheReader.List`:

- `InNamespace` + `MatchingLabels` takes the `case listOpts.Namespace != ""` branch → `c.indexer.ByIndex(NamespaceIndex, ns)` returns **every pod in the namespace**.
- It then loops over all of them applying `labelSel.Matches()` per pod — **label selectors are never indexed**.
- Only `FieldSelector` goes through `byIndexes` → `ByIndex(FieldIndexName(field), …)`, an actual index lookup.
- `DeepCopyObject()` happens *after* the label-match `continue`, so deep-copy is already bounded to matched pods — the **scan**, not the copy, is the cost that scales.

The operator registers **no field indexers** today (`grep -r IndexField` is empty), and the cache is filtered to `app.kubernetes.io/managed-by=grove-operator`, so the scan covers all grove pods of all cliques in the namespace.

## Impact (microbenchmark)

client-go's `cache.Indexer` (the structure backing the controller-runtime cache), label-scan vs indexed `ByIndex`:

| pods in ns | label scan (per call) | indexed (per call) |
|---|---|---|
| 80   | 14.6 µs  | 1.1 µs |
| 400  | 66.2 µs  | 1.1 µs |
| 800  | 124.6 µs | 1.1 µs |
| 3200 | 543.0 µs | 2.3 µs |

Full reconcile cycle (every clique lists once), 200×16 = 3200 pods: **117 ms vs 0.5 ms**, and **11.5 MB vs 77 KB** allocated per cycle. Label-scan is `O(C²·R)`; indexed is `O(C·R)`. Negligible on small namespaces, significant on large ones.

## Proposed fix

Register a Pod field index keyed on the **controller owner-reference UID** and query it with `client.MatchingFields{…: pclq.UID}`. The cache then serves the list via `ByIndex` in `O(pods-owned-by-this-clique)`. The index key is exactly what `metav1.IsControlledBy` verified, so the in-memory filter can be dropped.

kangclzjc:
## What

`GetPCLQPods` selected a PodClique's pods with a label selector, which controller-runtime's cache cannot index — every call scanned all grove-managed pods in the namespace and matched the label per-pod (`O(pods-in-namespace)` per call, `O(C²·R)` per reconcile cycle with `C` cliques sharing a namespace).

This registers a Pod field index keyed on the **controller owner-reference UID** and switches `GetPCLQPods` to `client.MatchingFields{…: pclq.UID}`, so the cache serves the list via `ByIndex` in `O(pods-owned-by-this-clique)`. The index captures the same ownership relation `metav1.IsControlledBy` verified, so the in-memory filter is removed.

Fixes #655.

## Correctness / behavior

- The index key is `metav1.GetControllerOf(pod).UID` — identical to what `IsControlledBy(pod, pclq)` checked — so the result set is unchanged for grove pods (which always carry both the owner ref and the label).
- Selection is now by ownership rather than the label proxy. A new unit test pins this down: an owned-but-unlabeled pod is returned; a labelled-but-other-owner pod is not.
- The index is registered once in `podclique` `RegisterWithManager`, before the cache starts.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt` — clean.
- Unit tests pass: `internal/controller/common/component/utils`, `internal/controller/podclique/...`.
- envtest integration passes against a real apiserver (`TestRegisterControllers`, `TestPodCacheFiltering`, `TestCreateManager`) — index registration verified end-to-end.

## Benchmark

Microbench over client-go's `cache.Indexer` (the structure backing the controller-runtime cache), full reconcile cycle (every clique lists its own pods once):

| clique × replica (= pods) | label scan | field index |
|---|---|---|
| 10×8 = 80    | 0.13 ms | 0.012 ms |
| 50×8 = 400   | 3.57 ms | 0.060 ms |
| 100×8 = 800  | 13.6 ms | 0.128 ms |
| 200×16 = 3200 | 117 ms | 0.51 ms |

### DIFF
--- a/operator/internal/controller/common/component/utils/pod.go
+++ b/operator/internal/controller/common/component/utils/pod.go
@@ -19,7 +19,6 @@ package utils
 import (
 	"context"
 
-	apicommon "github.com/ai-dynamo/grove/operator/api/common"
 	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
 
 	"github.com/samber/lo"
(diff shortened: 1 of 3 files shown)


Ronkahn21 (approved):
LGTM

danbar2 (approved):

PR #659, opened 2026-06-10
fix: shorten init-container SA token Secret name

### LINKED ISSUE #658: Make the init-container SA-token-secret name suffix configurable (or shorten the default)
### Summary

The fixed suffix `-initc-sa-token-secret` (22 bytes) appended by [`apicommon.GenerateInitContainerSATokenSecretName`](https://github.com/ai-dynamo/grove/blob/main/operator/api/common/namegen.go) is currently the binding constraint on usable PodCliqueSet (PCS) name length — tighter than the 45-char pod-name budget enforced by [PR #105](https://github.com/ai-dynamo/grove/pull/105). It would help downstream consumers if this suffix were either shorter by default or configurable at operator-install time.

### Why this is the binding constraint

The generated Secret name lands as a label value on init-container pods, and Kubernetes caps label values at 63 bytes. That gives:

```
PCS-name budget = 63 − len("-initc-sa-token-secret") = 63 − 22 = 41 chars
```

Compare with the pod-name budget (after using single-char clique / scaling-group names): roughly `63 − ~14 ≈ 49 chars`. So the secret-label budget wins by ~8 chars and is the one users actually hit.

### Impact

Any consumer whose PCS names embed a long stable identifier — UUIDs, multi-segment hierarchical names (e.g. `<group>-<deployment>-<variant>`), or per-tenant prefixes — runs out of budget well before they hit the documented 45-char limit. UUIDs without hyphens are 32 chars, leaving only 9 chars for any discriminator, namespace, or role suffix. Workarounds (hash-truncation, lossy compression of the original identifier) break the round-trip from PCS name back to the originating resource, complicating debugging and event correlation.

### Proposed solutions, least to most intrusive

1. **Shorten the default.** `-initc-sa` (9 bytes) or `-ic-sat` (7 bytes) preserves the semantic meaning and frees 13–15 chars. Backwards-incompatible on rolling upgrade (existing Secrets would be orphaned), so would need a migration path or a release-note callout.
2. **Move the cross-reference from a label to an annotation.** Annotations have no 63-byte cap, so the PCS-name budget would be bounded only by the pod-name math (~49 chars). Backwards-compatible if both the new annotation and the old label are written during a deprecation window.
3. **Make the suffix configurable via an operator install flag** (e.g. `--initc-secret-suffix=-ic-sat`), defaulting to today's value. Most flexible, fully backwards-compatible, and lets each cluster tune the trade-off between descriptiveness and name budget.

Happy to send a PR for any of these if a direction is agreed.

yankay:
#### What type of PR is this?

/kind bug

Shortens the init-container service account token Secret suffix from `-initc-sa-token-secret` to `-ic-sat`.

The old suffix could make generated Secret names too long when a long PodCliqueSet name was copied into labels such as `app.kubernetes.io/name`, causing Secret creation to fail. The shorter suffix keeps the generated name within Kubernetes label value limits for admitted PodCliqueSet names.

For compatibility, the controller still recognizes the legacy Secret name during cleanup:

- New Pods reference the shortened `<pcs>-ic-sat` Secret through the existing name generator.
- If the new Secret does not exist, the controller creates a new empty Secret with type `kubernetes.io/service-account-token`.
- No data is copied from the legacy Secret to the new Secret.
- Kubernetes issues a fresh token for the new Secret, avoiding reuse of a JWT bound to the legacy Secret name.
- Once the new Secret has a token, Sync deletes the owned legacy Secret when it is no longer referenced by Pods.
- If Pods still reference the legacy Secret, cleanup is deferred to future normal reconciles or PodCliqueSet deletion.
- PodCliqueSet deletion attempts to remove both the new and legacy Secret names.

Fixes #658

- The legacy Secret is retained only as a compatibility cleanup target; its data is never copied.
- The replacement `<pcs>-ic-sat` Secret is created empty so the Kubernetes token controller issues a fresh token.
- Legacy Secret cleanup waits until the replacement Secret has `data.token`.
- Cleanup does not continuously requeue while Pods still reference the legacy Secret.
- Only a legacy Secret controlled by the same PodCliqueSet is eligible for cleanup.
- If a `<pcs>-ic-sat` Secret already exists but is not controlled by the PodCliqueSet, Sync fails instead of adopting it.
- I also verified the upgrade path locally with a real kind upgrade test from `v0.1.0-alpha.8`; that e2e is intentionally not included in this minimal PR.

#### Does this PR introduce an API change?

```release-note
action required: Grove now uses the shorter `-ic-sat` suffix for init-container service account token Secrets. The legacy `-initc-sa-token-secret` Secret name is still recognized during cleanup, but support for the legacy name will be removed three releases from now. Existing PodCliqueSets should be reconciled or recreated before then.
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
+++ b/operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
@@ -36,6 +36,7 @@ import (
 
 const (
 	errCodeGetSecret              grovecorev1alpha1.ErrorCode = "ERR_GET_SECRET"
+	errCodeSecretConflict         grovecorev1alpha1.ErrorCode = "ERR_SECRET_CONFLICT"
 	errCodeSetControllerReference grovecorev1alpha1.ErrorCode = "ERR_SET_CONTROLLER_REFERENCE"
 	errCodeCreateSecret           grovecorev1alpha1.ErrorCode = "ERR_CREATE_SECRET"
 	errCodeDeleteSecret           grovecorev1alpha1.ErrorCode = "ERR_DELETE_SECRET"
@@ -92,7 +93,7 @@ func (r _resource) Sync(ctx context.Context, logger logr.Logger, pcs *grovecorev
 	if existingSecret != nil {
 		if !metav1.IsControlledBy(existingSecret, pcs) {
 			return groveerr.New(
-				errCodeGetSecret,
+				errCodeSecretConflict,
 				component.OperationSync,
 				fmt.Sprintf("Secret %v already exists but is not controlled by PodCliqueSet %v", objKey, pcsObjKey),
 			)
@@ -208,6 +209,8 @@ func (r _resource) cleanupLegacySecret(ctx context.Context, logger logr.Logger, 
 	if legacySecret == nil || !metav1.IsControlledBy(legacySecret, pcs) {
 		return nil
 	}
+	// Legacy cleanup is best-effort for Pods discoverable through current PCS-managed labels.
+	// Older Pods without these labels may keep the legacy Secret until they are replaced.
 	referencingPods, err := r.getPodsReferencingSecret(ctx, pcs, legacyObjKey.Name)
 	if err != nil {
 		return groveerr.WrapError(err,
--- a/operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
+++ b/operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
@@ -221,11 +221,9 @@ func (r _resource) cleanupLegacySecret(ctx context.Context, logger logr.Logger, 
 	}
 	if len(referencingPods) > 0 {
 		logger.Info("Skipping legacy Secret cleanup because Pods still reference it", "objectKey", legacyObjKey, "pods", referencingPods)
-		return groveerr.New(
-			groveerr.ErrCodeContinueReconcileAndRequeue,
-			component.OperationSync,
-			fmt.Sprintf("Secret %v is still referenced by Pods: %v", legacyObjKey, referencingPods),
-		)
+		// Cleanup is opportunistic on future reconciles or PodCliqueSet deletion.
+		// Referencing Pods may be long-lived, so do not poll while they remain.
+		return nil
 	}
 	if err = client.IgnoreNotFound(r.client.Delete(ctx, legacySecret)); err != nil {
 		return groveerr.WrapError(err,
--- a/operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
+++ b/operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
@@ -39,6 +39,7 @@ const (
 	errCodeSetControllerReference grovecorev1alpha1.ErrorCode = "ERR_SET_CONTROLLER_REFERENCE"
 	errCodeCreateSecret           grovecorev1alpha1.ErrorCode = "ERR_CREATE_SECRET"
 	errCodeDeleteSecret           grovecorev1alpha1.ErrorCode = "ERR_DELETE_SECRET"
+	errCodeListPods               grovecorev1alpha1.ErrorCode = "ERR_LIST_PODS"
 )
 
 type _resource struct {
@@ -107,17 +108,6 @@ func (r _resource) Sync(ctx context.Context, logger logr.Logger, pcs *grovecorev
 	}
 
 	secret := emptySecret(objKey)
-	legacySecret, err := r.getSecret(ctx, getLegacyObjectKey(pcs.ObjectMeta))
-	if err != nil {
-		return groveerr.WrapError(err,
-			errCodeGetSecret,
-			component.OperationSync,
-			fmt.Sprintf("Error getting legacy satokensecret for PodCliqueSet: %v", pcsObjKey),
-		)
-	}
-	if legacySecret != nil && metav1.IsControlledBy(legacySecret, pcs) {
-		secret.Data = copyMigratableSecretData(legacySecret.Data)
-	}
 	if err = r.buildResource(pcs, secret); err != nil {
 		return err
 	}
@@ -185,23 +175,6 @@ func hasServiceAccountToken(secret *corev1.Secret) bool {
 	return len(secret.Data[corev1.ServiceAccountTokenKey]) > 0
 }
 
-func copyMigratableSecretData(data map[string][]byte) map[string][]byte {
-	if len(data) == 0 {
-		return nil
-	}
-	copied := make(map[string][]byte, len(data))
-	for key, value := range data {
-		if key == corev1.ServiceAccountTokenKey {
-			continue
-		}
-		copied[key] = append([]byte(nil), value...)
-	}
-	if len(copied) == 0 {
-		return nil
-	}
-	return copied
-}
-
 func newWaitingForTokenError(objKey client.ObjectKey) error {
 	return groveerr.New(
 		groveerr.ErrCodeRequeueAfter,
@@ -235,6 +208,22 @@ func (r _resource) cleanupLegacySecret(ctx context.Context, logger logr.Logger, 
 	if legacySecret == nil || !metav1.IsControlledBy(legacySecret, pcs) {
 		return nil
 	}
+	referencingPods, err := r.getPodsReferencingSecret(ctx, pcs, legacyObjKey.Name)
+	if err != nil {
+		return groveerr.WrapError(err,
+			errCodeListPods,
+			component.OperationSync,
+			fmt.Sprintf("Error listing Pods referencing legacy satokensecret: %v for PodCliqueSet: %v", legacyObjKey, pcsObjKey),
+		)
+	}
+	if len(referencingPods) > 0 {
+		logger.Info("Skipping legacy Secret cleanup because Pods still reference it", "objectKey", legacyObjKey, "pods", referencingPods)
+		return groveerr.New(
+			groveerr.ErrCodeContinueReconcileAndRequeue,
+			component.OperationSync,
+			fmt.Sprintf("Secret %v is still referenced by Pods: %v", legacyObjKey, referencingPods),
+		)
+	}
 	if err = client.IgnoreNotFound(r.client.Delete(ctx, legacySecret)); err != nil {
 		return groveerr.WrapError(err,
 			errCodeDeleteSecret,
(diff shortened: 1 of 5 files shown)


operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
```diff
@@ -58,45 +58,60 @@ func New(client client.Client, scheme *runtime.Scheme) component.Operator[grovec
...
+			fmt.Sprintf("Error getting legacy satokensecret: %v for PodCliqueSet: %v", legacyObjKey, pcsObjKey),
+		)
+	}
+
 	secret := emptySecret(objKey)
 	if err = r.buildResource(pcs, secret); err != nil {
 		return err
 	}
+	if legacySecret != nil {
+		secret.Data = legacySecret.Data
```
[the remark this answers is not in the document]
yankay: Thanks, good catch. This is fixed now.

The migration now skips `data.token` and only copies non-token data from the legacy Secret, so Kubernetes mints a fresh token for the new `<pcs>-ic-sat` Secret. I added a regression test to cover that behavior.

operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
```diff
@@ -156,6 +172,20 @@ func getLabels(pcsName, secretName string) map[string]string {
...
+func (r _resource) getControlledSecret(ctx context.Context, pcsObjMeta metav1.ObjectMeta, objKey client.ObjectKey) (*corev1.Secret, error) {
+	secret := &corev1.Secret{}
+	if err := r.client.Get(ctx, objKey, secret); err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	if !metav1.IsControlledBy(secret, &pcsObjMeta) {
+		return nil, nil
```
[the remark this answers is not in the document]
yankay: Thanks, fixed. `Sync` now rejects an existing shortened Secret if it is not controlled by the current PodCliqueSet, and the create path no longer ignores `AlreadyExists`.

yankay:
Hi @renormalize and @houshengbo, 

could you please help review this PR when you have a chance? Thanks :-)

operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
```diff
@@ -58,72 +58,96 @@ func New(client client.Client, scheme *runtime.Scheme) component.Operator[grovec
...
+			return groveerr.New(
+				errCodeGetSecret,
+				component.OperationSync,
+				fmt.Sprintf("Secret %v already exists but is not controlled by PodCliqueSet %v", objKey, pcsObjKey),
+			)
+		}
+		if !hasServiceAccountToken(existingSecret) {
+			return newWaitingForTokenError(objKey)
+		}
+		logger.Info("Secret already exists, skipping creation", "existingSecret", client.ObjectKeyFromObject(existingSecret))
```
houshengbo, line 107: Consider adding a cleanup step in this `Sync` func to delete the legacySecret only after the new secret successfully acquires its token (i.e., when hasServiceAccountToken(existingSecret) == true). This ensures the migration is finalized and stale secrets don't linger indefinitely until the parent object is destroyed.

yankay: HI @houshengbo 

Thanks for the suggestion. I updated Sync to clean up the owned legacy Secret only after the replacement Secret has acquired its service account token. :-)

houshengbo (approved):
/lgtm

renormalize (changes requested):
I have one comment: perhaps we can just issue a new token when the new secret is created, instead of copying over the old token? That would probably simplify the flow and keep the same functionality? What do you think?

yankay:
> /lgtm

HI @houshengbo 

Thanks for pointing this out. While working on the legacy Secret cleanup, I realized there is a migration risk: existing Pods may still reference the old <pcs>-initc-sa-token-secret, so deleting it immediately could break those Pods.

I updated the cleanup flow to first list PCS-managed Pods and check whether any Pod volume still references the legacy Secret. If so, cleanup skips deletion and requests a follow-up reconcile; once no Pods reference it anymore, the legacy Secret is deleted.

yankay:
> I have one comment: perhaps we can just issue a new token when the new secret is created, instead of copying over the old token? That would probably simplify the flow and keep the same functionality? What do you think?

HI @renormalize 

Thanks for the suggestion. I updated the new shortened Secret to be created empty, so Kubernetes issues a fresh service account token instead of reusing data from the legacy Secret.

houshengbo:
@yankay Thanks for the updates.
@renormalize Could you give another review of the new change?

houshengbo (commented):
The migration flow is well-designed. The sequencing — create empty new Secret → wait for K8s to mint a fresh token → check pod references → delete legacy — is correct and avoids any window where a pod loses its credentials. A few observations:

**`errCodeGetSecret` reused for an ownership conflict**

In `Sync`, when the new Secret exists but isn't controlled by the PCS, the code returns `errCodeGetSecret`. That error code reads as "failed to fetch the Secret" to any operator reading the logs or error events, while the real situation is a resource ownership conflict. A dedicated `errCodeConflict` (or similar) would make this distinct and easier to triage in production.

**Pod-reference check relies on PCS-managed-resource labels**

`getPodsReferencingSecret` lists pods using `GetDefaultLabelsForPodCliqueSetManagedResources`. This is fine for steady-state, but pods that predate the label set or that were created by an older Grove version might be missed, leaving the legacy Secret stranded until those pods are replaced. Worth a short comment in `cleanupLegacySecret` documenting this assumption so future maintainers know the cleanup is best-effort for pre-label pods.

**Deprecation timeline tracking**

The `GenerateLegacyInitContainerSATokenSecretName` doc says "scheduled for removal three releases after the shortening" and links #658 — that is sufficient. One small nit: it might be worth cross-referencing the specific version or milestone in that issue so the removal is not forgotten.

**Everything else looks good**

- Fresh-token approach (empty new Secret, K8s mints) is correct and avoids JWT bound to the old name.
- Collision protection (error on unowned existing Secret, no `IgnoreAlreadyExists`) addresses the earlier Codex review point cleanly.
- `ContinueReconcileAndRequeue` for the "pods still reference legacy" path is the right error code.
- Test coverage is thorough: migration path, no-ownership legacy, unowned collision, pod-reference deferral, and the label-value budget test in `pod_test.go`.

operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
```diff
@@ -58,72 +59,88 @@ func New(client client.Client, scheme *runtime.Scheme) component.Operator[grovec
...
-			fmt.Sprintf("Error getting existing satokensecret names for PodCliqueSet: %v", pcsObjKey),
+			fmt.Sprintf("Error getting satokensecret: %v for PodCliqueSet: %v", objKey, pcsObjKey),
 		)
 	}
-	if len(existingSecretNames) > 0 {
-		logger.Info("Secret already exists, skipping creation", "existingSecret", existingSecretNames[0])
+	if existingSecret != nil {
+		if !metav1.IsControlledBy(existingSecret, pcs) {
+			return groveerr.New(
+				errCodeGetSecret,
```
houshengbo: Minor nit: `errCodeGetSecret` here describes a get-failure, but this branch is actually an ownership conflict — the Secret exists and was fetched successfully. A distinct code like `errCodeConflict` (or similar) would make it easier to distinguish this case in logs and error events.

operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
```diff
@@ -156,6 +173,98 @@ func getLabels(pcsName, secretName string) map[string]string {
...
+		return groveerr.WrapError(err,
+			errCodeGetSecret,
+			component.OperationSync,
+			fmt.Sprintf("Error getting legacy satokensecret for PodCliqueSet: %v", pcsObjKey),
+		)
+	}
+	if legacySecret == nil || !metav1.IsControlledBy(legacySecret, pcs) {
+		return nil
+	}
+	referencingPods, err := r.getPodsReferencingSecret(ctx, pcs, legacyObjKey.Name)
```
houshengbo, line 214: The pod list here is filtered by `GetDefaultLabelsForPodCliqueSetManagedResources`, so any pods that predate this label set or were created by an older Grove version won't be found — meaning the legacy Secret would not be deleted until those pods are replaced. This is acceptable as a best-effort approach, but worth a brief comment here so future readers understand the assumption.

yankay:
HI @houshengbo 

Thanks a lot for the review. I addressed the comments in the latest push and updated the deprecation note with the timeline.

houshengbo:
Thanks.
@yankay Are there any other reviewers you need to reach out to?
@renormalize Could you give another round of review?

houshengbo (approved):

yankay:
> Thanks. @yankay Are there any other reviewers you need to reach out to? @renormalize Could you give another round of review?

Thanks for checking. No other reviewers needed from my side, I just rebased the PR onto the latest main :-)

renormalize:
@houshengbo @yankay sorry for the delay! I will do one final pass of this and add my approval, but after a brief look I think everything looks good. We need another reviewer's approval since we require at least two approvals for a PR to be merged.

yankay:
Hi @unmarshall , sorry to bother you. Would you mind taking a look at it,  when you have some bandwidth?

renormalize (dismissed):
Looks good!

Checking the owner reference for the secret fetched is overkill in my opinion, since the name of the secret always contains the PCS name which is unique; but I don't have any issues with that being a part of the PR. (Even if a PCS with an identical name is created right after, cascade deletion will ensure cleanup of the secrets that were created for that particular PCS).

Let's wait for one more review and merge this.

houshengbo:
@shayasoolin @Ronkahn21
This PR is close to merge. Could you review this one, when you get a chance?

yankay:
ref to: #711

operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
```diff
@@ -156,6 +174,100 @@ func getLabels(pcsName, secretName string) map[string]string {
...
+	if err != nil {
+		return groveerr.WrapError(err,
+			errCodeListPods,
+			component.OperationSync,
+			fmt.Sprintf("Error listing Pods referencing legacy satokensecret: %v for PodCliqueSet: %v", legacyObjKey, pcsObjKey),
+		)
+	}
+	if len(referencingPods) > 0 {
+		logger.Info("Skipping legacy Secret cleanup because Pods still reference it", "objectKey", legacyObjKey, "pods", referencingPods)
+		return groveerr.New(
```
shayasoolin: Keeping the legacy secret while old pods still reference it makes sense, and we should not roll pods just for this rename. But because that can be a normal long-lived state, returning  ErrCodeContinueReconcileAndRequeue` here makes the controller wake up every 5s until those pods naturally disappear, which may be forever for stable workloads. Could this return `nil` instead and make legacy cleanup opportunistic on future normal reconciles/deletion?

shayasoolin (commented):
See one point, other than that LGTM.
Can you please update the PR description to reflect the updated design decision, to NOT copy secret data but to create a new empty secret?

houshengbo:
@yankay
Could you revisit this PR?

operator/internal/controller/podcliqueset/components/satokensecret/satokensecret.go
yankay: Thanks @shayasoolin. I updated cleanup to return nil while old Pods reference the legacy Secret, and corrected the PR description :-)

yankay:
> @yankay Could you revisit this PR?

Thanks for the reminder. I revisited the PR, addressed the latest feedback, and updated the PR description. :-)

houshengbo:
@renormalize @shayasoolin Could you review this PR again?

shayasoolin (approved):
LGTM, thanks @yankay!

houshengbo:
@renormalize We need your approve as well.

sanjaychatterjee (approved):
LGTM

julienmancuso (approved):

PR #663, opened 2026-06-10
feat: lpx scheduler backend

### LINKED ISSUE #664: add lpx-scheduler as an opt-in scheduler backend

register `lpx-scheduler` as a valid opt-in Grove scheduler backend, selected explicitly via `schedulerName`, so gang-scheduled workloads can delegate placement to an external DRA-native scheduler.

### Why is this needed?

we run gang-scheduled accelerator workloads whose placement requirements (exact device selection against DRA-published inventory, topology-aware gang binding) are owned by a dedicated scheduler. that scheduler is internal today, with the intent to open source it. 

grove's scheduler registry currently accepts only the default and KAI schedulers, so a `PodCliqueSet` selecting any other `schedulerName` fails validation. registering the backend name upstream lets us run unmodified upstream grove rather than maintaining a fork for a one-line registration plus validation. the backend is inert on clusters that don't deploy the external scheduler: explicit opt-in only, no runtime or build-time dependency on it existing.

kathleenfrench:
## summary

this adds `lpx-scheduler` as an explicit, opt-in Grove scheduler backend.
Grove still owns workload materialization and lifecycle; the external scheduler
owns topology solving, DRA device allocation, and final Pod binding.

the backend is intentionally narrow:

- `PreparePod` sets `spec.schedulerName: lpx-scheduler`;
- Grove preserves concrete Pod identity, PodGang membership, scheduling gates,
  and resource-claim references;
- `SyncPodGang` is a no-op because LPX consumes Grove's native PodGang directly;
- Grove topology constraints are rejected for LPX workloads so two schedulers
  cannot both claim topology authority; and
- existing Kubernetes, KAI, and Volcano backends remain unchanged.

```mermaid
flowchart LR
    A["workload opts in"] --> B["Grove PodCliqueSet"]
    B --> C["Grove Pods + PodGang + claims"]
    C --> D["lpx-scheduler"]
    D -->|"allocates claims + binds Pods"| C
```

## post-#516 integration

this branch is rebased onto `b3cf9caf`, after the scheduler backend interface
changes in #516. LPX now follows the same registry lifecycle as the other
backends:

- `Init(client.Client) error`;
- `PreparePod(*corev1.Pod) error`;
- the current base `Backend` interface, without the removed
  `OnPodGangDelete` hook; and
- coexistence with the new Volcano backend in the scheduler registry.

the LPX backend is not topology-aware from Grove's perspective. it therefore
does not register topology watches or create backend-specific scheduling
resources.

## why a named backend is needed

Grove validates scheduler names against enabled backends before creating Pods.
without this registration, an LPX workload cannot retain Grove's normal
PodCliqueSet, PodGang, claim-template, and Pod lifecycle while delegating
placement to `lpx-scheduler`.

the external scheduler is NVIDIA-internal today and is expected to be open
sourced. this change contains no dependency on that codebase or its APIs:
clusters that do not enable the profile are unaffected, and enabling the
profile without deploying the scheduler simply leaves opted-in Pods pending.

## scope

this PR does not:

- solve placement;
- inspect or mutate DRA allocations;
- bind Pods;
- prune candidate Pods;
- change `resourceSharing`; or
- add a Grove topology-affinity implementation for LPX.

## validation

the branch passes Grove generation, formatting, lint, full unit tests, build,
changed-package race tests, Helm lint, and chart rendering.

it also passes the campaign's isolated Kubernetes 1.35 Kind round trip against
this exact Grove commit. that proof deploys the real Grove and Dynamo
controllers, `lpx-scheduler`, and the real remote LPU DRA node plugin. it
covers legacy coexistence and upgrade, all four public LPX workload modes,
exact allocation and UID-safe binding, restart, replacement fencing, release,
driver restart recovery, and GPU inventory invalidation.

## issue

fixes #664

## release note

```release-note
adds lpx-scheduler as an opt-in scheduler backend while Grove retains workload materialization, PodGang, claim-reference, and lifecycle ownership.
```

### DIFF
--- a/operator/internal/scheduler/lpx/backend.go
+++ b/operator/internal/scheduler/lpx/backend.go
@@ -1,0 +1,88 @@
+// /*
+// Copyright 2026 The Grove Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// */
+
+package lpx
+
+import (
+	"context"
+	"errors"
+
+	configv1alpha1 "github.com/ai-dynamo/grove/operator/api/config/v1alpha1"
+	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
+	"github.com/ai-dynamo/grove/operator/internal/scheduler"
+
+	groveschedulerv1alpha1 "github.com/ai-dynamo/grove/scheduler/api/core/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	errTopologyConstraintsUnsupported = errors.New(
+		"lpx-scheduler does not support Grove topology constraints; " +
+			"placement topology must be expressed through the LPX workload contract",
+	)
+
+	_ scheduler.Backend = (*schedulerBackend)(nil)
+)
+
+type schedulerBackend struct {
+	name string
+}
+
+// New creates an LPX scheduler backend.
+func New(profile configv1alpha1.SchedulerProfile) scheduler.Backend {
+	return &schedulerBackend{name: string(profile.Name)}
+}
+
+func (b *schedulerBackend) Name() string {
+	return b.name
+}
+
+func (b *schedulerBackend) Init(_ client.Client) error {
+	return nil
+}
+
+func (b *schedulerBackend) SyncPodGang(
+	_ context.Context,
+	_ *groveschedulerv1alpha1.PodGang,
+) error {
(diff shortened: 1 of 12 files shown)


operator/internal/scheduler/lpx/backend.go
```diff
@@ -0,0 +1,88 @@
...
+
+	groveschedulerv1alpha1 "github.com/ai-dynamo/grove/scheduler/api/core/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	errTopologyConstraintsUnsupported = errors.New(
+		"lpx-scheduler does not support Grove topology constraints; " +
+			"placement topology must be expressed through the LPX workload contract",
```
unmarshall, line 35: is it not possible to map the Grove topology constraints to LPX workload contract?

unmarshall (commented):
Minor comments. Rest looks good.

sanjaychatterjee (approved):
lgtm

unmarshall (approved):

operator/internal/scheduler/lpx/backend.go
kathleenfrench: not completely. grove constraints express pod placement across k8s topology domains; lpx must also solve DRA devices, multi-tray LPU partitions, prop-sync connectivity, and LPU/GPU fabric locality. we can map the overlapping pod-level intent, but the device-level allocation contract still has to remain in lpx.

PR #666, opened 2026-06-11
Introduce `uv`

### LINKED ISSUE #665: Migrate to `uv` for Python package management

I would like Grove to depend on `uv` for all things Python. This solves the issues of having to deal with Python versions, `venv`s, and package management through `pip`.

### Why is this needed?

`uv` simplifies the Python tool chain significantly, and enables Grove contributors who are more familiar with Go in general to focus on code.

This would also align Grove with the general community standards; Grove's primary consumer [ai-dynamo/dynamo](https://github.com/ai-dynamo/dynamo/blob/main/pyproject.toml) also consumes `uv`.

renormalize:
#### What type of PR is this?

/kind feature

Instead of battling with missing Python versions, `pip`, and `venv`s, we use `uv` to simplify the Python experience for developers of Grove who need to run E2E tests.

`uv` needs to be installed once, and all Python+dependencies are fully handled by it.

The `e2e-cluster-up` target also installs the tool in `hack/tools/bin` for developers who simply want to run the E2E tests, and not modify any Python related code. This lowers the barrier further by not requiring the developer to install `uv` manually at all.

This also helps grove follow the current community standards, and aligns it closely with the practices in `ai-dynamo/dynamo`.

Fixes #665

```release-note
NONE
```

```docs
NONE
```


operator/Makefile
yankay: Hi @renormalize,

One small suggestion: since `infra-manager.py` now uses the `uv` shebang, it might be worth adding `$(UV)` to the other targets that invoke it as well. This would keep the entry points consistent and avoid failures on fresh machines where `uv` is not already installed:

- `e2e-cluster-down: $(UV)`
- `scale-cluster-up: $(UV)`
- `scale-cluster-down: $(UV)`

renormalize: Good catch, I was intending to do something like this, but it slipped my mind. Thanks!

renormalize: 33445177a8319f964a30d184783e5591249e2c58 should address this.

operator/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# */
+
+[project]
+name = "grove"
+version = "0.1.0-dev"
+description = "Kubernetes operator for disaggregated inference"
```
shayasoolin: Is this Grove's agreed formal description? I think it should be more generic than 'disagg inf'.

shayasoolin (dismissed):

operator/pyproject.toml
renormalize: Thanks for the review @shayasoolin. I checked the README and re-phrased the first sentence into this. Agreed, I shall make it more generic than the current description.

renormalize: @shayasoolin I've updated the description, and rebased the PR. Let me know if you require any more changes from this PR from me. Thanks!

shayasoolin (dismissed):

operator/pyproject.toml
```diff
@@ -17,7 +17,7 @@
 [project]
 name = "grove"
 version = "0.1.0-dev"
-description = "Kubernetes operator for disaggregated inference"
+description = "Kubernetes operator for AI inference workloads"
```
shayasoolin: Can you remove `inference` from here? See https://github.com/ai-dynamo/grove/pull/693

shayasoolin (dismissed):

Ronkahn21 (dismissed):

renormalize:
Thank you for the reviews @shayasoolin @Ronkahn21. Once the CI runs, let's go ahead and merge this PR.

Ronkahn21 (approved):

shayasoolin (approved):

PR #667, opened 2026-06-11
tighten pod and podclique predicate filtering

### LINKED ISSUE #409: Measure and potentially optimize number of enqueues of events across reconcilers

The scale at which Grove needs to operate requires that we carefully measure the number of enqueue events into each reconciler queue. This will serve two purposes:

1. We can get inputs if we need to reduce the number of events, filtering events that should not be reconciled.
2. Configure the number of concurrent workers for each reconciler queue.

### Why is this needed?

To help scale the reconcilers with increase in the number of enqueue events.

oleg-kushniriov:
#### What type of PR is this?

/kind cleanup

Tightens predicate filtering in two controllers to reduce unnecessary reconcile enqueues:

- **podclique controller**: `hasPodStatusChanged` no longer tracks per-container
  `Started`/`Ready` transitions when the pod is already fully ready. Those transitions
  are only relevant to the rolling-update state machine (`PodStartedButNotReady`
  classification); once a pod is Ready, any container churn is already captured by
  the existing `hasReadyConditionChanged` check. This reduced pod-source enqueues
  by ~68% under soak testing.

- **podcliqueset controller**: `hasAnyStatusReplicasChanged` now fires on `ReadyReplicas`
  only at the fully-ready boundary (not-fully-ready ↔ fully-ready), rather than on
  every intermediate pod-ready increment. The PodCliqueSet reconciler only cares
  whether a PodClique became fully available or dropped from fully available.

Fixes #409

The `hasStartedAndReadyChangedForAnyContainer` check in `podclique/register.go` is
still present and still fires — only guarded by `!k8sutils.IsPodReady(newPod)` so it
remains active during rolling updates when pods are transitioning.

The `ReadyReplicas` boundary logic uses `Replicas > 0 && ReadyReplicas == Replicas`
as the fully-ready predicate, which correctly handles scale-down to zero (not fully
ready when Replicas is 0).

```release-note

### DIFF
--- a/operator/internal/controller/podclique/register.go
+++ b/operator/internal/controller/podclique/register.go
@@ -153,10 +153,8 @@ func hasPodStatusChanged(updateEvent event.UpdateEvent) bool {
 		hasLastTerminationStateChanged(oldPod.Status.ContainerStatuses, newPod.Status.ContainerStatuses) {
 		return true
 	}
-	// Per-container started/ready transitions drive the rolling-update state machine
-	// (PodStartedButNotReady classification in rollingupdate.go). They are only
-	// meaningful while the pod is not yet fully ready; once Ready, container churn
-	// is already covered by the checks above.
+	// Per-container started/ready transitions are only meaningful while the pod is
+	// not yet fully ready; once Ready, container churn is already covered above.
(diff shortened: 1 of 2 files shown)


shayasoolin:
Looks legit but also risky, e.g. what if PCS controller does care about the exact number of ready replicas, for its status updates?
I think we should first asses the performance gain and then decide if it's worth the price that comes with event filtering.

PR #669, opened 2026-06-15
fix: harden topology sync against transient API server errors at startup

### LINKED ISSUE #654: Harden ClusterTopologyBinding startup synchronization against transient API server errors

Make Grove's startup-time synchronization of scheduler-specific topology resources resilient to transient Kubernetes API server errors, so that a single failed call does not crash the operator process.

Today, `clustertopology.SynchronizeTopology` (operator/internal/clustertopology/clustertopology.go) is invoked from operator/cmd/main.go before mgr.Start() and uses a direct (non-cached) client to:
  1. List all ClusterTopologyBinding resources.
  2. For each CT × auto-managed TopologyAwareBackend, call SyncTopology (which internally performs Get and conditionally Create / Delete on the backend topology CR — e.g. kai.Topology). The number of calls increase with more configured scheduler backends.

Any single error in any of these calls is propagated up to main.go, where the operator exits with `cli.ExitErrSynchronizeTopology`. There is no retry, no backoff, and no partial-success tolerance. Recovery currently relies entirely on the kubelet restarting the pod.

Concrete improvements proposed (in increasing order of disruption — we can pick one or layer them):
 1. **[Cheapest]** In-process bounded retry with backoff. Wrap `SynchronizeTopology` in `wait.ExponentialBackoff` (or `retry.OnError`) so transient List/Get/Create/Delete failures don't crash the operator.
  2. Best-effort startup & authoritative reconcile. Log-and-continue on startup sync errors, and rely on the `ClusterTopologyBinding` controller's steady-state reconcile loop (which already calls `SyncTopology`) to converge. To preserve the race-free guarantee, gate `PodCliqueSet` reconciliation on a CT status condition (e.g. SchedulerTopologyDrift==False) and requeue otherwise, turning the race into a self-healing wait. Of course this should only be done if `TopologyAwareSchedulingConfiguration.Enabled` is true and for the given PCS topology constraints are defined.
  3. **[Preferred]** Two-phase startup with readiness gating. Operator process stays up; mgr.Start() runs; but a custom `manager.Runnable` blocks PodCliqueSet controller startup until initial topology sync succeeds. The readiness probe reports "not ready" with a clear reason during this window, so the cluster sees the operator's actual state without restarting it.

Scope of this issue: agree on an approach, then implement. 
Out of scope: changes to the auto-managed vs. externally-managed split, or to the TopologyAwareBackend interface itself.

### Why is this needed?

The current design optimizes for one thing — eliminating the race where a `PodCliqueSet` reconcile could consume a topology reference before the backend topology CR exists — by making startup synchronization a hard precondition for the operator running at all. That tradeoff is too steep in practice:

- Transient API server unavailability (apiserver rolling restart, etcd hiccup, leader election churn, network blip) crashloops the Grove operator pod, even though Grove could safely come up and converge on its next reconcile.
- Thundering herd on control-plane recovery. When many operators start simultaneously after a control-plane event, retry behavior driven only by kubelet exponential backoff (no application-level jitter) amplifies load on the API server right when it's most fragile.
- Multiplied by replica count (leader election doesn't help here):  Grove uses `controller-runtime` leader election (operator/internal/controller/manager.go:101-107), but `SynchronizeTopology` runs from main.go:105 before mgr.Start() — and leader election only activates inside mgr.Start(). Therefore every replica (typically 2-3 in HA deployments) independently runs the full C × B × k sync (explained above) against the API server before any of them attempt to acquire leadership, multiplying every failure mode and load number above by the replica count.
- Cross-component coupling. If the KAI Topology CRD has a validating webhook and that webhook is unavailable, Grove cannot start — KAI's health becomes Grove's startup precondition, which is an unwanted dependency. 
 Let me explain this a bit further:
<details>
<summary>Detailed call sequence</summary>

For the auto-managed path, Grove will create OR delete + create the KAI topology CR (taking KAI as an example scheduler   backend CR, however the same is applicable for Volcano or any other scheduler backend CRs that are supported tomorrow)
The `Create` doesn't go straight to etcd. The Kubernetes API server runs the request through the admission chain first, which includes any validating or mutating webhooks registered for the `kai.scheduler/v1alpha1`, topologies resource. Those webhooks are HTTP endpoints served by the KAI scheduler's own controllers/webhook deployment, not by Grove.
So the actual sequence on a Create is:
  1. Grove → kube-apiserver: "create this KAI Topology"
  2. kube-apiserver → KAI's validating webhook (/validate-topologies): "is this OK?"
  3. KAI webhook → kube-apiserver: "yes / no"
  4. kube-apiserver → Grove: result
If step 3 fails — KAI's webhook pod is down, crashlooping, not yet scheduled, has expired certs, or its Service isn't routable — kube-apiserver returns an error to step 4. With the default webhook failurePolicy: Fail (which is what most webhooks ship with), the apiserver actively rejects the `Create`.
</details>
- Recoverable failure modes are treated as fatal. Permanent misconfiguration (missing CRD, RBAC, hijacked resource per IsControlledBy) deserves a hard failure; transient API errors do not. The current code does not distinguish between them.

#### Scalability
Now lets analyse the number of `kube-apiserver` (KAPI) calls that are required during startup. Since during the startup we are using the `uncached` client, all calls directly go to KAPI. The GET calls are also not serviced by the informer cache since there is no cache (why? - because the manager has not started yet).

```
Let:
    C = number of ClusterTopologyBinding resources
    B = number of TAS-capable backends
    E = average number of externally-managed bindings per CT (excluded from sync)

  API calls at startup:
    1                                  List ClusterTopologyBinding
    + C × (B - E) ×  k                 SyncTopology calls
                                       where k = API calls per SyncTopology

For KAI's SyncTopology (operator/internal/scheduler/kai/topology.go:56-96), k is:
  - Happy path, no drift: 1 (Get).
  - First reconcile, NotFound: 2 (Get + Create).
  - Drift: 3 (Get + Delete + Create).
```

So a cluster with C=50 CTs and B=3 backends (KAI + Volcano + Koordinator, hypothetically) hits the apiserver with ~150 sequential Get requests minimum at startup, before the manager has even started, before readiness probe can pass, and before any controller can do useful work. With drift on a fraction of resources, it climbs to 300–450.

  That math is conservative. Three additional multipliers:

  1. Webhook fan-out. Each Create and Delete on a backend topology CR triggers that backend's admission webhooks (the cross-component coupling point), so each "API call" is really
  apiserver → webhook → apiserver. From the apiserver's load perspective, that's two RPC hops per write.
  2. Auto-managed default. A CT with no schedulerTopologyBindings entry triggers SyncTopology for every TAS backend (reconciler.go:73-79, clustertopology.go:42-51). The default — and
  probably common — case is full B-fold fan-out, not partial.
  3. Sequential, not parallel. The current code iterates for backendName, tasBackend := range backends inside a for i := range ctList.Items (clustertopology.go:38-51). No concurrency, no
   batching. Wall-clock time grows linearly even when the apiserver could absorb parallel requests, which lengthens the startup window during which Grove is unavailable.

### Acceptance criteria
 - Grove operator process does not exit on transient API server errors (5xx, network timeouts, webhook unavailable) at startup. Verified by an integration test that injects a transient error and asserts the operator becomes ready after recovery.
- Permanent misconfiguration (missing CRD, RBAC denial, ownership conflict per IsControlledBy) still surfaces clearly — either as a hard exit with a distinct exit code, or via a readiness condition with a permanent reason.
- PodCliqueSet reconciliation does not produce stale or partially-bound workloads when a CT's backend topology has not yet been synchronized. Verified by a test that creates a PCS while the backend topology resource is absent and asserts the PCS reconcile requeues with a clear status condition rather than producing a partially-bound workload.
- ClusterTopologyBinding.status (drift condition + per-backend message) is updated within one reconcile interval of any backend topology becoming reachable again.
- No regression in startup time on the happy path (no transient errors).

#### Failure mode classification

  | # | Failure scenario | Root cause | Today's behavior | Proposed behavior (option 3) | Recoverable? |
  |---|---|---|---|---|---|
  | 1 | API server returns `5xx` / connection timeout | Transient — apiserver rolling restart, etcd hiccup, network blip | Operator process exits via `cli.ExitErrSynchronizeTopology`; kubelet crashloops the pod | Bounded retry with backoff inside the startup runnable; readiness reports "not ready, retrying topology sync"; process stays up | Yes |
  | 2 | Backend topology webhook unavailable (e.g. KAI's `/validate-topologies` endpoint not reachable) | Transient — webhook pod down, rolling, expired certs, Service routing issue | Operator process exits — Grove's startup is gated on KAI's admission path | Retry with backoff; readiness reports "not ready, waiting on KAI webhook"; CT controller surfaces backend-specific error in `status.message` once manager starts | Yes |
  | 3 | Backend topology CRD not installed in cluster | Permanent — KAI / Volcano / Koordinator not deployed | Operator process exits | Hard fail with distinct exit code **or** permanent not-ready condition with reason `BackendTopologyCRDMissing`; do not crashloop indefinitely | No |
  | 4 | RBAC denial on backend topology resource (`forbidden`) | Permanent — operator's ServiceAccount lacks the verbs | Operator process exits | Hard fail with distinct exit code and clear log pointing at the missing role/binding | No |
  | 5 | Existing backend topology resource not owned by this CT (`IsControlledBy` returns false) | Permanent for that one CT — name collision or hijack | Whole startup sync exits; **all** other CTs are also blocked | Per-CT failure recorded in `status.schedulerTopologyStatuses[].message`; controller continues processing other CTs and other backends; condition `SchedulerTopologyDrift=True` with reason `OwnershipConflict` | No (per CT); Yes (operator overall) |
  | 6 | `ct.spec.levels` changed since last reconcile (drift on auto-managed backend) | Expected — user updated the CT | Delete + Create on backend topology during startup sync; full `C × B × k` burst on every restart | Handled by steady-state CT controller reconcile via the workqueue; no startup cost when nothing changed since last leader's reconcile | N/A — not a failure |
  | 7 | Backend topology resource missing on cluster (auto-managed CT, fresh install or manual deletion) | Expected first-reconcile state | `Get` returns NotFound → `Create` issued during startup sync | Same `Create` issued by the steady-state CT controller reconcile, gated by leader election | N/A — not a failure |
  | 8 | `ClusterTopologyBinding` references a scheduler backend that is not enabled | Permanent — user typo or misconfiguration | Currently surfaced via `status.schedulerTopologyStatuses[].message` (this path already handled correctly in `reconciler.go:119-128`) | Unchanged | No (user must fix spec) |
  | 9 | Network partition between operator and apiserver lasting longer than retry budget | Transient but extended | Operator crashloops for the duration | Retry budget exhausted → readiness stays not-ready with reason `APIServerUnreachable`; pod stays up; converges automatically when partition heals | Yes |
  | 10 | Multiple operator replicas all running startup sync (leader election not yet active) | `SynchronizeTopology` runs in `main.go` before `mgr.Start()` | `C × B × k × replicas` burst on every restart of any replica | Move sync into a `manager.Runnable` so it runs only on the elected leader; non-leader replicas stand by | N/A — design fix |

AsadShahid04:
## Summary
- Add `SynchronizeTopologyWithRetry` wrapping `SynchronizeTopology` with bounded exponential backoff
- Permanent errors (Forbidden, Unauthorized) propagate immediately without retrying
- `main.go` updated to call the retry wrapper with a sensible default backoff (~63 s budget)

## Problem

`clustertopology.SynchronizeTopology` is called from `main.go` before `mgr.Start()`. Any single transient error — a 5xx from a rolling API server, a webhook pod not yet ready, a brief etcd hiccup — causes the operator to exit via `cli.ExitErrSynchronizeTopology`. The only recovery path is kubelet restarting the pod, amplifying load on a control plane that is already under stress.

See #654 for the full failure-mode analysis.

## Solution

Introduce `SynchronizeTopologyWithRetry(ctx, cl, logger, backends, backoff)` in `internal/clustertopology/clustertopology.go`:

- Uses `k8s.io/client-go/util/retry.OnError` with `wait.Backoff` (already a transitive dependency).
- `isTransientAPIError` classifies 5xx, service-unavailable, server-timeout, request-timeout, and 429 as retriable.
- Forbidden and Unauthorized (permanent RBAC/auth failures) are returned immediately — no retrying.
- Default backoff: 1 s initial, factor 2.0, 10% jitter, 6 steps ≈ 63 s total budget.
- Each retry logs `"Retrying topology synchronization after transient error"` with the attempt number.

`main.go` is updated to call `SynchronizeTopologyWithRetry` with `DefaultSyncRetryBackoff`.

## Testing
- `go build ./...`: passed
- Unit tests: 3 new tests added
  - `TestSynchronizeTopologyWithRetry_SucceedsAfterTransientError` — asserts success after 2 injected 500 errors
  - `TestSynchronizeTopologyWithRetry_PermanentErrorNotRetried` — asserts Forbidden stops immediately (1 call only)
  - `TestSynchronizeTopologyWithRetry_ExhaustsRetriesOnPersistentTransientError` — asserts last transient error returned after budget exhausted
- All existing `TestSynchronizeTopology*` tests continue to pass (no behaviour change to the base function)
- Full non-e2e suite: all packages pass

Closes #654

### DIFF
--- a/operator/internal/clustertopology/clustertopology.go
+++ b/operator/internal/clustertopology/clustertopology.go
@@ -18,7 +18,11 @@ package clustertopology
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
+	"net"
+	"syscall"
 	"time"
 
 	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
@@ -55,14 +59,25 @@ func SynchronizeTopologyWithRetry(ctx context.Context, cl client.Client, logger 
 	})
 }
 
-// isTransientAPIError reports whether err is a transient Kubernetes API server error
-// that is safe to retry. Permanent errors (Forbidden, Unauthorized) return false.
+// isTransientAPIError reports whether err is a transient error that is safe to retry.
+// It covers both Kubernetes API-level errors (5xx, timeouts) and transport-level
+// errors (connection refused, EOF, net timeout) that surface before a StatusError
+// is returned — e.g. during an API-server rolling restart or network partition.
+// Permanent errors (Forbidden, Unauthorized) return false.
 func isTransientAPIError(err error) bool {
-	return apierrors.IsServerTimeout(err) ||
+	if apierrors.IsServerTimeout(err) ||
 		apierrors.IsServiceUnavailable(err) ||
 		apierrors.IsInternalError(err) ||
 		apierrors.IsTimeout(err) ||
-		apierrors.IsTooManyRequests(err)
+		apierrors.IsTooManyRequests(err) {
+		return true
+	}
+	// Transport-level transient errors.
+	var netErr net.Error
+	if errors.As(err, &netErr) && netErr.Timeout() {
+		return true
+	}
+	return errors.Is(err, io.EOF) || errors.Is(err, syscall.ECONNREFUSED)
 }
 
 // SynchronizeTopology synchronizes scheduler-specific topology resources at operator startup.
--- a/operator/internal/clustertopology/clustertopology.go
+++ b/operator/internal/clustertopology/clustertopology.go
@@ -19,13 +19,51 @@ package clustertopology
 import (
 	"context"
 	"fmt"
+	"time"
 
 	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
 	"github.com/ai-dynamo/grove/operator/internal/scheduler"
 
 	"github.com/go-logr/logr"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+// DefaultSyncRetryBackoff is the retry backoff used by SynchronizeTopologyWithRetry.
+// Six steps with doubling delay gives roughly 60 seconds of retry budget (1+2+4+8+16+32).
+var DefaultSyncRetryBackoff = wait.Backoff{
+	Duration: time.Second,
+	Factor:   2.0,
+	Jitter:   0.1,
+	Steps:    6,
+}
+
+// SynchronizeTopologyWithRetry wraps SynchronizeTopology with a bounded exponential
+// backoff so that transient API-server errors (5xx, timeouts, service-unavailable)
+// at operator startup do not crash the operator process.
+// Permanent errors (Forbidden, Unauthorized) are returned immediately without retrying.
+func SynchronizeTopologyWithRetry(ctx context.Context, cl client.Client, logger logr.Logger, backends map[string]scheduler.TopologyAwareBackend, backoff wait.Backoff) error {
+	attempt := 0
+	return retry.OnError(backoff, isTransientAPIError, func() error {
+		if attempt > 0 {
+			logger.Info("Retrying topology synchronization after transient error", "attempt", attempt+1)
+		}
+		attempt++
+		return SynchronizeTopology(ctx, cl, logger, backends)
+	})
+}
+
+// isTransientAPIError reports whether err is a transient Kubernetes API server error
+// that is safe to retry. Permanent errors (Forbidden, Unauthorized) return false.
+func isTransientAPIError(err error) bool {
+	return apierrors.IsServerTimeout(err) ||
+		apierrors.IsServiceUnavailable(err) ||
+		apierrors.IsInternalError(err) ||
+		apierrors.IsTimeout(err) ||
+		apierrors.IsTooManyRequests(err)
+}
 
 // SynchronizeTopology synchronizes scheduler-specific topology resources at operator startup.
 // Lists all existing ClusterTopologyBinding resources and ensures backend topologies exist for each.
--- a/operator/cmd/main.go
+++ b/operator/cmd/main.go
@@ -102,7 +102,8 @@ func main() {
 
 	// Synchronize backend topologies for all existing ClusterTopologyBinding resources.
 	// This must be done before starting the controllers that may depend on the ClusterTopologyBinding resource.
-	if err = clustertopology.SynchronizeTopology(ctx, cl, logger, schedRegistry.AllTopologyAware()); err != nil {
+	// Uses a bounded exponential backoff so that transient API-server errors do not crash the operator.
+	if err = clustertopology.SynchronizeTopologyWithRetry(ctx, cl, logger, schedRegistry.AllTopologyAware(), clustertopology.DefaultSyncRetryBackoff); err != nil {
 		logger.Error(err, "failed to synchronize cluster topology")
 		handleErrorAndExit(err, cli.ExitErrSynchronizeTopology)
 	}
--- a/operator/internal/clustertopology/clustertopology_test.go
+++ b/operator/internal/clustertopology/clustertopology_test.go
@@ -18,6 +18,10 @@ package clustertopology
 
 import (
 	"context"
+	"fmt"
+	"io"
+	"net"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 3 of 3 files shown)


operator/internal/clustertopology/clustertopology.go
```diff
@@ -19,14 +19,52 @@ package clustertopology
...
+	Factor:   2.0,
+	Jitter:   0.1,
+	Steps:    6,
+}
+
+// SynchronizeTopologyWithRetry wraps SynchronizeTopology with a bounded exponential
+// backoff so that transient API-server errors (5xx, timeouts, service-unavailable)
+// at operator startup do not crash the operator process.
+// Permanent errors (Forbidden, Unauthorized) are returned immediately without retrying.
+func SynchronizeTopologyWithRetry(ctx context.Context, cl client.Client, logger logr.Logger, backends map[string]scheduler.TopologyAwareBackend, backoff wait.Backoff) error {
```
yankay, line 51: HI @AsadShahid04 

Nice PR.
(lower priority / optional): SynchronizeTopologyWithRetry puts startup retry/backoff/classification into the clustertopology package, which is otherwise focused on the sync itself. Would it make sense to keep the retry orchestration in main.go or a small startup helper, ideally with a context-aware backoff so it cancels cleanly on SIGTERM? Happy either way — just a thought.

operator/internal/clustertopology/clustertopology.go
```diff
@@ -19,14 +19,52 @@ package clustertopology
...
+			logger.Info("Retrying topology synchronization after transient error", "attempt", attempt+1)
+		}
+		attempt++
+		return SynchronizeTopology(ctx, cl, logger, backends)
+	})
+}
+
+// isTransientAPIError reports whether err is a transient Kubernetes API server error
+// that is safe to retry. Permanent errors (Forbidden, Unauthorized) return false.
+func isTransientAPIError(err error) bool {
```
yankay, line 67: Nice work on this

One thing on the transient classification: it currently only matches Kubernetes StatusErrors. But several of the recoverable startup cases in #654 — API server rolling restart, network partition, connection timeout — surface from the Go client as transport-level errors (e.g. connection refused, EOF, TLS handshake timeout, i/o timeout) rather than StatusErrors. Those go through reasonAndCodeForError as (Unknown, 0), so every apierrors.IsXxx returns false and the operator would still exit. Could we extend the predicate to cover transport-level transient errors (net.Error timeout, ECONNREFUSED, io.EOF) and add a unit test for at least one of them?

AsadShahid04:
Pushed fc5b7be addressing your feedback:

**Transport-level error coverage** — Extended `isTransientAPIError` to also match:
- `net.Error` with `Timeout() == true` (dial/read timeouts during API-server rolling restart)
- `io.EOF` (connection dropped mid-request)
- `syscall.ECONNREFUSED` (API server not yet accepting connections)

These all surface *before* the client wraps the error as a `StatusError`, so every `apierrors.IsXxx()` would return false on them and the operator would exit without retrying. All three error kinds are now caught and retried.

Added `TestIsTransientAPIError` (table-driven, 9 cases) and `TestSynchronizeTopologyWithRetry_TransportErrorRetried` to pin this behavior.

**Retry orchestration location** — I'll leave `SynchronizeTopologyWithRetry` in `clustertopology` for now; happy to move it to a startup helper in a follow-up if that direction is preferred.

operator/internal/clustertopology/clustertopology.go
```diff
@@ -18,15 +18,68 @@ package clustertopology
...
+	Factor:   2.0,
+	Jitter:   0.1,
+	Steps:    6,
+}
+
+// SynchronizeTopologyWithRetry wraps SynchronizeTopology with a bounded exponential
+// backoff so that transient API-server errors (5xx, timeouts, service-unavailable)
+// at operator startup do not crash the operator process.
+// Permanent errors (Forbidden, Unauthorized) are returned immediately without retrying.
+func SynchronizeTopologyWithRetry(ctx context.Context, cl client.Client, logger logr.Logger, backends map[string]scheduler.TopologyAwareBackend, backoff wait.Backoff) error {
```
Ronkahn21, line 51: This Retry functionality will be useful in other places 
I would like to make it more genric rather of hand coded wrapping each function


I would go with something more like 
```go
// Retrier holds retry configuration. Create one per logical operation or share
// a single instance across all calls with the same policy.
type Retrier struct {
	backoff     wait.Backoff
	isRetryable func(error) bool
	onRetry     OnRetryFunc
}
 
// New creates a Retrier. Pass nil for onRetry to skip logging.
func New(backoff wait.Backoff, isRetryable func(error) bool, onRetry OnRetryFunc) *Retrier {
	return &Retrier{backoff: backoff, isRetryable: isRetryable, onRetry: onRetry}
}
 
// Do retries fn using the configured backoff. Use a closure to capture any
// number of arguments — including ctx when it doesn't change between retries:
//
//	err := r.Do(func() error {
//	    return SynchronizeTopology(ctx, cl, logger, backends)
//	})
func (r *Retrier) Do(fn func() error) error {
	attempt := 0
	var lastErr error
	return retry.OnError(r.backoff, r.isRetryable, func() error {
		if attempt > 0 && r.onRetry != nil {
			r.onRetry(attempt, lastErr)
		}
		attempt++
		lastErr = fn()
		return lastErr
	})
}
----------
var topoRetrier = k8sretry.New(
    clustertopology.DefaultSyncRetryBackoff,
    k8sErrors.IsTransientAPIError,
    func(n int, _ error) {
        logger.Info("Retrying topology synchronization", "attempt", n+1)
    },
)

err = topoRetrier.Do(func() error {
    return clustertopology.SynchronizeTopology(ctx, cl, logger, schedRegistry.AllTopologyAware())
})
```

PR #661, opened 2026-06-10
Revert "fix: stabilize PodCliqueSet hashes against map slice reorder"

gflarity:
## Summary
- Remove the canonical PodCliqueSet/PodClique pod-template hashing behavior introduced by `94a0105cf9c26f94539b80d836610b5b1da2b0f9`.
- Keep Grove on the legacy/current single-hash model and do not restore canonical hash candidates, legacy hash migration helpers, or the legacy migration e2e fixture.
- Intentionally preserve the non-canonical controller fixes from `94a0105`: PCSG update completion safeguards, annotation-only PCS reconcile triggers, hash/status watch predicates, and stricter update accounting.

## Controller fixes preserved
- PCSG RollingRecreate now requeues after patching `UpdateEndedAt` so status convergence runs in a follow-up reconcile.
- PCSG generation-hash advancement waits for expected child PodCliques to exist, not be terminating, and have matching current label/status hashes.
- PCS and PCSG update accounting now requires child label hash, `CurrentPodTemplateHash`, `CurrentPodCliqueSetGenerationHash`, `ReadyReplicas`, and `UpdatedReplicas` convergence before counting a child as updated.
- PCS primary watch predicates allow spec generation changes and `grove.io/reconcile-trigger` annotation changes, while ignoring unrelated metadata-only updates.
- PCS/PCLQ/PCSG watch predicates compare `*string` status hashes by value and enqueue on update-progress, updated-replica, and hash-status changes that affect convergence.

## Testing
- `GOCACHE=/tmp/grove-go-cache GOMODCACHE=/tmp/grove-gomod-cache go test ./internal/controller/podclique/... ./internal/controller/podcliqueset/... ./internal/controller/podcliquescalinggroup/...`
- `GOCACHE=/tmp/grove-go-cache GOMODCACHE=/tmp/grove-gomod-cache make test-unit`

## Additional validation
- Ran a one-off k3d rollout matrix on Kubernetes/k3s `v1.31.5+k3s1`, using both `kubectl apply --server-side --force-conflicts` and `kubectl replace`.
- Deployment order-only pod-template changes replaced pods for 7 practical persisted fields under both write paths.
- LeaderWorkerSet `v0.8.0` order-only pod-template changes replaced pods for every practical accepted field under both write paths, in both the default worker-template shape and explicit leader-template shape.
- Result: native controllers already treat persisted order-only template changes as rollout-significant, so canonical Grove hashing is not needed for parity with upstream behavior.

Fixes #662
Closes #689
[...]

PR #671, opened 2026-06-15
podclique/status: converge updated replicas

### LINKED ISSUE #672: PodClique updatedReplicas can latch at zero after pod replacement
## Problem
A PodClique can get stuck with `status.updatedReplicas=0` even after its replacement pod is Ready and labeled with the desired `grove.io/pod-template-hash`.

Observed live state:
- A DynamoGraphDeployment owned a PodCliqueSet with a frontend PodClique
- After changing the frontend pod template and manually deleting the old evicted pod, Grove created a new frontend pod that was `1/1 Ready`
- PodClique metadata label `grove.io/pod-template-hash` was the new desired hash
- Running frontend pod had the same new desired hash and was `Ready`
- PodClique status had `replicas=1`, `readyReplicas=1`, `scheduledReplicas=1`, but `updatedReplicas=0`
- `status.currentPodTemplateHash` still held the old hash
- `status.updateProgress` was absent
- PodCliqueSet reported `availableReplicas=1` but `updatedReplicas=0`
- DynamoGraphDeployment stayed `Ready=False` with a message equivalent to `podclique/<name>: desired=1, updated=0`

## Consumer impact
Consumers such as DynamoGraphDeployment use Grove status to decide whether their workload is updated and ready. In this state the actual replacement pod is Ready and on the desired template, but Grove reports the PodClique and PodCliqueSet as not updated. The DynamoGraphDeployment therefore remains `Ready=False` indefinitely.

## Deadlock
The status reconciler first tried to promote `currentPodTemplateHash`, but refused because `updatedReplicas != replicas`. It then computed `updatedReplicas` from the stale status hash, not the desired hash already present on the converged PodClique metadata label, so the Ready replacement pod was counted as outdated. This left `updatedReplicas=0`, which continued to block hash promotion on every reconcile.

## Expected
If the PodClique metadata label and Ready pod labels already match the current desired template hash, status reconciliation should count the pod as updated and allow `currentPodTemplateHash` / generation hash convergence so parent resources and consumers see the workload as updated.

sttts:
Fixes #672

## Summary
- Count PodClique updated replicas against the current desired pod-template hash when the PodClique metadata label has already converged
- Reorder PodClique status reconciliation so fresh replica/update counts are available before promoting current hash fields
- Add a regression for the live latch: Ready pod and PCLQ label on the desired hash, stale status.currentPodTemplateHash, no UpdateProgress

## Why
We saw a status deadlock in a live DynamoGraphDeployment consumer flow after a template change and manual pod replacement. The PodClique object and the new Ready pod both had the desired `grove.io/pod-template-hash`, but `status.currentPodTemplateHash` still held the old hash and `status.updateProgress` was absent.

The old status flow first tried to promote `currentPodTemplateHash`, but refused because `updatedReplicas != replicas`. It then computed `updatedReplicas` using the stale status hash instead of the desired hash from the converged PodClique label, so the Ready replacement pod was counted as outdated. That left `updatedReplicas=0`, which kept blocking hash promotion on every reconcile.

As a result, the PodCliqueSet reported `availableReplicas=1` but `updatedReplicas=0`, and the owning DynamoGraphDeployment stayed `Ready=False` with the user-visible reason that its frontend PodClique had `desired=1, updated=0`, even though the live pod was Ready and on the desired template.

## Tests
- go test ./internal/controller/podclique -run TestReconcileStatusConvergesWhenReadyPodMatchesDesiredHash -count=1
- go test ./internal/controller/podclique -run TestMutateUpdatedReplica -count=1
- go test ./internal/controller/podclique -run TestMutateCurrentHashes -count=1
- go test ./internal/controller/podclique -count=1
- go test ./internal/controller/podcliqueset -count=1
- go test ./internal/controller/podcliquescalinggroup -count=1
- git diff --check

### DIFF
--- a/operator/internal/controller/podclique/reconcilestatus.go
+++ b/operator/internal/controller/podclique/reconcilestatus.go
@@ -62,14 +62,14 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 
 	podCategories := k8sutils.CategorizePodsByConditionType(logger, existingPods)
 
+	// mutate PodClique Status Replicas, ReadyReplicas, ScheduleGatedReplicas and UpdatedReplicas.
+	mutateReplicas(pclq, podCategories, len(existingPods))
+	mutateUpdatedReplica(pclq, existingPods)
 	// mutate PodClique.Status.CurrentPodTemplateHash and PodClique.Status.CurrentPodCliqueSetGenerationHash
 	if err = mutateCurrentHashes(logger, pcs, pclq); err != nil {
 		logger.Error(err, "failed to compute PodClique current hashes")
 		return ctrlcommon.ReconcileWithErrors("failed to compute PodClique current hashes", err)
 	}
-	// mutate PodClique Status Replicas, ReadyReplicas, ScheduleGatedReplicas and UpdatedReplicas.
-	mutateReplicas(pclq, podCategories, len(existingPods))
-	mutateUpdatedReplica(pclq, existingPods)
 
 	// mutate the conditions only if the PodClique has been successfully reconciled at least once.
 	// This prevents prematurely setting incorrect conditions.
@@ -117,7 +117,7 @@ func mutateCurrentHashes(logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet
 		if err != nil {
 			return err
 		}
-		if pclq.Status.CurrentPodTemplateHash == nil || *pclq.Status.CurrentPodTemplateHash == expectedPodTemplateHash {
+		if isPodCliqueTemplateHashCurrent(pclq, expectedPodTemplateHash) {
 			pclq.Status.CurrentPodTemplateHash = ptr.To(expectedPodTemplateHash)
 			pclq.Status.CurrentPodCliqueSetGenerationHash = pcs.Status.CurrentGenerationHash
 		}
@@ -127,6 +127,11 @@ func mutateCurrentHashes(logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet
 		pclq.Status.CurrentPodCliqueSetGenerationHash = ptr.To(pclq.Status.UpdateProgress.PodCliqueSetGenerationHash)
 	}
 	return nil
+}
+
+func isPodCliqueTemplateHashCurrent(pclq *grovecorev1alpha1.PodClique, expectedPodTemplateHash string) bool {
+	labelPodTemplateHash, ok := pclq.Labels[apicommon.LabelPodTemplateHash]
+	return ok && labelPodTemplateHash == expectedPodTemplateHash
 }
 
 // mutateReplicas updates the PodClique status with current replica counts based on pod categorization
@@ -146,6 +151,11 @@ func mutateUpdatedReplica(pclq *grovecorev1alpha1.PodClique, existingPods []*cor
 	// This covers both the active update phase and the window after completion before CurrentPodTemplateHash is synced.
 	if pclq.Status.UpdateProgress != nil {
 		expectedPodTemplateHash = pclq.Status.UpdateProgress.PodTemplateHash
+	} else if labelPodTemplateHash := pclq.Labels[apicommon.LabelPodTemplateHash]; labelPodTemplateHash != "" {
+		// The PodClique label is the desired pod template hash propagated by the
+		// owner sync. Prefer it over stale current-status bookkeeping so status can
+		// recover after a replacement pod already converged to the desired hash.
+		expectedPodTemplateHash = labelPodTemplateHash
 	} else if pclq.Status.CurrentPodTemplateHash != nil {
 		// Steady state: no rolling update tracking exists.
 		// Use the stable current hash for pods that have been reconciled.
--- a/operator/internal/controller/podclique/reconcilespec.go
+++ b/operator/internal/controller/podclique/reconcilespec.go
@@ -86,7 +86,7 @@ func (r *Reconciler) processUpdate(ctx context.Context, logger logr.Logger, pclq
 		return ctrlcommon.ContinueReconcile()
 	}
 
-	if pcsHasNoActiveRollingUpdate(pcs) {
+	if pcs.Status.CurrentGenerationHash == nil {
 		return ctrlcommon.ContinueReconcile()
 	}
 	shouldEvaluatePCLQForUpdates, err := shouldCheckPendingUpdatesForPCLQ(logger, pcs, pclq)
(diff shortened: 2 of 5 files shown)


yankay:
Thanks for the fix. Verified on a real kind cluster (main vs. this PR): injected the #672 deadlock state onto a Ready PodClique on the new hash, then reconciled. On main, currentPodTemplateHash stays latched on the stale value;

Ronkahn21:
Thanks @sttts for the fix, The main issue you found is we used the status as input for another status field, the fix seems good to me but I am less familiar with the rolling update logic and there is trade off of moving the mutate order. 

better @gflarity @unmarshall review it

gflarity:
@sttts after can you please provide more details around this: 

> We saw a status deadlock in a live DynamoGraphDeployment consumer flow after a template change and manual pod replacement.

Seems like there's a very small window after template change, and the "manual pod" replacement occurring before the RollingUpdate starts. That seems to be when this happened, which is a bit surprising. Any ideas how this happened during that window?

The code change itself looks good, I gave it a first pass. I'll circle back tomorrow and finish the review. The only feedback I have currently is that would be nice to add some concise documentation around these changes to understand them in the code later.

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -141,11 +141,7 @@ func mutateCurrentHashes(logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet
 
 func isPodCliqueTemplateHashCurrent(pclq *grovecorev1alpha1.PodClique, expectedPodTemplateHashes componentutils.HashCandidates) bool {
 	labelPodTemplateHash, ok := pclq.Labels[apicommon.LabelPodTemplateHash]
-	if !ok || !expectedPodTemplateHashes.Matches(labelPodTemplateHash) {
-		return false
-	}
-	return pclq.Status.CurrentPodTemplateHash == nil ||
-		expectedPodTemplateHashes.Matches(*pclq.Status.CurrentPodTemplateHash)
+	return ok && expectedPodTemplateHashes.Matches(labelPodTemplateHash)
```
unmarshall: I am not sure if this is the right change. `pclq.Labels[apicommon.LabelPodTemplateHash]` captures the desired state and that will be same as `expectedPodTemplateHashes` so this effective becomes a not so useful check.

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -62,14 +62,14 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
 
 	podCategories := k8sutils.CategorizePodsByConditionType(logger, existingPods)
 
+	// mutate PodClique Status Replicas, ReadyReplicas, ScheduleGatedReplicas and UpdatedReplicas.
```
unmarshall, line 65: I am just wondering if this is the root cause of what you see.
I suspect the issue lies in `reconcileSpec.go -> processUpdate`

```go
	if pcsHasNoActiveRollingUpdate(pcs) {
	}
	shouldEvaluatePCLQForUpdates, err := shouldCheckPendingUpdatesForPCLQ(logger, pcs, pclq)
	if err != nil {
		return ctrlcommon.ReconcileWithErrors("error checking if PodClique should be evaluated for pending updates", err)
	}
	if !shouldEvaluatePCLQForUpdates {
	}
```
If we see `pcsHasNoActiveRollingUpdate` - it seems incorrect to me at first glance. Ideally the code should have been:

```go
// This check prevents triggering of any update if the PCS status itself has never been reconciled.
if pcs.Status.CurrentGenerationHash == nil {
}
 // check if this PCLQ belongs to the replica currently being updated (if an update is in flight)
 shouldEvaluate, err := shouldCheckPendingUpdatesForPCLQ(...)
```
What this change does is that it will then evaluate if an update is required and then sets `pclq.Status.UpdateProgress` properly.

unmarshall (changes requested):

sttts:
Root cause summary from the live Dynamo consumer case:

The live state had already converged at the spec/object level, but not at the PodClique status level.

```yaml
kind: PodClique
metadata:
  name: vllm-agg-0-frontend
  labels:
    grove.io/pod-template-hash: 54b6d458c549c4c7454d
status:
  replicas: 1
  readyReplicas: 1
  scheduledReplicas: 1
  updatedReplicas: 0
  observedGeneration: 2
  currentPodTemplateHash: 58596fb5bd86d4f5589f
  # updateProgress was absent
```

```yaml
kind: Pod
metadata:
  labels:
    grove.io/podclique: vllm-agg-0-frontend
    grove.io/pod-template-hash: 54b6d458c549c4c7454d
status:
  conditions:
  - type: Ready
    status: "True"
```

```yaml
kind: PodCliqueSet
metadata:
  name: vllm-agg
status:
  availableReplicas: 1
  updatedReplicas: 0
```

```yaml
kind: DynamoGraphDeployment
status:
  conditions:
  - type: Ready
    status: "False"
    message: "podclique/vllm-agg-0-frontend: desired=1, updated=0"
```

So the desired hash had already propagated to the PodClique metadata and to the replacement pod, and the pod was Ready. The only stale value left was `status.currentPodTemplateHash`.

The deadlock was:

1. `UpdateProgress` was absent, so `mutateUpdatedReplica` fell back to `status.currentPodTemplateHash`.
2. That stale status hash was `5859...`.
3. The actual Ready pod had the desired hash `54b6...`, so it was not counted as updated.
4. `updatedReplicas` stayed `0`.
5. `mutateCurrentHashes` refused to promote `currentPodTemplateHash` because `updatedReplicas != replicas`.
6. The PodCliqueSet and DynamoGraphDeployment then kept reporting `updated=0`.

I agree `processUpdate` / `pcsHasNoActiveRollingUpdate` may be the reason `UpdateProgress` was not initialized in this live sequence. But the durable failure was that status reconciliation could not recover from that missing progress even after external state had converged.

sttts:
Pushed a follow-up addressing the review feedback:

- `processUpdate` no longer treats "no active PCS rolling update" as a reason to skip PCLQ update evaluation. It only waits when the PCS has not published `status.currentGenerationHash` yet.
- `shouldCheckPendingUpdatesForPCLQ` now distinguishes recovery from an active rolling-update selection window:
  - no PCS `UpdateProgress`, or completed PCS `UpdateProgress`: standalone PCLQs evaluate their own persisted generation/hash state;
  - active PCS `UpdateProgress` with no selected replica yet: PCLQs wait;
  - active PCS `CurrentlyUpdating`: only the matching PCS replica's standalone PCLQs evaluate.
- `updatedReplicas` now prefers the PCLQ's own `grove.io/pod-template-hash` label over stale `status.currentPodTemplateHash` when `UpdateProgress` is absent. The PCS lookup is kept only to expand canonical/legacy hash candidates for hash-format migration.

This keeps the fix split across the two failure modes from the live case: `processUpdate` should initialize missing PCLQ `UpdateProgress`, and status reconciliation should still recover if external state has already converged while `UpdateProgress` is absent.

operator/internal/controller/podclique/reconcilestatus.go
sttts: done

operator/internal/controller/podclique/reconcilestatus.go
sttts: Done. Agreed, this only needs to check that the PCLQ label reached the desired hash. Requiring `status.currentPodTemplateHash` here was the self-blocking stale-status dependency.

gflarity (approved):
LGTM

gflarity (changes requested):
Actually no issues, but we discussed holding off on merging this one until #661 lands.  Discussed review of #661 with @shayasoolin and @Ronkahn21. Will land soon I hope.

sttts:
https://github.com/ai-dynamo/grove/pull/661 has landed.

gflarity (dismissed):

danbar2 (dismissed):

unmarshall (dismissed):

unmarshall (approved):

danbar2 (approved):

Issue #676, opened 2026-06-16
Dynamo scale to zero of worker takes down frontend too


treydock:
### What happened?

Also posted here: https://github.com/ai-dynamo/dynamo/issues/10751

When KEDA scales my DGD's decode and prefill containers to zero, the frontend goes into `SchedulingGated` state and stops running:

```
qwen3-0-frontend-qvf62                     0/1     SchedulingGated   0          11s
```

This is after these scaledobject events:

```
Events:
  Type    Reason                      Age   From           Message
  ----    ------                      ----  ----           -------
  Normal  KEDAScalersStarted          5s    keda-operator  Scaler prometheus is built
  Normal  KEDAScalersStarted          5s    keda-operator  Started scalers watch
  Normal  ScaledObjectReady           5s    keda-operator  ScaledObject is ready for scaling
  Normal  KEDAScaleTargetDeactivated  5s    keda-operator  Deactivated nvidia.com/v1beta1.DynamoGraphDeploymentScalingAdapter dynamo/qwen3-vllmprefillworker from 1 to 0
```

Noticed this:

The issue might be the `podgroup` resource or kai-scheduler:

```
[root@kubecontroller-test ~]# kubectl describe podgroup -n dynamo pg-qwen3-0-9283a429-d640-4116-a3d7-c1f810ea98df
<SNIP>
Events:
  Type    Reason    Age                    From           Message
  ----    ------    ----                   ----           -------
  Normal  NotReady  24m (x14 over 24m)     kai-scheduler  Job is not ready for scheduling. Waiting for 1 pods for SubGroup qwen3-0-vllmprefillworker, currently 0 exist, 0 are gated.
  Normal  NotReady  14m (x133 over 23m)    kai-scheduler  Job is not ready for scheduling. Waiting for 1 pods for SubGroup qwen3-0-vllmprefillworker, currently 0 exist, 0 are gated. Waiting for 1 pods for SubGroup qwen3-0-vllmdecodeworker, currently 0 exist, 0 are gated.
  Normal  StaleJob  8m45s                  kai-scheduler  Job is stale. 2 pods are active, minMember is 3, subGroup qwen3-0-vllmdecodeworker minMember is 1 and 0 pods are active
[...]

PR #686, opened 2026-06-25
GREP-0677: Scale-to-Zero Component Hibernation

### LINKED ISSUE #677: Discuss minAvailable=0 for scale-to-zero workloads

Support `minAvailable: 0` for PodCliques or PodCliqueScalingGroups that are intentionally scaled to `replicas: 0` by an autoscaler.

The intended meaning is that the component is currently not a required gang member, rather than being considered unavailable.

## Why is this needed?

For scale-to-zero serving, an always-on frontend/router may need to stay schedulable while decode/prefill workers scale to zero. If zero-scaled workers still carry `minAvailable: 1`, the scheduler may wait for pods that intentionally do not exist and block the base PodGang.

This also matches the Dynamo autoscaling model where the frontend/router remains the metrics and request entry point. The Dynamo autoscaling docs describe frontend-exported metrics such as TTFT, ITL, and pending queue depth being used for worker autoscaling, and the mixed autoscaling example keeps the frontend at `minReplicas: 1` while scaling decode with KEDA.

Example:

```text
frontend: replicas=1, minAvailable=1
decode:   replicas=0, minAvailable=0
prefill:  replicas=0, minAvailable=0
```

Related:
- https://github.com/ai-dynamo/grove/issues/676
- https://github.com/ai-dynamo/dynamo/issues/10753
- https://github.com/ai-dynamo/dynamo/pull/10532
- https://docs.nvidia.com/dynamo/v1.1.1/kubernetes-deployment/deployment-guide/autoscaling

This is meant as a discussion starter. If maintainers think this direction makes sense, I would be happy to help implement it.

yankay:
#### What type of PR is this?

/kind documentation

Adds GREP-0677 to define hibernation for standalone `PodClique` and `PodCliqueScalingGroup` components at `replicas: 0`, without disrupting components that remain active. Idle components should not block gang scheduling, trigger gang termination, or stall rolling updates. Wake-up preserves gang-scheduling guarantees; `minAvailable` stays positive and immutable.

New positive replica targets below `minAvailable` are rejected. Existing below-quorum objects may update unrelated fields without changing `replicas` or `minAvailable`.

Fixes #677
Related to #676

Design proposal only; no implementation changes.

Runtime hibernation targets the derived PodClique or PodCliqueScalingGroup, not PodCliqueSet replicas. PCSG members idle through their group; non-zero scaling remains allowed when `replicas >= minAvailable`.

Detailed wake, backend, and autoscaler contracts are in the GREP.

```release-note
NONE
```

```docs
docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```

### DIFF
--- a/docs/proposals/0677-scale-to-zero-component-hibernation/README.md
+++ b/docs/proposals/0677-scale-to-zero-component-hibernation/README.md
@@ -25,7 +25,7 @@
 
 ## Summary
 
-Grove should treat standalone `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP leaves existing `minAvailable` semantics unchanged, makes gang logic tolerant of zero-replica members, and rejects any positive replica count below `minAvailable`.
+Grove should treat standalone `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP makes gang logic tolerant of zero-replica members, rejects any positive replica count below `minAvailable`, and treats a PCSG wake as ordinary scale-out with one scaled `PodGang` per replica.
 
 ## Motivation
 
@@ -43,11 +43,11 @@ External replica writers should target either `0` or a value at least `minAvaila
 
 ### Non-Goals
 
-- Change `minAvailable` semantics, including allowing `minAvailable: 0` or making it mutable.
+- Allow `minAvailable: 0` or make it mutable.
 
 ## Proposal
 
-When a standalone `PodClique` or `PodCliqueScalingGroup` has `replicas: 0`, Grove should not require it as a gang member. When replicas become positive again, its existing `minAvailable` should apply normally.
+When a standalone `PodClique` or `PodCliqueScalingGroup` has `replicas: 0`, Grove should not require it as a gang member. When replicas become positive again, `minAvailable` continues to define the component's availability and gang-termination threshold. For a waking `PodCliqueScalingGroup`, it does not require the first `minAvailable` replicas to be scheduled atomically with each other.
 
 An idle component should leave the `PodGang` rather than stay in it with a zero threshold: no `PodGroup` in `PodGang.spec.podGroups`, and no scaled `PodGang` for an idle `PodCliqueScalingGroup`. Shrinking a gang must not disturb the members still running.
 
@@ -71,7 +71,7 @@ Grove rejects positive replica counts below `minAvailable`:
 | --- | --- |
 | `0` | Intentional idle state. The component contributes no required gang members and does not breach `minAvailable`. |
 | `0 < replicas < minAvailable` | Rejected for create and update requests, including the `/scale` subresource. |
-| `replicas >= minAvailable` | Persisted as requested. Normal Grove gang behavior applies. |
+| `replicas >= minAvailable` | Persisted as requested. Waking components follow the [gang behavior](#gang-behavior) below. |
 
 For independent scale targets, the persisted invariant is:
 
@@ -99,7 +99,32 @@ For `PodClique`, omitted `replicas` defaults to `1`, while explicit `0` is prese
 
 While a component is idle, it contributes no `PodGroup` and zero observed scheduled, available, and updated replicas. It does not set `MinAvailableBreached` to `True`, counts as updated for rolling-update completion, and does not trigger gang termination.
 
-An idle component is omitted from `PodGangMap` entry membership. If all components are idle, Grove retains an empty current-generation anchor without materializing a `PodGang`. On wake, Grove populates that anchor if no `PodGang` is active; otherwise, it creates a new entry for the waking component without expanding an existing materialized `PodGang`.
+Idle components leave `PodGangMap` membership. Empty current-generation anchor and scale-out entries remain logical slots without materialized `PodGangs`.
+
+On `0 -> N`:
+
+1. **PCSG:** all replica indices enter the current-generation `ScaleOut` entry, creating one SPG per replica. Gang and PCSG topology constraints apply per replica, not across `minAvailable` replicas.
+2. **Standalone PodClique:** restore membership and the full `PodGroup` in the current-generation anchor (`A0`, or the highest `AnchorIndex` if several exist), materializing it if needed. No additional anchor is created.
+
+```text
+PCSG (minAvailable: 2):
+  Initial:  A0 [replica 0 + replica 1] + SPG [replica 2]
+  Wake:     A0, if active
+             +--> SPG [replica 0]
+             +--> SPG [replica 1]
+             +--> SPG [replica 2]
+
+Standalone Decode:
+  Before idle: A0 [Router + Decode]
+  Idle:        A0 [Router]
+  Wake:        A0 [Router + Decode]
+```
+
+An empty `ScaleOut` entry gets a fresh epoch, depending on a materialized anchor only if one exists. Non-empty entries retain their epoch and dependencies.
+
+Restoring a standalone PodClique's membership does not require it to be scheduled together with running anchor members or waking PCSGs. An already scheduled anchor keeps its epoch and `LastScheduled`, so dependent SPGs do not wait for this PodClique to wake. The restored `PodGroup` uses the PodClique's `minAvailable` as `MinReplicas`; backend tests must verify that this is enforced without disrupting running pods.
+
+With a running router and independent P/D PCSGs (`minAvailable: 1`), each `0 -> 1` wake creates one SPG, without another anchor or joint P/D scheduling.
 
 ### Autoscaler Integration
 
@@ -208,7 +233,8 @@ Prototype coverage should show:
 - rejected updates leave `spec.replicas` unchanged;
 - a KEDA-like integration transitions from `0` directly to `minAvailable` or above without writing a positive below-quorum value;
 - an HPA-like writer receives a validation error for a below-quorum recommendation and leaves `spec.replicas` unchanged, with repeated retry behavior documented;
-- scaling to `minAvailable` or above re-enters normal gang behavior.
+- repeated sleep/wake cycles, including all-idle and post-coherent-update cases, follow the gang behavior above without adding anchors on standalone wake;
+- scheduler-backed tests verify the restored standalone `MinReplicas` floor with `minAvailable > 1`, undisturbed running members, and SPG dependency behavior during concurrent wakes.
 
 ### Graduation Criteria
 
--- a/docs/proposals/0677-scale-to-zero-component-hibernation/README.md
+++ b/docs/proposals/0677-scale-to-zero-component-hibernation/README.md
@@ -1,4 +1,4 @@
-# GREP-0677: Zero-Replica Gang Membership
+# GREP-0677: Scale-to-Zero Component Hibernation
 
 <!-- toc -->
 - [Summary](#summary)
--- a/docs/proposals/0677-zero-replica-gang-membership/README.md
+++ b/docs/proposals/0677-zero-replica-gang-membership/README.md
@@ -18,7 +18,7 @@
 
 ## Summary
 
-Grove should treat `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP is only a direction-setting draft: it keeps `minAvailable` non-zero and immutable, makes future gang logic tolerant of zero-replica members, and intentionally leaves detailed implementation design to follow-up work.
+Grove should treat `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP is only a direction-setting draft: it keeps `minAvailable` non-zero and immutable, makes future gang logic tolerant of zero-replica members, and keeps positive replicas below `minAvailable` invalid.
 
 ## Motivation
 
@@ -29,6 +29,7 @@ Using `minAvailable: 0` would overload a field that already affects gang members
 ### Goals
 
 - Confirm `replicas: 0` as an intentional idle state for `PodClique` and `PodCliqueScalingGroup`.
+- Confirm that positive replicas below `minAvailable` remain invalid.
 - Keep `minAvailable` non-zero and immutable.
 - Give follow-up work one agreed direction.
 - Use a prototype PR only to show feasibility.
@@ -37,6 +38,7 @@ Using `minAvailable: 0` would overload a field that already affects gang members
 
 - Support `minAvailable: 0`.
 - Make `minAvailable` mutable.
+- Automatically inflate positive replicas below `minAvailable`.
 - Define the complete implementation in this draft.
 - Treat the prototype as production complete.
 - Define complete behavior differences between `PodClique` and `PodCliqueScalingGroup`.
@@ -45,13 +47,23 @@ Using `minAvailable: 0` would overload a field that already affects gang members
 
 When a `PodClique` or `PodCliqueScalingGroup` has `replicas: 0`, Grove should not require it as a gang member. When replicas become positive again, its existing `minAvailable` should apply normally.
 
-This GREP only special-cases `replicas: 0`; positive replicas below `minAvailable` remain invalid.
+This GREP only special-cases `replicas: 0`; positive replicas below `minAvailable` remain invalid. The intended contract is:
+
+| Desired replicas | Behavior |
+| --- | --- |
+| `replicas: 0` | Intentional idle state. The component contributes no required gang members and does not breach `minAvailable`. |
+| `0 < replicas < minAvailable` | Invalid desired state. Grove should reject it rather than gang-scheduling a partial component. |
+| `replicas >= minAvailable` | Normal Grove behavior. The component participates in gang scheduling and gang termination using its configured `minAvailable`. |
+
+Grove should not derive `max(minAvailable, replicas)` or overwrite the autoscaler's desired value.
 
 This draft does not define every controller or backend branch. Those details should be added only after reviewers agree on the direction.
 
 ### Limitations/Risks & Mitigations
 
 The main risk is confusing "idle" with "unavailable." The proposed direction ties idle state to desired replicas, not observed status.
+
+External autoscalers must follow the same contract: scale these components either to `0` or to `>= minAvailable`.
 
 The prototype must be described only as a feasibility artifact, not as the final design.
 
@@ -151,7 +163,8 @@ Prototype coverage should show:
 
 - zero-replica components do not block the base gang;
 - `minAvailable` remains non-zero;
-- scaling back above zero can re-enter normal gang behavior.
+- positive replicas below `minAvailable` are rejected;
+- scaling back to `minAvailable` or above can re-enter normal gang behavior.
 
 ### Graduation Criteria
 
@@ -163,6 +176,7 @@ Prototype coverage should show:
 
 - Allow `minAvailable: 0`: explicit, but overloads an existing availability contract.
 - Make `minAvailable` mutable: follows scale state, but changes the immutability contract.
+- Treat desired replicas as `max(minAvailable, replicas)`: avoids rejecting partial wake-ups, but hides the autoscaler's requested state.
 - Work around this outside Grove: avoids Grove changes, but leaves no consistent Grove semantics.
 
 ## Appendix
--- a/docs/proposals/0677-zero-replica-gang-membership/README.md
+++ b/docs/proposals/0677-zero-replica-gang-membership/README.md
@@ -1,0 +1,176 @@
+# GREP-0677: Zero-Replica Gang Membership
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Limitations/Risks &amp; Mitigations](#limitationsrisks--mitigations)
+- [Design Details](#design-details)
+  - [Example](#example)
+  - [Monitoring](#monitoring)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Alternatives](#alternatives)
+- [Appendix](#appendix)
+<!-- /toc -->
+
+## Summary
+
+Grove should treat `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP is only a direction-setting draft: it keeps `minAvailable` non-zero and immutable, makes future gang logic tolerant of zero-replica members, and intentionally leaves detailed implementation design to follow-up work.
+
+## Motivation
+
+Scale-to-zero serving commonly keeps a router running while workers scale to zero. In that state, worker pods are intentionally absent, not unhealthy.
+
+Using `minAvailable: 0` would overload a field that already affects gang membership, termination, rolling updates, and startup ordering. `replicas` is the mutable scale target, so `replicas: 0` is the cleaner signal.
+
+### Goals
+
+- Confirm `replicas: 0` as an intentional idle state for `PodClique` and `PodCliqueScalingGroup`.
+- Keep `minAvailable` non-zero and immutable.
+- Give follow-up work one agreed direction.
+- Use a prototype PR only to show feasibility.
+
+### Non-Goals
+
+- Support `minAvailable: 0`.
+- Make `minAvailable` mutable.
+- Define the complete implementation in this draft.
+- Treat the prototype as production complete.
+- Define complete behavior differences between `PodClique` and `PodCliqueScalingGroup`.
+
+## Proposal
+
+When a `PodClique` or `PodCliqueScalingGroup` has `replicas: 0`, Grove should not require it as a gang member. When replicas become positive again, its existing `minAvailable` should apply normally.
+
+This GREP only special-cases `replicas: 0`; positive replicas below `minAvailable` remain invalid.
+
+This draft does not define every controller or backend branch. Those details should be added only after reviewers agree on the direction.
+
+### Limitations/Risks & Mitigations
+
+The main risk is confusing "idle" with "unavailable." The proposed direction ties idle state to desired replicas, not observed status.
+
+The prototype must be described only as a feasibility artifact, not as the final design.
+
+## Design Details
+
+No new API fields are proposed.
+
+Implementation should derive effective gang membership from desired replicas. A zero-replica component contributes zero required members while idle.
+
+### Example
+
+Only the fields relevant to gang membership are shown.
+
+Initial serving shape: the router, standalone decode worker, and prefill worker group are running.
+
+```yaml
+apiVersion: grove.io/v1alpha1
+kind: PodCliqueSet
+metadata:
+  name: scale-to-zero-deepseek
+spec:
+  replicas: 1
+  template:
+    cliques:
+    - name: router
+      spec:
+        roleName: router
+        replicas: 1
+        minAvailable: 1
+        podSpec:
+          containers: [{name: router, image: nginx:latest}]
+    - name: pleader
+      spec:
+        roleName: pleader
+        replicas: 1
+        minAvailable: 1
+        podSpec:
+          containers: [{name: pleader, image: nginx:latest}]
+    - name: pworker
+      spec:
+        roleName: pworker
+        replicas: 1
+        minAvailable: 1
+        podSpec:
+          containers: [{name: pworker, image: nginx:latest}]
+    - name: decode
+      spec:
+        roleName: decode
+        replicas: 1
+        minAvailable: 1
+        podSpec:
+          containers: [{name: decode, image: nginx:latest}]
+    podCliqueScalingGroups:
+    - name: prefill
+      cliqueNames: [pleader, pworker]
+      replicas: 1
+      minAvailable: 1
+```
+
+After scale-to-zero: the components remain defined, and the autoscaler only changes the worker `replicas`.
+
+```yaml
+apiVersion: grove.io/v1alpha1
+kind: PodCliqueSet
+metadata:
+  name: scale-to-zero-deepseek
+spec:
+  template:
+    cliques:
+    - name: decode
+      spec:
+        replicas: 0
+        minAvailable: 1
+    podCliqueScalingGroups:
+    - name: prefill
+      cliqueNames: [pleader, pworker]
+      replicas: 0
+      minAvailable: 1
+```
+
+Under this GREP, the second shape means the standalone decode `PodClique` and prefill `PodCliqueScalingGroup` are idle and do not block the router. If they scale above zero, their existing `minAvailable` applies again.
+
+With current Grove validation, the full scale-to-zero shape is rejected because `PodCliqueScalingGroup.replicas: 0` is not yet allowed:
+
+```text
+spec.template.podCliqueScalingGroups[0].replicas: Invalid value: 0: must be greater than 0
+spec.template.podCliqueScalingGroups[0].minAvailable: Invalid value: 1: minAvailable must not be greater than replicas
+```
+
+### Monitoring
+
+No new metrics or status fields are required by this initial draft.
+
+### Test Plan
+
+Prototype coverage should show:
+
+- zero-replica components do not block the base gang;
+- `minAvailable` remains non-zero;
+- scaling back above zero can re-enter normal gang behavior.
+
+### Graduation Criteria
+
+- Alpha: direction accepted and initial implementation exists.
+- Beta: behavior documented and tested.
+- GA: semantics are stable and validated with real scale-to-zero workloads.
+
+## Alternatives
+
+- Allow `minAvailable: 0`: explicit, but overloads an existing availability contract.
+- Make `minAvailable` mutable: follows scale state, but changes the immutability contract.
+- Work around this outside Grove: avoids Grove changes, but leaves no consistent Grove semantics.
+
+## Appendix
+
+- Tracking issue: [ai-dynamo/grove#677](https://github.com/ai-dynamo/grove/issues/677)
+- Related bug: [ai-dynamo/grove#676](https://github.com/ai-dynamo/grove/issues/676)
+- Feasibility prototype: [yankay/grove#2](https://github.com/yankay/grove/pull/2)
+- Local YAML shape: [multinode-disaggregated-with-frontend.yaml](../../../operator/samples/user-guide/02_pod-and-resource-naming-conventions/multinode-disaggregated-with-frontend.yaml)
+- Related Dynamo issue: [ai-dynamo/dynamo#10753](https://github.com/ai-dynamo/dynamo/issues/10753)
+- Related Dynamo PR: [ai-dynamo/dynamo#10532](https://github.com/ai-dynamo/dynamo/pull/10532)
+- Dynamo autoscaling guide: [Dynamo autoscaling](https://docs.nvidia.com/dynamo/v1.1.1/kubernetes-deployment/deployment-guide/autoscaling)
--- a/docs/proposals/0677-zero-replica-gang-membership/README.md
+++ b/docs/proposals/0677-zero-replica-gang-membership/README.md
@@ -18,7 +18,7 @@
 
 ## Summary
 
-Grove should treat `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP is only a direction-setting draft: it keeps `minAvailable` non-zero and immutable, makes future gang logic tolerant of zero-replica members, and keeps positive replicas below `minAvailable` invalid.
+Grove should treat `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP is only a direction-setting draft: it keeps `minAvailable` non-zero and immutable, makes future gang logic tolerant of zero-replica members, keeps the default behavior for positive replicas below `minAvailable`, and defines an opt-in clamp policy for generic autoscalers.
 
 ## Motivation
 
@@ -26,36 +26,36 @@ Scale-to-zero serving commonly keeps a router running while workers scale to zer
 
 Using `minAvailable: 0` would overload a field that already affects gang membership, termination, rolling updates, and startup ordering. `replicas` is the mutable scale target, so `replicas: 0` is the cleaner signal.
 
+External autoscalers should be treated as producers of desired replicas, not as components that understand Grove's per-component `minAvailable`. If Grove needs gang-safe membership above the autoscaler-written value, Grove should derive an internal effective value without writing back to `spec.replicas`.
+
 ### Goals
 
-- Confirm `replicas: 0` as an intentional idle state for `PodClique` and `PodCliqueScalingGroup`.
-- Confirm that positive replicas below `minAvailable` remain invalid.
-- Keep `minAvailable` non-zero and immutable.
-- Give follow-up work one agreed direction.
-- Use a prototype PR only to show feasibility.
+- Treat `replicas: 0` as an intentional idle state without changing `minAvailable`.
+- Keep the default validation behavior for positive replicas below `minAvailable`.
+- Define an opt-in `Clamp` policy for generic autoscalers that wake below `minAvailable`.
+- Preserve autoscaler-written `spec.replicas` as desired scale.
 
 ### Non-Goals
 
-- Support `minAvailable: 0`.
-- Make `minAvailable` mutable.
-- Automatically inflate positive replicas below `minAvailable`.
-- Define the complete implementation in this draft.
-- Treat the prototype as production complete.
-- Define complete behavior differences between `PodClique` and `PodCliqueScalingGroup`.
+- Change `minAvailable` semantics, including allowing `minAvailable: 0` or making it mutable.
+- Require generic autoscalers to discover Grove component quorum.
+- Define the complete production implementation in this draft.
 
 ## Proposal
 
 When a `PodClique` or `PodCliqueScalingGroup` has `replicas: 0`, Grove should not require it as a gang member. When replicas become positive again, its existing `minAvailable` should apply normally.
 
-This GREP only special-cases `replicas: 0`; positive replicas below `minAvailable` remain invalid. The intended contract is:
+For `0 < replicas < minAvailable`, Grove should define an explicit below-quorum policy:
 
-| Desired replicas | Behavior |
-| --- | --- |
-| `replicas: 0` | Intentional idle state. The component contributes no required gang members and does not breach `minAvailable`. |
-| `0 < replicas < minAvailable` | Invalid desired state. Grove should reject it rather than gang-scheduling a partial component. |
-| `replicas >= minAvailable` | Normal Grove behavior. The component participates in gang scheduling and gang termination using its configured `minAvailable`. |
+| Desired replicas | Default policy `Block` | Opt-in policy `Clamp` |
+| --- | --- | --- |
+| `replicas: 0` | Intentional idle state. The component contributes no required gang members and does not breach `minAvailable`. | Same as `Block`. |
+| `0 < replicas < minAvailable` | Invalid, preserving the existing positive-replica validation behavior. Grove does not create partial gang members. | Grove preserves `spec.replicas` as desired scale, computes `effectiveReplicas = minAvailable`, and exposes the desired/effective split in status or events. |
+| `replicas >= minAvailable` | Normal Grove behavior. The component participates in gang scheduling and gang termination using its configured `minAvailable`. | Same as `Block`; desired and effective replicas are equal. |
 
-Grove should not derive `max(minAvailable, replicas)` or overwrite the autoscaler's desired value.
+Under the default `Block` policy, this GREP only changes the zero-replica case; positive replicas below `minAvailable` keep the existing validation behavior.
+
+Grove should not implement `Clamp` by writing the clamped value back to `spec.replicas`. HPA or KEDA may repeatedly recompute a desired value such as `1`; writing `minAvailable` back into the scale target can create a control-plane write fight. If `Clamp` is used, the autoscaler-written value remains desired scale and Grove uses a separate effective value internally.
 
 This draft does not define every controller or backend branch. Those details should be added only after reviewers agree on the direction.
 
@@ -63,15 +63,34 @@ This draft does not define every controller or backend branch. Those details sho
 
 The main risk is confusing "idle" with "unavailable." The proposed direction ties idle state to desired replicas, not observed status.
 
-External autoscalers must follow the same contract: scale these components either to `0` or to `>= minAvailable`.
+The default `Block` policy keeps the existing positive-replica validation behavior. Generic HPA/KEDA users can still fail to wake from zero to one if they do not configure an autoscaler floor. Users who prefer automatic wake-up can opt into `Clamp`.
+
+The `Clamp` policy makes generic HPA/KEDA scale-from-zero friendlier, but `spec.replicas: 1` may intentionally run `minAvailable` members. This desired/effective split must be visible in status or events.
 
 The prototype must be described only as a feasibility artifact, not as the final design.
 
 ## Design Details
 
-No new API fields are proposed.
+Implementation should derive gang membership from desired replicas and the below-quorum policy. A zero-replica component contributes zero required members while idle.
 
-Implementation should derive effective gang membership from desired replicas. A zero-replica component contributes zero required members while idle.
+This draft proposes `belowMinAvailablePolicy` as an optional enum field next to `replicas` and `minAvailable` on `PodClique`, `PodCliqueScalingGroup`, and their `PodCliqueSet` template configs:
+
+```go
+// +kubebuilder:validation:Enum=Block;Clamp
+// +kubebuilder:default=Block
+BelowMinAvailablePolicy *BelowMinAvailablePolicy `json:"belowMinAvailablePolicy,omitempty"`
+```
+
+The default value is `Block`. `Clamp` is opt-in for users who want Grove to translate a positive desired replica count below `minAvailable` into gang-safe effective replicas.
+
+For `Clamp`, the effective count is:
+
+```text
+effectiveReplicas = 0 if desiredReplicas == 0
+effectiveReplicas = max(desiredReplicas, minAvailable) otherwise
+```
+
+For `Block`, positive desired replicas below `minAvailable` remain invalid, preserving the existing behavior for positive replicas.
 
 ### Example
 
@@ -137,14 +156,26 @@ spec:
       spec:
         replicas: 0
         minAvailable: 1
+        belowMinAvailablePolicy: Clamp
     podCliqueScalingGroups:
     - name: prefill
       cliqueNames: [pleader, pworker]
       replicas: 0
       minAvailable: 1
+      belowMinAvailablePolicy: Clamp
 ```
 
 Under this GREP, the second shape means the standalone decode `PodClique` and prefill `PodCliqueScalingGroup` are idle and do not block the router. If they scale above zero, their existing `minAvailable` applies again.
+
+A canonical below-quorum wake-up looks like:
+
+```text
+minAvailable = 2
+replicas = 0
+external autoscaler writes replicas = 1
+```
+
+With `Block`, this remains invalid because positive replicas are below `minAvailable`. With `Clamp`, Grove preserves `replicas: 1`, internally uses `effectiveReplicas: 2`, and exposes that split.
 
 With current Grove validation, the full scale-to-zero shape is rejected because `PodCliqueScalingGroup.replicas: 0` is not yet allowed:
 
@@ -155,7 +186,7 @@ spec.template.podCliqueScalingGroups[0].minAvailable: Invalid value: 1: minAvail
 
 ### Monitoring
 
-No new metrics or status fields are required by this initial draft.
+For `Clamp`, status or events should show desired replicas and effective replicas so users can understand why the running member count is greater than `spec.replicas`.
 
 ### Test Plan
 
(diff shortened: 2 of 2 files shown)


docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+No new metrics or status fields are required by this initial draft.
+
+### Test Plan
+
+Prototype coverage should show:
+
+- zero-replica components do not block the base gang;
+- `minAvailable` remains non-zero;
+- scaling back above zero can re-enter normal gang behavior.
```
shayasoolin: This part of the design is not clear to me. Consider the following case: `minAvailable=2` and `replicas` scale to zero and then back from zero to 1. The external autoscaler took replicas to 1, as it's not supposed to be aware of the PCS's `minAvailable` value.
What is the expected behavior wrt gang scheduling/termination? Is Grove supposed to consider the desired number of replicas as 2 instead of 1 (`max(minAvailable, replicas)`) and take care of gang-scheduling those 2 replicas?
If it doesn't do it, the workload will gang-terminate sooner or later. If it does, should it overwrite `replicas` with the actual value being used (2)? Or is the autoscaler somehow forced to set `replicas` to either 0 or `>= minAvailable`?

yankay: HI @shayasoolin 
Thanks, I clarified this in the GREP at https://github.com/ai-dynamo/grove/pull/686/changes#diff-2931583a0e8e71574e767a4ddb4a4e3dbf171af1851c57a2963592c1940d6facR52 . The intended contract is that replicas: 0 is the only valid value below minAvailable; once replicas is positive, it must be >= minAvailable.
So for minAvailable=2 and 0 -> 1, 1 should be rejected rather than interpreted as 2 or written back as 2. Autoscaler integrations should scale to either 0 or >= minAvailable.

Does this direction sound reasonable? If so, I’ll keep refining the GREP and continue pushing the implementation PR along this contract.

shayasoolin: I'm afraid this is not a valid requirement from the autoscalers, which are external to Grove and are not expected to be aware of `minAvailable` of each component they are scaling. AFAIK from my experience with autoscalers, they usually don't have that setting and will naturally try to scale from 0 to 1 and not to a bigger number (unless the desired scale requires it, but talking about the case where the desired scale that is derived from the autoscaling metrics requires 1 replica.)
One exception I know of is Knative's own autoscaler - `KPA`, which has a setting of `activationScale` which is the number of replicas to spin up when scaling from zero. That's not the standard though.
I feel like we'll have to have Grove overwrite the scale from below `minAvailable` to `minAvailable`. Can you try and research this direction?

yankay: Hi @shayasoolin,

Thanks for the feedback. I looked more into KPA, and I agree that `activation-scale` is Knative/KPA-specific, not something generic autoscalers usually provide. That said, it is a useful reference point, and it suggests that Grove may need an explicit parameter to define this behavior.

I see two possible directions:

1. **Automatically clamp below `minAvailable`**

   If `desiredReplicas > 0`, Grove internally uses:

   ```text
   effectiveReplicas = max(desiredReplicas, minAvailable)
   ```

   So an autoscaler can write `replicas: 1`, while Grove runs `minAvailable` members. This is friendlier for HPA/KEDA scale-from-zero, but creates a desired/effective split.

2. **Add an explicit policy and keep the default behavior unchanged**

   ```yaml
   belowMinAvailablePolicy: Clamp or Block(default)
   ```

   By default, `replicas: 0` becomes idle, while `0 < replicas < minAvailable` keeps the existing validation behavior. Only opt-in `Clamp` uses the effective-replica rule above. I tried to capture this direction in the GREP here: https://github.com/ai-dynamo/grove/pull/686/changes#diff-2931583a0e8e71574e767a4ddb4a4e3dbf171af1851c57a2963592c1940d6facR50

Both options can work with HPA/KEDA-style autoscaling, where the autoscaler may naturally scale from 0 to 1. Option 1 is simpler for autoscalers; Option 2 is more explicit and preserves the current behavior by default, but adds a small API surface.

Which direction do you prefer? I would really appreciate your feedback.

danbar2 (commented):
The direction seems fine, but I think the main challenge here is to define the gang-scheduling, gang-termination and the pod gang structore in such case. how it handles updates (MVU specifically) and partial scale operations

docs/proposals/0677-zero-replica-gang-membership/README.md
shayasoolin: Hi @yankay apologize for the delayed response. @Ronkahn21 and I have discussed this today and got to the same conclusion as you actually :)
It basically says that:
1. 0 becomes a valid value for `replicas`. It's the only exception for the rule of `replicas >= minAvailable`.
2. Any value of `replicas` between 0 and `minAvailable` is being overridden to be `minAvailable`.
We said we'll need to check how common autoscalers like HPA and KEDA accept that, as their desired number of replicas is not being used. The concern is that they might try to enforce their desired number again and again.
Hopefully, we can motivate the autoscalers or contribute to them, introducing a concept of "min replicas" that can be > 0 while still allowing scale-to-zero. BTW AFAIK HPA doesn't support scale-to-zero (at least not out-of-the-box), so this point applies more to KEDA and perhaps others.
How would you like to take it from here? Maybe elaborate more on this approach and its implications on the controller, so we can review the GREP again.

cc @sanjaychatterjee @nvrohanv @unmarshall

Ronkahn21:
A few things we'd like to see addressed before moving forward:

1. **Autoscaler compatibility** — we'd like to understand how different autoscalers (HPA, KEDA, etc.) react to the mutation. If an autoscaler sets replicas to a value between 0 and minAvailable and we silently override it, it could cause a tight reconciliation loop.

2. **Mutation scope** — the defaulting/mutation should only apply in the scale-out direction: from 0 to minAvailable. For an invalid scale-in (e.g., a running PCS being scaled to a value between 0 and minAvailable), the webhook should return a validation error rather than mutating, so callers get explicit feedback.

3. **Replica 0 at every level** — we'd want to see replicas: 0 supported consistently across PodClique, PodCliqueScalingGroup, and PodCliqueSet. The GREP should clarify the expected behavior at each level before we move to implementation.

yankay:
> A few things we'd like to see addressed before moving forward:
> 
> 1. **Autoscaler compatibility** — we'd like to understand how different autoscalers (HPA, KEDA, etc.) react to the mutation. If an autoscaler sets replicas to a value between 0 and minAvailable and we silently override it, it could cause a tight reconciliation loop.
> 2. **Mutation scope** — the defaulting/mutation should only apply in the scale-out direction: from 0 to minAvailable. For an invalid scale-in (e.g., a running PCS being scaled to a value between 0 and minAvailable), the webhook should return a validation error rather than mutating, so callers get explicit feedback.
> 3. **Replica 0 at every level** — we'd want to see replicas: 0 supported consistently across PodClique, PodCliqueScalingGroup, and PodCliqueSet. The GREP should clarify the expected behavior at each level before we move to implementation.

HI @Ronkahn21 

Thanks for the clear feedback. I’ve updated the GREP to:

- document HPA/KEDA behavior without writing effective replicas back;
- limit `Clamp` to wake-up from `0`, while rejecting active scale-in below `minAvailable`; and
- define `replicas: 0` semantics at all three resource levels.

The remaining policy decision—whether wake-up clamping should be a fixed rule or remain opt-in—is documented under Open Questions. I’d appreciate your thoughts on the proposed direction. Thank you :)

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: HI @shayasoolin 

Thank you for taking the time to discuss this. I am glad we reached the same general direction. I expanded the GREP to cover zero-replica semantics, wake-up-only clamping, autoscaler validation, and the remaining controller implications.

I would appreciate your guidance on two points:

1. For Open Question 1, should wake-up `Clamp` be fixed behavior without a policy field, or opt-in with `Block` as the default?
2. For Open Question 2, Should /scale status.replicas report desired, effective, or actual observed replicas, and should all three resource levels use the same semantics?

I retained both policy options because fixed `Clamp` may change existing behavior. Once these points are confirmed, I will simplify the GREP accordingly. Thank you again. :)

shayasoolin: 1. I feel okay making `Clamp` the fixed and only behavior. Even that it changes the current behavior. I see the current behavior as buggy.. it may cause undesired gang-terminations.
2. The desired state is in `spec.replicas`. This is the effective desired number, after clamping. I'd use `status.replicas` for the observed replicas.

@Ronkahn21 WDYT?

yankay: Thank you. I updated the GREP:

1. `Clamp` is the fixed and only behavior.
2. `status.replicas` reports observed replicas at all three levels.

yankay:
> A few things we'd like to see addressed before moving forward:
> 
> 1. **Autoscaler compatibility** — we'd like to understand how different autoscalers (HPA, KEDA, etc.) react to the mutation. If an autoscaler sets replicas to a value between 0 and minAvailable and we silently override it, it could cause a tight reconciliation loop.
> 2. **Mutation scope** — the defaulting/mutation should only apply in the scale-out direction: from 0 to minAvailable. For an invalid scale-in (e.g., a running PCS being scaled to a value between 0 and minAvailable), the webhook should return a validation error rather than mutating, so callers get explicit feedback.
> 3. **Replica 0 at every level** — we'd want to see replicas: 0 supported consistently across PodClique, PodCliqueScalingGroup, and PodCliqueSet. The GREP should clarify the expected behavior at each level before we move to implementation.

HI @Ronkahn21 
Thanks for the clear feedback. I’ve updated the GREP to:

1. document HPA/KEDA behavior without writing effective replicas back;
2. limit Clamp to wake-up from 0, while rejecting active scale-in below minAvailable; and
3. define replicas: 0 semantics at all three resource levels.

The remaining policy decision—whether wake-up clamping should be a fixed rule or remain opt-in—is documented under Open Questions. I’d appreciate your thoughts on the proposed direction.

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+Grove should treat `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP keeps `minAvailable` non-zero and immutable, makes gang logic tolerant of zero-replica members, and uses a fixed `Clamp` rule to wake a component at `minAvailable` when an autoscaler writes a positive value below quorum.
+
+## Motivation
+
+Scale-to-zero serving commonly keeps a router running while workers scale to zero. In that state, worker pods are intentionally absent, not unhealthy.
+
+Using `minAvailable: 0` would overload a field that already affects gang membership, termination, rolling updates, and startup ordering. `replicas` is the mutable scale target, so `replicas: 0` is the cleaner signal.
+
+External autoscalers should be treated as producers of desired replicas, not as components that understand Grove's per-component `minAvailable`. If Grove needs gang-safe membership above the autoscaler-written value, Grove should derive an internal effective value without writing back to `spec.replicas`.
```
shayasoolin: IMHO `spec.replicas` should reflect a **valid** desired number, i.e. either 0 or >= minAvailable. What's the pro of keeping an invalid value in `spec.replicas` and the effective valid one in memory?
Keeping `spec.replicas` constantly different than actual number of pods seems like the resource is not yet reconciled, while in fact it is.
It may also look like a bug - an invalid value is allowed at `spec.replicas`.
It also raises the question whether a value of `0<spec.replicas<minAvailable` is valid for a newly submitted PCS yaml? (It shouldn't be)
I don't see a reason to go this way, I believe the clamping solution we discussed in the issue should work for us without this change which is breaking an important invariant.

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+### Goals
+
+- Treat `replicas: 0` as an intentional idle state without changing `minAvailable`.
+- Make `0` the only value below `minAvailable` that Grove runs as written.
+- Clamp a wake-up request in `0 < replicas < minAvailable` to an effective count of `minAvailable`.
+- Preserve autoscaler-written `spec.replicas` as desired scale, never writing an effective value back.
```
shayasoolin: Again, I disagree. Let's bring the autoscalers to where we believe they should be, rather than assuming they won't play nicely if we do what we prefer. KEDA is already there - which is great news (with `idleReplicaCount` vs `minReplicaCount`) - and to me it shows we can safely use this approach.

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+| Desired replicas | Behavior |
+| --- | --- |
+| `replicas: 0` | Intentional idle state. The component contributes no required gang members and does not breach `minAvailable`. |
+| `0 < replicas < minAvailable` | Accepted only when waking from `0`. Grove preserves `spec.replicas` as desired scale and computes `effectiveReplicas = minAvailable`. |
+| `replicas >= minAvailable` | Normal Grove behavior. The component participates in gang scheduling and gang termination using its configured `minAvailable`; desired and effective replicas are equal. |
+
+Existing behavior changes in two cases: a `PodClique` with `replicas: 0` becomes idle instead of defaulting to one, and an object already in `(0, minAvailable)` is reconciled at `effectiveReplicas = minAvailable`.
+
+`Clamp` applies only on the wake-up edge: a write from `0` into the range is accepted, one from `minAvailable` or above is rejected, so `0` is the only legal scale-in target below `minAvailable`. An explicit scale-in deserves an error, not a silent promotion. Comparing against the previous value puts that check in a validating webhook on the resource and its `scale` subresource, not in defaulting.
```
shayasoolin: "An explicit scale-in deserves an error" - what do you mean?
Also - consider the case of `minAvailable=3`, `replicas=4` and during scale-in the autoscaler wants to set `replicas=2`. Are we going to reject it, or clamp it to 3? I think the latter is the right thing to do. We should allow the closest scale-in possible while keeping the service alive.

shayasoolin (commented):
Thanks @yankay several comments around the same topic actually.

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Hi @shayasoolin , thank you very much for taking the time to review this. I agree that `spec.replicas` should always represent a valid desired state: either `0` or at least `minAvailable`. I updated the GREP so invalid creates are rejected, while resource and `/scale` updates persist the clamped value. This is an important change because it preserves a clear resource invariant.

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Thank you, I see your point. The GREP now asks replica writers to emit either `0` or at least `minAvailable`. KEDA can already express this directly, while Native HPA, Knative KPA, Dynamo Planner are listed as future compatibility targets. I have updated the proposal accordingly.

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Apologies, the wording “An explicit scale-in deserves an error” was unclear. I was referring to the earlier wake-only design: with `minAvailable: 3`, a scale-in request from `4` to `2` would be rejected so the caller received explicit feedback instead of having the request clamped.

That direction followed an earlier review suggestion, so previous drafts used wake-only Clamp and left direction-agnostic behavior as an open question. Your example clarified the trade-off and convinced me that clamping to `3` is more useful and consistent. I have updated the GREP to use direction-agnostic Clamp and preserved the earlier approach under Historical Discussion.

Your comments resolved a major uncertainty in the proposal. Thank you again for the thoughtful review.

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -218,49 +201,43 @@ A `PodClique` produces no error there only because defaulting rewrites `replicas
...
-- The implementation surface in one place. Neither `PodClique` nor `PodCliqueScalingGroup` has a webhook today, so this GREP implies new ones on both and their `scale` subresource, plus dropping the `PodClique` `replicas: 0` defaulting, the `effectiveMinAvailable` readers, looser `scaleConfig` rules, observed `status.replicas` at all three levels, and upgrade handling for pre-existing below-quorum objects. Listing them is in scope; designing them is not.
+- The implementation surface in one place. Neither `PodClique` nor `PodCliqueScalingGroup` has a webhook today, so this GREP implies new ones on both and their `scale` subresources, updating the `PodClique` and `PodCliqueScalingGroup` template validation to allow `replicas: 0` while rejecting positive below-quorum values, dropping the `PodClique` `replicas: 0` defaulting, the `effectiveMinAvailable` readers, observed `status.replicas` at all three levels, and upgrade handling for pre-existing below-quorum objects. Listing them is in scope; designing them is not.
```
shayasoolin: I think the term `effectiveMinAvailable` was dropped, but it still appears twice in the doc, can you replace it with the up to date terminology?

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -218,49 +201,43 @@ A `PodClique` produces no error there only because defaulting rewrites `replicas
...
+- KEDA with `idleReplicaCount: 0` and `minReplicaCount: minAvailable` writes only valid values;
+- scaling to `minAvailable` or above re-enters normal gang behavior.
 
 ### Graduation Criteria
 
 - Alpha: direction accepted and initial implementation exists.
 - Beta: behavior documented and tested.
 - GA: semantics are stable and validated with real scale-to-zero workloads.
 
-## Open Questions
```
shayasoolin: As I see it, we still have an open question about that potential "write loop" you mention multiple times in the doc. We're not completely sure how things will look like when an autoscaler may insist on a value that for Grove is invalid (between 0 and minAvailable).
I'd keep it as an open question, ready for a POC, e.g. with HPA having its `minReplicas` value below `minAvailable`. Note that scale-to-zero is not required for this test.

shayasoolin (dismissed):
LGTM - thanks for the proposal and for considering the feedback! Left some small comments.

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Thanks for catching this. Both references have been updated. :)

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: HI @shayasoolin 

Thanks for suggesting this. I ran the POC against [[PR #788](https://github.com/ai-dynamo/grove/pull/788)](https://github.com/ai-dynamo/grove/pull/788), with the results recorded [[here](https://github.com/yankay/grove/commit/db4be97ad42da99929b07d375f5f2354fc8cf5d1)](https://github.com/yankay/grove/commit/db4be97ad42da99929b07d375f5f2354fc8cf5d1).

With `minAvailable: 2`, HPA using `minReplicas: 1` produced a `WRITE_LOOP`. The `minReplicas: 2` control produced `NO_LOOP`, but cannot scale to zero. All 14 floor-respecting controls produced `NO_LOOP`, while all 14 repeated below-quorum treatments produced `WRITE_LOOP`.

If we retain persisted `Clamp`, supported writers must emit either `0` or at least `minAvailable`. Native HPA would require an adapter for below-quorum recommendations, while KEDA supports this contract through `idleReplicaCount: 0` and `minReplicaCount: minAvailable`.

Would you be comfortable retaining persisted `Clamp`, defining this writer contract, documenting the Native HPA limitation, and treating the KEDA configuration as a supported scale-to-zero integration?

shayasoolin: Hi, yes I personally feel comfortable with this contract. We should document it well of course, but since KEDA already respects it, and HPA doesn't support scale-to-zero out of the box anyway, it sounds reasonable to me.
I'd love to hear @unmarshall and @renormalize opinion as well here :)

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -223,13 +223,28 @@ Prototype coverage should show:
...
+  | Grove `scaleConfig` | `minReplicas: 2` | `minReplicas: 0`, recommending `1` | `Value` and `AverageValue`; PCLQ and PCSG; control cannot scale to zero |
+  | KEDA | `idleReplicaCount: 0`, `minReplicaCount: 2` | `idleReplicaCount: 0`, `minReplicaCount: 1` | `Value` and `AverageValue`; PCLQ and PCSG |
+  | Custom controller | Repeatedly requests `2` | Repeatedly requests `1` | PCLQ and PCSG |
+  | Knative KPA | One adapter request produced no repeat | Not tested | Continuous-writer behavior not established |
+  | Dynamo Planner | Not established | Not established | `INCONCLUSIVE`; full Planner metric loop was not driven |
+
+  These columns describe experimental controls and treatments, not recommended scale-to-zero configurations. All 14 floor-respecting controls completed with `NO_LOOP`. All 14 treatments that repeatedly requested `1` completed with `WRITE_LOOP`, each issuing seven post-stabilization requests. The remaining design question is whether Grove should require supported replica writers to enforce the active `minAvailable` floor, provide an adapter for writers that cannot express it, or revise persisted `Clamp`.
```
shayasoolin: > All 14 floor-respecting controls completed

What are those 14 controls? Something in this section - from the table above to this description below - is not very clear. Consider simplifying significantly. I think the message will be clearer if you stick to KEDA vs HPA only, and focus on HPA POC alone. I believe all the others resemble each of these two anyways.

shayasoolin (dismissed):

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Thank you, this makes sense. The 14 controls were combinations of replica writers, metrics, and scale targets, which made the result harder to follow. I simplified the section to KEDA-like versus HPA-like behavior, kept only the HPA POC, and mapped the other writers to those two models.

I also removed the leftover `/scale.status.replicas` discussion because it is not directly related to the scale-to-zero contract and can be addressed separately.

shayasoolin:
Had a discussion with @sanjaychatterjee and @unmarshall. @unmarshall thinks that the cleaner API semantics is to reject anything between 0 and minAvailable, rather than accept and clamp it. Let's try to converge on an agreement 🙏🏼

yankay:
Thanks for helping converge on this. I agree that rejecting any positive value below `minAvailable` gives us cleaner API semantics.

The GREP now uses reject-only behavior , with Clamp retained only as historical context. Would you mind taking another look when convenient? :)

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -10,45 +10,44 @@
...
 - [Historical Discussion](#historical-discussion)
 - [Appendix](#appendix)
+  - [Historical HPA Clamp Experiment](#historical-hpa-clamp-experiment)
   - [Upstream Follow-ups](#upstream-follow-ups)
 <!-- /toc -->
 
 ## Summary
 
-Grove should treat `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP keeps `minAvailable` non-zero and immutable, makes gang logic tolerant of zero-replica members, and uses a fixed, direction-agnostic `Clamp` rule that persists any positive scaling update below quorum as `minAvailable`.
+Grove should treat standalone `PodClique` or `PodCliqueScalingGroup` components with `replicas: 0` as an intentional idle state. This GREP leaves existing `minAvailable` semantics unchanged, makes gang logic tolerant of zero-replica members, and rejects any positive replica count below `minAvailable` instead of clamping it.
```
shayasoolin: The "instead of clamping it" part references a previous commit, but is out of context for the reader who sees this as a standalone latest version. I suggest to drop it (and check for other similar references).

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -181,19 +177,23 @@ A writer wakes prefill by requesting one replica:
...
 
-| Request | Persisted `spec.replicas` |
+| Request | Behavior |
 | --- | --- |
-| `0` to `1` | `2` |
-| `2` to `1` | `2` |
-| `2` to `3` | `3` |
-| `3` to `2` | `2` |
-| `2` to `0` | `0` |
+| `0` to `1` | Rejected |
```
shayasoolin: Consider removing this table, which is now pretty trivial..

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -226,33 +226,12 @@ Prototype coverage should show:
...
-
 ## Alternatives
 
 - Allow `minAvailable: 0`: explicit, but overloads an existing availability contract.
 - Keep idle members in the `PodGang` with `minReplicas: 0`: keeps the gang spec stable across idle transitions, but that value already signals released constraints during a coherent update, and backends map it back to a positive threshold.
 - Make `minAvailable` mutable: follows scale state, but changes the immutability contract.
-- Only support `Block` and never add `Clamp`: keeps `spec.replicas` valid, but writers without an active floor cannot wake from zero or repeatedly receive rejections.
+- Clamp positive below-quorum requests to `minAvailable`: lets writers without an active floor wake from zero, but silently rewrites caller intent and does not make HPA-like writers converge.
```
shayasoolin: Add a link to the below dedicated section about this alternative.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -288,6 +267,21 @@ An intermediate draft left active scale-in into `(0, minAvailable)` open: should
...
+Can persisted `Clamp` make an HPA-like writer converge when its recommendation falls below `minAvailable`?
+
+An HPA POC against [PR #788](https://github.com/ai-dynamo/grove/pull/788) used `minAvailable: 2` and observed each case for 35 seconds after stabilization. Both configurations were tested with `Value` and `AverageValue` metrics against `PodClique` and `PodCliqueScalingGroup`.
+
+| HPA configuration | Result for both metrics and scale targets |
+| --- | --- |
+| `minReplicas: 2` | `0` repeated writes; `NO_LOOP` |
+| `minReplicas: 1`, recommending `1` | `7` repeated writes per case; `WRITE_LOOP` |
+
+The experiment demonstrates that persisted `Clamp` does not make an HPA-like writer converge: HPA keeps requesting `1` while Grove stores `2`. Reject preserves the stored replica count and surfaces the invalid recommendation instead of silently rewriting it, but the experiment did not test whether HPA retries after a rejected update. The reject implementation must verify and document that behavior. KEDA-like writers avoid the invalid range by emitting only `0` or at least `minAvailable`.
```
shayasoolin, line 299: Nice, thanks for mentioning this POC here.

shayasoolin (dismissed):
LGTM, please see a few nits. I'll ask @unmarshall to review as well.

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Thanks, removed “instead of clamping it”

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Thanks, I’ve removed the table.

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: Thanks, I’ve added the link.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
yankay: Thanks! Glad this context is helpful.

shayasoolin (dismissed):

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Design Details
+
+For `PodClique`, omitted `replicas` defaults to `1`, while explicit `0` is preserved. Omitted `minAvailable` defaults to `max(1, replicas)`.
+
+### Gang Behavior
+
+While a component is idle, it contributes no `PodGroup` and zero observed scheduled, available, and updated replicas. It does not set `MinAvailableBreached` to `True`, counts as updated for rolling-update completion, and does not trigger gang termination.
+
+An idle component is omitted from `PodGangMap` entry membership. If all components are idle, Grove retains an empty current-generation anchor without materializing a `PodGang`. On wake, Grove populates that anchor if no `PodGang` is active; otherwise, it creates a new entry for the waking component without expanding an existing materialized `PodGang`.
```
sanjaychatterjee: Can you please add some details on the gang semantics and PodGangMap update on waking? I think it is better for a PCSG waking up to be in a new SPG (scaled PodGang) since that'll preserve the gang scheduling requirement. However, a standalone PodClique waking up can be added to the anchor PodGang? @yankay @unmarshall wdyt?

docs/proposals/0677-zero-replica-gang-membership/README.md
```diff
@@ -0,0 +1,281 @@
+# GREP-0677: Zero-Replica Gang Membership
```
sanjaychatterjee: I think calling it `Zero-Replica Gang Membership` is a bit misleading. Instead, this looks more like a hibernation feature and so, why not call it that? Something like: `Zero-Replica Component Hibernation`

yankay: Thanks @sanjaychatterjee 

Good point, agreed. `Zero-Replica Gang Membership` is misleading . I’ve renamed it to `Scale-to-Zero Component Hibernation`, 😊

docs/proposals/0677-zero-replica-gang-membership/README.md
yankay: HI @sanjaychatterjee 

Thanks for the suggestion. I've added a **Gang Behavior** section to clarify the wake semantics and `PodGangMap` updates:

- PCSG replica indices enter the `ScaleOut` entry, with one SPG per replica. This preserves gang scheduling within each replica.
- A standalone PodClique rejoins the existing anchor by restoring its membership and full `PodGroup`, without creating another anchor.

One trade-off is that independent P/D PCSGs may wake at different times. Gang atomicity is preserved per replica, not across components.

sanjaychatterjee: @yankay Unfortunately, there is still an open question. We have to ensure gang-scheduling semantics for at least MinAvailable replicas. So, for standalone PCLQs, the wake-up action can add the newly created pod replicas into the anchor PodGang a new PodGroup. In the case of a PCSG, the MinAvailable number of replicas need to be added to a single PodGang so that they are gang-scheduled. The remaining replicas can be added as SPGs. 

Now the open question is how to add the MinAvailable replicas of the PCSG. If they go back into the anchor PodGang, then it is mutation of the PodGang which may or may not be honored by a downstream scheduler. I think KAI should be able to handle it as these new Grove PodGroups would be manifested as new SubGroups in the KAI PodGroup. However, I am not sure if a mutated Volcano PodGroup can still maintain gang semantics for the new additions. I would suggest we verify this behavior with both KAI and Volcano. @enoodle and @kangclzjc can you please help resolve this?

The other option is that Grove will creates a new PodGang for the MinAvailable replicas of a PCSG. However, this needs to be logically treated as a new role in the PodGangMap, since this new PodGang does not align with any of the existing roles: anchor, TPG or SPG. I would request to please resolve this issue with @unmarshall first. Maybe we discuss this in next week's China community sync. @kangclzjc please note.

kangclzjc: > However, I am not sure if a mutated Volcano PodGroup can still maintain gang semantics for the new additions

Sorry for the late response. If you mutate MinMember, it will not maintain gang semantics. For the other fields, I need to check this.

yankay: HI @kangclzjc 

Thanks for clarifying. I agree that changing `minMember` alone is insufficient. I tested updating the full `subGroupPolicy` together with `minMember` on Volcano v1.15.2, and the tested scenarios passed.

For the minimal case, each group requires one Pod. Using `p` and `d` as illustrative names for the two required groups:

```yaml
spec:
  minMember: 3
  subGroupPolicy:
    - name: router
      labelSelector:
        matchLabels: {backendcontract.grove.io/role: router}
      matchLabelKeys: [backendcontract.grove.io/role]
      subGroupSize: 1
      minSubGroups: 1
    - name: p
      labelSelector:
        matchLabels: {backendcontract.grove.io/role: p}
      matchLabelKeys: [backendcontract.grove.io/role]
      subGroupSize: 1
      minSubGroups: 1
    - name: d
      labelSelector:
        matchLabels: {backendcontract.grove.io/role: d}
      matchLabelKeys: [backendcontract.grove.io/role]
      subGroupSize: 1
      minSubGroups: 1
```

This passed three wake cycles; Router recreation during idle also passed. A `MinMember`-only control with surplus existing Router Pods allowed partial admission.
I hope these results help clarify the behavior.

yankay: > @yankay Unfortunately, there is still an open question. We have to ensure gang-scheduling semantics for at least MinAvailable replicas. So, for standalone PCLQs, the wake-up action can add the newly created pod replicas into the anchor PodGang a new PodGroup. In the case of a PCSG, the MinAvailable number of replicas need to be added to a single PodGang so that they are gang-scheduled. The remaining replicas can be added as SPGs.
> 
> Now the open question is how to add the MinAvailable replicas of the PCSG. If they go back into the anchor PodGang, then it is mutation of the PodGang which may or may not be honored by a downstream scheduler. I think KAI should be able to handle it as these new Grove PodGroups would be manifested as new SubGroups in the KAI PodGroup. However, I am not sure if a mutated Volcano PodGroup can still maintain gang semantics for the new additions. I would suggest we verify this behavior with both KAI and Volcano. @enoodle and @kangclzjc can you please help resolve this?
> 
> The other option is that Grove will creates a new PodGang for the MinAvailable replicas of a PCSG. However, this needs to be logically treated as a new role in the PodGangMap, since this new PodGang does not align with any of the existing roles: anchor, TPG or SPG. I would request to please resolve this issue with @unmarshall first. Maybe we discuss this in next week's China community sync. @kangclzjc please note.

HI @shayasoolin 

Thanks for the detailed analysis. I agree with restoring the first `minAvailable` PCSG replicas to the existing anchor, with the remaining replicas in SPGs, and have updated the GREP accordingly.

I tested the active → idle → wake lifecycle on KAI v0.17.1 and Volcano v1.15.2 using complete subgroup constraints, with stale-gang eviction disabled for KAI. In the tested scenarios, insufficient capacity blocked the waking minimum set while the existing Router remained running.  :-)

tzulingk:
The scale-to-zero design makes sense and I'm not proposing changes to it — the
`replicas == 0 || replicas >= minAvailable` invariant is clean, and idling at the
group level is right for the router/worker case it targets.

My question is about a different level, and I don't think the two overlap.

The per-level table gives `replicas: 0` idle semantics to `PodCliqueSet`,
`PodCliqueScalingGroup`, and standalone `PodClique`, but excludes a PCSG member:

> A `PodClique` owned by a `PodCliqueScalingGroup` must not be scaled
> independently, including to zero.

For engine-world resize the member clique is the level that matters, and the group
itself is never idled:

    PodCliqueScalingGroup  replicas: 1      <- one expert-parallel world, stays up
      +- PodClique "leader"  replicas: 1
      +- PodClique "worker"  replicas: 3    <- one DP rank per pod

Growing that world is `worker: 3 -> 4`; shrinking it is `worker: 4 -> 3`. The PCSG
stays at 1 throughout and is never scaled to zero, so this doesn't touch the idle
path the GREP defines. Scaling the PCSG instead would create a second complete
world, which is a different operation.

This is the contract #793 asks Grove to confirm as stable, and what
ai-dynamo/dynamo#13121 builds on. #823 (GREP-0793) has an e2e that does it.

cc @julienmancuso

yankay:
> The scale-to-zero design makes sense and I'm not proposing changes to it — the `replicas == 0 || replicas >= minAvailable` invariant is clean, and idling at the group level is right for the router/worker case it targets.
> 
> My question is about a different level, and I don't think the two overlap.
> 
> The per-level table gives `replicas: 0` idle semantics to `PodCliqueSet`, `PodCliqueScalingGroup`, and standalone `PodClique`, but excludes a PCSG member:
> 
> > A `PodClique` owned by a `PodCliqueScalingGroup` must not be scaled
> > independently, including to zero.
> 
> For engine-world resize the member clique is the level that matters, and the group itself is never idled:
> 
> ```
> PodCliqueScalingGroup  replicas: 1      <- one expert-parallel world, stays up
>   +- PodClique "leader"  replicas: 1
>   +- PodClique "worker"  replicas: 3    <- one DP rank per pod
> ```
> 
> Growing that world is `worker: 3 -> 4`; shrinking it is `worker: 4 -> 3`. The PCSG stays at 1 throughout and is never scaled to zero, so this doesn't touch the idle path the GREP defines. Scaling the PCSG instead would create a second complete world, which is a different operation.
> 
> This is the contract #793 asks Grove to confirm as stable, and what [ai-dynamo/dynamo#13121](https://github.com/ai-dynamo/dynamo/issues/13121) builds on. #823 (GREP-0793) has an e2e that does it.
> 
> cc @julienmancuso

HI @tzulingk 

Nice catch, thanks! I've updated the GREP to clarify that a `PodClique` owned by a `PodCliqueScalingGroup` can still be scaled independently to non-zero values when `replicas >= minAvailable`.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+A `PodClique` owned by a `PodCliqueScalingGroup` must not be independently scaled to zero; idle it by setting the owning group's `replicas` to `0`. Non-zero scaling remains allowed when `replicas >= minAvailable`.
+
+### Below-Quorum Behavior
+
+Grove rejects positive replica counts below `minAvailable`:
+
+| Requested replicas | Behavior |
+| --- | --- |
+| `0` | Intentional idle state. The component contributes no required gang members and does not breach `minAvailable`. |
```
kangclzjc, line 72: `PodCliqueSpec.Replicas` currently is int32, if we want to support replicas: 0` as an intentional idle state and default as 1, we may need to change the type to *int32. Otherwise, we can't tell that users forget to set the replica or set replica to 0 intentionally? Shall we state that in the API change?

yankay: Thanks @kangclzjc 

It has been added to the GREP design details: https://github.com/ai-dynamo/grove/pull/686/changes#diff-d988166db613f46bd0b59f73b5f855bfd0238df55f00c03fc7a710edfb848ba3R98

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+| `0 < replicas < minAvailable` | Rejected for create and update requests, including the `/scale` subresource. |
```
unmarshall, line 73: There is currently no admission webhook on PodClique / PodCliqueScalingGroup objects or their /scale subresource. So today it allows you to go below `minAvailable`.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+| `replicas >= minAvailable` | Persisted as requested. Waking components follow the [gang behavior](#gang-behavior) below. |
+
+For independent scale targets, the persisted invariant is:
+
+```text
+spec.replicas == 0 || spec.replicas >= minAvailable
+```
+
+Reject is independent of the previous replica count. With `minAvailable: 3`, requests from `0` to `1` and from `4` to `2` are both rejected.
```
unmarshall, line 82: Not clear. Perhaps you should create a table where you have rows:
For minAvailable: 3
0 -> 1 rejected
4 -> 2 rejected (this is not there today)
4 -> 0 allowed
0 -> 3 allowed

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+The main risk is confusing "idle" with "unavailable." The proposed direction ties idle state to desired replicas, not observed status.
+
+An HPA may repeatedly request an invalid below-quorum value. The [historical HPA Clamp experiment](#historical-hpa-clamp-experiment) shows that accepting and clamping such requests does not make the writer converge; the reject path requires verification once implemented.
+
+Rolling updates already stall on a zero-replica standalone `PodClique`: completion requires `minAvailable` updated and ready pods, which zero pods never reach. A `PodCliqueScalingGroup` at `replicas: 0` does not stall, because its completion check compares only generation hashes. That asymmetry is why idle needs an explicit definition rather than a per-level accident.
+
+## Design Details
+
+For `PodClique`, change `PodCliqueSpec.Replicas` from `int32` to `*int32` with `json:"replicas,omitempty"` to distinguish omission from explicit zero. Omitted `replicas` defaults to `1`, while explicit `0` is preserved. Omitted `minAvailable` defaults to `max(1, replicas)`.
```
unmarshall, line 98: This creates asymmetry.
Today we already have `+kubebuilder:default=1` on replicas field (int32). The CRD structural default fires when the field is absent from the submitted manifest and preserves an explicit 0 if specified.
From the looks of it changing this to `*int32` is not required.

But if you should do it then please do it uniformly for PCS, PCSG and PCLQ resources.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+- Related bug: [ai-dynamo/grove#676](https://github.com/ai-dynamo/grove/issues/676)
+- Feasibility prototype: [yankay/grove#2](https://github.com/yankay/grove/pull/2)
+- Local YAML shape: [multinode-disaggregated-with-frontend.yaml](../../../operator/samples/user-guide/02_pod-and-resource-naming-conventions/multinode-disaggregated-with-frontend.yaml)
+- Related Dynamo issue: [ai-dynamo/dynamo#10753](https://github.com/ai-dynamo/dynamo/issues/10753)
+- Related Dynamo PR: [ai-dynamo/dynamo#10532](https://github.com/ai-dynamo/dynamo/pull/10532)
+- Dynamo autoscaling guide: [Dynamo autoscaling](https://docs.nvidia.com/dynamo/v1.1.1/kubernetes-deployment/deployment-guide/autoscaling)
+
+### Historical HPA Clamp Experiment
+
+This historical, non-normative experiment evaluated persisted `Clamp`, not the current Reject proposal. It shows that Clamp did not eliminate repeated writes, but provides no evidence about retry behavior after rejection.
```
unmarshall, line 291: Difficult to read and understand. Simplify the language please.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+Can persisted `Clamp` make an HPA-like writer converge when its recommendation falls below `minAvailable`?
```
unmarshall, line 293: You have repeatedly used the word `Clamp` without clarifying what it means precisely.
I think what you want to say is that the HPA can set replicas to below the minAvailable and Grove will `clamp` it back to `minAvailable` silently. HPA sees that the value is higher than what it set in the first place and writes back the value only to be clamped back (silently) by Grove to minAvailable.

Now when i read the code i do not find this clamping anywhere. As told before today Grove allows going below the minAvailable. Can you give the code reference where this clamping happens today, maybe i missed it?

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+## Design Details
+
+For `PodClique`, change `PodCliqueSpec.Replicas` from `int32` to `*int32` with `json:"replicas,omitempty"` to distinguish omission from explicit zero. Omitted `replicas` defaults to `1`, while explicit `0` is preserved. Omitted `minAvailable` defaults to `max(1, replicas)`.
+
+### Gang Behavior
+
+While a component is idle, it contributes no `PodGroup` and zero observed scheduled, available, and updated replicas. It does not set `MinAvailableBreached` to `True`, counts as updated for rolling-update completion, and does not trigger gang termination.
+
+On `N -> 0`, idle components leave `PodGangMap` membership and `A0`'s required groups without disrupting surviving pods or their recreation. Empty current-generation anchor and scale-out entries remain logical slots without materialized `PodGangs`.
```
unmarshall, line 104: Except scale-out entry all empty entries are removed from the PGM. Can you double check your assertion here?

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+On `0 -> N`:
```
unmarshall, line 106: you need to be explicit and state clearly that N >= minAvailable.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+1. **PCSG:** restore replicas `[0, minAvailable)` into the current-generation anchor for joint admission. Remaining replicas enter the current-generation `ScaleOut` entry, creating one SPG per replica.
```
unmarshall, line 108: 1. `restore` is inaccurate terminology. When you say restore then you are going back to a previous state that existed prior to setting replicas to 0. There is no where this state is stored and the GREP also does not suggest anything as well. 
2. Why should replicas above minAvailable go to `ScaleOut` entry. When you do the initial deployment of the PCS, anything above minAvailable goes to a single `Tail` entry. Nothing is there in the scale-out entry. Post initial deployment, if there is an explicit scale-out then only it goes into the scale-out entry.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+2. **Standalone PodClique:** restore membership and the full `PodGroup` in the current-generation anchor (`A0`, or the highest `AnchorIndex` if several exist), materializing it if needed. No additional anchor is created.
+
+```text
+PCSG (minAvailable: 2):
+  Initial:  A0 [Router + replica 0 + replica 1] + SPG [replica 2]
```
unmarshall, line 113: That is no longer accurate. After migration there is no SPG. There is only Anchor, Tail and ScaleOut.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+  Idle:     A0 [Router]
+  Wake:     A0 [Router + replica 0 + replica 1]
```
unmarshall, line 115: Confused: when you use the term wake up what does that actually mean? Who sets the replicas back to 3? If spec.replicas is 0 while idle, there is no persisted per-replica identity to "restore." Wake is a fresh scale-out from 0 to N, not a restoration of prior replica indices.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+             +--> SPG [replica 2], after this wake's quorum is admitted
+
+Standalone Decode:
+  Before idle: A0 [Router + Decode]
+  Idle:        A0 [Router]
+  Wake:        A0 [Router + Decode]
+```
+
+An empty `ScaleOut` entry gets a fresh epoch, depending on a materialized anchor only if one exists. Non-empty entries retain their epoch and dependencies.
```
unmarshall, line 124: If a PCS defines at least a single PCSG, then it will always have an empty ScaleOut entry in the PGM. So what do you mean by `An empty `ScaleOut` entry gets a fresh epoch`?
Also `only if one exists` expects the ScaleOut entry to be missing.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+An existing `A0` keeps its epoch, but its historical `LastScheduled` alone must not release new SPGs: extras wait for their PCSG's current wake quorum. Backends must enforce the restored `MinReplicas` and subgroup minima without disrupting surviving pods or triggering Grove gang termination.
```
unmarshall, line 126: `Quorum` is typically used as a consensus threshold for voting in a distributed consensus use cases. `minAvailable` is a gang-scheduling and gang-termination threshold. So using this term in this context is a bit misleading.

Also lets simplify the language a bit - its too dense for me.

A reused anchor carries a stale `LastScheduled` timestamp from before the component went idle. Upon wakeup, that timestamp must not be read as evidence that the anchor's minAvailable replicas are freshly
scheduled, otherwise the backend would release the scaled PodGangs (the PCSG replicas above minAvailable) before the wakeup's minAvailable replicas (anchor PG) are actually co-scheduled. Required behavior: the scaled PodGangs stay gated until the reused anchor's minAvailable replicas schedule in this wake, not on the value of the old LastScheduled.

unmarshall: But there is a bigger concern this opens up.
An anchor PG can consist of both standalone PCLQs and PCSGs. If only some PCLQs / PCSGs are scaled-in to 0, then the anchor PG continuous to exist since there are some parts that are scheduled and active. The anchor PodGang was admitted/scheduled long ago and keeps its LastScheduled.

Adding a PodGroup back into an already-admitted PodGang is a mutation of a live PodGang, not a fresh gang admission/scheduling. In other words the scaled-out component replicas and re-association to the existing anchor PG, will not be gang scheduled.

The only case where a wakeup actually gets gang scheduling is when the anchor was fully drained and removed. So it re-materializes as a fresh PG. But that's not the router-stays-up scenario the GREP is built around.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+- **KAI (v0.17.0+):** Update full subgroup policies; set `spec.stalenessGracePeriod: "-1s"` only on Grove-managed PodGroups, not globally.
```
unmarshall, line 128: This contradicts to what https://github.com/ai-dynamo/grove/pull/831/changes/68d94b48d27455ef764f8de01576945c7b5478b0 does. Please align with that decision. Gang termination at the backend scheduler level is never switched off.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+- **Kubernetes WAS:** Add native hierarchical mapping, pending support for [MinGroupCount mutable for CompositePodGroups](https://github.com/kubernetes/kubernetes/pull/141023).
+
+### Autoscaler Integration
+
+Grove-compatible replica writers must converge to `0` while idle and at least `minAvailable` while active without repeatedly writing a positive below-quorum value.
+
+| Model | Behavior | Compatibility |
+| --- | --- | --- |
+| KEDA-like | Separate idle and active floors, such as KEDA `idleReplicaCount: 0` with `minReplicaCount: minAvailable`. | Compatible. |
+| HPA-like | May repeatedly write a positive below-quorum value or prevent scale-to-zero. | Incompatible. |
```
unmarshall, line 139: Grove's own `scaleConfig` generates an HPA, and HPA-like writers are declared incompatible with this design, then Grove's own autoscaling path is incompatible with its own scale-to-zero proposal. This is a fundamental gap that should be a first-class "Limitations" item or a Non-Goal with a migration plan.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
```diff
@@ -0,0 +1,310 @@
...
+
+[KEDA](https://keda.sh/docs/2.20/reference/scaledobject-spec/) emits only `0` or at least `minAvailable`. Native [HPA](https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/#scaling-to-and-from-zero) is HPA-like. Knative Serving v1.23.0 KPA is also HPA-like because [`autoscaling.knative.dev/activation-scale`](https://knative.dev/docs/serving/autoscaling/scale-bounds/#scale-up-minimum) is not a persistent active floor. Dynamo Planner has not established Grove-compatible scale-to-zero behavior. Grove `scaleConfig` is HPA-like because it generates an HPA. New custom controllers should use KEDA-like semantics to support scale-to-zero.
```
unmarshall, line 141: > Dynamo Planner has not established Grove-compatible scale-to-zero behavior.

Without support from Planner, would you really want to provide this option? This entire GREP now looks like a KEDA specific solution.

docs/proposals/0677-scale-to-zero-component-hibernation/README.md
unmarshall: In case of KEDA this would perhaps map to `minReplicaCount` which will be equal to `minAvailable`, right?

yankay:
Hi @sanjaychatterjee 

Following up on our last meeting, I ran the native scheduler tests and prepared this [summary](https://github.com/yankay/grove/blob/5ff8c0aea4a16249ad17ad501048651f264cbed6/docs/testing/native-scheduler-capabilities-summary-20260924.md).
Volcano eventually adopts the updated policy, but new Pods can bind against the old policy during propagation. I reported this in [volcano-sh/volcano#6010](https://github.com/volcano-sh/volcano/issues/6010) for upstream discussion.
I think the overall approach remains viable, although strict wake-up guarantees still need to be addressed. What do you think?

Issue #405, opened 2026-02-08
☂️ Reconciler performance issues


unmarshall:
### What happened?

A few hot spots have been identified in the reconciler flow which results in significant spike in CPU consumption. This is an umbrella issue which will identify and provide fixes for the performance bottlenecks.

### Environment
- Grove version: Since v0.1.0-alpha.1

PR #692, opened 2026-06-28
perf(operator): suppress success events for no-op syncs

cparadis-nvidia:
## Summary

- add a shared helper that records `CreateOrPatch` success events only when an object changed
- suppress success events for unchanged PodCliques, PodCliqueScalingGroups, PodGangs, and ComputeDomains
- preserve warning events and success events for actual creates and updates

Steady-state reconciles commonly return `OperationResultNone`; emitting a Kubernetes Event for those no-op syncs creates avoidable event API writes and control-plane noise.

Related scalability context: #405 .

## Testing

- `go test ./... -count=1`
- `make --directory=operator lint`


operator/internal/controller/common/component/events.go
```diff
@@ -0,0 +1,33 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// RecordCreateOrPatchSuccessEvent records a Normal event when CreateOrPatch changed the object.
+// Steady-state reconciles return OperationResultNone and must not generate success events.
+func RecordCreateOrPatchSuccessEvent(eventRecorder record.EventRecorder, object runtime.Object, opResult controllerutil.OperationResult, reason, messageFmt string, args ...any) {
+	if opResult == controllerutil.OperationResultNone {
+		return
+	}
+	eventRecorder.Eventf(object, corev1.EventTypeNormal, reason, messageFmt, args...)
```
Ronkahn21: Rather than an all-or-nothing approach, we should consider a more balanced mechanism for rate-limiting these events, one that prevents API server spam without silencing them entirely.

Ronkahn21 (changes requested):

Ronkahn21 (commented):
I like the apprch but I would go farther creating a new event submitter client that impl the same interface as the original but add lmiting the event type as config build in
its clear approch also it should sit in new packge

cparadis-nvidia:
hey @Ronkahn21 here is the latest implementation following your latest suggestion: 
https://github.com/ai-dynamo/grove/pull/692/changes/b6804628666b01595ac5cc1cba9bf12b202dbd30

PR #694, opened 2026-06-30
Make `default-scheduler` the only enabled profile by default.

### LINKED ISSUE #695: Operator `CrashLoopBackOff`s when the default charts are used in a vanilla cluster
### What happened?

The operator `CrashLoopBackOff`s when the default charts are used either through Helm, or through the make targets used for development through `make kind-up deploy`. 

This is because enabling all scheduler backends by default will make the operator ensure that all relevant CRDs for these backends are present in the cluster.

We can not expect that all these relevant CRDs are installed by default in any cluster, and should have an opt-in approach where the user consciously chooses which profiles are to be enabled, and therefore is responsible for installing the relevant CRDs.

The operator should run successfully in any cluster through the default charts we have, *without any modification*. User should modify the charts according to their own requirements.

### Environment

- Grove version d7cb1bc22ca1f93b0e37394c11b03228ab23d8d2

renormalize:
#### What type of PR is this?

/kind bug

* The operator `CrashLoopBackOff`s when brought up using the development make targets `make kind-up deploy`. This is because the CRDs required by Volcano (and other backends) are not installed in this cluster by default.

* Commented out all profiles other than the `default-scheduler` profile in `charts/values.yaml`, users can configure this as needed.

Specifying all the supported backends by Grove as an uncommented list in the profiles will make consumption harder; all these profiles are *expected* to be in the cluster when the users are using the default, unmodified charts.

Commenting these additional profiles in the default charts will force consumers to make a conscious decision to enable a specific profile that they want to use, setting the expectation that they are responsible to installing the relevant CRDs into the cluster.

Fixes #695

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/charts/values.yaml
+++ b/operator/charts/values.yaml
@@ -93,12 +93,13 @@ config:
       concurrentSyncs: 3
   # Scheduler configures which scheduler backends are active. default-scheduler is always available.
   # List profiles to enable backends; set defaultProfileName to the profile that is the default backend.
+  # Enable the corresponding profiles when required.
   scheduler:
     profiles:
       - name: default-scheduler
-      - name: kai-scheduler
-      - name: volcano
-      - name: lpx-scheduler
+      # - name: kai-scheduler
+      # - name: volcano
+      # - name: lpx-scheduler
   logLevel: info
   logFormat: json
   topologyAwareScheduling:


danbar2 (approved):

Ronkahn21 (approved):

renormalize:
https://github.com/ai-dynamo/grove/issues/695#issuecomment-4889512805

PR #696, opened 2026-06-30
docs: GREP-0285 job support proposal

shayasoolin:
Adds GREP-0285 — the Grove Enhancement Proposal for adding job support to Grove: finite, completion-oriented workloads with bottom-up completion and failure semantics, gang restart, and new API fields (`completedNames`, `maxRestarts`).

Related to #285


docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,377 @@
...
+
+Non-job-mode `PodClique`s (`restartPolicy: Always`) never set `Completed` or `Failed`.
+
+**PodCliqueScalingGroup**
+
+Evaluation is two-level:
+
+1. *Replica state*: a replica is **Completed** when all of its job-mode child `PodClique`s are `Completed` — or, if `completedNames` is set, when all named children are `Completed`. A replica is permanently **Failed** when a required child `PodClique` fails and no restart budget remains (`maxRestarts` exhausted, or `maxRestarts: 0`). A failure of a `PodClique` not listed in `completedNames` triggers a gang restart and consumes budget, but does not immediately mark the replica as `Failed` — the named children can still complete on the next attempt.
+
+2. *PCSG state*: the PCSG is **Completed** when all replicas are `Completed`. It is **Failed** when enough replicas have exhausted their budget that the remaining replicas cannot satisfy the completion criterion. If `maxRuntime` is exceeded, the PCSG is marked `Failed` immediately.
```
danbar2, line 216: what we do with the running PCLQ's when the `maxRuntime` is exceeded ?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,372 @@
...
+          restartPolicy: Never
+```
+
+### Completion and Failure Evaluation
+
+Completion and failure are evaluated independently at each level, using only the observed state of direct children.
+
+**PodClique**
+
+A `PodClique` is **Completed** when all of its pods have exited with code 0 (`phase=Succeeded`). It is **Failed** when any pod exits with a non-zero code (`phase=Failed`), since pod-level retry is not supported and a single failure makes the all-pods completion criterion unreachable. If `maxRuntime` is exceeded, the `PodClique` is marked `Failed` immediately regardless of pod states.
```
sanjaychatterjee, line 206: Is it possible to relax this behavior with a configurable completion count?

docs/proposals/0285-job-support/README.md
shayasoolin: Thanks, added notes about that - mostly in the 'cleanup' section, please see two new commits `b945f5c68d711515534734a9c9ecdc60b802c662` and `a7b36c7392e6c8ecddfc77a784d080f6aae78820`

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,387 @@
...
+
+**`maxRuntime` at the `PodCliqueScalingGroup` / `PodCliqueSet` level is not reset on restarts.**
+`maxRuntime` on a `PodCliqueScalingGroup` or `PodCliqueSet` is measured from the first start of that resource and does not reset when replicas are restarted. Users who set a tight deadline may find the budget exhausted before all retry attempts have run. Mitigation: documentation should clearly state this behavior; users should size `maxRuntime` to account for the full expected duration across all attempts, not just one. Note: `maxRuntime` at the `PodClique` level does reset per instance, since the `PodClique` itself is recreated on gang restart.
+
+**Log loss on retry.**
+When a failed `PodClique` is deleted and recreated during a gang restart, terminal pods from the previous attempt are cascade-deleted along with it. Logs from failed attempts are not durably retained across retries. Mitigation: users who need per-attempt logs should rely on a cluster-level logging stack (e.g. Fluentd, Loki) rather than `kubectl logs`.
+
+## Design Details
+
+### Job Mode Signal
```
Ronkahn21, line 106: **Design concern: `restartPolicy: Never` is not a safe job mode signal**

Using `restartPolicy: Never` as the sole signal for job mode ties a workload classification (finite completion semantics) to a pod runtime policy (kubelet does not restart containers in-place). These are orthogonal concerns, and coupling them means the API cannot express a future mode where one applies without the other.

The GREP also does not include an explicit implementation checklist for the required webhook changes. The Test Plan mentions that `restartPolicy: OnFailure` is rejected at admission, but the defaulting webhook today warns on any explicit `restartPolicy` and treats it as unsupported — that behavior also needs to change for `Never` to work as the job mode signal. Without naming these as discrete implementation tasks (defaulting webhook update, validation webhook update), they are at risk of being missed or handled inconsistently.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,387 @@
...
+
+// MaxRestarts is the maximum number of times a replica may be restarted after
+// failure. Each restart consumes one unit from this per-replica budget. A replica
+// that exhausts its budget is considered failed. Defaults to 0 — the first failure
+// immediately marks the replica failed with no restart attempt.
+// +optional
+// +kubebuilder:default=0
+MaxRestarts *int32 `json:"maxRestarts,omitempty"`
+
+// MaxRuntime is the maximum wall-clock duration from when the first replica of
```
Ronkahn21: **Design concern: `maxRuntime` has different reset semantics at different hierarchy levels**

`maxRuntime` is defined at three levels, but its reset behavior differs:

- **PodClique**: resets per attempt — the PodClique CRD object is recreated on gang restart, so each new instance starts a fresh clock.
- **PCSG / PCS**: does not reset — the timer runs from the first replica start across all attempts.

Same field name, opposite reset semantics. A user setting `maxRuntime: 2h` on a PCLQ gets "each attempt may run for 2 hours." A user setting `maxRuntime: 2h` on a PCSG gets "the entire job, across all retry attempts, must finish within 2 hours." These are fundamentally different concepts and the distinction is not surfaced at the point of configuration.

A user familiar with PCLQ-level `maxRuntime` will reasonably assume the same per-attempt semantics at PCSG/PCS level. Setting `maxRuntime: 2h` with `maxRestarts: 3` looks like budgeting 8 hours total — it actually caps the whole job at 2 hours, making retries 2–4 impossible if each attempt approaches that limit. This class of misconfiguration will be silent: the job simply never gets its retry attempts.

shayasoolin: I hear you. But let me suggest a slightly different way to look at it, and tell me if it still looks confusing.
We can say that `maxRuntime` is always on the whole resource level, rather than per-replica.
For PCS and PCSG this is already clear - it starts with the first scheduled pod and resets only when the resource is recreated. The PCS is never recreated, but a PCSG can be recreated, when a PCS replica is restarted.
For PCLQ it's also reset when the PCLQ is recreated, which is when the parent PCS/PCSG replica is restarted.
I think the confusion comes from the fact that as of the first stage of this feature, we don't allow pod restarts in the PCLQ level - any single pod failure causes the PCQL to fail. But this may change in the future, and then `maxRuntime` will still be cross-replica for PCLQ, just like it is for PCSG and PCS.
Makes more sense?

Ronkahn21: PCSG and PCS arent the owner for pods, so how you defined it on that level ?
I am a bit confugege with the data flow is it bottom top or top bottom?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+Completion and failure are evaluated bottom-up. At each level, the resource is **Completed** when all job-mode children have succeeded, and **Failed** when enough children have failed that the completion criterion is unreachable — regardless of how remaining children resolve. Once a resource reaches **Failed**, it cannot subsequently become **Completed**.
+
+When a `PodClique` fails, its parent (`PodCliqueScalingGroup` or `PodCliqueSet`) deletes it and recreates the full gang as a unit. This consumes one restart from the parent's per-replica budget. A `PodCliqueScalingGroup` or `PodCliqueSet` fails when enough of its replicas have exhausted their budget that the completion target is unreachable.
```
sanjaychatterjee, line 64:
```suggestion
When a `PodClique` fails, its parent (`PodCliqueScalingGroup` or `PodCliqueSet`) deletes it and recreates the full resource as a unit. This consumes one restart from the parent's per-replica budget. A `PodCliqueScalingGroup` or `PodCliqueSet` fails when enough of its replicas have exhausted their budget that the completion target is unreachable.
```

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+Three new API fields control job-mode policy:
+
+- **`completedNames`** *(PodCliqueScalingGroup / PodCliqueSet only)*: Named children within a replica that must complete for that replica to count as Completed. If omitted, all job-mode children must complete. At the `PodClique` level, all pods must complete for the `PodClique` to be considered Completed.
+- **`maxRestarts`** *(PodCliqueScalingGroup / PodCliqueSet only)*: Per-replica restart budget. A replica that exhausts its budget is considered failed.
```
sanjaychatterjee, line 69: Can you please clarify this point a bit more? 

1. IIUC `maxRestarts` for a PCSG is the max number of restarts tolerated per replica, and not the aggregated value over all PCSG replicas, correct?

2. In case a PCLQ (e.g. job-worker) within a PCSG replica fails, then the entire PCSG replica would restart and consume one restart from its restart budget, right? And given the semantics of PCLQ restart, does this now also imply that any pod failure would trigger the PCLQ failure and which would also consequently cause the PCSG replica to restart?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,403 @@
...
+- **`maxRuntime`** *(all levels)*: Wall-clock deadline from `status.jobStartTime`, which is set on first start. Exceeding it fails the resource immediately and permanently, regardless of any subsequent child outcomes. At the `PodCliqueScalingGroup` / `PodCliqueSet` level, `maxRuntime` is not reset when replicas are restarted. At the `PodClique` level, the timer resets per instance — the `PodClique` is recreated on gang restart, so each new instance starts a fresh clock.
```
sanjaychatterjee, line 70: It is not clear if `maxRuntime` is per-resource or per-replica? If it is per-resource, then exceeding the `maxRuntime` would imply failure of all the resource replicas? What happens if a PCS is deployed with 2 replicas, but one of the replicas gets scheduled much later than the other replica?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+#### Story 3: Leader-driven completion
+
+As a machine learning engineer running a leader-worker training job, I want the job to be considered complete when the leader pod exits successfully — regardless of whether worker pods have exited — so that I can use the leader's exit code as the authoritative signal of job success, while workers remain running until the leader finishes.
+
+### Limitations/Risks & Mitigations
+
+**Gang termination depends on application-level failure propagation.**
+In job mode, `MinAvailableBreached`-based gang termination is disabled. If pods are evicted (e.g. due to a node failure) but the remaining running pods do not exit — for example because the application hangs rather than fails — Grove will not detect the partial gang and will not trigger a restart. Gang termination relies on the application observing the failure (e.g. a rendezvous timeout or broken collective) and exiting non-zero. Workloads that do not fail fast on peer loss may stall indefinitely.
```
sanjaychatterjee, line 93: This should never happen when `restartPolicy` is set as `Never`. If such as pod is evicted, the pod should transition to a `Failed` state afaik. Then that should trigger the PodClique failure as well?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+**API overhead and topology placement loss at scale.**
+Job mode uses `restartPolicy: Never`, which disables kubelet's in-place container restart. Grove is solely responsible for recreating pods on failure. At scale, this means every gang restart triggers a full pod deletion and recreation cycle — incurring Kubernetes API overhead and requiring the scheduler to re-place all pods from scratch. Re-scheduling at scale can take meaningful time and may not recover the same topology placement that the previous attempt had. This is a known limitation of the design.
```
sanjaychatterjee, line 96: You could also argue that this "limitation" is a consequence of a node or pod failure, which is beyond Grove's control.

sanjaychatterjee (commented):
1/n reviews

docs/proposals/0285-job-support/README.md
shayasoolin: It's basically bottom-up but once a resource reaches terminal states, its children should be handled so it's also top-down. You can see the last changes I made in the GREP to explain that.

docs/proposals/0285-job-support/README.md
shayasoolin: We considered this feature in our discussions (with you) and decided for MVP to go with the simplest definition where all the pods should complete successfully. 
Both `completions` and `completedIndexes` can be added later as an extension. It's mentioned in the GREP, but marked as out of scope for now.

docs/proposals/0285-job-support/README.md
shayasoolin: I intentionally use the term `gang` here. A gang may be constructed of several resources. Can you explain the change you suggest here? Are we calling a PCS/G replica a `resource`?
The emphasis here is that the behavior acts exactly as gang-termination - same termination scope as the current gang-termination mechanism.

docs/proposals/0285-job-support/README.md
shayasoolin: 1. Correct
2. Correct, exactly

docs/proposals/0285-job-support/README.md
shayasoolin: It's per-resource (_"Exceeding it fails the resource immediately"_).
If you have two replicas, it starts counting as soon as the first replica gets scheduled.
The idea is for the user to be able to say "this workload should be completely done max X hours after it started consuming any resources." (That's for the PCS level. Same idea for PCSG and PCLQ)

docs/proposals/0285-job-support/README.md
shayasoolin: Yes I think you're right, thanks. I'll update this section later on, once we decide on `MinAvailableBreached`-based gang termination since as discussed we may bring it back (with an adjusted `MinAvailable` interpretation).

docs/proposals/0285-job-support/README.md
shayasoolin: If we put node failure aside - pod failures require recreating pods, which may be many at scale - this is the issue I mention here. The pods failure itself is beyond Grove's control but the handling is very much its control, so I'm just mentioning here that our choice to not let the kubelet handle failures has these consequences.

yankay:
Thanks for the proposal. Could you help me understand the intended behavior when `completedNames` is satisfied while a non-required child fails in the same reconciliation? If the leader is `Completed` and a worker is `Failed`, should the replica complete, or should Grove restart the gang?

shayasoolin:
@yankay 
> Thanks for the proposal. Could you help me understand the intended behavior when `completedNames` is satisfied while a non-required child fails in the same reconciliation? If the leader is `Completed` and a worker is `Failed`, should the replica complete, or should Grove restart the gang?

Good question. I think there are two coherent options here:

**Option A:** allow non-required completion-aware children, but define precedence clearly.
A child not listed in `completedNames` can still fail the current gang attempt and trigger a gang restart while the required completion criterion is not yet satisfied. If the restart budget is exhausted before the required children complete, the replica fails. However, if `completedNames` is already satisfied in the same reconciliation snapshot, the replica completes and no restart budget is consumed for non-required children.

**Option B:** require children outside `completedNames` to be regular, non-completion-aware PCLQs.
In this model, only children listed in `completedNames` may be completion-aware. Non-required children must be regular long-running PCLQs, so their failures do not participate in completion/failure evaluation or consume job restart budget.

I prefer Option A. It is less strict, fits better with the rest of the current design, and keeps more scenarios valid. I tend to think it may be problematic to disable gang-terminations for such leader-worker combinations, requiring the workers to always restart individually.

What do you think? What will better serve the applications running on top of this?

yankay:
> @yankay
> 
> > Thanks for the proposal. Could you help me understand the intended behavior when `completedNames` is satisfied while a non-required child fails in the same reconciliation? If the leader is `Completed` and a worker is `Failed`, should the replica complete, or should Grove restart the gang?
> 
> Good question. I think there are two coherent options here:
> 
> **Option A:** allow non-required completion-aware children, but define precedence clearly. A child not listed in `completedNames` can still fail the current gang attempt and trigger a gang restart while the required completion criterion is not yet satisfied. If the restart budget is exhausted before the required children complete, the replica fails. However, if `completedNames` is already satisfied in the same reconciliation snapshot, the replica completes and no restart budget is consumed for non-required children.
> 
> **Option B:** require children outside `completedNames` to be regular, non-completion-aware PCLQs. In this model, only children listed in `completedNames` may be completion-aware. Non-required children must be regular long-running PCLQs, so their failures do not participate in completion/failure evaluation or consume job restart budget.
> 
> I prefer Option A. It is less strict, fits better with the rest of the current design, and keeps more scenarios valid. I tend to think it may be problematic to disable gang-terminations for such leader-worker combinations, requiring the workers to always restart individually.
> 
> What do you think? What will better serve the applications running on top of this?

Thanks for the thoughtful reply. I agree that Option A seems like the better fit, since it preserves gang-level failure handling while supporting leader-driven completion.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+**Ordering guarantee.** In all cases, terminal conditions and updated `replicaRestartCounts` are persisted to status before any deletion begins. If the controller restarts mid-cleanup, it can resume from the persisted state without double-counting restarts or re-creating resources that were already deleted. Cleanup of active (non-terminal) pods when a resource reaches a terminal state is described in [Cleanup Behavior](#cleanup-behavior).
+
+**Gang scheduling on restart.** Recreated pods are placed by the scheduler as a complete gang, consistent with the initial launch behavior.
+
+### Condition and Status Model
+
+**New conditions.** Two terminal conditions are added across all three resource types:
+
+- `Completed` — all required pods or completion-aware children have succeeded.
```
sanjaychatterjee, line 327: Can we call this condition as `Succeeded`?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+Regular-mode `PodClique`s never set `Completed` or `Failed`.
+
+**PodCliqueScalingGroup**
+
+For a completion-aware `PodCliqueScalingGroup`, evaluation is two-level:
+
+1. *Replica state*: a replica is **Completed** when all required completion-aware child `PodClique`s are `Completed` — either all completion-aware children, or the named children in `policy.completion.success.completedNames` when it is set. A replica is permanently **Failed** when a required child `PodClique` fails and no restart budget remains (`maxRestarts` exhausted, or `maxRestarts: 0`). A failure of a completion-aware `PodClique` not listed in `completedNames` triggers a gang restart and consumes budget while budget remains; it does not mark the replica as `Failed` until no restart budget remains.
+
+2. *PCSG state*: the PCSG is **Completed** when all replicas are `Completed`. It is **Failed** when enough replicas have exhausted their budget that the remaining replicas cannot satisfy the completion criterion.
```
sanjaychatterjee: How does this work with `MinAvailable`? Or should the semantics be driven by `MinAvailable`, i.e. a PCSG is considered successfully completed when at least `MinAvailable` replicas have completed with success.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+**PodCliqueSet**
+
+For a completion-aware `PodCliqueSet`, evaluation follows the same two-level pattern as PCSG:
+
+1. *Replica state*: a replica is **Completed** when all required completion-aware direct children are `Completed` — either all completion-aware `PodClique`s and `PodCliqueScalingGroup`s, or the named children in `policy.completion.success.completedNames` when it is set. A replica is permanently **Failed** when a required direct child fails and no restart budget remains. A failure of a completion-aware direct child not listed in `completedNames` follows the same gang-restart behavior as PCSG: it consumes budget while budget remains, and fails the replica when no budget remains.
+
+2. *PCS state*: the PCS is **Completed** when all replicas are `Completed`. It is **Failed** when enough replicas have exhausted their budget that the remaining replicas cannot satisfy the completion criterion.
```
sanjaychatterjee: Again, there needs to be a consideration of `MinAvailable` number of replicas to complete with succeed for the PCS to be considered as succeeded.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+    // +optional
+    Success *CompletionSuccessPolicy `json:"success,omitempty"`
+
+    // Failure defines restart and failure behavior for this resource.
+    // +optional
+    Failure *CompletionFailurePolicy `json:"failure,omitempty"`
+}
+
+type CompletionSuccessPolicy struct {
+    // CompletedNames lists direct child names that must reach Completed for a
```
sanjaychatterjee:
```suggestion
    // CompletedNames lists direct child names that must complete with success within a
```

docs/proposals/0285-job-support/README.md
shayasoolin: I checked the closest APIs. My recommendation is to rename the success condition from `Completed` to `Complete`, but not `Succeeded`:
- Kubernetes Job - the terminal condition types are `Complete` and `Failed`.
- Kubernetes API conventions allow `Succeeded`, but that is generic guidance, the concrete batch Job API uses `Complete`.
- JobSet currently uses `Completed`, but that looks like an inconsistency: it also uses `Complete` in related places, and there is an open discussion about renaming `Completed` to `Complete`.
- Kubeflow is mixed: older Training Operator uses `Succeeded`, while newer Trainer uses `Complete`.
- KubeRay also distinguishes app-level job result (SUCCEEDED/FAILED) from Kubernetes deployment lifecycle (`Complete`/`Failed`).

I’d keep `spec.policy.completion` and `failure.maxRestarts`. I’d consider renaming `success.completedNames`, since it sounds like observed status. `success.requiredNames` or `success.targetNames` will be clearer.

What do you think? cc @nvrohanv

docs/proposals/0285-job-support/README.md
shayasoolin: What is the concern exactly with `minAvailable`? What kind of conflict or issue do you see?
For jobs, I expect `minAvailable` to be equal to `replicas`, but since `replicas` is the higher number, and we do not support partial completion at this first release (as agreed), I think the right thing to do is to follow `replicas` as the deciding value here. I don't see a semantic issue but maybe I'm missing something so please elaborate more.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+**Gang termination for completion-aware resources.** Gang scheduling is unchanged: `minAvailable` continues to gate pod launch until the full gang can be placed simultaneously, on initial start and after each restart. `minAvailable` also continues to be passed to scheduler backends as the gang's minimum member count. For gang termination, completion-aware resources use the `Failed` condition as the termination signal instead of `MinAvailableBreached`, and fire immediately without `terminationDelay`. A pod failure or eviction in a completion-aware `PodClique` moves the pod to `pod phase=Failed`, causing the owning `PodClique` to set the `Failed` condition and triggering this gang-termination path. Crucially, the termination scope is identical to the existing gang-termination implementation: the same gang boundary that would have been torn down by a `MinAvailableBreached` trigger is torn down by a `Failed` trigger. Completion-aware behavior changes the runtime trigger, not the scheduling contract or termination scope.
```
unmarshall: > A pod failure or eviction in a completion-aware `PodClique` moves the pod to `pod phase=Failed`, causing the owning `PodClique` to set the `Failed` condition and triggering this gang-termination path.

A pod eviction causing PCLQ to transition to FAILED state is a bit too harsh. Also this will then cause a restart/recreate subtracting from the restart/recreate budget. Slurm distinguishes infrastructure loss or scheduler preemption from application failure (container exiting with non-zero exit code). `JOB_NODE_FAIL` is a separate state from `JOB_FAILED`. `JobRequeue` requeues batch jobs on node failure/preemption instead of failing them. Automatic requeues are capped by `MaxBatchRequeue`.  Similarly if a job is preempted then its quite visible as `PREEMPTED` (never `FAILED`).  As far as i know preemption-driven restarts/reschedules are effectively uncapped in Slurm.

In k8s changes of node failures and preemption very much exists and therefore categorizing all of these under `FAILED` is IMO not correct.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+This distinction is important for partial-completion policies. For example, a `PodCliqueScalingGroup` replica may be considered `Completed` because the `PodClique`s listed in `completedNames` completed successfully, while other child `PodClique`s are still running. Once the replica is terminal, the PCSG controller deletes the non-terminal child `PodClique`s or active pods in that replica so they stop consuming resources. Similarly, once a `PodCliqueSet` replica or the whole PCS reaches a terminal state, the PCS controller cleans up active child `PodClique`s and `PodCliqueScalingGroup`s in that completed or failed scope.
+
+**On `PodClique` terminal state (`Completed` or `Failed`):**
+- Delete all active pods (`Pending`, `Running`) in the `PodClique`.
+- Retain terminal pods (`Succeeded`, `Failed`) for log access via `kubectl logs`, unless the `PodClique` is deleted as part of a gang restart.
+
+**On `PodCliqueScalingGroup` replica terminal state:**
+- Delete active pods and non-terminal child `PodClique`s belonging to that replica.
```
unmarshall, line 375: This is perhaps in line with the user story 3 that mentions that if the leader completes while workers are still running, you will start the cleanup. Slurm provides `WaitTime` and the default wait time is 0 (which means wait indefinitely). These workers are termed as `stragglers`. For training this is important. A worker receiving a SIGTERM in the middle of a checkpoint flush or a final `all_gather` produces corrupt artifacts. 

The only saving grace is `terminationGracePeriodSeconds` that are set on the Pod but this is always optional. So either make this explicit in the GREP or provide an explicit wait time.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+- Introduce `spec.policy.completion` on `PodClique` to signal completion-aware behavior, and on `PodCliqueScalingGroup` / `PodCliqueSet` to configure parent completion criteria and restart budgets.
+- Define `policy.completion.failure.maxRestarts` on `PodClique`, `PodCliqueScalingGroup`, and `PodCliqueSet`; `PodClique` supports only `0` in this release, while `PodCliqueScalingGroup` and `PodCliqueSet` use it as a per-replica gang restart budget.
+- Define `policy.completion.success.completedNames` on `PodCliqueScalingGroup` and `PodCliqueSet` to support named-child completion criteria.
+- Support gang restart for completion-aware parent scopes: when a required `PodClique` or `PodCliqueScalingGroup` fails, the parent deletes and recreates the affected scope as a unit, consuming from a per-replica restart budget.
+- Guarantee that terminal states are persisted to status before any pod cleanup, and that terminal pods (`Succeeded`, `Failed`) from final terminal scopes are retained for log access until the workload is deleted.
+
+### Non-Goals
```
unmarshall, line 54: Will you support for rolling updates for running Jobs?
If yes then how will this impact restart budgets and how will you do the updates? If no, then what happens when a PCS is changed, thus changing its hash? Will you reject the change in the webhook itself? There is no mention of that in this GREP. If you do not wish to support that then please mention this as a non-goal and explicitly state how will this prevented.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+    // +optional
+    Failure *CompletionFailurePolicy `json:"failure,omitempty"`
+}
+
+type CompletionSuccessPolicy struct {
+    // CompletedNames lists direct child names that must reach Completed for a
+    // PodCliqueScalingGroup or PodCliqueSet replica to be considered Completed.
+    // If omitted, all completion-aware direct children must complete.
+    // +optional
+    CompletedNames []string `json:"completedNames,omitempty"`
```
unmarshall: This is not an accurate name and IMO should be changed. `completedNames` is past tense and reads like  its an observed state when its actually part of the desired-state (spec) and signals intent.
**Alternatives:** : `targetNames`, `requiredNames`
I would recommend:
```yaml
  success:
    targetNames: [leader]
```
This reads as "success targets all of these names"

So tomorrow when you support  `Any` (that JobSet also supports) then it becomes:

```yaml
  success:
    targetNames: [leader, backup-leader]
    operator: Any
```
This reads as "success targets any of these names"

unmarshall (commented):
1/n review - will post the comments in iterations

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+- Introduce `spec.policy.completion` on `PodClique` to signal completion-aware behavior, and on `PodCliqueScalingGroup` / `PodCliqueSet` to configure parent completion criteria and restart budgets.
+- Define `policy.completion.failure.maxRestarts` on `PodClique`, `PodCliqueScalingGroup`, and `PodCliqueSet`; `PodClique` supports only `0` in this release, while `PodCliqueScalingGroup` and `PodCliqueSet` use it as a per-replica gang restart budget.
+- Define `policy.completion.success.completedNames` on `PodCliqueScalingGroup` and `PodCliqueSet` to support named-child completion criteria.
+- Support gang restart for completion-aware parent scopes: when a required `PodClique` or `PodCliqueScalingGroup` fails, the parent deletes and recreates the affected scope as a unit, consuming from a per-replica restart budget.
+- Guarantee that terminal states are persisted to status before any pod cleanup, and that terminal pods (`Succeeded`, `Failed`) from final terminal scopes are retained for log access until the workload is deleted.
```
unmarshall: >Guarantee that terminal states are persisted to status before any pod cleanup

There are case like API initiated eviction (node-drain, eviction sub resource) or scheduler pre-emption or a node loss (force-delete) or a taint-based eviction. In these case kubelet will not update Pod phase to Failed. So in these cases failure must be inferred from absence of the Pod and not from a pod in failed phase. This GREP's whole failure story is written around observing phase=Failed. For such pods the retention promise will break. Since this GREP also promises that terminal pods are retained for log access - this should be corrected. 

IMO, terminal pod retention is `best-effort` (not a guarantee) and must be documented clearly.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+- **No scaling for completion-aware workloads.** Completion-aware workloads use fixed replica counts. Grove rejects autoscaling configuration and manual replica changes for resources with `policy.completion` and for parent scopes that contain completion-aware direct children.
+- `completions` and `completedIndexes` — configurable completion counts and index-based filtering at the `PodCliqueScalingGroup` and `PodCliqueSet` levels. In this release, all replicas must complete successfully for a resource to be considered Completed.
+- Runtime deadline support (`maxRuntime`). Deadline semantics require a separate design for resource-level versus replica-level or attempt-level limits, and whether deadlines reset on gang restart. Since runtime deadlines are orthogonal to completion tracking and gang restart, they are deferred to keep the first release focused.
```
unmarshall: >Since runtime deadlines are orthogonal to completion tracking and gang restart

Choice of words is not accurate. Runtime deadline is separable but not orthogonal to `gang restart` but it is orthogonal  to `completion tracking`.

Why its not orthogonal to `gang restart` - upon gang restart do we reset the stopwatch? So the deadline could be per-attempt deadline or deadline across restarts. So you see it's not orthogonal.

Perhaps it can be reworded to below:
```
A runtime deadline does not affect how we detect completion. Completion is decided by pod exit codes, not by a clock. So we can add `maxRuntime` later without changing completion tracking.
A runtime deadline does interact with gang restart, but only in one way: when a gang restarts, we must decide whether the deadline clock resets or keeps running. This is a small, self-contained question. It does not change the core completion or restart logic. Because the core does not depend on it, and its only interaction with gang restart is limited and well-scoped, we defer maxRuntime to keep the first release focused.
```

docs/proposals/0285-job-support/README.md
unmarshall: >It is **Failed** when enough replicas have exhausted

what does `enough replicas` mean here? This is very vague. Either there is a configurable value or for this iteration it defaults to 1 (so even if 1 exhausts its restart budget, it's considered failed). Can you please be precise here?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+- Pod cleanup policies other than the fixed default (retain terminal pods from final terminal scopes, delete active pods on terminal state).
```
unmarshall: This sentence is not easy for me to read. Would it make sense to rephrase it?
My attempt at rewording:
```
Pod cleanup uses a single fixed policy: when a resource reaches a terminal state, still-active pods (Running/Pending) in that scope are deleted, while already-finished pods (Succeeded/Failed) in a final terminal scope are retained for log access. Configurable cleanup policies and TTL-based deletion are out of scope
```

Also no where in the GREP I could find the definition of `terminal scope`.  Also "final" in `final terminal scopes` is also undefined and is left open for user interpretation. 

My recommendation:
```
Scope: A unit that can independently reach a terminal state. e.g. `PodClique`, a `PodCliqueScalingGroup` replica,  `PodCliqueScalingGroup` resource, a `PodCliqueSet` replica or the `PodCliqueSet` resource.
Terminal Scope: A scope whose status is `Completed` or `Failed`.
Final Terminal Scope:  A terminal scope that will not be retried (success or failure with no restart budget remaining). A scope that is failed but still has a restart budget is `terminal-but-not-final` - it is deleted and recreated by gang restart, so its pods (including finished ones) are not retained.
```

unmarshall (commented):
2/n review. More will follow

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+The new fields are nested under `spec.policy.completion`. The API shape is:
+
+**PodCliqueSpec**, **PodCliqueScalingGroupSpec**, and **PodCliqueSetSpec** each gain:
+
+```go
+// Policy describes optional workload behavior controlled by Grove.
+// +optional
+Policy *WorkloadPolicy `json:"policy,omitempty"`
+
+type WorkloadPolicy struct {
```
unmarshall: First issue: `Workload` is IMO not accurate. 
Reasons:
1. Collision with `Workload` API (k8s scheduling API), Kueue's `Workload` API object (that is the unit of admission/queueing). This creates confusion.
2. `Workload` - binds to no specific level. The same type is embedded in PodCliqueSpec, PodCliqueScalingGroupSpec and PodCliqueSetSpec.  So now workload is an undefined concept that is smuggled in via a type.

Alternative: `RunPolicy` - kubeflow uses this and its quite accurate.

Second observation: one shared type across levels.
- `success.completedNames/targetNames` - invalid on PCLQ, valid on PCSG and PCS.
- failure.maxRestarts — valid everywhere, but on PodClique only 0 is legal.

So additional validations needs to be introduced to prevent setting fields inside WorkloadSpec at different levels. Creates a cognitive overload.

unmarshall: My recommendation that you could consider:

```go
type RunPolicy struct {
  // Completion configures completion-aware behaviour.
  // +optional
  Completion *CompletionPolicy `json:"completion,omitempty"`
  // (future: MaxRuntime, TTL, CleanupPolicy, ... live here)
}

type CompletionPolicy struct {
  // +optional
  Success *CompletionSuccessPolicy `json:"success,omitempty"`
  // +optional
  Failure *CompletionFailurePolicy `json:"failure,omitempty"
}

 // +kubebuilder:validation:XValidation:rule="self.type == 'NamedChildren' ? has(self.targetNames) :
 !has(self.targetNames)",message="targetNames must be set iff type is NamedChildren"
// +union
type CompletionSuccessPolicy struct {
   // +unionDiscriminator
   // +kubebuilder:default=AllChildren
   Type  SuccessType           `json:"type"`

   // TargetNames lists the direct children that must complete when type is NamedChildren.
   // +optional
   // +listType=atomic
   // +kubebuilder:validation:MinItems=1
   TargetNames []string `json:"targetNames,omitempty"`
}
```

Benefits: 
* Explicit `SuccessType`: With the flat completedNames, "field absent" silently overloaded to "all children must complete," which is ambiguous: did the author mean all, or did they forget to set it? The discriminator forces the choice to be stated (type: AllChildren vs type: NamedChildren)

It now reads as:
```yaml
  success:
    type: NamedChildren
    targetNames: [leader]
```
```yaml
  success:
    type: AllChildren
```

The `job mode` is currently decided by `presence of leaf`. First I could not really understand what is `leaf` as it's a term that is not even defined in this GREP. But then i finally understood that it only means that somewhere deep inside `completion` is set (at the PCLQ level).

That raises another API consumption question:
In the GREP the choice is made to force the user to specify:
```yaml
        policy:
          completion:
            failure:
              maxRestarts: 0
```
even though the default value of `maxRestarts` is 0 for PCLQ. This is not so nice. 
You should simply allow definition of an empty completion and use it as a marker which signals 2 things:
1. PCLQ is completion-aware
2. Default values must apply

So this allows:
```yaml
policy:
  completion: {}
```
to be a valid construct. I find this a LOT more intuitive and easy to reason than forcing the user to specify the entire defaults.

unmarshall (commented):
3/n review. More will follow

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+- A terminal parent scope is a stop condition for descendants: child controllers must not recreate pods or child resources when their owning `PodCliqueScalingGroup` replica, `PodCliqueSet` replica, or `PodCliqueSet` resource is already terminal.
+
+### Gang Restart Flow
+
+When a completion-aware `PodClique` reaches `Failed`, its completion-aware parent evaluates whether to restart or mark the replica as terminally failed.
+
+**PCLQ failure handled by completion-aware PCSG:**
+
+1. The PCSG increments `replicaRestartCounts[replicaIndex]`.
+2. If the budget is not exhausted: the PCSG deletes the failed `PodClique` and recreates it from the template. The new `PodClique` is placed by the scheduler as a complete gang before any pods run.
```
unmarshall: A PCSG can comprise of one more PCLQs. All the PCLQs are gang scheduled together. Unit of gang scheduling for a PCSG is a single PCSG replica. So your statement that a PCLQ in a PCSG is recreated and it will be gang scheduled is incorrect.

So if you only delete a single PCLQ and then it gets scheduled not-as-a-gang but as pods wherever there is capacity then you lose:
* Atomicity: You can only get partial placement while the surviving leader idles at a barrier. No all-or-nothing guarantee is applicable here.
* Topology: the whole reason for gang + topology-aware placement is co-locating the ranks. New pods scheduled against fragmented leftover capacity may land topologically far from the survivors, degrading the collective — exactly the scenario a training job restart is supposed to fix, not create

So you must delete all PCLQs for this PCSG replica.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+Two nested policy fields control completion-aware behavior:
+
+- **`policy.completion.success.completedNames`** *(PodCliqueScalingGroup / PodCliqueSet only)*: Named direct children within a replica that must complete for that replica to count as Completed. If omitted, all completion-aware direct children must complete. At the `PodClique` level, all pods must complete for the `PodClique` to be considered Completed.
+- **`policy.completion.failure.maxRestarts`** *(PodClique / PodCliqueScalingGroup / PodCliqueSet)*: Restart budget. On `PodCliqueScalingGroup` and `PodCliqueSet`, this is a per-replica gang restart budget. On `PodClique`, only `0` is supported in this release, meaning any pod failure makes the `PodClique` Failed.
+
+Regular-mode `PodClique`s within a `PodCliqueScalingGroup` or `PodCliqueSet` are excluded from completion evaluation. A resource can be Completed even if some of its children remain running in regular mode.
```
unmarshall, line 77: Is it possible to have a regular-mode PCLQ and a completion-aware PCLQ within the same PCSG? I do not see the GREP mentioning that it's disallowed. So for now I will assume that this is indeed allowed. There are now repurcursions.
So for example lets take a PCSG that has 2 PCLQs:
**pclq-a** : this is completion-aware
**pclq-b**: this is regular-mode

* If pclq-a fails, then to ensure gang-scheduling we must recreate the PCSG replica. This now impacts the pclq-b which is now swept with pclq-b. Is this intended? If yes what is the use case?
* GREP's own cleanup says "once the replica is terminal, the PCSG controller deletes the non-terminal child PodCliques or active pods in that replica." → the regular PCLQ is deleted.

My view:
* If two PCLQs are tightly coupled enough to require co-scheduling and topo-packing → they share a fate, so both should be finite (or both long-running).
* If they genuinely have independent lifecycles (one finite, one long-running) → they shouldn't be in the same gang in the first place; they should be separate PodGangs (e.g. a separate standalone PCLQ / separate PCSG) so their scheduling and lifecycle aren't fused.

So I am not aware of any use case where a long-running service is co-scheduled/gang-scheduled and topo-packed  with a finite training job in the same PCSG.

If you want to have - "a member that runs in the gang and shares its fate, but does not gate success" then the API already has provision for that via `targetNames`. Do not list it in there. That will express: `in the gang but does not gate success` coherently.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+#### Story 3: Leader-driven completion
+
+As a machine learning engineer running a leader-worker training job, I want the job to be considered complete when the leader pod exits successfully — regardless of whether worker pods have exited — so that I can use the leader's exit code as the authoritative signal of job success, while workers remain running until the leader finishes.
+
+### Limitations/Risks & Mitigations
+
+**Application-level hangs without pod failure.**
+For completion-aware resources, `MinAvailableBreached`-based gang termination is disabled. Ordinary pod failures and evictions are still handled: the pod reaches `pod phase=Failed`, the owning `PodClique` sets the `Failed` condition, and Grove terminates or restarts the gang through the failure path. The remaining risk is narrower: if the cluster does not surface a failed pod and the application keeps running despite a lost peer or broken collective, Grove cannot infer the application-level deadlock from availability alone. Workloads should use framework-level failure detection, such as rendezvous timeouts, and exit non-zero when peer loss makes progress impossible.
```
unmarshall: >`MinAvailableBreached`-based gang termination is disabled.

We need to be precise here. What does disabled mean here?
Does it mean that this condition is never set or does it mean that it is set but ignored?
If it is never set then it's fine. If its set but ignored, then it will have True during successful completion which is very misleading as its still set in the status and potentially risky if later code evolution opens a closed door which reacts on it.

docs/proposals/0285-job-support/README.md
unmarshall: > Crucially, the termination scope is identical to the existing gang-termination implementation: the same gang boundary that would have been torn down by a `MinAvailableBreached` trigger is torn down by a `Failed` trigger

Well, unfortunately that is really true. Today the gang-termination is only done at the PCS level. While this needs to be corrected your claim of `same gang boundary` is not really true today.

unmarshall (commented):
4/n review comments. More will follow

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+**Log loss on retry.**
+When a failed `PodClique` is deleted and recreated during a gang restart, terminal pods from the previous attempt are cascade-deleted along with it. Logs from failed attempts are not durably retained across retries. Mitigation: users who need per-attempt logs should rely on a cluster-level logging stack (e.g. Fluentd, Loki) rather than `kubectl logs`.
+
+## Design Details
+
+### Completion Policy
+
+Completion-aware behavior is rooted at the `PodClique` level. A `PodClique` becomes completion-aware by setting at least one concrete supported leaf under `spec.policy.completion`. Omitting `policy.completion` keeps the `PodClique` in regular mode.
```
unmarshall: > A `PodClique` becomes completion-aware by setting at least one concrete supported leaf under `spec.policy.completion`

Please define what a `leaf` means here. You have this term on more than one occasion but not really defined it clearly.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+- Guarantee that terminal states are persisted to status before any pod cleanup, and that terminal pods (`Succeeded`, `Failed`) from final terminal scopes are retained for log access until the workload is deleted.
+
+### Non-Goals
+
+- **No pod-level retry.** A failed pod within a `PodClique` is not replaced in isolation. `PodClique` `policy.completion.failure.maxRestarts` may only be `0` in this release; values greater than `0` are future work. This is acceptable for the first release because current distributed training frameworks generally do not tolerate replacing a single failed worker independently. A single worker failure usually requires restarting the whole group, so gang restart covers the common recovery path.
+- **No scaling for completion-aware workloads.** Completion-aware workloads use fixed replica counts. Grove rejects autoscaling configuration and manual replica changes for resources with `policy.completion` and for parent scopes that contain completion-aware direct children.
```
unmarshall: Can you also clarify that as a consequence of this Grove at least in this iteration will not support elastic training. e.g. `PyTorch elastic (torchelastic)`, Elastic horovod, Ray train. Kubeflow PytorchJob exposes `elasticPolicy`

docs/proposals/0285-job-support/README.md
unmarshall: Let's assume a PCSG has a completion-aware PCLQ and a regular-mode PCLQ.
If `MinAvailableBreached` condition is disabled for this PCS - then what happens when the regular-mode PCLQ goes below minAvailable? This PCLQ also will never go to `Failed` state since its `restartPolicy` is set to `Always`.

 - If MinAvailableBreached is disabled scope-wide → a regular PCLQ can lose all its pods and the gang silently ignores it. If the completion-aware siblings depend on that helper (and they're co-scheduled in one gang, so they're coupled), the job keeps running against a dead dependency with no signal. This is quite unexpected.
 - If you keep MinAvailableBreached alive only for regular mode PCLQs then now you have 2 termination triggers that coexist together in one scope. Now if due to MinAvailableBreached the PCSG replica is recreated will this count towards maxRestarts?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+2. *PCSG state*: the PCSG is **Completed** when all replicas are `Completed`. It is **Failed** when enough replicas have exhausted their budget that the remaining replicas cannot satisfy the completion criterion.
+
+**PodCliqueSet**
+
+For a completion-aware `PodCliqueSet`, evaluation follows the same two-level pattern as PCSG:
+
+1. *Replica state*: a replica is **Completed** when all required completion-aware direct children are `Completed` — either all completion-aware `PodClique`s and `PodCliqueScalingGroup`s, or the named children in `policy.completion.success.completedNames` when it is set. A replica is permanently **Failed** when a required direct child fails and no restart budget remains. A failure of a completion-aware direct child not listed in `completedNames` follows the same gang-restart behavior as PCSG: it consumes budget while budget remains, and fails the replica when no budget remains.
```
unmarshall: >A failure of a completion-aware direct child not listed in `completedNames` follows the same gang-restart behavior as PCSG: it consumes budget while budget remains, and fails the replica when no budget remains.

If the PCLQs listed in the `targetNames/completedNames` are satisfied but a completion-aware PCLQ that is not part of `targetNames/completedNames`  fails then why should it fail the replica if the restart budget has expired? So even if the success gate is met it still fails the replica. This conflicts with "leader driven user story" - So if the leader completes successfully but one or more workers fails the will this now fail the replica?

unmarshall (commented):
5/n review comments. More to come.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+**PCLQ failure handled by completion-aware PCSG:**
+
+1. The PCSG increments `replicaRestartCounts[replicaIndex]`.
+2. If the budget is not exhausted: the PCSG deletes the failed `PodClique` and recreates it from the template. The new `PodClique` is placed by the scheduler as a complete gang before any pods run.
+3. If the budget is exhausted: the PCSG marks that replica as failed and re-evaluates its own terminal conditions.
+
+**PCLQ or PCSG failure handled by completion-aware PCS:**
+
+When a constituent completion-aware `PodClique` or `PodCliqueScalingGroup` within a PCS replica fails, the PCS treats it as a gang-level failure for the whole replica:
```
unmarshall, line 313: the usage of the term `gang-level` failure for the whole PCS replica is incorrect. There is no concept of a Gang at the PCS replica level.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+1. The PCS increments `replicaRestartCounts[replicaIndex]`.
+2. If the budget is not exhausted: the PCS deletes all constituents of that replica (all `PodClique`s and `PodCliqueScalingGroup`s) and recreates them together from the template.
+3. If the budget is exhausted: the PCS marks that replica as failed and re-evaluates its own terminal conditions.
+
+**Ordering guarantee.** In all cases, terminal conditions and updated `replicaRestartCounts` are persisted to status before any deletion begins. If the controller restarts mid-cleanup, it can resume from the persisted state without double-counting restarts or re-creating resources that were already deleted. Cleanup of active (non-terminal) pods when a resource reaches a terminal state is described in [Cleanup Behavior](#cleanup-behavior).
```
unmarshall: > terminal conditions and updated `replicaRestartCounts` are persisted to status before any deletion begins

Persisting the updated replicaRestartCount will prevent double counting but it does nothing to tell Grove whether the  delete + recreated work is pending or it's already done. If the status update is done with the increased replicaRestartCount and Grove crashes. When it comes back up Grove cannot know whether there is a pending delete + recreate.

So the count alone is insufficient.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+**Gang scheduling on restart.** Recreated pods are placed by the scheduler as a complete gang, consistent with the initial launch behavior.
```
unmarshall, line 321: The same issue - if PCLQ pods are recreated in isolation then its pods will not be gang scheduled.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+### Condition and Status Model
+
+**New conditions.** Two terminal conditions are added across all three resource types:
+
+- `Completed` — all required pods or completion-aware children have succeeded.
+- `Failed` — enough pods or children have failed that the completion criterion is permanently unreachable. Once set, this condition is irreversible.
```
unmarshall: `enough pods` - vague - what is enough?

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+These conditions are represented in the standard `conditions` list on each resource's status. Regular-mode resources never set these conditions. Grove does not introduce `Pending` or `Running` states for this feature.
+
+Terminal conditions are not propagated top-down to children that did not independently complete or fail. For example, if a `PodCliqueScalingGroup` replica completes because the `PodClique`s listed in `completedNames` completed successfully, any other child `PodClique`s may not have a `Completed` or `Failed` condition. Instead, the parent terminal state makes them no longer desired, and top-down cleanup removes their active pods and non-terminal child resources.
```
unmarshall: > if a `PodCliqueScalingGroup` replica completes because the `PodClique`s listed in `completedNames` completed successfully, any other child `PodClique`s may not have a `Completed` or `Failed` condition. Instead, the parent terminal state makes them no longer desired,

I have raised a similar issue before where you have mentioned that if a PCLQ that is not part of `targetNames/completedNames` fails then it will PCSG replica will be considered fail. But here this is reversed. So when you make the correction there, perhaps this will then be harmonised.

unmarshall (commented):
final review.

docs/proposals/0285-job-support/README.md
sanjaychatterjee: Agree with this API suggestion.

docs/proposals/0285-job-support/README.md
sanjaychatterjee: Can we assume in the job mode these components run to completion?

docs/proposals/0285-job-support/README.md
shayasoolin: It's intentional. As described in the GREP, the majority/all of the distributed training frameworks cannot tolerate a single worker's failure. They expect the whole group to be restarted, and re-sync with the leader. Please correct me if this is a wrong or inaccurate observation. Therefore, due to this application-level expectation, we prefer to treat the whole PCLQ as failed rather than restart a single failed or evicted pod.
This is also not the final word here. We will be able to add a more nuanced restart policy for pods within PCLQ in later releases.
If it's important for this first release, we can consider having a dedicated phase for preemption, which is different than failure, but still keep the consequences of it as a failed PCLQ. Otherwise it's a bigger change while we really prefer to keep this first release minimal.

docs/proposals/0285-job-support/README.md
shayasoolin: The answer is in the second half of the sentence you've quoted:
> that the remaining replicas cannot satisfy the completion criterion

Will be a bit clearer if I use "such that" instead of "that".

docs/proposals/0285-job-support/README.md
shayasoolin: @sanjaychatterjee I'm surprised you agree, as we had a very specific meeting exactly about that, and agreed with @Ronkahn21 and @nvrohanv that an empty completion section `{}` is a poor signal for a "job-mode resource". You preferred (and I agreed) that the user will explicitly specify at least one member field e.g. `maxRestarts`, even that it has a default value.

Second point, I don't see the value in `success.type`. The default is an empty `targetNames` field, which means ALL names should complete. If a subset is set, then this is the subset. I think it's cleaner and easier. It also saves the case of specifying a list of children but using `AllChildren` as the type.

docs/proposals/0285-job-support/README.md
shayasoolin: I see that the phrasing can be better and I'll correct it. 
As stated elsewhere in the GREP, a PCS/G is in "completion aware" if ANY of its children is. 
A failed PCLQ within a PCSG will cause the whole PCSG replica to restart - just as you'd expect.

docs/proposals/0285-job-support/README.md
shayasoolin: Please see the previous reply:

As stated elsewhere in the GREP, a PCS/G is in "completion aware" if ANY of its children is.
A failed PCLQ within a PCSG will cause the whole PCSG replica to restart - just as you'd expect.

docs/proposals/0285-job-support/README.md
shayasoolin: The condition will not be set. I'll clarify that in the doc.

docs/proposals/0285-job-support/README.md
shayasoolin: It should be `field`, sorry, I'll correct that.

docs/proposals/0285-job-support/README.md
shayasoolin: Can you clarify? Pods are not recreated in isolation - any single pod failure takes the whole PCLQ down (together with the rest of the PCS/PCSG replica).

docs/proposals/0285-job-support/README.md
shayasoolin: Duplicate comment? The definition of "enough" is in the second half of the sentence. Resource can no longer reach completion.

shayasoolin:
> Thanks for the thoughtful reply. I agree that Option A seems like the better fit, since it preserves gang-level failure handling while supporting leader-driven completion.

Clarified in f2afa6b. Thanks

docs/proposals/0285-job-support/README.md
shayasoolin: Clarified in cdc54ad.

docs/proposals/0285-job-support/README.md
shayasoolin: Not supported.
Clarified in ea281e9:
  - Added rolling updates as a non-goal.
  - Stated that the validation webhook rejects edits that would update running job pods or change generated child-resource hashes.

docs/proposals/0285-job-support/README.md
shayasoolin: Agree and this indeed was the plan following another discussion. Renamed `completedNames` to `targetNames`, see 6e362e0.

docs/proposals/0285-job-support/README.md
shayasoolin: Updated GREP to use `RunPolicy` instead of `WorkloadPolicy`, with separate types:

  - `PodCliqueRunPolicy` for `PodClique`
  - `RunPolicy` for `PodCliqueScalingGroup` and `PodCliqueSet`

See 916affe.

docs/proposals/0285-job-support/README.md
shayasoolin: Done in 7b34ee5.

docs/proposals/0285-job-support/README.md
shayasoolin: Done in 65af329.

docs/proposals/0285-job-support/README.md
shayasoolin: Got it, thanks, see 41df52f.

docs/proposals/0285-job-support/README.md
shayasoolin: Done in 9053c09.

docs/proposals/0285-job-support/README.md
shayasoolin: Thanks, indeed we previously assumed that an evicted/failed node will lead to a failed pod, but this is not necessarily the case. cc @sanjaychatterjee 
Fixed in 322106c.

docs/proposals/0285-job-support/README.md
shayasoolin: Ok. See 8e4cb62.

docs/proposals/0285-job-support/README.md
shayasoolin: Revised. I removed the "final terminal" wording everywhere, it doesn't sound right.
See a05504f.

docs/proposals/0285-job-support/README.md
shayasoolin: Done 2457914.

docs/proposals/0285-job-support/README.md
shayasoolin: Please see my discussion with @yankay [in this PR](https://github.com/ai-dynamo/grove/pull/696#issuecomment-5323310560) and check out the updated wording of this section.

shayasoolin: See f2afa6b3b18ca2a17b001e3507bafc5ec7431ee0.

docs/proposals/0285-job-support/README.md
unmarshall: >@sanjaychatterjee I'm surprised you agree, as we had a very specific meeting exactly about that, and agreed with @Ronkahn21 and @nvrohanv that an empty completion section {} is a poor signal for a "job-mode resource".

I am not sure why is that considered a poor choice. I can quote some examples in k8s space:
- emptyDir: {} — VolumeSource.EmptyDir *EmptyDirVolumeSource. Every field in EmptyDirVolumeSource is optional, so emptyDir: {} is the idiomatic way to request the volume with defaults. Presence selects the type.
- rollingUpdate: {} — DeploymentStrategy.RollingUpdate *RollingUpdateDeployment (same for StatefulSet/DaemonSet). Both maxUnavailable and maxSurge are optional and default to 25%, so an empty block is valid and simply means "rolling update with defaults."
- securityContext: {} — PodSpec.SecurityContext *PodSecurityContext; all fields optional, presence-based.
- podFailurePolicy / successPolicy on batch/v1 Job — optional struct pointers whose presence turns on that evaluation path.

Maybe there are more such cases.

>You preferred (and I agreed) that the user will explicitly specify at least one member field e.g. maxRestarts, even that it has a default value.

That is not a clean design IMHO.  Completion is a pointer field.
If it is nil for a component then its not completion-aware.
if it is not-nil then it is completion aware.

Now given this - If the user does not really wish to configure any values and is happy with the defaults then he should not be asked to explicitly re-mention the defaults. An empty struct != nil and thus makes the component completion aware.

docs/proposals/0285-job-support/README.md
shayasoolin: Following this comment as well as [this one](//github.com/ai-dynamo/grove/pull/696#discussion_r3915496376), I'm thinking to refrain from using "gang-termination" at all, but replace it with "failure handling". I feel that "gang-termination" is already a loaded term, plus it's not fully implemented or working as planned. WDYT?

docs/proposals/0285-job-support/README.md
unmarshall: >Second point, I don't see the value in success.type. The default is an empty targetNames field, which means ALL names should complete. If a subset is set, then this is the subset. I think it's cleaner and easier. It also saves the case of specifying a list of children but using AllChildren as the type.

Sure we can go with this. If in future you have a need for specifying % of workers to succeed then you can add another field `MinSucceeded`. You can then model `JobSet`s `Any` operator by using the int value of (1) and if there is a need can also specify that at least 50% of the workers should succeed. I think this might only be needed when there is elastic training workloads, right?

shayasoolin: Re `completion: {}` - okay, I agree with that. I wasn't familiar with the common usage of this pattern. That changes the picture for me.
Glad we agree to defer `successType`. It's relevant for elastic training but not only. We have a customer use-case of MoE training where they define a threshold of replicas that need to succeed (more precisely, an index range). It's not exactly a list of names but it goes along the same lines.

docs/proposals/0285-job-support/README.md
shayasoolin: I agree that `replicaRestartCounts` alone is not enough.
I suggest to encode the restart attempt on recreated child resources.

For example, after `replicaRestartCounts[replicaIndex]` is incremented to `1`, Grove creates the resources for that replica with an attempt label, e.g. `grove.io/restart-attempt: "1"`.

On reconcile, Grove compares the desired attempt from the parent status with the attempt labels on existing child resources:
- children still have attempt `0` → the restart is still pending, so Grove should delete/recreate them
- children have attempt `1` → the restart already happened, so Grove should not consume budget again

This keeps the parent status relatively small and fits the usual reconciliation model: desired attempt is stored on the parent, realized attempt is visible on the children.

An alternative is to add an explicit restart operation state to the parent status, for example `replicaRestarts[]` with `replicaIndex`, `attempt`, and `phase: Deleting|Recreating`. That is more explicit and easier to debug, but it adds more status/API surface and another state machine to maintain.

docs/proposals/0285-job-support/README.md
shayasoolin: Thanks, good points, my POV here:
1. We have to support those mixed regular/completion-aware PCS/PCSGs. The main use-cases have things like parameter-server or router etc. that do not run to completion but are part of the training/RL composite.
2. It feels wrong to disable `MinAvailableBreached` on regular PCLQs, exactly for the reasons you mentioned.
3. It's hard to decide whether to consume `MaxRestarts` upon gang-termination or not.

**I tend NOT to consume. Why?**
a. `MaxRestarts` refers to failure-triggered restarts. That's what this budget is about. When provided by the user, the user doesn't expect it to count for restarts due to PCLQs that are not part of the job itself.
b. If we consume `MaxRestarts` for gang-terminations, why not do that also for regular PCS that has no completion-aware children at all? Then this field turns to be a generic one, not just for completion-aware, and this takes us to a different journey, as now we assume that everything under `completion.failure` is related to completion-aware resources only.

Done in d8d01c8.

docs/proposals/0285-job-support/README.md
```diff
@@ -0,0 +1,478 @@
...
+
+**Gang termination for completion-aware resources.** Gang scheduling is unchanged: `minAvailable` continues to gate pod launch until the full gang can be placed simultaneously, on initial start and after each restart. `minAvailable` also continues to be passed to scheduler backends as the gang's minimum member count. Completion-aware failures do not use `MinAvailableBreached`; the `Failed` condition is the termination signal, and the parent reacts immediately without waiting for `terminationDelay`. Regular-mode direct children keep `MinAvailableBreached`. If a regular-mode child breaches `minAvailable` in a mixed parent, Grove handles it through availability-based recovery, and that action does not increment `replicaRestartCounts` or consume `policy.completion.failure.maxRestarts`. A failed pod in a completion-aware `PodClique` causes the owning `PodClique` to set the `Failed` condition. The completion-aware parent then evaluates the affected replica: if restart budget remains, Grove deletes and recreates the replica as a gang; if the budget is exhausted, Grove marks the replica failed and re-evaluates the parent terminal conditions.
```
unmarshall, line 79: > Completion-aware failures do not use MinAvailableBreached; the Failed condition is the termination signal, and the parent reacts immediately without waiting for terminationDelay. Regular-mode direct children keep MinAvailableBreached

So the non-completion-aware PCLQ is independently recovered while the job runs, but terminated the moment the job finishes. What is not clear is the statement `Regular-mode direct children keep MinAvailableBreached` - does this mean that it can result in termination of the entire PCS replica which also has completion-aware PCLQs?

docs/proposals/0285-job-support/README.md
unmarshall: Thanks, the mechanics are now clearly written and `targetNames` reads much better. But I don't think this resolves the original concern - it defines how a mixed gang behaves without establishing why it should be allowed.

Two things I'd like nailed down before this is settled:

1. **Use case:** What concrete workload co-schedules a finite job and a genuinely long-running service in the same PCSG/gang? Story 1 is "jobs alongside inference," which is separate workloads, not one mixed gang. Without a real scenario it becomes difficult to even comprehend and hence evaluate the design choices for such a combination.
2. **A regular-mode child in a completion-aware parent isn't actually long-running:** Per the current text it is deleted+recreated when a completion-aware sibling triggers a gang restart. In another place you mention `deleted by top-down cleanup when the parent completes while it is still running`.  So its lifecycle is fused to the job yet you also treat it as independently recoverable via MinAvailableBreached condition.

docs/proposals/0285-job-support/README.md
unmarshall: The next task on my list is to fix gang termination, post which you might be able to use it correctly. But here you seem to be talking about a Gang at the level of PCS replica. A PCS replica can have one or more PodGangs but there is no one single gang representing the whole PCS replica. So the wording needs to be corrected

docs/proposals/0285-job-support/README.md
unmarshall: The isolated-recreation problem is resolved now that both PCSG and PCS restart recreate the full replica. One correction on "placed by the scheduler as a complete gang": a PCS replica isn't a single gang. Standalone PCLQs and PCSG replicas [0, minAvailable) are in the anchor PodGang, but PCSG replicas above minAvailable are separate PodGangs. For a PCS replica with a scaled PCSG, restart reproduces multiple PodGangs, not one. Suggest rewording to "placed with the same gang boundaries (PodGangs) as the initial launch" to avoid implying a single gang.

docs/proposals/0285-job-support/README.md
unmarshall: Thanks for making that explicit. 
`terminationGracePeriodSeconds` maps only to Slurm's `KillWait` which is the last resort grace period before SIGKILL. 
Slurm handles the "leader finished, workers still running" case with two distinct, layered timers:
- **WaitTime:** how long to wait after the first task terminates before terminating all remaining tasks. During this window the stragglers run undisturbed - no signal is sent - so a worker can finish its checkpoint flush naturally and exit 0.
- **KillWait:** once Slurm has actually decided to terminate a task, it sends SIGTERM, waits KillWait, then SIGKILL. This is the grace window after the process has been told to stop.

SIGTERM itself is the interruption. The moment the replica is terminal, Grove deletes the worker Pod and kubelet sends a SIGTERM. A well behaved worker Pod treats a SIGTERM as "stop and cleanup". It is a race against a fixed clock, not "until done". WaitTime=0 waits however long the work naturally takes. A grace period is always a guess, and underestimating is a silent SIGKILL mid-flush

So either we add a minimal grove-level straggler wait even for MVP or explicitly scope this to workers that are safe to kill. We explicitly mention that a first-class straggler wait is a non-goal or future work.

docs/proposals/0285-job-support/README.md
unmarshall: Adding a label is fine and is crash-safe. One correction to the rule as you suggest: `children have attempt 1 → the restart already happened, so Grove should not consume budget again` -  this is only correct if the attempt-1 child has not failed again. The trigger for consuming budget can't be label == count. It has to be "a child at the current attempt has failed. So:
- Increment only when a Failed child's attempt label equals the current count (the current attempt failed) - then recreate at count+1
- A child whose attempt label is less than the count is pending recreation from an already-counted increment - recreate it at count, do not increment.

Adding a phase would not really help. Post crash the phase could be stale and you must still re-derive the state from the existing/actual children. So at best its a debugging aid which costs an additional status field and you have stated an unnecessary state machine.

Perhaps adding it as an annotation would be more appropriate than a label (these are indexed) since you are not really going to do any selection based on this label.

docs/proposals/0285-job-support/README.md
unmarshall: > If a regular-mode child breaches minAvailable in a mixed parent, Grove handles it through availability-based recovery, and that action does not increment replicaRestartCounts or consume maxRestarts

So there are two triggers in one gang with two different accountings:

1. completion-aware PCLQ failed. This will delete + recreate the replica as a Gang and it will consume `maxRestarts` budget.
2. regular-mode PCLQ (uses MinAvailableBreached). This will be handled via `availability-based-recovery` and it does not consume any budget and is therefore as a consequence `unbounded`.

The completion-aware path explicitly ties "recreate the replica as a gang" (which deletes the regular-mode child too) to consuming budget. If the regular-mode path's "availability-based recovery" also recreates the whole replica gang which is how `MinAvailableBreached` gang-termination works today, and the GREP doesn't say otherwise then it's the same action (tear down and re-place the whole replica, completion-aware siblings included).

So a flaky regular helper can recreate the replica gang (destroying the job's in-progress work each time) without limit, while the job's own failures are capped. So the open question is does regular-mode availability recovery recreate the whole replica gang or just that child?

docs/proposals/0285-job-support/README.md
unmarshall: The "such that" rephrase helps the logic, but it still doesn't answer the precise question and reads like a tunable threshold. Since this release requires all replicas to complete (partial completion via completions/completedIndexes is out of scope), the threshold is always exactly 1 - the PCSG/PCS fails the moment a single replica exhausts its restart budget, immediately. Please state that explicitly. You can keep the general "cannot satisfy the completion criterion" wording as the forward-looking form, but add that it currently reduces to 1 so it doesn't read as configurable.

docs/proposals/0285-job-support/README.md
unmarshall: I think this is answering a different question. My concern wasn't the restart granularity - whole-gang restart for distributed training is fine and I'm not disputing it. The concern is classification and budget accounting: an infrastructure-caused termination (preemption, API eviction, node loss) is treated as Failed and subtracts from the finite maxRestarts budget.

A maxRestarts: 3 job on preemptible/spot nodes can be permanently failed after 3 preemptions with zero application failures. Slurm avoids this deliberately: JOB_NODE_FAIL/PREEMPTED requeues are separate and effectively uncapped, distinct from JOB_FAILED retries.

This also isn't a large change - Kubernetes already exposes the distinction and Job already acts on it. Pods terminated by disruption get a DisruptionTarget condition (reasons PreemptionByScheduler, EvictionByEvictionAPI, DeletionByTaintManager, DeletionByPodGC), and Job's podFailurePolicy matches onPodConditions: DisruptionTarget to Ignore them for backoffLimit, versus onExitCodes for real failures.

So i would consider the following:
- A pod that reaches a non-Succeeded terminal state due to DisruptionTarget should not consume maxRestarts. It still triggers a gang restart but does not consume from the failure budget, or at most consume a separate, uncapped/larger disruption budget (which you can also introduce later).
- An application failure (phase=Failed with a non-zero container exit and no DisruptionTarget) consumes maxRestarts as today.

If that's genuinely out of scope for the first release, then at minimum this needs to be called out as an explicit limitation/risk: infrastructure churn consumes the finite restart budget, so a job on preemptible infrastructure can fail without any application error. Users need to know that before relying on maxRestarts.

PR #698, opened 2026-07-01
fix(podclique): prevent rolling update from deleting all serving replicas

### LINKED ISSUE #697: Rolling update can delete all PodClique replicas in a single reconcile, violating minAvailable
## Summary

A `PodCliqueSet` spec change (generation-hash flip) can cause the rolling update to delete **every** replica of a `PodClique` in one reconcile, driving ready replicas to 0 and violating `minAvailable`. Observed in production: a `replicas: 2`, `minAvailable: 1` PodClique lost both serving pods simultaneously → full outage until a replacement became ready (~2.5 min).

Code references are against `main` @ `21c24c6`.

## Impact

- `minAvailable` is not honored during rolling updates under the conditions below.
- All replicas of a role can be terminated at once → service outage, not a rolling update.

## Root cause

Two independent defects in the podclique rolling-update path combine.

### 1. A currently-Ready serving pod is misclassified as non-ready and hard-deleted

`computeUpdateWork` (`operator/internal/controller/podclique/components/pod/rollingupdate.go`) evaluates the unhealthy predicate **before** readiness:

```go
switch {
case k8sutils.IsPodPending(pod):
    work.oldTemplateHashPendingPods = append(...)
case k8sutils.HasAnyStartedButNotReadyContainer(pod) || k8sutils.HasAnyContainerExitedErroneously(logger, pod): // checked first
    work.oldTemplateHashUnhealthyPods = append(...)
case k8sutils.IsPodReady(pod):
    work.oldTemplateHashReadyPods = append(...)
...
}
```

`HasAnyContainerExitedErroneously` inspects `ContainerStatus.LastTerminationState` (`operator/internal/utils/kubernetes/pod.go`, `GetContainerStatusIfTerminatedErroneously`), which stays set after any prior non-zero container exit (e.g. an earlier OOM/liveness kill, exit 137) until the container restarts and stabilizes. It is **not** a statement about current readiness.

Consequently a pod that is **currently Ready and serving**, but carries a stale non-zero `LastTerminationState` from an earlier restart, lands in `oldTemplateHashUnhealthyPods`. `deleteOldNonReadyPods` then deletes it immediately, with **no `minAvailable` guard**, out-of-band from the one-at-a-time rolling path.

This diverges from the status-side categorization `CategorizePodsByConditionType` (`operator/internal/utils/kubernetes/pod.go`), which checks `IsPodReady` **first** and therefore counts the same pod in `Status.ReadyReplicas` (`reconcilestatus.go`, `ReadyReplicas = len(podCategories[corev1.PodReady])`). So the two code paths disagree about whether the pod is ready.

### 2. Off-by-one in the minAvailable guard

`processPendingUpdates` (`rollingupdate.go`):

```go
if pclq.Status.ReadyReplicas < *pclq.Spec.MinAvailable {
    return requeue
}
nextPodToUpdate = updateWork.getNextPodToUpdate() // deletes one Ready pod
```

Selecting the next pod deletes one Ready pod, so preserving `minAvailable` afterwards requires `ReadyReplicas - 1 >= MinAvailable`. The condition uses `<`, which permits deleting the last Ready pod when `ReadyReplicas == MinAvailable`, dropping ready below `minAvailable`.

## Reproduction

PodClique with `replicas: 2`, `minAvailable: 1`, both pods Ready and serving, where one pod carries a stale non-zero `LastTerminationState` (any prior container restart with a non-zero exit code). Trigger a rolling update (spec / PCS generation-hash change). In a single reconcile:

1. `deleteOldNonReadyPods` deletes the Ready-but-previously-errored pod (misclassified as unhealthy — defect 1).
2. Guard `ReadyReplicas(2) < MinAvailable(1)` is false → proceeds and deletes the other Ready pod.

Result: both pods deleted, 0 ready.

Defect 2 also fires independently when `replicas > minAvailable` and exactly `minAvailable` pods are Ready while the rest are genuinely not-ready: the not-ready pods are deleted, then the last Ready pod is deleted, breaching `minAvailable`.

## Evidence (production incident)

- Operator log during update: `container previously exited with a non-zero exit code`, `exitCode: 137` for the pod that was still serving.
- Both replicas served steady traffic until update start, then dropped to 0 QPS together; kubelet issued `Killing container … gracePeriod=60` for both pods within ~30 ms.
- `PodClique` status showed `MinAvailableBreached` and a 0-ready window matching `rollingUpdateProgress` start/end.

## Suggested fix

1. In `computeUpdateWork`, check `IsPodReady` **before** the unhealthy predicates, aligning with `CategorizePodsByConditionType` so a currently-Ready pod is always routed to the protected one-at-a-time rolling path.
2. Change the guard from `ReadyReplicas < MinAvailable` to `ReadyReplicas <= MinAvailable`.

Note: with fix (2), `minAvailable == replicas` (no surge) correctly blocks the rolling update instead of breaching. Proper surge-based updates (create-before-delete) are a separate, larger change.

I have a fix with unit + full-flow regression tests (reproduces the incident and the guard case) ready and will open a PR referencing this issue.

_No response_

### Environment

- Kubernetes version: 1.35
- Grove version - 0.1.0-alpha8
- Scheduler details 
- Cloud provider or hardware configuration - B200s 
- Tools that you are using Grove together with - Dynamo Platform

uthark:
A PodCliqueSet spec change triggered a rolling update that deleted both replicas of a PodClique at once, causing a full outage despite minAvailable=1.

Two independent defects:

1. computeUpdateWork classified a currently-Ready, serving pod as "unhealthy" whenever it carried a stale non-zero LastTerminationState from an earlier restart, because the erroneous-exit predicate was checked before IsPodReady. Such a pod was then hard-deleted by deleteOldNonReadyPods, out-of-band from the minAvailable-protected rolling path. This diverged from the status-side categorization in CategorizePodsByConditionType (which counts it in ReadyReplicas). Fix: check IsPodReady before the unhealthy predicates.

2. The minAvailable guard used ReadyReplicas < MinAvailable, an off-by-one that permitted deleting the last Ready pod when ReadyReplicas == MinAvailable. Fix: use <=.

Adds one unit test (categorization) and two full-flow tests (incident reproduction + minAvailable guard).

#### What type of PR is this?

/kind bug

Fixes https://github.com/ai-dynamo/grove/issues/697

```release-note

```

```docs

```

### DIFF
--- a/operator/internal/controller/podclique/components/pod/rollingupdate.go
+++ b/operator/internal/controller/podclique/components/pod/rollingupdate.go
@@ -94,11 +94,15 @@ func (r _resource) processPendingUpdates(logger logr.Logger, sc *syncContext) er
 	// In either of these cases we should pick up next pod to update if there are any pending pods to update.
 	var nextPodToUpdate *corev1.Pod
 	if podNamesPendingUpdate := updateWork.getPodNamesPendingUpdate(r.expectationsStore.GetDeleteExpectations(sc.pclqExpectationsStoreKey)); len(podNamesPendingUpdate) > 0 {
-		if pclq.Status.ReadyReplicas < *pclq.Spec.MinAvailable {
+		// Selecting the next pod deletes one Ready pod. To keep at least MinAvailable Ready pods
+		// afterwards we need ReadyReplicas-1 >= MinAvailable, i.e. block while ReadyReplicas <= MinAvailable.
+		// Using '<' here is off-by-one: it permits deleting the last Ready pod when ReadyReplicas == MinAvailable,
+		// dropping Ready to below MinAvailable (contributed to the harbinger-24b 0-ready outage window).
+		if pclq.Status.ReadyReplicas <= *pclq.Spec.MinAvailable {
 			return groveerr.New(
 				groveerr.ErrCodeContinueReconcileAndRequeue,
 				component.OperationSync,
-				fmt.Sprintf("ready replicas %d lesser than minAvailable %d, requeuing", pclq.Status.ReadyReplicas, *pclq.Spec.MinAvailable),
+				fmt.Sprintf("ready replicas %d not greater than minAvailable %d, requeuing", pclq.Status.ReadyReplicas, *pclq.Spec.MinAvailable),
 			)
 		}
 		nextPodToUpdate = updateWork.getNextPodToUpdate()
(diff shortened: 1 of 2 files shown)


uthark:
I found some other issue (not related to the failed tests) - if pod had any restarts it is not being properly update if there was a spec update on a podclique, will update PR with a fix.

operator/internal/controller/podclique/components/pod/rollingupdate.go
```diff
@@ -94,11 +94,15 @@ func (r _resource) processPendingUpdates(logger logr.Logger, sc *syncContext) er
 	// In either of these cases we should pick up next pod to update if there are any pending pods to update.
 	var nextPodToUpdate *corev1.Pod
 	if podNamesPendingUpdate := updateWork.getPodNamesPendingUpdate(r.expectationsStore.GetDeleteExpectations(sc.pclqExpectationsStoreKey)); len(podNamesPendingUpdate) > 0 {
-		if pclq.Status.ReadyReplicas < *pclq.Spec.MinAvailable {
+		// Selecting the next pod deletes one Ready pod. To keep at least MinAvailable Ready pods
+		// afterwards we need ReadyReplicas-1 >= MinAvailable, i.e. block while ReadyReplicas <= MinAvailable.
+		// Using '<' here is off-by-one: it permits deleting the last Ready pod when ReadyReplicas == MinAvailable,
+		// dropping Ready to below MinAvailable (contributed to the harbinger-24b 0-ready outage window).
+		if pclq.Status.ReadyReplicas <= *pclq.Spec.MinAvailable {
```
danbar2, line 101: this could be a blocker for the whole PCS. Lets say you have replicas=1 and minAvailable=1, you can never update this workload.
There is an FR for minAvailable/maxSurge configuration that is more suitable for your needs

PR #703, opened 2026-07-08
GREP-704: add phase 1 for Kueue integration

### LINKED ISSUE #704: Kueue integration

Kueue is the community defacto project for batch workloads. It has
quota management semantics along with topology-aware-scheduling(TAS).

With Groves scheduling framework, it would be natural to add kueue as 
a separate backend scheduler so users can integrate inference workloads
like dynamo in the training clusters where kueue is already deployed.

### Why is this needed?

Currently there is a gap where admins manage cluster with kueue are not
able to run grove workloads. This integration bridges the gap and allows 
grove's inference workloads to be widely available to Kueue users.

alaypatel07:
#### What type of PR is this?

/kind documentation

This PR adds GREP for Kueue/Grove integration. The integration proposes the following:
1. Grove creates a pre-built Kueue Workload object to reflect a PodGang
2. Grove uses the necessary Kueue labels on pods created for PCS for Kueue quota admission and Kueue TAS 

Fixes #704

```release-note
Add GREP for Kueue integration
```

```docs

```

### DIFF
--- a/docs/proposals/704-kueue-tas-integration/README.md
+++ b/docs/proposals/704-kueue-tas-integration/README.md
@@ -1,0 +1,553 @@
+# GREP-9999: Kueue TAS Integration
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Kueue Quota and TAS Semantics](#kueue-quota-and-tas-semantics)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Shared Cluster Capacity for Training and Inference](#story-1-shared-cluster-capacity-for-training-and-inference)
+    - [Story 2: Kueue TAS for Grove Pods](#story-2-kueue-tas-for-grove-pods)
+  - [Limitations/Risks &amp; Mitigations](#limitationsrisks--mitigations)
+    - [Phase 1 <code>minCount</code> Mapping](#phase-1-mincount-mapping)
+    - [CPU-Only Validation](#cpu-only-validation)
+- [Design Details](#design-details)
+  - [Kueue Scheduler Backend](#kueue-scheduler-backend)
+  - [Kueue Queue Selection](#kueue-queue-selection)
+  - [Kueue Topology Synchronization](#kueue-topology-synchronization)
+  - [PodSet Topology Request Flow](#podset-topology-request-flow)
+  - [Kueue Cleanup](#kueue-cleanup)
+  - [Monitoring](#monitoring)
+  - [Dependencies](#dependencies)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Alternatives](#alternatives)
+  - [Custom Kueue Job Framework Integration](#custom-kueue-job-framework-integration)
+  - [Kueue-Constructed Workloads from Plain Pods](#kueue-constructed-workloads-from-plain-pods)
+  - [Kueue-Specific Topology Field Only](#kueue-specific-topology-field-only)
+<!-- /toc -->
+
+## Summary
+
+This GREP proposes integrating Grove with Kueue so Grove-managed inference workloads can use Kueue for queue-based quota admission and topology-aware scheduling. Grove will continue to describe application structure through `PodCliqueSet`, `PodClique`, and `PodGang`, while Kueue will admit those workloads against `LocalQueue` and `ClusterQueue` policy and assign topology-aware placement through its existing `Workload` and plain Pod integration.
+
+## Motivation
+
+Kueue is a Kubernetes-native way to express queueing, quota, and admission policy through `LocalQueue`, `ClusterQueue`, `ResourceFlavor`, and `Workload` resources. Many Grove users run on shared Kubernetes clusters where quota and capacity are managed by Kueue. Grove should integrate with Kueue so users can use the existing quota management system for Grove-managed workloads.
+
+Kueue also provides topology-aware scheduling (TAS) for admitted workloads. Grove already has a topology model through `ClusterTopologyBinding`, `PodCliqueSet` topology constraints, and translated `PodGang` constraints. Integrating these systems allows Grove users to express inference topology requirements once, while allowing cluster administrators to keep Kueue as the source of truth for quota, admission, and topology-aware capacity.
+
+The integration should preserve the separation of responsibilities: Grove describes the inference workload and its clique relationships; Kueue decides whether and where that workload can run based on queue, quota, resource flavor, and topology policy.
+
+### Kueue Quota and TAS Semantics
+
+Kueue separates workload submission from workload execution. A user or controller submits work to a namespace-scoped `LocalQueue`. The `LocalQueue` points to a cluster-scoped `ClusterQueue`, which defines the quota and admission policy for that work. The `ClusterQueue` admits workloads against one or more `ResourceFlavor`s. A `ResourceFlavor` represents a class of capacity, such as a pool of GPU nodes with specific labels, taints, tolerations, or topology characteristics.
+
+Quota is evaluated at the `ClusterQueue` and `ResourceFlavor` level. For example, if an inference workload requests eight GPUs and is admitted with the `h100` flavor, Kueue reserves eight GPUs from the quota configured for that flavor in the selected `ClusterQueue`. This quota reservation is independent of where those GPUs are located topologically. Unless the administrator models racks, zones, or other domains as separate queues or flavors, the quota is not divided by topology domain.
+
+Topology-aware scheduling adds a second check after flavor quota is considered. A `ResourceFlavor` can reference a Kueue `Topology` through `spec.topologyName`. The `Topology` describes the node label hierarchy Kueue should use for placement decisions, such as rack and host labels. When a `Workload` includes a PodSet topology request, Kueue asks whether the requested resources can fit inside an allowed topology domain for the selected flavor. A required topology request must fit within the requested level; a preferred request asks Kueue to try that level first and relax placement if needed.
+
+A workload is admitted only when both decisions succeed: Kueue can reserve quota for the selected flavor, and Kueue can find enough available capacity in a topology domain that satisfies the PodSet topology request. In other words, quota answers "is there enough capacity in this flavor for this workload?", while TAS answers "is there a suitable topology domain in that capacity where this workload can fit?"
+
+The following example shows the two layers. The `ClusterQueue` defines quota for the `h100-tas` flavor. The `ResourceFlavor` links that quota to a node pool and, separately, to the topology Kueue should use for TAS decisions:
+
+```yaml
+apiVersion: kueue.x-k8s.io/v1beta2
+kind: Topology
+metadata:
+  name: gb200-topology
+spec:
+  levels:
+  - nodeLabel: nvidia.com/gb200-rack
+  - nodeLabel: kubernetes.io/hostname
+---
+apiVersion: kueue.x-k8s.io/v1beta2
+kind: ResourceFlavor
+metadata:
+  name: gb200-tas
+spec:
+  nodeLabels:
+    accelerator: gb200
+  topologyName: gb200-topology
+---
+apiVersion: kueue.x-k8s.io/v1beta2
+kind: ClusterQueue
+metadata:
+  name: inference-gpu
+spec:
+  namespaceSelector: {}
+  resourceGroups:
+  - coveredResources:
+    - nvidia.com/gpu
+    flavors:
+    - name: gb200-tas
+      resources:
+      - name: nvidia.com/gpu
+        nominalQuota: "720"
+---
+apiVersion: kueue.x-k8s.io/v1beta2
+kind: LocalQueue
+metadata:
+  namespace: serving
+  name: inference
+spec:
+  clusterQueue: inference-gpu
+```
+
+A workload submitted to `LocalQueue/inference` consumes `nvidia.com/gpu` quota from `ClusterQueue/inference-gpu`. If the workload also requests `nvidia.com/gb200-rack` as its PodSet topology level, Kueue uses `Topology/gb200-topology` to choose a rack domain where the admitted Pods can fit.
+
+After admission, Kueue records the placement decision in `Workload.status.admission.podSetAssignments[].topologyAssignment`. Kueue does not call a special `kube-scheduler` API to communicate this decision. Instead, Kueue updates the admitted Pods using normal Kubernetes scheduling fields, such as `nodeSelector`, and removes the Kueue scheduling gates. Once the gates are removed, `kube-scheduler` schedules the Pods normally, constrained by the selectors that Kueue applied.
+
+### Goals
+
+* Provide an API for Grove workloads to select which Kueue quota they use.
+* Translate Grove topology constraints into Kueue PodSet topology annotations for Kueue TAS admission.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+  - [Graduation Criteria](#graduation-criteria)
+- [Alternatives](#alternatives)
+  - [Custom Kueue Job Framework Integration](#custom-kueue-job-framework-integration)
+  - [Kueue-Constructed Workloads from Plain Pods](#kueue-constructed-workloads-from-plain-pods)
+  - [Kueue-Specific Topology Field Only](#kueue-specific-topology-field-only)
+<!-- /toc -->
+
+## Summary
+
+This GREP proposes integrating Grove with Kueue so Grove-managed inference workloads can use Kueue for queue-based quota admission and topology-aware scheduling. Grove will continue to describe application structure through `PodCliqueSet`, `PodClique`, and `PodGang`, while Kueue will admit those workloads against `LocalQueue` and `ClusterQueue` policy and assign topology-aware placement through its existing `Workload` and plain Pod integration.
```
danbar2, line 34: nit: please remove the 'inference' word, it should work for all kinds

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+
+## Motivation
+
+Kueue is a Kubernetes-native way to express queueing, quota, and admission policy through `LocalQueue`, `ClusterQueue`, `ResourceFlavor`, and `Workload` resources. Many Grove users run on shared Kubernetes clusters where quota and capacity are managed by Kueue. Grove should integrate with Kueue so users can use the existing quota management system for Grove-managed workloads.
+
+Kueue also provides topology-aware scheduling (TAS) for admitted workloads. Grove already has a topology model through `ClusterTopologyBinding`, `PodCliqueSet` topology constraints, and translated `PodGang` constraints. Integrating these systems allows Grove users to express inference topology requirements once, while allowing cluster administrators to keep Kueue as the source of truth for quota, admission, and topology-aware capacity.
```
danbar2, line 40: Same, remove the 'inference' please validate it acrros the GREP

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+* Replacing Kueue queue, quota, preemption, or topology policies with Grove-owned policy.
+* Automatically discovering physical cluster topology or GPU interconnect topology.
+
+## Proposal
+
+Grove will support Kueue as an admission and topology-aware scheduling backend for `PodCliqueSet` workloads. Users will be able to select a Kueue `LocalQueue` from the Grove API and use Grove topology constraints to request Kueue TAS placement, while Grove continues to manage the workload structure and Pods.
+
+Kueue will remain responsible for queue selection, quota reservation, admission, and topology assignment. Grove will remain responsible for describing the inference workload, creating and reconciling the Pods, and translating portable Grove topology constraints into scheduler-specific placement requests.
+
+The proposed API change is intentionally small: Grove workloads can select the Kueue `LocalQueue` they submit to, and existing Grove topology constraints can be used for Kueue TAS. This allows platform administrators to keep Kueue as the source of truth for quota and topology-aware capacity while Grove users continue to express inference workload structure through `PodCliqueSet`.
```
danbar2, line 120: it is not clear what is the actual API change, is it on the CRD or Config ? which fields are changed/added ?

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
+# GREP-9999: Kueue TAS Integration
+
+<!-- toc -->
+- [Summary](#summary)
```
amy, line 4: Could you also add a section that covers things that this integration won't support in terms of grove feature? Like if there's grove features that don't cleanly port over to Kueue. 

(feels different from non-goals section)

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+
+GA:
+
+* The integration is used by multiple Grove workloads without known cleanup or admission correctness issues.
+* Upgrade and rollback behavior is documented.
+* Monitoring and status conditions are sufficient for operators to diagnose queue, quota, and topology failures.
+
+## Alternatives
+
+### Custom Kueue Job Framework Integration
```
kannon92, line 543: Basically every project I know starts with what you propose and then realizes the downsides are large.

And then they end up implementing a first class controller for this.

Users lose the ability to see why their grove workload is pending. They would have to read the pods to learn why they are not being placed.

kannon92: KServe is in the process of introducing proper support for this due to the challenges of doing gang admission based on pod integration when the workload can create many different types of Kinds. (job, multiple lws).

https://docs.google.com/document/d/1Bjv91MfGBvr0NRL8c2m1bNEfa8ZXA5YFSvZ-ZP5QtBI/edit?usp=sharing

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+
+Grove could implement a first-class Kueue job framework integration for Grove `PodGang` resources. This would give Kueue direct knowledge of Grove workload structure and lifecycle, but it requires Kueue to import or understand Grove APIs and significantly increases the integration surface. This proposal targets the same `PodGang -> Workload` admission unit while using Kueue's existing plain Pod integration.
```
kannon92, line 545: This is also not true at all. An external framework can be implemented out of Kueue entirely. See https://github.com/tektoncd/tekton-kueue for a production example of how you can use Kueue with tekton pipelines without needing to import anything related to tekton in Kueue.

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+metadata:
+  labels:
+    kueue.x-k8s.io/queue-name: inference
+spec:
+  template:
+    cliques:
+    - name: worker
+      spec:
+        podSpec:
+          schedulerName: kueue
```
kannon92, line 219: This API is quite confusing. Kueue is not a scheduler plugin or a secondary scheduler.

For example, we are looking at piloting an idea where Kueue can admit a workload and then slurm handles scheduling. In this case, Kueue can still admit the workload but the podSpec itself uses slurmBridge. 

The other area is if someone uses their own schedulerPlugin and needs to patch pods to use that with Kueue. This is allowed but I am not clear if this is possible for grove.

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+spec:
+  template:
+    cliques:
+    - name: decode
+      spec:
+        podSpec:
+          schedulerName: kueue
+```
+
+A more representative inference workload can include standalone prefill capacity and a decode scaling group:
```
kannon92, line 298: This API brings out a point. What happens if someone forgets to specify `schedulerName: kueue` on a clique.

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: good point, let me explore what the feedback loop looks like with the current documentation.

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: ack, I will take a look.

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: Thanks, I investigated this further. I agree that Grove needs to surface why a workload is pending, but I do not think this requires a GenericJob integration.

Grove already has a first-class PodGang scheduling object and creates one prebuilt Kueue Workload for each PodGang. The integration can watch that owned Workload and propagate its QuotaReserved and Admitted conditions, including the reason and message, into PodGang status. Grove can then expose those PodGang conditions through PodCliqueSet.status.podGangStatuses. This provides the diagnostic path from Kueue to the Grove object users submitted without changing Workload lifecycle ownership to GenericJob.

I also found a pre-existing Grove status gap: PodGang.status.phase is initialized to Pending but is never advanced to Starting or Running, including when using KAI. Similarly, PodCliqueSet.status.podGangStatuses is defined but not populated. I will track these as general Grove status issues.

For this integration, I will update the proposal to require Workload admission status propagation into PodGang and PodCliqueSet. GenericJob remains a possible lifecycle integration approach, but it is not a prerequisite for solving the visibility issue.

alaypatel07: > I also found a pre-existing Grove status gap: PodGang.status.phase is initialized to Pending but is never advanced to Starting or Running, including when using KAI. Similarly, PodCliqueSet.status.podGangStatuses is defined but not populated. I will track these as general Grove status issues.

Created https://github.com/ai-dynamo/grove/issues/762

kannon92: > Thanks, I investigated this further. I agree that Grove needs to surface why a workload is pending, but I do not think this requires a GenericJob integration.

If you don't use GenericJob integration, you lose the ability to use Grove with MultiKueue or ProvisionRequests/Autoscaling. 

Pretty much every project I worked with has been fine with implementing GenericJob. I don't quite understand the push back on implementing this interface and getting all the Kueue features it provides.

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+
+#### Story 2: Kueue TAS for Grove Pods
+
+As a cluster operator using Kueue TAS, I want Grove-created Pods to become Kueue `Workload`s with topology requests so that Kueue can reserve capacity and apply topology-aware placement before Pods are scheduled.
+
+### Limitations/Risks & Mitigations
+
+#### Phase 1 `minCount` Mapping
+
+The phase 1 design assumes that standalone `PodClique`s can preserve Grove `minAvailable` semantics by mapping `PodGroup.MinReplicas` to Kueue `PodSet.minCount`. This is only applied when `minReplicas` is smaller than the PodSet `count`.
```
kannon92, line 138: Are you sure you don't mean PodSet.Count?

```
	// count is the number of pods for the spec.
	// +kubebuilder:default=1
	// +kubebuilder:validation:Minimum=0
	// +optional
	Count int32 `json:"count"`
```

MinCount is an alpha feature for partial admission.

```yaml
	// minCount is the minimum number of pods for the spec acceptable
	// if the workload supports partial admission.
	//
	// If not provided, partial admission for the current PodSet is not
	// enabled.
	//
	// Only one podSet within the workload can use this.
	//
	// This is an alpha field and requires enabling PartialAdmission feature gate.
	//
	// +optional
	// +kubebuilder:validation:Minimum=1
	MinCount *int32 `json:"minCount,omitempty"`
```

kannon92:
I think this is good for phase 1 integration.

sanjaychatterjee (dismissed):

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
+# GREP-9999: Kueue TAS Integration
```
kangclzjc: Can you rename it to GREP-704?

docs/proposals/704-kueue-tas-integration/README.md
```diff
@@ -0,0 +1,553 @@
...
+
+## Motivation
+
+Kueue is a Kubernetes-native way to express queueing, quota, and admission policy through `LocalQueue`, `ClusterQueue`, `ResourceFlavor`, and `Workload` resources. Many Grove users run on shared Kubernetes clusters where quota and capacity are managed by Kueue. Grove should integrate with Kueue so users can use the existing quota management system for Grove-managed workloads.
+
+Kueue also provides topology-aware scheduling (TAS) for admitted workloads. Grove already has a topology model through `ClusterTopologyBinding`, `PodCliqueSet` topology constraints, and translated `PodGang` constraints. Integrating these systems allows Grove users to express inference topology requirements once, while allowing cluster administrators to keep Kueue as the source of truth for quota, admission, and topology-aware capacity.
+
+The integration should preserve the separation of responsibilities: Grove describes the inference workload and its clique relationships; Kueue decides whether and where that workload can run based on queue, quota, resource flavor, and topology policy.
```
kangclzjc: Still exist inference workload

alaypatel07: My understanding is that running inference workload for kueue administered clusters is the primary use case for this effort.

For my curiosity what is the hang up with the usage of this language?

kangclzjc: One reason could be that grove and kueue should also benefit other workloads, but I agree running inference workload for kueue administered clusters is the primary use case for this effort.

kangclzjc:
It look good to me. After you rename it to GREP-704 for the line 1, I will approve it.

alaypatel07:
@kangclzjc Updated, PTAL

kangclzjc (dismissed):

sanjaychatterjee (dismissed):

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: removed

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: fixed

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: Added a section.

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: I updated the language here to reflect the changes with clarity

danbar2 (dismissed):

kangclzjc (approved):

sanjaychatterjee (approved):

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: yes I'm sure we are using minCount to make the workload analogous to minAvailable in PodGang.

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: Unfortunately we dont have any other mechanism to trigger a backend, we will bring this question in phase 2. For now resolving

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: We have added a mapping at Grove configuration level. This mapping will provide value of scheduler name to substitute in podSpec. With that all the cases mentioned here with kueue will be possible

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: I looked at this implementation, we decided to push this to phase 2 and rely on plan pod for now.

docs/proposals/704-kueue-tas-integration/README.md
alaypatel07: See here, this will be explored in the future, https://github.com/ai-dynamo/grove/pull/703#discussion_r4037757840

PR #682, opened 2026-06-23
chore: align Go toolchain pins to 1.26.3

yankay:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Aligns Grove's Go baseline to `1.26.3` across the remaining modules, CI workflows, e2e setup, and related documentation.

This completes the follow-up work after:
- #603 updated the operator module Go directive while upgrading Kubernetes dependencies.
- #681 updated the operator Dockerfile builder image.

The e2e setup also pins `GOROOT` to the active Go toolchain after `setup-go`, so self-hosted runners do not leak a stale `GOROOT` from their preinstalled Go version into `skaffold`/`ko` builds.

#### Which issue(s) this PR fixes:

Fixes: #683

#### Special notes for your reviewer:

NONE

#### Does this PR introduce a API change?

```release-note
Upgrade Go baseline to 1.26.3
```

#### Additional documentation e.g., enhancement proposals, usage docs, etc.:

```docs
NONE
```

PR #705, opened 2026-07-09
chore: move e2e runner dependencies into workflow setup

### LINKED ISSUE #702: Replace pre-baked runner image with cache-based dependency management

**Background**

CI currently uses the pre-baked runner image [prod-grove-e2e-v1](https://github.com/ai-dynamo/grove/blob/b3a10ef0628b8a78541c504db0cc7d44312c029c/.github/workflows/build-check-test.yaml#L112), which speeds up workflows by having required tools pre-installed. However, this approach introduces several problems.

**Problems**

- **Opaque versioning:** The image is built outside the Grove repository. External contributors and the broader community have no visibility into which tool versions are baked in.
- **Version drift:** Tool versions are defined in [tools.mk](https://github.com/ai-dynamo/grove/blob/main/hack/tools.mk) and used by Makefile targets. As these versions are updated in the repo, they diverge from what's actually installed in the runner image — and we have no way to update the image to match.

**Proposed Solution**

Replace the pre-baked runner image with standard GitHub Actions tooling (actions/cache, actions/setup-go, etc.) so that all dependency versions are resolved at runtime from within the Grove repo itself.

### Why is this needed?

**Benefits**
  - Single source of truth for all tool/dependency versions, living inside the codebase
  - Developer environments stay in sync with CI
  - Easy version bumps to track upstream releases, new features, and security fixes
  - Full transparency for community contributors

yankay:
#### What type of PR is this?

/kind cleanup

Moves e2e CI from the pre-baked `prod-grove-e2e-v1` runner to the standard GitHub-hosted `ubuntu-24.04` runner.

The e2e setup now installs Go, Python dependencies, and CLI tools explicitly from workflow configuration and `hack/tools.mk`, with caches for Go, pip, and `hack/tools/bin`.

Fix: #702

Inventory showed `prod-grove-e2e-v1` is about 16 vCPU / 61 GiB RAM / 300 GB disk, while standard `ubuntu-24.04` is 4 vCPU / 16 GB RAM / 14 GB SSD. GitHub offers comparable larger runners, for example 16 vCPU / 64 GB RAM, but they require org configuration and a configured runner label. This PR starts with the standard runner; heavy e2e jobs may need a larger runner or further splitting if they become flaky.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/.github/workflows/build-check-test.yaml
+++ b/.github/workflows/build-check-test.yaml
@@ -46,6 +46,7 @@ jobs:
           filters: |
             e2e-relevant:
               - 'operator/**'
+              - 'hack/**'
               - '.github/**'
 
   test:
@@ -108,8 +109,7 @@ jobs:
       github.event_name == 'pull_request' &&
       needs.changes.outputs.e2e-relevant == 'true' &&
       (github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'run-e2e'))
-    # use NVIDIA self-hosted runner setting is on Velonix repository
-    runs-on: prod-grove-e2e-v1
+    runs-on: ubuntu-24.04
     timeout-minutes: 60
     strategy:
       fail-fast: false
--- a/.github/workflows/build-check-test.yaml
+++ b/.github/workflows/build-check-test.yaml
@@ -39,8 +39,8 @@ jobs:
     outputs:
       e2e-relevant: ${{ steps.filter.outputs.e2e-relevant }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: dorny/paths-filter@v3
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      - uses: dorny/paths-filter@d1c1ffe0248fe513906c8e24db8ea791d46f8590 # v3.0.3
         id: filter
         with:
           filters: |
@@ -52,9 +52,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@7b8cf10d4e4a01d4992d18a89f4d7dc5a3e6d6f4 # v4.3.0
         with:
           go-version: "1.26.3"
           cache: true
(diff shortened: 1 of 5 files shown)


danbar2 (dismissed):

yankay:
ref to: #711

.github/workflows/build-check-test.yaml
```diff
@@ -108,8 +109,7 @@ jobs:
       github.event_name == 'pull_request' &&
       needs.changes.outputs.e2e-relevant == 'true' &&
       (github.event.pull_request.draft == false || contains(github.event.pull_request.labels.*.name, 'run-e2e'))
-    # use NVIDIA self-hosted runner setting is on Velonix repository
-    runs-on: prod-grove-e2e-v1
+    runs-on: ubuntu-24.04
```
Ronkahn21: changing Runner from the internal one is something that need to be discuss with the dynamo operation team
as we have uniq setup and Nvidia requirement ( we moved to dedicated runners after availability, capability, etc   issues)


I would rather prefer create a more clean version of grove runner without all the tools installed than use random runner from github

Ronkahn21 (changes requested):
overall the PR looks good, but the runner change is something that need to be discuss 
we moved from public to private after many e2e issues

gflarity (changes requested):
Could you add some context on the motivation for switching runners? I believe one reason we moved to dedicated runners was E2E turnaround: having enough capacity to start all ten shards quickly and run them massively in parallel.  The tests maybe quicker now and maybe this not an issue, but I recall waiting quite a while for queued E2E tests when multiple people had PRs to be merged. 

Also please pin the github actions, I'm not sure the dorny one is allowed.

.github/workflows/build-check-test.yaml
yankay: Thanks, that makes sense. I restored prod-grove-e2e-v1 and kept this PR focused on repo-managed tool versions, installation, caches, and pinned Action SHAs. This preserves the dedicated runner setup while allowing its image to be simplified separately.

yankay:
> Could you add some context on the motivation for switching runners? I believe one reason we moved to dedicated runners was E2E turnaround: having enough capacity to start all ten shards quickly and run them massively in parallel. The tests maybe quicker now and maybe this not an issue, but I recall waiting quite a while for queued E2E tests when multiple people had PRs to be merged.
> 
> Also please pin the github actions, I'm not sure the dorny one is allowed.

HI @gflarity 
Thanks for raising this. The motivation is documented in #702 , and I ran into the problem firsthand in  https://github.com/ai-dynamo/grove/pull/682/changes#diff-fb79935a661639544d948d703447c34442c8b3cb56caa72a3fd28b13563c0c87R67 : despite Go being preinstalled, CI still needed another Go version, and the stale GOROOT affected skaffold/ko builds.
I restored prod-grove-e2e-v1 to preserve its capacity and kept this PR focused on repository-managed tools and caches. All Actions, including the existing dorny/paths-filter, are now pinned to immutable SHAs.

yankay:
Thank you both for the helpful feedback. I’ve kept prod-grove-e2e-v1, pinned the workflow Actions, and focused the PR on repository-managed tools and caches. @gflarity and @Ronkahn21, when you have a chance, would you mind taking another look? Thank you! :)

PR #707, opened 2026-07-09
feat: add Kueue prebuilt workload backend

### LINKED ISSUE #704: Kueue integration

alaypatel07:
Add a Kueue scheduler backend path that creates prebuilt Workloads from Grove PodGangs so PodGroup minReplicas can map to Kueue PodSet minCount. Include focused demo manifests for standalone minCount and PCSG all-or-nothing behavior.

This PR adds Kueue integration. This PR:
1. adds kueue scheduler backend
2. the backend pre-creates the workload for the pods
3. the operator then creates the pod with specific labels that is needed 
    for kueue to admit the workload.

Fixes #704 

```release-note
GREP-704: add kueue integration
```

```docs
TBD
```


shayasoolin:
@enoodle would you like to review this? Thanks

operator/internal/controller/podclique/components/pod/pod.go
```diff
@@ -176,6 +183,20 @@ func (r _resource) buildResource(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grov
 			"failed to prepare pod spec with scheduler backend",
 		)
 	}
+	if backend.Name() == string(configv1alpha1.SchedulerNameKueue) {
+		if podGang == nil && componentutils.HasAnyTopologyConstraint(pcs) {
```
enoodle: Shouldn't this fail also if the podGang is `nil` and PCS doesn't have topology because it means the Workload was not created yet?

enoodle (commented):
I see that a lot of the code changes are outside of the kueue scheduler backend packages (internal/scheduler/kueue) - I think it is not a good practice and it is against the idea of that separation. I do understand why it is done - For the Kueue backend you need some information from the PCS or PodGang that you can't access from the current hooks in the scheduler backend.
I think that the solution should be to extend the API to include the extra infromation that you need instead of leaking the logic outside.
Something like 
```
 PreparePod(ctx context.Context, pod *corev1.Pod, preparationContext PodPreparationContext) error                                                                          
```                                                                                      
where the context can expose the PCS, PodClique, and PodGang—or equivalent backend-neutral derived data.

operator/internal/scheduler/kueue/backend.go
```diff
@@ -0,0 +1,412 @@
...
+			Name:     kueuev1beta2.NewPodSetReference(podGroup.Name),
+			Count:    count,
+			Template: template,
+		}
+		isPartiallyAdmittedStandalonePodGroup := !isCliqueInScalingGroup(pcs, cliqueTemplate.Name) && minCount < count
+		if isPartiallyAdmittedStandalonePodGroup {
+			if minCountUsed {
+				return nil, fmt.Errorf("PodGang %s/%s has more than one partially admitted standalone PodGroup", podGang.Namespace, podGang.Name)
+			}
+			podSet.MinCount = &minCount
```
enoodle, line 181: When Kueue removes the scheduling gate for this PodSet, will all pods be ungated even though Kueue only reserved resources for minCount of them? I am not sure if it is a problem or not, they will just stay pending, But is it Kueue's only way to control them? What if Kueue didn't want them to consume resources?

operator/internal/scheduler/kueue/backend.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+	pod.Annotations[podGroupTotalCountAnnotation] = strconv.Itoa(totalPodCount)
+	pod.Annotations[roleHashAnnotation] = podCliqueName
+	// Grove-managed pods are deliberately NOT marked as a Kueue "serving" pod group. A serving group is never
+	// considered finished, so Kueue would never remove the kueue.x-k8s.io/managed finalizer on teardown (the
+	// prebuilt Workload has no Kueue finalizer, so Workload deletion cannot trigger finalization either). As a
+	// non-serving unretriable group, Kueue finalizes the Pods once they all terminate, letting them drain.
+	if pod.Annotations[retriableInGroupAnnotation] == "" {
+		pod.Annotations[retriableInGroupAnnotation] = "false"
+	}
```
enoodle, line 294: AI review raised an issue that might come up when Kueue is preempting this workload due to this annotation:

```
  1. Workload is admitted.
  2. Kueue preempts it:
     admission cleared → quota already released → Workload should requeue.
  3. Kueue deletes the old Pods.
  4. Grove intends to recreate them.
  5. Before replacements appear, Kueue observes:
     no active Pods + an old Pod marked retriable-in-group=false.
  6. Kueue sets Workload Finished=True.
  7. Grove creates replacements referencing the same prebuilt Workload.
  8. Kueue does not reopen it.

  Kueue’s Pod-group implementation reports the group finished when:

  !isActive && unretriableGroup

  Kueue Pod-group completion logic (https://github.com/kubernetes-sigs/kueue/blob/v0.17.8/pkg/controller/jobs/pod/pod_controller.go#L349-L391)

  Once Finished=True, the generic reconciler finalizes the job and returns before reaching eviction/requeue/admission handling. There is no transition back to Pending. Kueue terminal handling (https://github.com/kubernetes-sigs/kueue/blob/v0.17.8/pkg/controller/jobframework/reconciler.go#L425-L467)
```

operator/internal/scheduler/kueue/backend.go
```diff
@@ -0,0 +1,412 @@
...
+// assumption that PCSG.minAvailable == PCSG.replicas.
+func (b *schedulerBackend) SyncPodGang(ctx context.Context, podGang *groveschedulerv1alpha1.PodGang) error {
+	pcsName := podGang.Labels[apicommon.LabelPartOfKey]
+	if pcsName == "" {
+		return fmt.Errorf("PodGang %s/%s must set label %q", podGang.Namespace, podGang.Name, apicommon.LabelPartOfKey)
+	}
+
+	err := b.client.Get(ctx, client.ObjectKey{Namespace: podGang.Namespace, Name: podGang.Name}, &kueuev1beta2.Workload{})
+	if err == nil {
+		// Kueue Workload podSets are immutable once admitted, so the Workload is only created and not updated in this POC.
```
enoodle, line 116: This is probably something we will have to fix in the future because Grove's PCS can be updated. Maybe we can extend `ValidatePodCliqueSet` to also be called on update or give oldPCS, newPCS pair so that this backend can block updates when needed.

operator/internal/scheduler/kueue/backend.go
```diff
@@ -0,0 +1,412 @@
...
+	if requiredKey != "" {
+		request.Required = &requiredKey
+		annotations[podSetRequiredTopologyAnnotation] = requiredKey
+	}
+	if preferredKey != "" {
+		request.Preferred = &preferredKey
+		annotations[podSetPreferredTopologyAnnotation] = preferredKey
+	}
+	return &resolvedTopologyRequest{request: request, annotations: annotations}
+}
```
enoodle, line 322: Kueue will reject a pod with multiple topology annotations:
https://github.com/kubernetes-sigs/kueue/blob/v0.19.4/pkg/controller/jobframework/tas_validation.go#L34-L58
(also true for v0.17.8)
So maybe this should be validated in ValidatePodCliqueSet ?

operator/internal/scheduler/kueue/backend.go
alaypatel07: For now this is a good tradeoff. I will look into it.

> Maybe we can extend ValidatePodCliqueSet to also be called on update or give oldPCS, newPCS pair so that this backend can block updates when needed.

operator/internal/scheduler/kueue/backend.go
alaypatel07: > When Kueue removes the scheduling gate for this PodSet, will all pods be ungated even though Kueue only reserved resources for minCount of them?

Not when the PodSet gets a TAS assignment. Quota is reserved for the reduced count either way, but Grove's Pods can carry two gates:
- kueue.x-k8s.io/admission — Pod.Run removes this from every Pod in the group, with no limit on how many.
- kueue.x-k8s.io/topology — only added when the PodSet got a TopologyAssignment, and assignGatedPodsToDomainsGreedy removes it from only as many Pods as that assignment covers.

So under TAS the extra Pods keep the topology gate and stay Pending — which is what the minCount demo shows: 4 replicas, minAvailable: 2, Kueue admits 2, so 2 run and 2 stay gated. Without a TopologyAssignment there is no second gate, so all Pods become schedulable even though quota was reserved for fewer.

That over-admission needs both:

1. The PodSet gets no TopologyAssignment — Grove emits no topology request (TAS off via Kueue's feature gate or Grove's topologyAwareScheduling.enabled, or the PCS declares no topologyConstraint, or the topologyName / ClusterTopologyBinding / domain fails to resolve — the last three degrade silently) and the ClusterQueue isn't TAS-only, since a TAS-only queue implies TAS even without an explicit request.
2. A standalone clique with minAvailable < replicas (PCSG cliques never get minCount), where quota fits ≥ minCount but < replicas so partial admission actually reduces the count.

Agree, but IMO it is not a blocker

operator/internal/scheduler/kueue/backend.go
enoodle: OK

operator/internal/scheduler/kueue/backend.go
```diff
@@ -0,0 +1,412 @@
...
+	return total, nil
+}
+
+// ValidatePodCliqueSet rejects PodCliqueSets that cannot be represented as a valid Kueue Workload.
+// Grove builds one prebuilt Kueue Workload per PodGang and maps each standalone PodClique with
+// minAvailable < replicas to a Kueue podSet that sets minCount. Kueue permits at most one podSet per
+// Workload to set minCount, so at most one standalone PodClique may declare partial-gang
+// (minAvailable < replicas) semantics. PodCliqueScalingGroups, and the PodCliques that belong to
+// them, are always all-or-nothing and must not declare minAvailable < replicas.
+func (b *schedulerBackend) ValidatePodCliqueSet(_ context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error {
```
alaypatel07, line 342: As per @sanjaychatterjee 

Need to Validated if PodGang is being mutated. 

Also, we need to list down all the PCS changes that can lead to pod gang mutations and reject those in the webhook

alaypatel07: Also, need to look at LWS implementation for how to support scale out and scale in.

Also add the limitation around this in the GREP

operator/internal/scheduler/kueue/backend.go
alaypatel07: I have added the validation at pod clique level, we dont have any remaining cases where old and new PCS is required for validation. The existing validation covers everything.

operator/internal/scheduler/kueue/backend.go
alaypatel07: added a validation for this.

PR #710, opened 2026-07-13
fix: store revisions to ensure grove upgrades do not cause new rollouts

### LINKED ISSUE #709: Grove upgrade causes all pods to be recreated
### What happened?

When upgrading from v0.1.0-alpha.9 to v0.1.0-alpha.11, pods that existed before the upgrade were recreated:

```
kind create cluster --name rollout-repro --wait 3m

helm upgrade --install grove oci://ghcr.io/ai-dynamo/grove/grove-charts \
  --version v0.1.0-alpha.9 \
  --namespace grove-system \
  --create-namespace \
  --set config.leaderElection.enabled=false \
  --set 'config.scheduler.profiles[0].name=default-scheduler' \
  --set config.scheduler.defaultProfileName=default-scheduler \
  --set webhookServerSecret.enabled=false \
  --wait \
  --timeout 3m

kubectl apply -f - <<'EOF'
apiVersion: grove.io/v1alpha1
kind: PodCliqueSet
metadata:
  name: busybox
spec:
  replicas: 1
  template:
    cliques:
      - name: worker
        spec:
          roleName: worker
          replicas: 1
          minAvailable: 1
          podSpec:
            schedulerName: default-scheduler
            terminationGracePeriodSeconds: 1
            containers:
              - name: busybox
                image: busybox:1.36
                imagePullPolicy: IfNotPresent
                command: [sh, -c, "sleep 3600"]
                resources:
                  requests:
                    cpu: 1m
                    memory: 1Mi
EOF

kubectl get pod -l app.kubernetes.io/managed-by=grove-operator,app.kubernetes.io/part-of=busybox -o jsonpath='{..metadata.uid}'
kubectl get pod -l app.kubernetes.io/managed-by=grove-operator,app.kubernetes.io/part-of=busybox -o jsonpath='{..metadata.labels.grove\.io/pod-template-hash}'

helm upgrade grove oci://ghcr.io/ai-dynamo/grove/grove-charts \
  --version v0.1.0-alpha.11 \
  --namespace grove-system \
  --reset-values \
  --set config.leaderElection.enabled=false \
  --set 'config.scheduler.profiles[0].name=default-scheduler' \
  --set config.scheduler.defaultProfileName=default-scheduler \
  --set webhookServerSecret.enabled=false \
  --wait \
  --timeout 3m

kubectl get pod -l app.kubernetes.io/managed-by=grove-operator,app.kubernetes.io/part-of=busybox -o jsonpath='{..metadata.uid}'
kubectl get pod -l app.kubernetes.io/managed-by=grove-operator,app.kubernetes.io/part-of=busybox -o jsonpath='{..metadata.labels.grove\.io/pod-template-hash}'
```

The root cause seems to be the dependency upgrade to K8s v1.35 (https://github.com/ai-dynamo/grove/commit/60c2613fa48e2cc3990de5b68af2e11202f6739d) changing the underlying PodSpec representation that then changes the hash for the semantically-equal PCS.

A grove upgrade does not cause all managed pods to be recreated.

### Environment

_No response_

steved:
#### What type of PR is this?

/kind bug

Implements a [ControllerRevision](https://kubernetes.io/docs/reference/kubernetes-api/apps/controller-revision-v1/)-backed solution for tracking the desired state of pod cliques to ensure that Grove upgrades (and underlying dependency changes) do not cause pods to be updated unnecessarily.

Fixes #709

```release-note
NONE
```

```docs
TBD
```


docs/installation.md
```diff
@@ -118,6 +118,26 @@ Follow the instructions in the [quickstart guide](quickstart.md) to deploy a Pod
...
+   and no automatic update is active.
+3. Start the new operator. It records the already-observed desired workload as
+   a baseline ControllerRevision while preserving the existing rollout hashes;
+   it does not update or replace workload children during adoption.
+
+The next rollout-relevant spec change creates a new ControllerRevision and
+uses the normal update strategy. An active or otherwise unobserved legacy
+revision is left unchanged and reports a reconciliation error until the
+migration precondition is resolved. Downgrading afterward to an operator that
+does not understand ControllerRevision-backed selection is unsupported.
```
shayasoolin: Preventing the option to rollback is a big deal IMHO. Can you think of ways to support it, even if it comes with some price, including even pods being restarted? At least so we can discuss the tradeoffs. Being able to rollback is important, especially given the complexity of changes like this one and other major ones that are in the pipeline.

docs/api-reference/operator-api.md
```diff
@@ -491,7 +491,8 @@ _Appears in:_
...
 | `podGangStatuses` _[PodGangStatus](#podgangstatus) array_ | PodGangStatuses captures the status for all the PodGang's that are part of the PodCliqueSet. |  |  |
-| `currentGenerationHash` _string_ | CurrentGenerationHash is a hash value generated out of a collection of fields in a PodCliqueSet.<br />Since only a subset of fields is taken into account when generating the hash, not every change in the PodCliqueSetSpec will<br />be accounted for when generating this hash value. A field in PodCliqueSetSpec is included if a change to it triggers<br />a rolling recreate of PodCliques and/or PodCliqueScalingGroups.<br />Only if this value is not nil and the newly computed hash value is different from the persisted CurrentGenerationHash value<br />then an update needs to be triggered. |  |  |
+| `currentGenerationHash` _string_ | CurrentGenerationHash is the opaque identity selected for the current rollout-relevant revision. |  |  |
+| `currentRevision` _string_ | CurrentRevision names the ControllerRevision containing the selected rollout revision. |  |  |
```
shayasoolin, line 495: Is there any kind of cleanup for old revisions or are they all kept for the whole PCS lifetime?

operator/internal/controller/podcliqueset/reconcilespec.go
```diff
@@ -45,7 +43,7 @@ func (r *Reconciler) reconcileSpec(ctx context.Context, logger logr.Logger, pcs
 	rLog := logger.WithValues("operation", "spec-reconcile")
 	reconcileStepFns := []ctrlcommon.ReconcileStepFn[grovecorev1alpha1.PodCliqueSet]{
 		r.ensureFinalizer,
-		r.processGenerationHashChange,
+		r.processRevision,
 		r.syncPodCliqueSetResources,
```
shayasoolin, line 47: After selecting a new ControllerRevision, this reconcile continues directly into child sync, and child sync reloads the selected revision through the controller-runtime client/cache. In a real manager this can race informer cache propagation: the Create has reached the API server, but the cache may not yet contain the new ControllerRevision. Fake-client tests will not catch this. Consider requeueing after revision selection, using APIReader for this read, or passing the selected revision through the sync path.

operator/internal/controller/common/revision/data.go
```diff
@@ -0,0 +1,105 @@
...
+
+// DataVersion is the current ControllerRevision data schema version.
+const DataVersion = 1
+
+// Data is the persisted ControllerRevision.Data schema. Revision construction code populates it;
+// reconciliation code should consume a validated SelectedRevision instead.
+type Data struct {
+	Version        int                        `json:"version"`
+	Desired        json.RawMessage            `json:"desired"`
+	Cliques        map[string]json.RawMessage `json:"cliques"`
```
shayasoolin: The revision payload stores the rollout template data twice: Desired contains all pod templates, and Cliques stores each clique template again. For large PodCliqueSets this can make each ControllerRevision much larger than the PCS template itself, and rollout can temporarily retain more than one revision. Can we avoid the duplication, or add a guard/test/documentation around the maximum supported object size?

steved: Yeah - this was to preserve clique ordering and still retain the map. Let me see if there's a better way to do this.

docs/api-reference/operator-api.md
steved: Right now all but the active one are cleaned up: https://github.com/steved/grove/blob/7cb40f286add7cd4006e736a1f03dec202e0c6c5/operator/internal/controller/podcliqueset/revision.go#L300-L321

docs/installation.md
steved: Let me take a look. It should work theoretically as long as the actual template hash doesn't change, but I didn't  test it.

kangclzjc:
hi @steved , did you have the time to finish this PR?

steved:
> hi @steved , did you have the time to finish this PR?

Hey, yeah I'll give this another look this week. We've been running it internally for a bit without issues.

PR #715, opened 2026-07-14
chore: test grove operator upgrades

### LINKED ISSUE #712: Add upgrade E2E coverage
### What would you like to be added?

Add release-to-release upgrade E2E coverage that verifies a Grove upgrade does not recreate existing workload Pods.

### Why is this needed?

This would prevent regressions such as https://github.com/ai-dynamo/grove/issues/709. Unit tests using only the current dependency version cannot fully detect cross-version behavior.

steved:
#### What type of PR is this?

/kind cleanup

Adds an E2E upgrade test from the latest release version and validates that pods created before the upgrade survive.

To reproduce #709: 
```
GROVE_UPGRADE_FROM_VERSION=v0.1.0-alpha.9 make run-upgrade-e2e
```

Fixes #712

_N/A_

```release-note
NONE
```

```docs

```


danbar2:
Hey @steved , is it ready for review or blocked on the upstream commit ?

steved:
> Hey @steved , is it ready for review or blocked on the upstream commit ?

This should be all set to review independently. Thanks!

operator/e2e/tests/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,607 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	githubOwner           = "ai-dynamo"
```
Ronkahn21: Project convention is to use string literals in tests rather than constants. Please inline these at their use sites.

operator/e2e/tests/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,607 @@
...
+		expectedInitImage := "grove-initc:" + currentImageTag
+		require.Truef(t, podUsesInitImage(pods, expectedInitImage),
+			"scaled workload does not contain a pod created with current init image %q", expectedInitImage)
+	}) {
+		t.FailNow()
+	}
+}
+
+func findOperatorDir(t *testing.T) string {
+	t.Helper()
```
Ronkahn21: All functions including internal helpers need at least a one-line docstring per project convention. This applies to all helpers in this file: `findOperatorDir`, `requireCommands`, `durationFromEnv`, `envBool`, `latestGitHubRelease`, `firstNonEmpty`, `newKubernetesClients`, `applyManifestEventually`, `applyObjectEventually`, `patchPodCliqueSetEventually`, `waitForWorkload`, `podContainersReady`, `assertDeployment`, `findContainer`, `imageOf`, `containerEnvValue`, `podUsesInitImage`.

operator/e2e/tests/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,607 @@
...
+}
+
+func applyManifestEventually(t *testing.T, ctx context.Context, client dynamic.Interface, mapper meta.RESTMapper, manifest string) {
+	t.Helper()
+	decoder := yaml.NewYAMLOrJSONDecoder(strings.NewReader(manifest), 4096)
+	for {
+		object := &unstructured.Unstructured{}
+		if err := decoder.Decode(object); err != nil {
+			if err == io.EOF {
+				return
```
Ronkahn21: Prefer `errors.Is(err, io.EOF)` over direct equality — handles wrapped errors correctly.

Ronkahn21 (changes requested):
Thanks for the PR — the upgrade test coverage is really valuable and the Helm lifecycle (install → workload → upgrade → assert) is well structured!

One structural ask: the cluster setup and teardown is embedded inside the Go test via `exec.Command("k3d", ...)`, but every other E2E test in the repo delegates this to `infra-manager.py` + the Makefile layer:

```
# current pattern for all other tests
make run-e2e-full
  → infra-manager.py setup           # creates cluster + installs Grove
  → go test -tags=e2e ./tests/...
  → infra-manager.py delete k3d-cluster
```

The upgrade test should follow the same pattern — the Go test owns the Helm lifecycle, not the k3d lifecycle. `infra-manager.py` already supports `--set grove.enabled=false` for a bare cluster without Grove:

```
make run-upgrade-e2e
  → infra-manager.py setup --set grove.enabled=false
  → go test -tags=e2e,e2eupgrade
  → infra-manager.py delete k3d-cluster
```

This avoids duplicating the k3d config (k3s image, `--no-lb`, `--disable=traefik`, etc.) and ensures the upgrade test picks up any future infra changes automatically.

steved:
> This avoids duplicating the k3d config (k3s image, --no-lb, --disable=traefik, etc.) and ensures the upgrade test picks up any future infra changes automatically.

Updated to use the helpers much more, thanks for the review!

operator/e2e/tests/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,223 @@
...
+	})
+	require.NoError(t, err, "upgrade Grove chart to current checkout")
+
+	Logger.Info("reconciling pre-upgrade workload")
+
+	err = tc.ScalePCS(workloadName, 2)
+	require.NoError(t, err, "patching workloads")
+
+	err = podsManager.WaitForReadyInNamespace(t.Context(), workloadNamespace, 4, defaultPollTimeout, defaultPollInterval)
+	require.NoError(t, err, "waiting for workload to become ready")
```
steved: I thought about adding a wait (sleep?) + UID check in addition to this test, since I think they could technically be considered separate compatibility targets: if the PCS is updated after the upgrade I think it could be valid to replace all the pods.

operator/e2e/setup/helm.go
oleg-kushniriov, line 91: These are mutually exclusive Helm semantics (`--reuse-values` vs `--reset-values`). `HelmInstallConfig.Validate()` currently doesn't guard against this. 


```suggestion
if c.ReuseValues && c.ResetValues {
    return fmt.Errorf("ReuseValues and ResetValues are mutually exclusive")
}
```

operator/e2e/setup/grove.go
```diff
@@ -219,3 +219,15 @@ func GetGroveChartDir() (string, error) {
 	chartDir := filepath.Join(filepath.Dir(currentFile), "../../charts")
 	return filepath.Abs(chartDir)
 }
+
+// GetOperatorRootDir returns the absolute path to the Grove operator directory.
+// It uses runtime.Caller to find the path relative to this source file.
+func GetOperatorRootDir() (string, error) {
```
oleg-kushniriov, line 240: GetOperatorRootDir() duplicates nearly all of `GetGroveChartDir() `- only the path suffix differs. Consider extracting the shared `runtime.Caller / filepath `boilerplate so future additions don't repeat it again.

operator/Makefile
```diff
@@ -129,6 +132,17 @@ run-e2e-verbose: $(GOTESTSUM)
 	@echo "> Running e2e tests (verbose)..."
 	@cd e2e && $(GOTESTSUM) --format short-verbose -- -count=1 -tags=e2e ./tests/... -timeout 45m $(if $(TEST_PATTERN),-run '$(TEST_PATTERN)')
 
+# Run the standalone upgrade test from the latest published GitHub release to
+# the operator built from the current checkout. The test owns its k3d cluster.
+.PHONY: run-upgrade-e2e
+run-upgrade-e2e: export GROVE_E2E_DIAG_DIR = $(DIAG_DIR)
+run-upgrade-e2e: export GROVE_E2E_DIAG_MODE = $(DIAG_MODE)
+run-upgrade-e2e: $(GOTESTSUM)
```
oleg-kushniriov, line 140: `run-upgrade-e2e` should follow the same pattern as `run-e2e-full` which tears down the cluster even on failure; otherwise a test failure leaves a dangling cluster on the self-hosted runner.

operator/e2e/setup/skaffold.go
```diff
@@ -130,6 +130,48 @@ func InstallWithSkaffold(ctx context.Context, config *SkaffoldInstallConfig) (fu
 	return cleanup, nil
 }
 
+// BuildWithSkaffold builds images using Skaffold CLI.
+// Note: we're using Skaffold CLI  instead of go libraries because there are
+// depdenency conflicts between grove and Skaffold at the time of implentaiton.
+func BuildWithSkaffold(ctx context.Context, config *SkaffoldInstallConfig) (map[string]string, func(), error) {
```
oleg-kushniriov: `BuildWithSkaffold` copy-pastes the entire setup body of `InstallWithSkaffold`:
- `config.Validate()`
- `filepath.Abs(config.SkaffoldYAMLPath)`
- `writeTemporaryKubeconfig()`
- `defer cleanup on err`
- `filepath.Dir(absSkaffoldPath)`
- `runSkaffoldBuild()`

The only difference is `BuildWithSkaffold` skips `runSkaffoldDeploy` and returns the image map.
Consider extract a private `skaffoldSetup()` that both functions call:

```go
func skaffoldSetup(ctx context.Context, config *SkaffoldInstallConfig) (absPath, skaffoldDir, kubeconfigPath string, cleanup func(), err error) { ... }
```

oleg-kushniriov: `BuildWithSkaffold` requires `RestConfig` for a build-only operation.

`SkaffoldInstallConfig.Validate()` returns an error if `RestConfig == nil`. 

`BuildWithSkaffold` is documented as build-only (no deploy), yet callers must supply a real `*rest.Config` - which `BuildWithSkaffold` only uses to write a temporary `kubeconfig` that is then passed as an env var 
to `skaffold build`. Skaffold build pushes images to a registry; it does not contact the cluster.
The kubeconfig is dead weight here.

Callers of the upgrade test must wire up a cluster REST config even when they just want to build images before the cluster has Grove installed - which is the exact use case this function was created for.

Please consider following options:
- Make `RestConfig` optional in `Validate()` when `PushRepo` is set (build-only path).
- Or have `BuildWithSkaffold` skip kubeconfig setup entirely and not pass `KUBECONFIG` to the build command.
or any other valid option

operator/e2e/tests/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,223 @@
...
+	// defaultPollTimeout is the default timeout for polling operations
+	defaultPollTimeout = 5 * time.Minute
+	// defaultPollInterval is the interval for most polling conditions
+	defaultPollInterval = 5 * time.Second
+)
+
+// TestUpgradeFromLatestGitHubRelease owns the complete upgrade lifecycle. It deliberately
+// does not reuse the shared E2E cluster lifecycle: upgrade coverage should begin with the
+// published chart and images, not with a cluster that already contains checkout resources.
+func TestUpgradeFromLatestGitHubRelease(t *testing.T) {
```
oleg-kushniriov, line 45: The test covers one scenario well: latest release -> HEAD with a fixed workload. As the suite grows, two dimensions will naturally need to vary:

**Update strategy.** Another Update Strategies are coming, and verifying that a workload under those strategies survives an operator upgrade is a distinct assertion from the same test under RollingUpdate. 
The existing `tests/update/` package handles this cleanly with a `testConfig` struct - the upgrade package would benefit from the same pattern rather than having each strategy require a full copy of the test.

**From/to versions.** Upgrading from the latest release is the most important path, and it's a good starting point. 
Worth designing with N-2 and specific pinned versions in mind, since those paths tend to surface different compatibility issues.

operator/e2e/setup/skaffold.go
```diff
@@ -130,6 +130,48 @@ func InstallWithSkaffold(ctx context.Context, config *SkaffoldInstallConfig) (fu
...
+	skaffoldDir := filepath.Dir(absSkaffoldPath)
+
+	// Build and push images to localhost:port
+	config.Logger.Debugf("🏗️ Building and pushing images...")
+	builtImages, err := runSkaffoldBuild(ctx, absSkaffoldPath, skaffoldDir, kubeconfigPath, config)
+	if err != nil {
+		return nil, cleanup, fmt.Errorf("skaffold build failed: %w", err)
+	}
+
+	return builtImages, cleanup, nil
```
oleg-kushniriov: The deferred `if err != nil { cleanup() }` fires on error, but `cleanup` is also returned to the caller who may call it again. Two callers, one responsibility. 
The same pattern exists in the pre-existing `InstallWithSkaffold`; the PR replicates it rather than fixing it. 
`cleanup()` should have a single owner: either the function handles it internally on all error paths, or it always delegates to the caller - not both.

oleg-kushniriov (commented):
The test is currently hardcoded to one scenario: 
- latest release -> HEAD
- fixed workload
- fixed update strategy. 
Those parameters should be an input to reuse the test with different strategies/to-from-versions and workloads

operator/e2e/setup/helm.go
steved: This was added but ended up unused, removed.

danbar2 (dismissed):
LGTM

oleg-kushniriov (dismissed):
👍

Ronkahn21 (dismissed):

yankay:
Thanks. No blocking issues found. LGTM

yankay (approved):

enoodle (dismissed):
:+1: This is great! Lets merge this!

steved:
Sorry to trigger a re-review. Got hit by #749, so I put in a fix for it: 1ebaf3c7bc8ff2ea767d6debf1d9d7d7ac1256f4. It passed 10 retries locally, where it was previously failing 5 / 10 tries.

danbar2 (approved):

oleg-kushniriov (approved):

PR #716, opened 2026-07-16
fix: reconcile PodCliqueSet from live API state

### LINKED ISSUE #711: Flaky E2E: RU-11 rolling update stops converging after scale-out
### What happened?

`Test_RU11_RollingUpdateWithPCSScaleOut` intermittently fails after scale-out has completed. The Pods become ready, but the rolling update does not converge before the two-minute timeout.

Examples across unrelated PRs:

- https://github.com/ai-dynamo/grove/actions/runs/29320066677/job/87043942428
- https://github.com/ai-dynamo/grove/actions/runs/29225997082/job/86740528337
- https://github.com/ai-dynamo/grove/actions/runs/29000280661/job/86059729814
- https://github.com/ai-dynamo/grove/actions/runs/29331756906/job/87081798851

The same PR 614 run failed twice, then passed in 38 seconds on the third attempt:

- Failed: https://github.com/ai-dynamo/grove/actions/runs/29000280661/job/86059729814
- Failed: https://github.com/ai-dynamo/grove/actions/runs/29000280661/job/86063876172
- Passed: https://github.com/ai-dynamo/grove/actions/runs/29000280661/job/86065641958

The failure appears as either `condition not met within timeout` or `client rate limiter Wait ... context deadline exceeded`. Capturing `UpdatedReplicas`, `CurrentlyUpdating`, and child hashes at timeout should help identify where status convergence stops.

Diagnostics are currently not uploaded because the collector and workflow use different output paths.

After scale-out reaches three replicas and 30 Pods, the rolling update should reliably complete.

### Environment

- Runner: `prod-grove-e2e-v1`
- Test: `Test_RU11_RollingUpdateWithPCSScaleOut`
- Cluster: k3d with 30 KWOK nodes

yankay:
#### What type of PR is this?

/kind bug

RU-11 diagnostics showed that a stale informer-cached PodCliqueSet could make the status no-op optimization skip a required convergence write, leaving the API status frozen after all children had updated.

This change reads the parent PodCliqueSet directly from the API Server at the start of each reconcile. Child PCLQ, PCSG, and Pod reads remain cache-backed. The existing semantic no-op check is unchanged, but now compares against persisted API state.

This adds one single-object GET per PodCliqueSet reconcile. Child read paths and public APIs are unchanged.

#### Why this approach:

The lost write happens in the no-op path, where no API write or conflict exists to trigger a retry. Reading the parent PCS directly makes the no-op baseline authoritative for every status writer in the reconcile, without maintaining a controller-local resourceVersion protocol.

Benefits:
- removes UID/resourceVersion expectations and lifecycle edge cases;
- keeps PCLQ, PCSG, and Pod reads cache-backed;
- retains the status no-op optimization and avoids watch cascades;
- makes the convergence behavior simpler to reason about and test.

Trade-offs:
- adds one single-object API GET per PCS reconcile;
- the live PCS can briefly be ahead of cached children, which is handled by the existing level-triggered reconciliation and generation-hash expectations;
- child reads remain eventually consistent.

The API cost is explicit and bounded. Further optimization should be driven by measured scale-test data.

Fixes https://github.com/ai-dynamo/grove/issues/711

A real-API-server envtest reproduces `API status=2, cached children=3` and verifies convergence to `3`, preservation of progress/error fields, one live PCS GET per reconcile, and no steady-state resourceVersion bump.

Validated with Kubernetes 1.35.0 and 1.36.2 envtest, related package tests, a targeted race test, and golangci-lint. The full RU-11 E2E was not rerun.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podcliqueset/reconciler.go
+++ b/operator/internal/controller/podcliqueset/reconciler.go
@@ -39,6 +39,7 @@ type Reconciler struct {
 	config                        configv1alpha1.PodCliqueSetControllerConfiguration
 	tasConfig                     configv1alpha1.TopologyAwareSchedulingConfiguration
 	client                        ctrlclient.Client
+	apiReader                     ctrlclient.Reader
 	reconcileStatusRecorder       ctrlcommon.ReconcileErrorRecorder
 	operatorRegistry              component.OperatorRegistry[grovecorev1alpha1.PodCliqueSet]
 	pcsGenerationHashExpectations sync.Map
@@ -52,6 +53,7 @@ func NewReconciler(mgr ctrl.Manager, controllerCfg configv1alpha1.PodCliqueSetCo
 		config:                        controllerCfg,
 		tasConfig:                     topologyAwareSchedulingConfig,
 		client:                        client,
+		apiReader:                     mgr.GetAPIReader(),
 		reconcileStatusRecorder:       ctrlcommon.NewReconcileErrorRecorder(client),
 		operatorRegistry:              pcscomponent.CreateOperatorRegistry(mgr, eventRecorder, topologyAwareSchedulingConfig, networkConfig, schedRegistry),
 		pcsGenerationHashExpectations: sync.Map{},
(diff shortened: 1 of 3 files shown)


yankay (commented):
Withdrawn: posted by mistake.

danbar2:
Please keep WIP PR's in draft mode (for running e2e you can set the label of 'run-e2e')

yankay:
#757  fixes the directly reproduced RU-11 trigger, and the full RU-11 E2E passed 6/6 locally. This PR therefore appears unnecessary for resolving #711. I’ll close it as superseded once #757 is merged.

PR #724, opened 2026-07-17
fix: preserve scheduling gates owned by other controllers

### LINKED ISSUE #723: Grove removes scheduling gates owned by other controllers
## What happened?

When Grove releases a Pod after its PodGang dependencies are satisfied, the PodClique controller clears the Pod's entire `spec.schedulingGates` list.

A Pod can have gates owned by multiple controllers. For example, Grove may add `grove.io/podgang-pending-creation` while Kueue adds `kueue.x-k8s.io/admission` or `kueue.x-k8s.io/topology`. Clearing the entire list removes the Kueue gates before Kueue has released them, allowing the Pod to reach the scheduler prematurely.

## Expected behavior

Grove should remove only `grove.io/podgang-pending-creation` and preserve all scheduling gates owned by other controllers.

## Reproduction

1. Create a Grove-managed Pod that has the Grove PodGang gate.
2. Have another controller add a second scheduling gate to the same Pod.
3. Allow the PodGang dependencies to become ready.
4. Observe that Grove removes every entry from `spec.schedulingGates`, including the foreign gate.

## Impact

This breaks safe controller composition and can bypass admission or topology placement decisions enforced through external scheduling gates.

alaypatel07:
## Summary

- remove only Grove's `grove.io/podgang-pending-creation` scheduling gate when PodGang dependencies are ready
- preserve admission and topology gates owned by other controllers
- add unit coverage for selective gate removal and foreign-gate preservation

Fixes #723

## Test plan

- [x] `cd operator && go test ./internal/controller/podclique/components/pod/`

### DIFF
--- a/operator/internal/controller/podclique/components/pod/syncflow_test.go
+++ b/operator/internal/controller/podclique/components/pod/syncflow_test.go
@@ -320,6 +320,81 @@ func TestCheckAndRemovePodSchedulingGates_ConcurrentExecution(t *testing.T) {
 	}
 }
 
+func TestCheckAndRemovePodSchedulingGates_PreservesForeignGates(t *testing.T) {
+	const kueueAdmissionGate = "kueue.x-k8s.io/admission"
+
+	pod := createTestPod("simple1-0", true, true)
+	pod.Spec.SchedulingGates = append(pod.Spec.SchedulingGates, corev1.PodSchedulingGate{Name: kueueAdmissionGate})
+
+	basePodGang := &groveschedulerv1alpha1.PodGang{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "simple1-0",
+			Namespace: "default",
+		},
+		Spec: groveschedulerv1alpha1.PodGangSpec{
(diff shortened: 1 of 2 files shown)


operator/internal/controller/podclique/components/pod/syncflow_test.go
```diff
@@ -320,6 +320,81 @@ func TestCheckAndRemovePodSchedulingGates_ConcurrentExecution(t *testing.T) {
 	}
 }
 
+func TestCheckAndRemovePodSchedulingGates_PreservesForeignGates(t *testing.T) {
+	const kueueAdmissionGate = "kueue.x-k8s.io/admission"
```
alaypatel07: this should probabaly be a foo admission gate

alaypatel07: Updated to use generic `foo.io/admission` / `foo.io/topology` placeholders instead of Kueue-specific gate names.

danbar2 (approved):
LGTM

Ronkahn21 (approved):

PR #728, opened 2026-07-22
feat(operator): add finite PodClique phase handling

shayasoolin:
## Summary

Related to #285.

This PR introduces the first implementation slice for finite/job-style PodCliques. A PodClique is treated as finite when its pod template uses `restartPolicy: Never`.

The branch is currently stacked on the GREP branch from #696, so until #696 is merged this PR also includes the proposal/documentation commits in its diff.

The implementation scope is intentionally limited to the PCLQ level:
- add `PodClique.status.phase` with `Completed` / `Failed`
- mark a finite PCLQ `Completed` when all pods succeed
- mark a finite PCLQ `Failed` when any pod fails
- stop pod/resource recreation once the PCLQ is terminal
- delete non-terminal pods after terminal phase is persisted
- retain terminal pods for logs/debugging

## API / Validation

- Adds `JobPhase` and optional `status.phase` to `PodClique`.
- Allows `restartPolicy: Never` for finite PCLQs.
- Keeps default/empty restart policy behavior as `Always`.
- Rejects `restartPolicy: OnFailure`.
- Rejects `autoScalingConfig` for finite PCLQs.
- Rejects replica changes for finite PCLQs after creation.

## Controller Behavior

- Adds finite-PCLQ helpers behind generic naming so the finite/job-mode decision can later move away from `restartPolicy` if needed.
- Updates PCLQ status bottom-up from pod phases.
- Requeues immediately when status reconciliation transitions a PCLQ to terminal, so the next reconcile performs terminal cleanup with persisted phase.
- Updates the pod watch predicate to react directly to `pod.status.phase` changes.

## Tests

- Unit coverage for finite detection, validation, phase calculation, terminal sync skipping, active pod cleanup, and terminal-transition requeue.
- E2E coverage for:
  - all pods succeed -> PCLQ `Completed`, terminal pods retained
  - one pod fails -> PCLQ `Failed`, terminal pods retained, active pods deleted and not recreated

## Verification

```sh
GOFLAGS=-mod=mod GOCACHE=/private/tmp/grove-go-build-cache GOTOOLCHAIN=auto go test ./internal/controller/podclique
GOFLAGS=-mod=mod GOCACHE=/private/tmp/grove-go-build-cache GOTOOLCHAIN=auto go test -tags=e2e ./e2e/tests -c -o /private/tmp/grove-e2e-tests.test
```


docs/api-reference/operator-api.md
```diff
@@ -604,6 +622,7 @@ _Appears in:_
 | `updatedReplicas` _integer_ | UpdatedReplicas is the number of Pods that have been updated and are at the desired revision of the PodClique. | 0 |  |
 | `scheduleGatedReplicas` _integer_ | ScheduleGatedReplicas is the number of Pods that have been created with one or more scheduling gate(s) set.<br />Sum of ReadyReplicas and ScheduleGatedReplicas will always be <= Replicas. | 0 |  |
 | `scheduledReplicas` _integer_ | ScheduledReplicas is the number of Pods that have been scheduled by the backend scheduler. | 0 |  |
+| `phase` _[JobPhase](#jobphase)_ | Phase is the terminal phase of a finite PodClique. It is empty for non-finite PodCliques and finite PodCliques that are still running. |  | Enum: [Completed Failed] <br /> |
```
danbar2, line 625: why not supporting it for not finite PCLQ ? "pending"/ "running"

operator/internal/controller/common/component/utils/podclique.go
```diff
@@ -209,6 +209,23 @@ func ComputePCLQPodTemplateHash(pclqTemplateSpec *grovecorev1alpha1.PodCliqueTem
...
+// centralized so it can move from restartPolicy to a dedicated API field later.
+func IsFinitePCLQ(pclq *grovecorev1alpha1.PodClique) bool {
+	if pclq == nil {
+		return false
+	}
+	return pclq.Spec.PodSpec.RestartPolicy == corev1.RestartPolicyNever
+}
+
+// IsPCLQTerminal reports whether a PodClique has reached a finite terminal phase.
+func IsPCLQTerminal(pclq *grovecorev1alpha1.PodClique) bool {
```
danbar2, line 222: nit: rename to IsPCLQTerminated

operator/internal/controller/podclique/reconcilespec.go
```diff
@@ -66,6 +78,31 @@ func (r *Reconciler) ensureFinalizer(ctx context.Context, logger logr.Logger, pc
...
+		}
+		logger.Info("Deleting active pod for terminal PodClique", "pod", client.ObjectKeyFromObject(pod), "PodClique", client.ObjectKeyFromObject(pclq), "phase", pclq.Status.Phase)
+		if err := client.IgnoreNotFound(r.client.Delete(ctx, pod)); err != nil {
+			return ctrlcommon.ReconcileWithErrors("failed to delete active pod for terminal PodClique", err)
+		}
+	}
+	return ctrlcommon.ContinueReconcile()
+}
+
+func isTerminalPod(pod *corev1.Pod) bool {
```
danbar2, line 102: nit: rename to isPodTerminated

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -588,12 +592,14 @@ func validateScaleConfig(scaleConfig *grovecorev1alpha1.AutoScalingConfig, minAv
...
 
 // validatePodSpec validates the PodSpec ensuring certain fields are not set that would conflict with operator management.
+//
+//nolint:unparam // Keep warning plumbing aligned with peer validators; future PodSpec checks may emit admission warnings.
 func (v *pcsValidator) validatePodSpec(spec corev1.PodSpec, fldPath *field.Path) ([]string, field.ErrorList) {
 	allErrs := field.ErrorList{}
 	var warnings []string
 
-	if !utils.IsEmptyStringType(spec.RestartPolicy) {
-		warnings = append(warnings, "restartPolicy will be ignored, it will be set to Always")
```
danbar2, line 596: this is actually breaking, cause we supported any value before that and now it will return an error "OnFailure" for example

docs/api-reference/operator-api.md
shayasoolin: We defined that we'll set phases only for "job-mode" resources. If we support pending/running we should put them on any resource even in regular non-job mode, and I'm not sure there's value in doing that.
Unless you suggest to have pending/running but only for finite resources?
Can we discuss this as part of the GREP PR?

PR #732, opened 2026-07-24
Pass Grove context to scheduler Pod preparation

### LINKED ISSUE #731: Pass Pod preparation context to scheduler backends

Extend the scheduler backend `PreparePod` method to receive the Grove resources associated with the Pod being prepared.

Introduce a typed context:

```go
type PodPreparationContext struct {
    PodCliqueSet *grovecorev1alpha1.PodCliqueSet
    PodClique    *grovecorev1alpha1.PodClique
    PodGang      *groveschedulerv1alpha1.PodGang
    PodGangName  string
}
```

Update the backend interface to:

```go
PreparePod(pod *corev1.Pod, preparation PodPreparationContext) error
```

The PodClique controller should construct this context and pass it to the selected scheduler backend. Scheduler-specific Pod labels, annotations, topology configuration, and related validation should be implemented by the backend instead of the generic controller.

### Why is this needed?

The current `PreparePod` method receives only the Pod. Backends that require information from the PodCliqueSet, PodClique, or PodGang cannot implement their preparation logic without adding scheduler-specific behavior to generic controllers.

For example, the Kueue backend needs this information to configure:

- Queue selection
- Pod group and prebuilt Workload references
- Pod group total count
- PodSet role identity
- Required topology
- The underlying scheduler

Passing a typed preparation context keeps scheduler-specific behavior within each backend, removes backend-name conditionals from generic controllers, and avoids propagating backend-specific metadata through intermediate Grove resources.

alaypatel07:
## Summary
- add `PreparePodContext` to provide PodCliqueSet, PodClique, and PodGang data to scheduler backends
- pass the PodGang already loaded by the PodClique sync flow without adding another API request
- update backend implementations, test fakes, framework documentation, and controller coverage

Fixes #731

## Test plan
- [x] `go test ./... -count=1` from `operator/`
- [x] focused scheduler, PodClique Pod controller, and ClusterTopology controller tests
- [ ] CI end-to-end tests

### DIFF
--- a/operator/internal/scheduler/types.go
+++ b/operator/internal/scheduler/types.go
@@ -46,10 +46,15 @@ type Backend interface {
 
 	// PreparePod adds scheduler-backend-specific configuration to the given Pod object
 	// prior to its creation (schedulerName, annotations, etc.).
-	PreparePod(pod *corev1.Pod) error
+	PreparePod(pod *corev1.Pod, preparation PreparePodContext) error
 
 	// ValidatePodCliqueSet runs scheduler-specific validations on the PodCliqueSet (e.g. TAS required but not supported).
 	ValidatePodCliqueSet(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error
+}
+
+// PreparePodContext contains the PodGang associated with a Pod being prepared.
+type PreparePodContext struct {
+	PodGang *groveschedulerv1alpha1.PodGang
 }
 
 // TopologyAwareBackend is an optional interface that Backend
--- a/docs/proposals/375-scheduler-backend-framework/README.md
+++ b/docs/proposals/375-scheduler-backend-framework/README.md
@@ -157,6 +157,10 @@ The core of the framework is the `Backend` interface, which defines all operatio
 ```go
 package scheduler
 
+type PreparePodContext struct {
+	PodGang *groveschedulerv1alpha1.PodGang
+}
+
 // SchedulerBackend defines the interface that different scheduler backends must implement.
 //
 // Architecture: Backend validates PodCliqueSet at admission, converts
@@ -192,7 +196,7 @@ type Backend interface {
 	// prior to its creation. This includes setting schedulerName,
 	// annotations, etc.
 	// This is called during Pod creation in the PodClique controller.
-	PreparePod(pod *corev1.Pod)
+	PreparePod(pod *corev1.Pod, preparation PreparePodContext) error
 
 	// ValidatePodCliqueSet provides an ability to the scheduler backends to run additional
 	// validations on the PodCliqueSet resource. For example - if a scheduler does not yet support
--- a/operator/internal/controller/clustertopology/reconciler_test.go
+++ b/operator/internal/controller/clustertopology/reconciler_test.go
@@ -83,7 +83,9 @@ func (f *fakeBackend) SyncPodGang(_ context.Context, _ *groveschedulerv1alpha1.P
 	return nil
 }
 
-func (f *fakeBackend) PreparePod(_ *corev1.Pod) error { return nil }
+func (f *fakeBackend) PreparePod(_ *corev1.Pod, _ scheduler.PreparePodContext) error {
+	return nil
+}
 
 func (f *fakeBackend) ValidatePodCliqueSet(_ context.Context, _ *grovecorev1alpha1.PodCliqueSet) error {
 	return nil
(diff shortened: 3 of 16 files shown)


danbar2 (dismissed):
LGTM

danbar2:
I think @enoodle will be glad to review that as well

operator/internal/scheduler/types.go
```diff
@@ -46,12 +46,19 @@ type Backend interface {
...
 	// prior to its creation (schedulerName, annotations, etc.).
-	PreparePod(pod *corev1.Pod) error
+	PreparePod(pod *corev1.Pod, preparation PreparePodContext) error
 
 	// ValidatePodCliqueSet runs scheduler-specific validations on the PodCliqueSet (e.g. TAS required but not supported).
 	ValidatePodCliqueSet(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) error
 }
 
+// PreparePodContext contains the Grove resources associated with a Pod being prepared.
+type PreparePodContext struct {
```
unmarshall, line 56: Why is this required? In this PR is it not clear on how this is going to be used.
None of the scheduler backends that  this PR modifies, namely - `kai`, `kube`, `lpx` and `volcano` ever use this. This entire PR is hinged around this new type as most of the change are just to propagate this new type around. 

In the issue description you mention:
```
For example, the Kueue backend needs this information to configure:

Queue selection
Pod group and prebuilt Workload references
Pod group total count
PodSet role identity
Required topology
The underlying scheduler
```
Can you please help map the above to PCS, PCLQ and PodGang that are included in this type?

Also except Keueue none of the backends need this, so why wire this through the API at all?

unmarshall (changes requested):

operator/internal/scheduler/types.go
enoodle: @unmarshall The context is https://github.com/ai-dynamo/grove/pull/707  - this is needed to prevent spreading the Kueue-specific code all over.
The other alternative is to use the labels on the pod and api "calls" (because they will be cached hopefully) to get these objects, but I think it is much more awkward than just passing them to the API.
KAI's scheduler backend could use that too

unmarshall: @enoodle when I look at #707 i spot something quite weird:
It says it is a backend scheduler and also defines `SchedulerNameKeueue` but then it goes on and creates `KueueSchedulerConfiguration` which has `UnderlyingSchedulerName` - which then immediately nullifies `SchedulerNameKeueue`. 

When i look at `PreparePod` and `SyncPodGang` then i can gather that you need the following information:
* Queue name - this is a label that is added on PCS.
* PodGang name - this is the label that is present on the Pod
* PodClique name - this is present as an owner reference on the Pod
* PodGang resource - to get the effective required topology constraint for each PodGroup within a PodGang.
* Total number of pods for a PodGang - but shouldn't this be per `PodGroup` as per https://github.com/kubernetes-sigs/kueue/blob/main/keps/976-plain-pods/README.md? - for this you rely only on PCS which i am not sure it is correct. Today there are BPG and one or more SPGs. So either the function name `podGangTotalPodCount` is not named correctly or its implementation is incorrect. Need to understand this better to comment further.

What has been introduced is:
```go
type PreparePodContext struct {
	PodCliqueSet *grovecorev1alpha1.PodCliqueSet
	PodClique    *grovecorev1alpha1.PodClique
	PodGang      *groveschedulerv1alpha1.PodGang
}
```
* You are already making a GET call in SyncPodGang to get the PCS, so why do you need to have PCS as part of `PreparePodContext`, you can make another GET call (just a local call to an informer cache)?
* For PCLQ - you only need its name - its there in the Pod resource metadata, take it from there.
* This leaves PodGang resource.

alaypatel07: @unmarshall Thanks for the detailed explnation, I agree, we can keep the context small and use cached get inside the PreparePod call for getting PCS.

Updated to just pass the pod gang. Still kept it in PreparePodContext struct if it needs expansion in the future.

alaypatel07: POC: https://github.com/ai-dynamo/grove/pull/707/changes/e251dec76dac2586dea85e4d39830303aaa40050

PR 707 is also updated to reflect this.

alaypatel07:
Closing this, I can get PodGang using a local call get call same as PCS this PR is not needed anymore.

PR #178, opened 2025-09-08
Rolling update for `PodGangSet` and its child resources.

unmarshall:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:
Introduces rolling update for `PodGangSet` resources.
> NOTE: Process of rolling update will be described in a separate doc and not as part of the issue/PR description as it is a bit involved.

#### Which issue(s) this PR fixes:

Fixes #177 

#### Does this PR introduce a API change?

```release-note
Introduces rolling update for PodGangSet resources.
```

#### Additional documentation e.g., enhancement proposals, usage docs, etc.:

```docs

```

PR #725, opened 2026-07-20
feat: add KAI scheduler backend with one PodGroup per PodGang

enoodle:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Continuation of [Daisy's original KAI scheduler backend implementation (PR #524)](https://github.com/ai-dynamo/grove/pull/524).

Adds the KAI scheduler backend and moves KAI PodGroup lifecycle ownership into it.

The initial strategy deliberately preserves current behavior: each Grove PodGang is synchronized to exactly one KAI PodGroup. Pods are configured to join that PodGroup, and standard owner-reference garbage collection cleans it up.

#### Which issue(s) this PR fixes:

Partly implements #525 

#### Special notes for your reviewer:

This is PR1 of a stacked rollout. It intentionally excludes:

- persisted layout/version markers and upgrade coexistence behavior;
- one-PodGroup-per-PCS-replica aggregation;
- deletion finalizers for aggregate layouts; and
- runtime KAI version enforcement.

#### Does this PR introduce a API change?

```release-note
NONE
```

#### Additional documentation e.g., enhancement proposals, usage docs, etc.:

```docs
NONE
```

PR #734, opened 2026-07-25
test(e2e): watch RU-12 ordinal update transition

### LINKED ISSUE #733: [BUG] RU-12 polling misses transient CurrentlyUpdating transition
### What happened?

`Test_RU12_RollingUpdateWithPCSScaleInDuringUpdate` intermittently times out while waiting for:

```go
pcs.Status.UpdateProgress.CurrentlyUpdating[0].ReplicaIndex == 1
```

The test polls every 5 seconds, but `CurrentlyUpdating` is a transient, non-monotonic status. If ordinal 1 completes between polls, the status is cleared and a longer timeout cannot recover the missed transition.

### Provenance

This is a pre-existing test synchronization bug, not a regression introduced by #595:

- #178 (`ef01f6ea`, September 2025) introduced rolling update behavior. From its first implementation, healthy replicas were selected in ascending ordinal order and `CurrentlyUpdating` was cleared when the update completed.
- #280 (`1db7d083`, January 2026) introduced RU-12 and `waitForOrdinalUpdating`. The test incorrectly documented the order as highest-to-lowest, waited for ordinal 1, and observed the transient status using the default 5-second polling interval.
- #438, #526, and #576 changed status representation or refactored the waiter but did not introduce the underlying behavior.
- #595 does not change RU-12. It adds a second `rolling_updates_default-scheduler` matrix entry, increasing exposure of the existing flake.

RU-12 has therefore been structurally racy since it was added in #280.

### Reproduction

Tested #595 at `771efc2` on L20 with the default-scheduler profile, 30 KWOK workers, and the deployed Grove operator:

- 25 sequential RU-12 runs
- 17 PASS / 8 FAIL
- 32% observed failure rate
- all 8 failures reached the 60-second waiter timeout

Across the failed samples:

- ordinal 1 entered `CurrentlyUpdating` successfully;
- its observable window lasted only 0.318–0.434 seconds;
- the controller then completed the update and cleared `CurrentlyUpdating`;
- controller activity ended 54–59 seconds before the test timeout.

There were no leader-election flaps, operator restarts, API-server warnings, controller backoff/rate-limit messages, or Kubernetes Warning events.

The `object has been modified` reconcile conflict is not sufficient to explain the failures: it occurred in all 8 failures but also in most analyzed passing runs, and the controller recovered immediately.

### Why this differs from #503

#503 failed while `triggerPodCliqueRollingUpdate` was issuing GET/UPDATE operations. In this case, all three triggers succeed. The timeout happens afterward in `waitForOrdinalUpdating`, while controller logs show that the rolling update completed successfully.

### Root cause

Two assumptions introduced with RU-12 in #280 combine to create the flake:

1. The test says replicas update from highest to lowest, but the controller has always selected healthy replicas in ascending ordinal order (`0 -> 1`).
2. The test uses a 5-second GET polling interval to observe a status transition that can disappear in less than half a second.

The final `client rate limiter Wait returned an error: context deadline exceeded` message is misleading. It is emitted after the waiter has already missed the transition.

### Expected behavior

RU-12 should reliably synchronize with ordinal 1 entering `CurrentlyUpdating` and initiate scale-in without relying on polling timing.

### Proposed fix

Establish a field-selected PodCliqueSet watch from a known `resourceVersion` before triggering the update, then consume events until ordinal 1 enters `CurrentlyUpdating`.

A local implementation was validated on the same L20 environment:

- watch waiter unit tests: PASS
- update E2E package compile: PASS
- scale E2E package compatibility compile: PASS
- targeted RU-12 smoke run: PASS
- repeated RU-12 validation: **25 PASS / 0 FAIL**
- ordinal observation wait: 0.40–5.59 seconds

The watch captured two sub-second samples that a 5-second poll could readily miss. Increasing the timeout alone does not address the synchronization defect.

brluobt:
## Summary

- establish a field-selected PodCliqueSet watch before RU-12 triggers the rolling update
- consume every watch event until ordinal 1 enters `CurrentlyUpdating`
- add a reusable watch waiter with cancellation, closure, API error, and type-safety handling
- preserve the existing E2E client interface while adding controller-runtime watch support
- correct the RU-12 comment to reflect ascending healthy-replica order

## Root cause

RU-12 has been structurally racy since it was introduced in #280:

- #178 implemented rolling updates with healthy replicas selected in ascending ordinal order and cleared `CurrentlyUpdating` when the update completed.
- #280 documented the order as highest-to-lowest, waited for ordinal 1, and observed that transient status with the default 5-second polling interval.
- On L20 failure samples, ordinal 1 remained observable for only 0.318–0.434 seconds. Once the poll missed that transition, increasing the timeout could not recover it.

#595 does not modify RU-12. It adds a second rolling-update scheduler matrix entry and therefore increases exposure of the pre-existing flake.

## Impact

The test now observes the transition itself instead of relying on polling phase. This removes the misleading 60-second waiter timeout after the controller has already completed the rolling update.

## Validation

Original L20 reproduction using the #595 default-scheduler profile:

- before: **17 PASS / 8 FAIL** across 25 sequential RU-12 runs
- after watch fix: **25 PASS / 0 FAIL**
- two observed waits were sub-second (0.40s and 0.48s), which a 5-second poll could readily miss

Clean branch from latest `upstream/main` (`19905d5`) using the native KAI profile:

- `git diff upstream/main...HEAD --check`: PASS
- `go test -tags=e2e ./e2e/waiter`: PASS
- update E2E package compile: PASS
- scale E2E package compatibility compile: PASS
- targeted RU-12 smoke run: PASS (39.26s)
- repeated RU-12 validation: **25 PASS / 0 FAIL**
- ordinal observation wait: min 2.87s, mean 7.12s, median 7.24s, max 8.00s
- runs reaching the 30-second diagnostic threshold: 0

This branch contains one commit on top of `upstream/main` and intentionally excludes the QPS/burst and RU-11 timeout changes from #595.

Fixes #733

### DIFF
--- a/operator/e2e/tests/update/utils.go
+++ b/operator/e2e/tests/update/utils.go
@@ -502,6 +502,16 @@ func waitForRollingUpdate(tc *testctx.TestContext, expectedReplicas int32) <-cha
 // ordinalUpdateObserver watches a PodCliqueSet for a specific ordinal entering
 // CurrentlyUpdating. The watch must be established before triggering the update
 // so a short-lived transition cannot occur between observations.
+//
+// Lifecycle note: newOrdinalUpdateObserver deliberately hands ownership of the
+// timeout context and its cancel func to the returned observer. Callers keep
+// the observer alive across the mutation window and later call Wait() (which
+// reads from a channel bound to the same context) before finally calling
+// Stop(). Do NOT collapse the constructor's cancel into a `defer cancel()` at
+// the top of the function: that would fire cancel at constructor return time,
+// leaving the caller with an already-Done context and a watcher whose result
+// channel is closed before Wait() ever runs. See observer_defer_cancel_test.go
+// for a runnable proof of this invariant.
 type ordinalUpdateObserver struct {
 	ctx     context.Context
 	cancel  context.CancelFunc
@@ -512,6 +522,11 @@ type ordinalUpdateObserver struct {
 
 func newOrdinalUpdateObserver(tc *testctx.TestContext, ordinal int32) (*ordinalUpdateObserver, error) {
 	pcsName := tc.Workload.Name
+	// NOTE: cancel is intentionally NOT deferred here. It is either invoked on
+	// the error paths below, or handed to the returned observer to be run from
+	// Stop() after Wait() completes. Adding `defer cancel()` here would break
+	// Wait(): the caller only invokes Wait() after triggering the update, and
+	// by then this context (and the watcher bound to it) would already be dead.
 	timeoutCtx, cancel := context.WithTimeout(tc.Ctx, tc.Timeout)
 
 	var pcs grovev1alpha1.PodCliqueSet
--- a/operator/e2e/tests/update/utils.go
+++ b/operator/e2e/tests/update/utils.go
@@ -35,6 +35,7 @@ import (
 	"github.com/ai-dynamo/grove/operator/e2e/waiter"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/util/retry"
@@ -498,41 +499,78 @@ func waitForRollingUpdate(tc *testctx.TestContext, expectedReplicas int32) <-cha
 	return errCh
 }
 
-// waitForOrdinalUpdating waits for a specific ordinal to start being updated during rolling update.
-// Uses tc.Workload.Name as the PCS name and tc.Timeout for the timeout (use a modified tc if a different timeout is needed).
-func waitForOrdinalUpdating(tc *testctx.TestContext, ordinal int32) error {
+// ordinalUpdateObserver watches a PodCliqueSet for a specific ordinal entering
+// CurrentlyUpdating. The watch must be established before triggering the update
+// so a short-lived transition cannot occur between observations.
+type ordinalUpdateObserver struct {
+	ctx     context.Context
+	cancel  context.CancelFunc
+	watcher watch.Interface
+	pcsName string
+	ordinal int32
+}
+
+func newOrdinalUpdateObserver(tc *testctx.TestContext, ordinal int32) (*ordinalUpdateObserver, error) {
 	pcsName := tc.Workload.Name
+	timeoutCtx, cancel := context.WithTimeout(tc.Ctx, tc.Timeout)
 
-	pollCount := 0
-	fetchPCS := waiter.FetchByName(pcsName, k8sclient.Getter[*grovev1alpha1.PodCliqueSet](tc.Client, tc.Namespace))
-	predicate := waiter.Predicate[*grovev1alpha1.PodCliqueSet](func(pcs *grovev1alpha1.PodCliqueSet) bool {
-		pollCount++
+	var pcs grovev1alpha1.PodCliqueSet
+	if err := tc.Client.Get(
+		timeoutCtx,
+		types.NamespacedName{Name: pcsName, Namespace: tc.Namespace},
+		&pcs,
+	); err != nil {
+		cancel()
+		return nil, fmt.Errorf("get PodCliqueSet %s before watching: %w", pcsName, err)
+	}
 
-		// Log status every few polls for debugging
-		if pollCount%3 == 1 {
+	pcsWatch, err := tc.Client.Watch(
+		timeoutCtx,
+		&grovev1alpha1.PodCliqueSetList{},
+		&client.ListOptions{
+			Namespace:     tc.Namespace,
+			FieldSelector: fields.OneTermEqualSelector("metadata.name", pcsName),
+			Raw:           &metav1.ListOptions{ResourceVersion: pcs.ResourceVersion},
+		},
+	)
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("watch PodCliqueSet %s: %w", pcsName, err)
+	}
+
+	return &ordinalUpdateObserver{
+		ctx:     timeoutCtx,
+		cancel:  cancel,
+		watcher: pcsWatch,
+		pcsName: pcsName,
+		ordinal: ordinal,
+	}, nil
+}
+
+func (o *ordinalUpdateObserver) Wait() error {
+	_, err := waiter.WaitForWatchEvent(
+		o.ctx,
+		o.watcher.ResultChan(),
+		func(pcs *grovev1alpha1.PodCliqueSet) bool {
 			currentOrdinal := int32(-1)
 			if pcs.Status.UpdateProgress != nil && len(pcs.Status.UpdateProgress.CurrentlyUpdating) > 0 {
 				currentOrdinal = pcs.Status.UpdateProgress.CurrentlyUpdating[0].ReplicaIndex
 			}
-			tests.Logger.Debugf("[waitForOrdinalUpdating] Poll #%d: waiting for ordinal %d, currently updating ordinal: %d",
-				pollCount, ordinal, currentOrdinal)
-		}
-
-		// Check if the target ordinal is currently being updated
-		if pcs.Status.UpdateProgress != nil &&
-			len(pcs.Status.UpdateProgress.CurrentlyUpdating) > 0 &&
-			pcs.Status.UpdateProgress.CurrentlyUpdating[0].ReplicaIndex == ordinal {
-			tests.Logger.Debugf("[waitForOrdinalUpdating] Ordinal %d started updating after %d polls", ordinal, pollCount)
-			return true
-		}
-
-		return false
-	})
-	w := waiter.New[*grovev1alpha1.PodCliqueSet]().
-		WithTimeout(tc.Timeout).
-		WithInterval(tc.Interval)
-	err := w.WaitUntil(tc.Ctx, fetchPCS, predicate)
+			tests.Logger.Debugf(
+				"[ordinalUpdateObserver] Waiting for %s ordinal %d, currently updating ordinal: %d",
+				o.pcsName,
+				o.ordinal,
+				currentOrdinal,
+			)
+			return currentOrdinal == o.ordinal
+		},
+	)
 	return err
+}
+
+func (o *ordinalUpdateObserver) Stop() {
+	o.watcher.Stop()
+	o.cancel()
 }
 
 // getPodIdentifier returns a stable identifier for a pod based on its logical position in the workload.
--- a/operator/e2e/tests/update/utils.go
+++ b/operator/e2e/tests/update/utils.go
@@ -506,6 +506,16 @@ func waitForRollingUpdate(tc *testctx.TestContext, expectedReplicas int32) <-cha
 // ordinalUpdateObserver watches a PodCliqueSet for a specific ordinal entering
 // CurrentlyUpdating. The watch must be established before triggering the update
 // so a short-lived transition cannot occur between observations.
+//
+// Lifecycle note: newOrdinalUpdateObserver deliberately hands ownership of the
+// timeout context and its cancel func to the returned observer. Callers keep
+// the observer alive across the mutation window and later call Wait() (which
+// reads from a channel bound to the same context) before finally calling
+// Stop(). Do NOT collapse the constructor's cancel into a `defer cancel()` at
+// the top of the function: that would fire cancel at constructor return time,
+// leaving the caller with an already-Done context and a watcher whose result
+// channel is closed before Wait() ever runs. See observer_defer_cancel_test.go
+// for a runnable proof of this invariant.
 type ordinalUpdateObserver struct {
 	ctx     context.Context
 	cancel  context.CancelFunc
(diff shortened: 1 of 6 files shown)


kangclzjc:
Is this because that kai scheduler use gang to group schedule and then kai cost more time than default scheduler? If that is the case, then probably we should modify test or let default scheduler support gang too?

brluobt:
Thanks for the question. This is not caused by KAI gang scheduling taking longer, and it does not require adding gang scheduling behavior to the default scheduler.

The race is in the RU-12 E2E waiter: it polls the transient `CurrentlyUpdating` status every 5 seconds, while that status was observed to remain visible for only 0.3–0.4 seconds. Both scheduler profiles execute the same racy waiter; scheduler timing only changes the probability of hitting it.

#595 exposed the pre-existing race by adding the default-scheduler matrix entry. This PR fixes the E2E synchronization by replacing polling with a watch; it does not change either scheduler.

Validation was 17/25 → 25/25 on the #595 default-scheduler environment, and 25/25 on a clean main-based KAI environment.

brluobt:
## Post-rebase L20 validation — 25/25 PASS

Rebased onto latest `upstream/main` and revalidated on L20 against the new base (which includes the KAI backend rework in `ddc2e60` / #725). While that change has no textual conflict with this PR, it can affect RU-12 event timing, so the previous 25/25 on base `19905d5` is no longer authoritative.

### Environment
- HEAD: `3e4df18`
- Base: `92707b8` (`operator/helm: enable health probes by default`, #743)
- Profile: native KAI E2E
- Cluster: L20, 30 KWOK workers
- Operator image built from `3e4df18` and confirmed running in the cluster

### Static & compile
- `git diff upstream/main...HEAD --check`: PASS
- `go test -tags=e2e ./e2e/waiter`: PASS
- update E2E package compile: PASS
- scale E2E package compatibility compile: PASS
- targeted RU-12 smoke run: PASS

### RU-12 distribution (25 sequential runs)
- **25 PASS / 0 FAIL**
- Ordinal observation wait: min **6.45s**, mean **7.01s**, median **6.94s**, max **7.91s**
- Runs with wait < 20s: **25 / 25**
- Runs reaching the 30s diagnostic threshold: **0 / 25**
- Per-run wall time: 39–46s

### Operator behavior
- No unexpected operator restarts during the 25-run distribution.
- One `exit code 0` restart at initial deploy, from the first-time webhook certificate generation (expected).
- Cleanup phase produced the known optimistic-lock / not-found reconcile noise, which converged in every run and did not affect results.
- No residual Grove resources after each run.

### Delta vs previous validation
Prior 25/25 was measured on base `19905d5` under the #595 default-scheduler profile. This round targets the new KAI backend on latest main and reproduces the same 25/25 outcome, confirming the watch-based waiter remains stable across the new backend rework.

brluobt:
### CI status update: `E2E - auto_mnnvl` failure is unrelated to this PR

Latest full CI on this PR (after flipping to Ready For Review, run [30869041746](https://github.com/ai-dynamo/grove/actions/runs/30869041746)):

- ✅ `test` / `build` / `check` / `changes` / `validate-pr-issues`
- ✅ `E2E - rolling_updates` **and** `E2E - rolling_updates_default-scheduler` (this is what this PR actually fixes)
- ✅ `E2E - ondelete_updates`, `gang_scheduling`, `gang_termination`, `startup_ordering`, `Topology_Aware_Scheduling`, `cert_management`, `crd_installer`, `resource_sharing`
- ❌ `E2E - auto_mnnvl` — fails in `Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing` with the exact 2-minute preflight-log timeout from #449

A targeted rerun of only that job on the same commit reproduces the same failure ([job 91923153142](https://github.com/ai-dynamo/grove/actions/runs/30869041746/job/91923153142)), so this is not a one-off CI flake — it looks like a regression / incomplete fix of #449 / #451.

This PR only touches `operator/internal/controller/podcliqueset_component/podclique/sync.go` (the RU-12 ordinal watch fix). It does not touch operator startup, preflight, auto-MNNVL, ComputeDomain, or the e2e suite. The AutoMNNVL failure is on a completely orthogonal code path.

Filed as a separate issue so it can be fixed independently and does **not** get bundled into this PR:

- #749 — Flaky e2e (regression): `Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing` fails deterministically on main; diagnostics artifact path also broken

Summary for reviewers: RU-12 rolling-update fix is green in CI on both `rolling_updates` and `rolling_updates_default-scheduler`. The only red job is an independent AutoMNNVL regression tracked in #749.

brluobt:
### CI status update (fa0a5bd)

Quick note on the current CI on `fa0a5bd` since it looks unstable at first glance:

**Rolling updates job — red, but not RU-12:**

The only failing test in `E2E - rolling_updates` is **`Test_RU11_RollingUpdateWithPCSScaleOut`**, which is a known flake tracked in **#711** (open, same signature `client rate limiter Wait returned an error: context deadline exceeded`, filed by @yankay with 4 prior CI reproductions on unrelated PRs). RU-11 uses the pre-existing polling path — this PR does not modify RU-11 or the code it exercises.

**RU-12 (the test this PR fixes) passed:**

```
--- PASS: Test_RU12_RollingUpdateWithPCSScaleInDuringUpdate (37.77s)
```

Job: [E2E - rolling_updates @ 2026-08-06 07:34 UTC](https://github.com/ai-dynamo/grove/actions/runs/31081045736/job/92550557365) — RU-7/8/9/10 PASS → RU-11 FAIL (#711) → **RU-12 PASS** → RU-16/18/19/20 PASS.

**auto_mnnvl (from #749) — passed this run:**

Previously red on the pre-rebase HEAD (3e4df18). After rebasing onto current main (base `814f57b`), the run at 07:34 UTC on `fa0a5bd` came back green. No auto-mnnvl code changed upstream between the two bases, so this may still be the intermittent flake described in #749; but at least it is not a determinstic failure on this branch.

**Summary:** rebase-only from `3e4df18` (5 source files SHA-identical to the previously L20-verified 25/25 PASS commit), no code changes since review-ready. I'll trigger a rerun of `rolling_updates` to clear the #711 flake and confirm.

brluobt:
### L20 stability distribution — rebased head `e79c5e6`

Following @shayasoolin's `/ok to test` (thanks!), official CI is now green on this SHA including `E2E - rolling_updates`. In addition, I ran a 25× stability distribution on L20 to confirm the watch-based fix is not just a lucky single pass.

**Setup**
- HEAD: `e79c5e63c4346608fd800580507d03efc27913c0`
- Base: `447a5fe5e4767610aa56bc9ae9cdd5195c73d37d` (upstream/main, includes #757)
- Cluster: shared-e2e-test-cluster, 30 KWOK workers
- Scheduler profile: KAI
- Operator: built from this PR's source

**Sanity**

| Check | Result |
|---|---|
| `go build ./...` | ✅ PASS |
| `go vet ./e2e/...` | ✅ PASS |
| waiter unit tests | ✅ PASS |
| RU-12 smoke (1×) | ✅ PASS, 38.40s |

**RU-12 distribution (25× with `-count=1`)**
- **PASS: 25/25** — FAIL: 0/25 — flakes: 0
- Per-run wall time: min 38.26s / mean 42.86s / max 44.32s

**`ordinalObserver.Wait()` latency distribution (n=25)**

| stat | value |
|---|---|
| min | 1.764s |
| median | 6.734s |
| mean | 6.600s |
| max | 7.509s |
| ≤ 0.5s | 0 |
| ≤ 1.0s | 0 |
| ≥ 5.0s | 24 |

The observer reliably catches ordinal 1's `CurrentlyUpdating` transition — the original polling race (which relied on catching a ~0.3–0.4s transient with a 5s poll) is not reproducible across 25 consecutive runs on the rebased head.

Ready for review whenever a maintainer has a moment.

shayasoolin (dismissed):
LGTM, thanks!

operator/e2e/tests/update/utils.go
```diff
@@ -498,43 +499,80 @@ func waitForRollingUpdate(tc *testctx.TestContext, expectedReplicas int32) <-cha
...
+	ctx     context.Context
+	cancel  context.CancelFunc
+	watcher watch.Interface
+	pcsName string
+	ordinal int32
+}
+
+func newOrdinalUpdateObserver(tc *testctx.TestContext, ordinal int32) (*ordinalUpdateObserver, error) {
 	pcsName := tc.Workload.Name
+	timeoutCtx, cancel := context.WithTimeout(tc.Ctx, tc.Timeout)
```
Ronkahn21, line 535:
```suggestion
	timeoutCtx, cancel := context.WithTimeout(tc.Ctx, tc.Timeout)
	defer cancel()

```

brluobt: Thanks for the review, @Ronkahn21! I actually tried this simplification first and it turned out to break `Wait()` in a subtle way — flagging it here in case it's useful for future readers.

The observer's `Wait()` reads from a `watch.Interface` channel that is bound to `timeoutCtx`. The caller pattern is:

```go
obs, _ := newOrdinalUpdateObserver(&tcOrdinalTimeout, 1)
defer obs.Stop()
triggerPodCliqueUpdate(...)   // takes real time
obs.Wait()                    // <-- reads the ctx-bound channel here
```

If we `defer cancel()` inside the constructor, cancel fires the instant the constructor returns, so by the time the caller reaches `Wait()`:

- `timeoutCtx.Err()` is already `context.Canceled`, and
- the underlying watcher channel is closed by apimachinery.

`Wait()` then returns immediately with either a canceled-context error or `"watch closed before condition was met"`, instead of observing the ordinal transition.

To make this concrete I pushed a self-contained unit test in `operator/e2e/tests/update/observer_defer_cancel_test.go` (no e2e build tag, runs under plain `go test`). It exercises both patterns side by side. Locally:

```
$ go test -v -run '^TestObserver_' ./operator/e2e/tests/update/
=== RUN   TestObserver_CancelOnStop_ExhibitsCorrectBehavior
--- PASS: TestObserver_CancelOnStop_ExhibitsCorrectBehavior (0.05s)
=== RUN   TestObserver_DeferCancel_BreaksWait
    defer-cancel variant failed as expected: watch closed before condition was met
--- PASS: TestObserver_DeferCancel_BreaksWait (0.02s)
=== RUN   TestObserver_DeferCancel_ContextAlreadyDoneOnReturn
    as expected, obs.ctx.Err() = context canceled right after constructor return
--- PASS: TestObserver_DeferCancel_ContextAlreadyDoneOnReturn (0.00s)
=== RUN   TestObserver_CancelOnStop_ContextAliveOnReturn
--- PASS: TestObserver_CancelOnStop_ContextAliveOnReturn (0.00s)
PASS
```

I also added a doc comment on `ordinalUpdateObserver` and an inline note above the `context.WithTimeout` call explaining why cancel is intentionally handed to `Stop()` rather than deferred. Commit: fdd3bd4.

If you'd prefer a different structural approach (e.g. splitting the constructor so the cancel/ctx are constructed by the caller and passed in), happy to explore that direction.

Ronkahn21 (dismissed):

Ronkahn21:
/ok to test https://github.com/ai-dynamo/grove/pull/734/changes/fdd3bd42e6d228bc194345e7bfd25a57444ee5b6

shayasoolin (dismissed):

brluobt:
Rebased onto latest `upstream/main` (`af2df1f`) to clear `mergeable_state=dirty` before merge.

**Rebase scope (minimal, tests-only):**

- Trivial 3-way merge marker in `operator/e2e/tests/update/utils.go` around `newOrdinalUpdateObserver` — resolved to keep the watch-based observer this PR introduces (no semantic change).
- Adapted a new call site in `Test_RU14_RollingUpdateWithPCSGScaleOutDuringUpdate` (introduced by #778 after this PR was approved) that still called the old `waitForOrdinalUpdating` function this PR deletes — switched it to `newOrdinalUpdateObserver` + `Wait()` for consistency, preserving the **watch-before-trigger** invariant this PR relies on (see the lifecycle doc comment on `ordinalUpdateObserver` at line ~506 of `utils.go`).
- **No changes to production code, no changes to observer semantics, no changes to the 4 unit tests that back the earlier discussion with @Ronkahn21.**

**Diff vs the previously-approved `fdd3bd4`:**

```
git diff fdd3bd4..HEAD -- operator/e2e/tests/update/rolling_recreate_test.go
```

...shows only the ~10 line RU-14 adaptation described above.

**Verified before pushing (exact tree, not just "equivalent rebase"):**
- Tree SHA validated: `efe076b7d7de4151799e1dd6a9a67d59cc9e7f24` (HEAD), `bca2fcde2e8af13b2b56b1bb0a5bde827a04c802` (HEAD~1)
- `go build ./...` clean
- `go build -tags=e2e ./e2e/...` clean
- 4 `TestObserver_*` unit tests still PASS on the rebased tree
- `go vet -tags=e2e ./e2e/tests/update/...` clean

@Ronkahn21 @shayasoolin — sorry to reset the approval; would appreciate a re-stamp when convenient 🙏

@danbar2 @unmarshall — 2 approves were in on the previous SHA with all E2E matrix jobs green (rolling_updates, auto_mnnvl, gang_scheduling, gang_termination, startup_ordering, ondelete_updates, Topology_Aware_Scheduling, resource_sharing, cert_management, crd_installer, upgrade_from_latest_release). Once re-approve lands this should be a clean rebase-and-merge.

shayasoolin (approved):

Ronkahn21 (approved):

Ronkahn21:
/ok to test https://github.com/ai-dynamo/grove/commit/2b315d3cae32b777aaa3d5f8cf8377dfcec4c57a

PR #736, opened 2026-07-27
fix: prevent panic in ExtractScalingGroupNameFromPCSGFQN and reuse

### LINKED ISSUE #737: ExtractScalingGroupNameFromPCSGFQN may panic on unexpected label values
### Describe the bug

`ExtractScalingGroupNameFromPCSGFQN` performs direct string slicing on the FQN without validating whether the expected prefix actually matches. When the function is called with label values that do not form the expected prefix, it can cause a **slice bounds out of range panic**.

### To Reproduce

1. A `PodClique` has a `PodCliqueScalingGroup` name or replica index label that does not match the expected prefix pattern in its resource name
2. `GetPodCliqueNameFromPodCliqueFQN` calls `ExtractScalingGroupNameFromPCSGFQN` with the mismatched values
3. The direct string slicing `pcsgFQN[len(prefix):]` panics if `len(pcsgFQN) <= len(prefix)` or produces a silently incorrect result

### Expected behavior

The function should validate the prefix before slicing and return an error if the FQN does not have the expected format, instead of panicking or returning incorrect data.

### Actual behavior

- Potential **panic**: slice bounds out of range when FQN is shorter than the prefix
- Potential **silent data corruption**: returns a substring that is not the actual scaling group name when the prefix does not match

### Additional context

The fix has been implemented in PR #736.

wzshiming:
#### What type of PR is this?

/kind bug

- prevent panic in ExtractScalingGroupNameFromPCSGFQN with input validation and resue
- simplify function signature and reuse in componentname.go

Fixes #737

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/api/common/namegen.go
+++ b/operator/api/common/namegen.go
@@ -16,6 +16,7 @@ package common
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
 )
@@ -124,7 +125,10 @@ func GeneratePodGangNameForPodCliqueOwnedByPCSG(pcs *v1alpha1.PodCliqueSet, pcsR
 
 // ExtractScalingGroupNameFromPCSGFQN extracts the scaling group name from a PodCliqueScalingGroup FQN.
 // For example, "simple1-0-sga" with pcsNameReplica="simple1-0" returns "sga".
-func ExtractScalingGroupNameFromPCSGFQN(pcsgFQN string, pcsNameReplica ResourceNameReplica) string {
+func ExtractScalingGroupNameFromPCSGFQN(pcsgFQN string, pcsNameReplica ResourceNameReplica) (string, error) {
 	prefix := fmt.Sprintf("%s-%d-", pcsNameReplica.Name, pcsNameReplica.Replica)
-	return pcsgFQN[len(prefix):]
+	if !strings.HasPrefix(pcsgFQN, prefix) {
+		return "", fmt.Errorf("FQN %q does not have expected prefix %q", pcsgFQN, prefix)
+	}
+	return pcsgFQN[len(prefix):], nil
 }
--- a/operator/api/common/namegen_test.go
+++ b/operator/api/common/namegen_test.go
@@ -112,7 +112,8 @@ func TestExtractScalingGroupNameFromPCSGFQN(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			result := ExtractScalingGroupNameFromPCSGFQN(tt.pcsgName, tt.pcsNameReplica)
+			result, err := ExtractScalingGroupNameFromPCSGFQN(tt.pcsgName, tt.pcsNameReplica)
+			assert.NoError(t, err)
 			assert.Equal(t, tt.expected, result)
 		})
 	}
@@ -175,7 +176,8 @@ func TestExtractScalingGroupNameFromPCSGFQN_Consistency(t *testing.T) {
 			generatedPCSGName := GeneratePodCliqueScalingGroupName(tc.pcsNameReplica, tc.scalingGroupName)
 
 			// Extract scaling group name back
-			extractedScalingGroupName := ExtractScalingGroupNameFromPCSGFQN(generatedPCSGName, tc.pcsNameReplica)
+			extractedScalingGroupName, err := ExtractScalingGroupNameFromPCSGFQN(generatedPCSGName, tc.pcsNameReplica)
+			assert.NoError(t, err)
 
 			// They should match
 			assert.Equal(t, tc.scalingGroupName, extractedScalingGroupName)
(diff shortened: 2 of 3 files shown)


oleg-kushniriov:
Thanks for the PR. The defensive guard in `componentname.go` is a reasonable improvement, but the panic is not a real production risk - the controller always stamps labels consistent with the generated name, and Grove does not truncate resource names, so if an object exists its name is always longer than the constructed prefix. The only scenario that could trigger the panic is external label tampering.                                                                                                                                                        

Regarding the signature change in `ExtractScalingGroupNameFromPCSGFQN` — moving from `ResourceNameReplica` to loose `(name, index string)` is a type-safety regression. `ResourceNameReplica` already models this concept precisely and the rest of the naming API uses it consistently with a typed int replica. The reuse in `componentname.go` does not justify weakening the type contract across the board.

operator/api/common/namegen.go
```diff
@@ -124,7 +124,10 @@ func GeneratePodGangNameForPodCliqueOwnedByPCSG(pcs *v1alpha1.PodCliqueSet, pcsR
 
 // ExtractScalingGroupNameFromPCSGFQN extracts the scaling group name from a PodCliqueScalingGroup FQN.
 // For example, "simple1-0-sga" with pcsNameReplica="simple1-0" returns "sga".
-func ExtractScalingGroupNameFromPCSGFQN(pcsgFQN string, pcsNameReplica ResourceNameReplica) string {
-	prefix := fmt.Sprintf("%s-%d-", pcsNameReplica.Name, pcsNameReplica.Replica)
-	return pcsgFQN[len(prefix):]
+func ExtractScalingGroupNameFromPCSGFQN(pcsgFQN, name, index string) (string, error) {
```
oleg-kushniriov: `ResourceNameReplica` already models this concept — the reuse in `componentname.go` does not justify weakening the type contract here

wzshiming: ~In fact, this function has not been used yet~

~[I saw a possible panic risk in GetPodCliqueNameFromPodCliqueFQN](https://github.com/ai-dynamo/grove/blob/19905d530d48bdabe3f1de6a8a175544f941d81e/operator/internal/utils/componentname.go#L39-L64), and looked for an existing function, and found this one, which looks like there was a change from `ResourceNameReplica` to `metav1.ObjectMeta` at some point, and the function wasn't cleaned up.~

~So if type assertions are not used, then this function can no longer use ResourceNameReplica now.~

~Then I should use assertions to keep this parameter type, or add a private function in `componentname.go`?~

wzshiming: Done, I misunderstood just now, now I've changed it according to what you said.

operator/api/common/namegen.go
```diff
@@ -124,7 +124,10 @@ func GeneratePodGangNameForPodCliqueOwnedByPCSG(pcs *v1alpha1.PodCliqueSet, pcsR
 
 // ExtractScalingGroupNameFromPCSGFQN extracts the scaling group name from a PodCliqueScalingGroup FQN.
 // For example, "simple1-0-sga" with pcsNameReplica="simple1-0" returns "sga".
-func ExtractScalingGroupNameFromPCSGFQN(pcsgFQN string, pcsNameReplica ResourceNameReplica) string {
+func ExtractScalingGroupNameFromPCSGFQN(pcsgFQN string, pcsNameReplica ResourceNameReplica) (string, error) {
 	prefix := fmt.Sprintf("%s-%d-", pcsNameReplica.Name, pcsNameReplica.Replica)
-	return pcsgFQN[len(prefix):]
+	if len(pcsgFQN) <= len(prefix) || pcsgFQN[:len(prefix)] != prefix {
```
oleg-kushniriov: consider using `strings.HasPrefix(pcsgFQN, prefix)` - it is cleaner and more readable than the two-condition check

operator/api/common/namegen_test.go
```diff
@@ -112,7 +112,7 @@ func TestExtractScalingGroupNameFromPCSGFQN(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			result := ExtractScalingGroupNameFromPCSGFQN(tt.pcsgName, tt.pcsNameReplica)
+			result, _ := ExtractScalingGroupNameFromPCSGFQN(tt.pcsgName, tt.pcsNameReplica)
```
oleg-kushniriov: Please assert `err == nil`. Silently discarding the error defeats the purpose of adding it.

operator/internal/utils/componentname.go
```diff
@@ -59,6 +62,9 @@ func GetPodCliqueNameFromPodCliqueFQN(pclqObjectMeta metav1.ObjectMeta) (string,
 	if !ok {
 		return "", fmt.Errorf("missing label %s on PodClique: %v", apicommon.LabelPodCliqueSetReplicaIndex, pclqObjectKey)
 	}
-	pclqNamePrefix := fmt.Sprintf("%s-%s-", pcsName, pcsReplicaIndex)
-	return pclqObjectMeta.Name[len(pclqNamePrefix):], nil
+	pcsReplicaIndexInt, err := strconv.Atoi(pcsReplicaIndex)
```
oleg-kushniriov, line 65: 👍

oleg-kushniriov (commented):
well done. several small comments

wzshiming:
It's done

danbar2 (approved):

oleg-kushniriov (approved):

PR #738, opened 2026-07-27
fix: avoid redundant secret create request when secret already exists

### LINKED ISSUE #739: Avoid redundant secret create request when secret already exists
### What happened?

https://github.com/wzshiming/grove/blob/2197ac698a10dcfbdfbceb5e65aa40b2071e17d7/operator/internal/controller/cert/cert.go#L71-L108

In `createPlaceholderSecretIfNotExists()`, when the webhook TLS secret already
exists (e.g., created by another replica in an HA deployment), the function
currently falls through to `Create()` after a successful `Get()`, causing a
redundant API request and a log error (even though the `AlreadyExists` error
is handled gracefully).

Add an early return after a successful `Get()` to skip the `Create()` call
when the secret already exists.

In HA deployments (`replicaCount > 1`) or during operator restarts, two
replicas can race between `Get` (not found) and `Create`. Currently:

1. Replica A: `Get` → NotFound → `Create` → success
2. Replica B: `Get` → NotFound → `Create` → AlreadyExists (handled, but
   produces an error log and an unnecessary API call)

With this fix, the second replica's `Get` will find the secret created by
the first replica and return early, avoiding the redundant `Create` call
entirely. This reduces API server load and eliminates spurious error logs
in HA deployments.

### Environment

- Kubernetes version
- Grove version
- Scheduler details

wzshiming:
#### What type of PR is this?

/kind bug

Fixes #739

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/cert/cert.go
+++ b/operator/internal/controller/cert/cert.go
@@ -148,6 +148,12 @@ func getWebhooks(authorizerEnabled bool) []cert.WebhookInfo {
 func createPlaceholderSecretIfNotExists(ctx context.Context, cl client.Client, namespace, secretName string) error {
 	secret := &corev1.Secret{}
 	err := cl.Get(ctx, types.NamespacedName{Namespace: namespace, Name: secretName}, secret)
+	if err == nil {
+		// In HA deployments (replicaCount > 1) or restarts, two replicas can race between
+		// Get (not found) and Create. If the secret already exists,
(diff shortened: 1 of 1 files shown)


oleg-kushniriov:
Redundant change - `if !apierrors.IsNotFound(err) `already returns nil when err == nil. The PR adds a check that the existing logic already covers.

wzshiming:
Sorry for the mistake

PR #745, opened 2026-08-03
operator/helm: support imagePullSecrets on the operator pod

### LINKED ISSUE #746: operator/helm: support imagePullSecrets on the operator pod

Add an optional `imagePullSecrets` list to the operator Helm chart
(`operator/charts`), wired into the operator Deployment pod spec. It would
default to empty so public installs are unaffected, and accept a list of
`LocalObjectReference` (e.g. `[{name: my-registry-creds}]`) when set.

### Why is this needed?

The operator and crd-installer init-container images are pulled from
`grove-operator` / `grove-install-crds` repositories. When those images are
mirrored to or built into a private registry, the operator pod has no way to
authenticate the pull: the chart exposes no `imagePullSecrets` field, so the
Deployment cannot reference a dockerconfigjson Secret and the pull fails with
`ImagePullBackOff`. Public installs from `ghcr.io/ai-dynamo` are unaffected,
but any downstream running Grove from a private registry currently has to
patch the rendered manifest out of band.

kulkarnisamr:
#### What type of PR is this?
/kind feature

Adds an optional `imagePullSecrets` list to the operator Helm chart, wired
into the deployment pod spec, so the operator (and crd-installer init
container) can be pulled from a private registry. Defaults to empty, so
public installs are unaffected.

Fixes #746

Follows the existing `deployment_test.go` helm-unittest pattern; added
`TestOperatorImagePullSecrets` covering the empty default and a set value.

```release-note
The operator Helm chart now supports `imagePullSecrets` for pulling the
operator image from a private registry.
```

### DIFF
--- a/operator/charts/deployment_test.go
+++ b/operator/charts/deployment_test.go
@@ -23,6 +23,7 @@ import (
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/engine"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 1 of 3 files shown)


unmarshall (approved):
We need to relook at why there is a deployment_test.go inside a helm chart. But this PR did not introduce this file and therefore lets not block this PR.

sanjaychatterjee (approved):

Issue #449, opened 2026-02-25
Flaky e2e: Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing


shmuel-runai:
### What happened?

The Config2_UnsupportedButEnabled e2e test (Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing) is flaky — it intermittently fails with a 2-minute timeout waiting for the expected preflight-failure log message, but passes on re-run.
```
=== RUN   Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing
    unsupported_but_enabled_test.go:101:
        Error Trace:  unsupported_but_enabled_test.go:101
                      unsupported_but_enabled_test.go:61
        Error:        Received unexpected error:
                      condition not met within timeout of 2m0s
        Test:         Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing
        Messages:     Operator logs should show preflight failure due to missing CRD
--- FAIL: Test_AutoMNNVL_UnsupportedButEnabled (120.39s)
    --- FAIL: Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing (120.15s)
```

### What did you expect to happen?

The test should reliably find the crashing operator pod and verify that its logs contain the preflight failure message `("MNNVL preflight check failed" + "ComputeDomain CRD").`

**RCA**
The flakiness is a race condition in pod selection during a rolling deployment.
Before `Config2_UnsupportedButEnabled` runs, `config-cluster.py` performs a helm upgrade that changes the operator ConfigMap (enabling autoMNNVLEnabled). 

Because the Helm chart uses a content-hashed ConfigMap name (grove-operator-cm-<hash>), the deployment spec changes, triggering a rolling update.

With the default Kubernetes rolling update strategy for a single-replica Deployment (maxUnavailable=0, maxSurge=1):
A new pod is created with the updated config — it crashes immediately on the MNNVL preflight check (CRD is missing).
The old pod (from Config1) stays Running because the new pod never becomes Ready.
The old pod may have RestartCount > 0 from a cert-refresh restart during Config1.
The waitForOperatorPod() function iterates over all operator pods and returns the first one with RestartCount > 0 or a terminated container — this is non-deterministic because List ordering is not guaranteed:
[...]

PR #747, opened 2026-08-04
fix: reconcile Grove resources from live API state

### LINKED ISSUE #711: Flaky E2E: RU-11 rolling update stops converging after scale-out

danbar2:
#### What type of PR is this?
/kind bug

`reconcileStatus` skips its status write when the recomputed status equals the status the reconciled object was loaded with, treating that as "already persisted". The object was loaded from the informer cache, which can still be serving a copy from before a write the controller already made — so the recomputed status could match a **stale baseline**, and the skip would drop a write the apiserver still needed.

Nothing recovered from that. A status write does not change the generation, so it never re-enqueued the controller, and once the children went quiet no other event arrived either. That is the wedge @yankay diagnosed in #711.

This PR reads the reconciled object through `mgr.GetAPIReader()`, so the baseline the skip relies on is authoritative by construction. Child PodClique / PodCliqueScalingGroup / Pod reads stay cache-backed — they are level-triggered by their own watches, and a stale child read is corrected by the next event.

**Applied to all three controllers.** The skip is identical in `podcliqueset/reconcilestatus.go`, `podclique/reconcilestatus.go` and `podcliquescalinggroup/reconcilestatus.go`, and so is the reasoning that makes a cached baseline unsound. PodCliqueSet is where it was caught, not where it is unique.

**Removes the PodCliqueSet generation-hash expectations.** That `sync.Map` and its requeue gate existed only to wait for the informer cache to catch up with a `CurrentGenerationHash` this controller had just written — exactly the staleness the live read now eliminates, so the gate can no longer fire. Removing it beats leaving compensation next to a live read telling future readers the cache still needs waiting on.

#### Why this approach:

The alternative is to keep the cached read and compensate: track what was persisted, or requeue after a skip. Both leave the unsound assumption in place and add controller-local state that has to be kept correct, bounded, and cleaned up. Reading the object live deletes the assumption instead, and it deletes code on net in the reconcile path.

Trade-off: one single-object GET per reconcile per controller. Child read paths, watches and public APIs are unchanged. If that GET turns out to matter at scale, `make run-scale-test` is the place to show it — I have not run it, so the cost is stated rather than measured.

Fixes #711

Verified against the flake itself. RU-11 had been failing at the full 2-minute convergence timeout (148.38s and 148.40s on two consecutive attempts of an unrelated PR). With the fix it converges in ~43s, in line with its neighbours, and all ten rolling-update scenarios pass.

Each controller gets a regression test that gives `client` and `apiReader` deliberately divergent contents, so only a reconcile that reads through `apiReader` behaves correctly. All three were confirmed to fail when the read is reverted to `r.client`.

`E2E - auto_mnnvl` is red and unrelated — `Test_AutoMNNVL_UnsupportedButEnabled/operator_exits_when_CD_CRD_is_missing`, same 120.07s timeout signature on unrelated branches. It is #449, closed 2026-03-03 and evidently regressed; worth reopening separately.

`make lint` cannot run locally here (golangci-lint v2.6.1 is built with go1.24, repo targets 1.26.3), so CI covers it. `go build`, `go vet`, `gofmt`, the full unit suite and `-race` are green.

Supersedes the earlier approach on this branch, which compensated for the stale baseline instead of removing it. Note this overlaps with draft #716, which fixes the same bug for PodCliqueSet only — only one of the two should land.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podclique/reconciler.go
+++ b/operator/internal/controller/podclique/reconciler.go
@@ -37,8 +37,12 @@ import (
 
 // Reconciler reconciles PodClique objects.
 type Reconciler struct {
-	config                  configv1alpha1.PodCliqueControllerConfiguration
-	client                  ctrlclient.Client
+	config configv1alpha1.PodCliqueControllerConfiguration
+	client ctrlclient.Client
+	// apiReader reads straight from the apiserver, bypassing the informer cache. It is used only to
+	// fetch the PodClique being reconciled. See the Reconcile godoc for why that read cannot be
+	// served from the cache.
+	apiReader               ctrlclient.Reader
 	eventRecorder           record.EventRecorder
 	reconcileStatusRecorder ctrlcommon.ReconcileErrorRecorder
 	expectationsStore       *expect.ExpectationsStore
@@ -52,6 +56,7 @@ func NewReconciler(mgr ctrl.Manager, controllerCfg configv1alpha1.PodCliqueContr
 	return &Reconciler{
 		config:                  controllerCfg,
 		client:                  mgr.GetClient(),
+		apiReader:               mgr.GetAPIReader(),
 		eventRecorder:           eventRecorder,
 		reconcileStatusRecorder: ctrlcommon.NewReconcileErrorRecorder(mgr.GetClient()),
 		expectationsStore:       expectationsStore,
@@ -60,6 +65,19 @@ func NewReconciler(mgr ctrl.Manager, controllerCfg configv1alpha1.PodCliqueContr
 }
 
 // Reconcile reconciles the `PodClique` resource.
+//
+// The PodClique itself is read through apiReader rather than the informer cache. reconcileStatus
+// skips its write when the recomputed status equals the status this object was loaded with, treating
+// that as "already persisted". A cached read makes that claim unsound: the cache can still be serving
+// a copy from before a write this controller already made, so the recomputed status can match a stale
+// baseline and the skip drops a write the apiserver still needs. Nothing recovers from that - a status
+// write does not change the generation, so it is filtered by this controller's own
+// GenerationChangedPredicate and never re-enqueues, and once the Pods go quiet no other event arrives
+// either. Reading the object live makes the baseline authoritative by construction, which is what the
+// skip assumes.
+//
+// Child Pod reads and the parent PodCliqueSet lookup stay cache-backed; they are level-triggered by
+// their own watches and a stale read of them is corrected by the next event.
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := ctrllogger.FromContext(ctx).WithName(controllerName)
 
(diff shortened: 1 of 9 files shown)


danbar2:
Follow-up on the memory footprint of `lastPersistedStatuses`, since "one status per PodCliqueSet held in memory" deserves a number rather than a hand-wave.

**Measured size.** `PodCliqueSetStatus` is fixed-shape: counters, two short `*string`s, a condition or two, and a small `UpdateProgress`. `PodGangStatutes` is declared but never populated by the operator, so nothing in it scales with the spec. Heap measured over 10k deep copies of a fully reconciled status built from `e2e/yaml/scale-test-1000.yaml`:

| | per PodCliqueSet |
|---|---|
| recorded status + map key | **420 B** |
| the whole PCS the informer cache already holds | 3,288 B |

So this adds ~13% on top of what the cache already keeps per object, and it is constant in the size of the spec — a 50-clique PCS records the same 420 B as a 1-clique one. At 5k PodCliqueSets that is ~2 MB against the chart's 1Gi limit; at 20k, ~8 MB.

**The growth risk was real, though, and is now closed.** Steady-state size was never the danger — an unbounded map was. Releasing the record only in the finalizer flow left an entry behind for any PodCliqueSet that disappeared without running it (force delete, finalizer stripped out of band), so the map could creep up with PodCliqueSet churn over the operator's lifetime rather than tracking live objects.

`82762827` moves the release to where `GetPodCliqueSet` reports the object is gone. Every deletion the operator observes produces exactly one such reconcile, so that is a single choke point that no deletion path bypasses. A failed read is deliberately left alone — dropping the record on a transient error would hand the next reconcile back the cached-only baseline this PR exists to distrust.

Two tests pin it: `TestReconcileForgetsPersistedStatusOfDeletedPodCliqueSet` (verified failing without the cleanup) and `TestReconcileKeepsPersistedStatusWhenTheGetFails`.

If even 420 B/object is unwelcome, storing a hash of the status instead of the status would cut it to ~50 B. I did not do that: a hash mismatch on semantically equal statuses only costs a redundant write, but a hash *collision* silently reinstates exactly the lost write this PR fixes, and `equality.Semantic.DeepEqual` against the real status has no such failure mode. Happy to switch if you would rather have the bytes.

PR #753, opened 2026-08-05
fix(operator): allow deriving environment variables from Grove values

### LINKED ISSUE #752: Template environment variables cannot derive values from Grove-injected variables
## Description

Grove injects its pod-discovery environment variables by appending them to the environment variable lists copied from the PodClique template. This prevents a higher-level API from defining an abstract environment variable whose value is derived from Grove variables.

Kubernetes expands `$(VAR_NAME)` references in an `env[].value` using variables that were defined earlier in the list. Because the `GROVE_*` variables are appended after the template-provided variables, this does not work:

```yaml
env:
  - name: CLUSTER_LEADER_ADDRESS
    value: "$(GROVE_PCSG_NAME)-$(GROVE_PCSG_INDEX)-worker-ldr-0.$(GROVE_HEADLESS_SERVICE)"
args:
  - "--leader-address=$(CLUSTER_LEADER_ADDRESS)"
```

When the template is authored, the concrete values of the Grove variables are not available yet. The higher-level controller knows only the expression that should be evaluated for the final Grove-created Pod. Since the expansion of environment-variable values is order-dependent and not recursive, `CLUSTER_LEADER_ADDRESS` retains unresolved `$(GROVE_*)` references.

This makes it impossible for higher-level APIs to expose backend-independent command or argument contracts while deriving their concrete values from Grove's pod-discovery variables.

The same issue applies to both regular containers and init containers.

## Expected behavior

Grove should provide a documented and stable way for template-provided environment variables to derive values from the Grove-injected variables before the final container command and arguments are expanded.

Possible solutions could include:

- making the Grove variables available before template-provided environment variables are evaluated;
- supporting derived environment variables that Grove appends after its own variables; or
- another explicit late-binding mechanism.

The exact mechanism is less important than allowing a higher-level controller to define an abstract variable without rewriting the user's command or depending on undocumented injection ordering.

## Current behavior

Grove appends its injected variables:

```go
containers[i].Env = append(containers[i].Env, envVars...)
```

Consequently, template-provided environment variables cannot reference them using Kubernetes environment-variable expansion.

sttts:
#### What type of PR is this?

/kind bug

Grove currently appends its pod-discovery environment variables after the environment variables copied from PodClique templates. Kubernetes expands references in environment variable values in list order, so a template-defined variable cannot derive its value from a later `GROVE_*` variable.

This change prepends Grove's injected variables for containers and init containers at both injection stages. It also removes colliding template entries so Grove's reserved variables remain authoritative and duplicates are not passed to the kubelet.

The environment-variable reference documentation now describes the ordering contract and shows how to define a derived leader address.

Fixes #752

The final order for PCSG-backed pods is the Pod-level Grove variables, the PCSG-level Grove variables, and then the remaining template variables. This makes all injected Grove variables available when Kubernetes expands template-defined environment variable values.

Validation:

- `make test-unit`
- `make lint`

```release-note
Grove now injects pod-discovery environment variables before template-defined environment variables, allowing templates to derive values from Grove variables.
```

```docs
docs/user-guide/03_environment-variables-for-pod-discovery/02_env_var_reference.md
```

### DIFF
--- a/operator/internal/controller/podclique/components/pod/pod_test.go
+++ b/operator/internal/controller/podclique/components/pod/pod_test.go
@@ -338,6 +338,15 @@ func TestAddGroveEnvironmentVariables_NoDuplicates(t *testing.T) {
 								},
 							},
 						},
+						InitContainers: []corev1.Container{
+							{
+								Name:  "test-init-container",
+								Image: "test-image",
+								Env: []corev1.EnvVar{
+									{Name: "GROVE_POD_INDEX", Value: "old-pod-index"},
+								},
+							},
+						},
 					},
 				},
 			},
@@ -460,12 +469,29 @@ func TestAddGroveEnvironmentVariables_NoDuplicates(t *testing.T) {
 
 			addEnvironmentVariables(pod, tt.pclq, "test-pcs", 0)
 
-			// Check that all containers have the expected environment variables
-			for _, container := range pod.Spec.Containers {
+			expectedPrefix := []string{
+				constants.EnvVarPodCliqueSetName,
+				constants.EnvVarPodCliqueSetIndex,
+				constants.EnvVarPodCliqueName,
+				constants.EnvVarHeadlessService,
+				constants.EnvVarPodIndex,
+			}
+			assertContainer := func(container corev1.Container) {
 				assertExpectedEnvVars(t, container, tt.expectedEnvVars)
 				assertReplacedEnvVars(t, container, tt.shouldReplace)
 				assertPreservedEnvVars(t, container, tt.shouldPreserve)
 				assertNoDuplicateEnvVars(t, container)
+				require.GreaterOrEqual(t, len(container.Env), len(expectedPrefix))
+				for i, envVarName := range expectedPrefix {
+					assert.Equal(t, envVarName, container.Env[i].Name)
+				}
+				assertEnvVarUsesFieldRef(t, container, constants.EnvVarPodIndex, fmt.Sprintf("metadata.labels['%s']", common.LabelPodCliquePodIndex))
+			}
+			for _, container := range pod.Spec.Containers {
+				assertContainer(container)
+			}
+			for _, container := range pod.Spec.InitContainers {
+				assertContainer(container)
 			}
 		})
 	}
--- a/operator/internal/controller/podclique/components/pod/pod_test.go
+++ b/operator/internal/controller/podclique/components/pod/pod_test.go
@@ -465,6 +465,7 @@ func TestAddGroveEnvironmentVariables_NoDuplicates(t *testing.T) {
 				assertExpectedEnvVars(t, container, tt.expectedEnvVars)
 				assertReplacedEnvVars(t, container, tt.shouldReplace)
 				assertPreservedEnvVars(t, container, tt.shouldPreserve)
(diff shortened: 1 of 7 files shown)


danbar2 (dismissed):

operator/internal/controller/podclique/components/pod/pod_test.go
```diff
@@ -465,6 +465,7 @@ func TestAddGroveEnvironmentVariables_NoDuplicates(t *testing.T) {
 				assertExpectedEnvVars(t, container, tt.expectedEnvVars)
 				assertReplacedEnvVars(t, container, tt.shouldReplace)
 				assertPreservedEnvVars(t, container, tt.shouldPreserve)
+				assertNoDuplicateEnvVars(t, container)
```
oleg-kushniriov, line 468: Test Coverage - No init-container scenario in `TestAddGroveEnvironmentVariables_NoDuplicates` .
The newly added assertNoDuplicateEnvVars call iterates over pod.Spec.Containers only - none of the table test cases in `TestAddGroveEnvironmentVariables_NoDuplicates` include InitContainers. The same PrependEnvVarsToContainers path is exercised for init containers by the addEnvironmentVariables function, so a collision in an init container (e.g., a template init-container predefining GROVE_POD_INDEX) goes untested. 

Consier: Add at least one table case with an InitContainers entry that has a colliding variable, and assert both `assertNoDuplicateEnvVars` and correct ordering for it.

oleg-kushniriov (dismissed):
Test Coverage - 
PCSG collision scenario not tested operator/internal/controller/podcliquescalinggroup/components/podclique/podclique_test.go (file unchanged in PR, existing TestAddEnvironmentVariablesToPodContainerSpecs at line 334) `TestAddEnvironmentVariablesToPodContainerSpecs` only verifies presence of PCSG vars. There is no test case where the PodClique template predefines a var that collides with a PCSG-injected var (e.g., GROVE_PCSG_NAME set to a stale value). The dedup/replace behavior of PrependEnvVarsToContainers is exercised in the utility tests, but the PCSG-component integration path is not. 

Consider: Add a case where a container has {Name: "GROVE_PCSG_NAME", Value: "stale"} in its Env, call addEnvironmentVariablesToPodContainerSpecs, and assert no duplicates and that the injected value wins.

renormalize (dismissed):
thanks!

operator/internal/controller/podclique/components/pod/pod_test.go
sttts: Addressed in db9575c. The PodClique test now includes an init container with a colliding `GROVE_POD_INDEX` and applies the same duplicate, injected-prefix ordering, and authoritative fieldRef assertions to regular and init containers. Targeted tests and `make lint` pass.

sttts:
@oleg-kushniriov Addressed both test-coverage suggestions in db9575c: the PodClique path now covers an init-container collision and verifies injected ordering/no duplicates, and the PCSG integration test now starts with a stale `GROVE_PCSG_NAME` and verifies the full authoritative, deduplicated env result. Validation: targeted Go tests for both controllers and the shared utility, plus `make lint`.

oleg-kushniriov (approved):
Looks good to me

renormalize (approved):

shayasoolin (approved):

PR #755, opened 2026-08-05
feat(operator): expose PCSG-wide pod index

### LINKED ISSUE #754: Expose a group-wide pod index for PodCliqueScalingGroup replicas
## Motivation

LeaderWorkerSet exposes a zero-based pod identity across each leader/worker group: the leader receives index 0 and workers receive indices 1 through `size - 1`. This lets higher-level APIs map the backend-specific value to a concrete rank without rewriting a container command.

Grove currently exposes `GROVE_PCLQ_POD_INDEX`, but that index is scoped to a single PodClique and restarts at zero for every clique. A common distributed topology uses two PodCliques in one PodCliqueScalingGroup replica:

```yaml
podCliqueScalingGroups:
  - name: engine
    cliqueNames:
      - engine-leader # replicas: 1
      - engine-worker # replicas: 2
```

The current indices are:

| PodClique | `GROVE_PCLQ_POD_INDEX` |
| --- | ---: |
| engine-leader | 0 |
| engine-worker, first pod | 0 |
| engine-worker, second pod | 1 |

A higher-level controller therefore cannot define a concrete rank such as:

```yaml
- name: DYNAMO_RANK
  value: "$(GROVE_PCSG_POD_INDEX)"
```

Instead, worker commands currently need shell arithmetic such as `$((GROVE_PCLQ_POD_INDEX + 1))`.

## Proposed behavior

Expose a new `GROVE_PCSG_POD_INDEX` environment variable for pods whose PodClique belongs to a PodCliqueScalingGroup.

The value should be:

- zero-based within one concrete PodCliqueScalingGroup replica;
- assigned by flattening member PodCliques in `PodCliqueScalingGroupConfig.cliqueNames` order; and
- concrete and numeric when injected into containers and init containers.

For a clique at position `i`, its offset when the concrete scaling element is created is:

```text
offset(i) = sum(replicas of cliqueNames before i)

GROVE_PCSG_POD_INDEX =
    offset(i)
    + GROVE_PCLQ_POD_INDEX
```

The offset belongs to the generated member PodClique and must be retained for its lifetime. Replacing a pod then reuses the concrete scaling element's layout instead of recalculating ranks from a potentially changed PodCliqueSet template.

The example above produces indices 0, 1, and 2. Scaling the PCSG creates another independent scaling element whose indices start at 0 again. Together with the existing variables, a pod is uniquely identified within the PodCliqueSet by:

```text
GROVE_PCSG_NAME
GROVE_PCSG_INDEX
GROVE_PCSG_POD_INDEX
```

Standalone PodCliques should not receive this variable, matching the behavior of the existing `GROVE_PCSG_*` variables.

## API contract

This makes the order of `PodCliqueScalingGroupConfig.cliqueNames` semantically significant and should be documented. The field is already immutable after creation.

The variable should be injected before template-defined environment variables, as proposed in #752, so a higher-level API can define a backend-neutral variable such as `DYNAMO_RANK` through Kubernetes environment-variable expansion.

## Benefits

- LWS-analogous group rank semantics for Grove.
- No shell arithmetic or command rewriting for rank-aware applications.
- Stable ranks across replacement of individual pods in a scaling element.
- A generic group-wide pod identity that does not require Grove to understand application-specific leader or worker roles.

sttts:
## Summary

- expose a concrete `GROVE_PCSG_POD_INDEX` in containers and init containers for PodCliqueScalingGroup members
- flatten member PodCliques in immutable `cliqueNames` order and retain each generated PodClique's offset across pod replacement
- document the LWS-analogous semantics and demonstrate defining `DYNAMO_RANK` as `$(GROVE_PCSG_POD_INDEX)`

A leader PodClique with one pod followed by a worker PodClique with two pods receives group-wide indices `0`, `1`, and `2`. Each PCSG replica starts its own index space at zero. Standalone PodCliques do not receive the new variable.

This PR is stacked on and depends on #753, which prepends Grove-injected environment variables so template-defined variables can reference them.

Fixes #754.

## Validation

- `make test-unit`
- generation, formatting, API docs, and all golangci-lint targets from `make check`
- `make verify-toc`
- `go test ./internal/controller/common/component/utils ./internal/controller/podcliquescalinggroup/components/podclique ./internal/controller/podclique/components/pod`


danbar2 (dismissed):
LGTM

docs/user-guide/03_environment-variables-for-pod-discovery/02_env_var_reference.md
```diff
@@ -97,7 +110,30 @@ If a pod belongs to a PodClique that is part of a PodCliqueScalingGroup, these a
...
+podCliqueScalingGroups:
+  - name: engine
+    cliqueNames:
+      - engine-leader # replicas: 1
+      - engine-worker # replicas: 2
+```
+
+assigns index `0` to the leader pod and indices `1` and `2` to the worker pods. A higher-level API can use the concrete value to define an application-specific rank without shell arithmetic:
+
+```yaml
```
shayasoolin, line 131: Nit - I'd drop this example. The description that comes before the yaml is good enough IMHO. The example is a bit out of context and requires to be familiar with Dynamo, otherwise `RANK` is just another name for `POD_INDEX`.

docs/user-guide/03_environment-variables-for-pod-discovery/03_hands-on-examples.md
```diff
@@ -141,6 +141,9 @@ spec:
           containers:
           - name: leader
             image: busybox:latest
+            env:
```
shayasoolin: Same here - I'd drop and avoid Dynamo-specific reference here.

operator/samples/user-guide/03_environment-variables-for-pod-discovery/pcsg-env-vars.yaml
```diff
@@ -22,6 +23,9 @@ spec:
           containers:
           - name: leader
             image: busybox:latest
+            env:
+            - name: DYNAMO_RANK
```
shayasoolin: And again here

shayasoolin (commented):
Looks great, thanks for the important PR. Left a minor comment please LMK if you agree. I prefer not to involve Dynamo in the examples, which are currently pretty generic and neutral.

docs/user-guide/03_environment-variables-for-pod-discovery/02_env_var_reference.md
sttts: Addressed in 48d7395. Removed the Dynamo-specific rank example here and the related references from the hands-on guide and sample, while retaining the generic GROVE_PCSG_POD_INDEX semantics.

docs/user-guide/03_environment-variables-for-pod-discovery/03_hands-on-examples.md
sttts: Addressed in 48d7395. Removed the DYNAMO_RANK environment variables, log output, and explanatory text from the hands-on example.

operator/samples/user-guide/03_environment-variables-for-pod-discovery/pcsg-env-vars.yaml
sttts: Addressed in 48d7395. Removed the Dynamo-specific environment variables and output from the sample; the sample now demonstrates only Grove variables and leader discovery.

shayasoolin (dismissed):

operator/internal/controller/podclique/components/pod/pod.go
```diff
@@ -201,6 +211,25 @@ func (r _resource) buildResource(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grov
...
+// getPCSGPodIndex returns the pod's zero-based index within its PodCliqueScalingGroup replica.
+// Standalone PodCliques return nil because they have no group-wide index.
+func getPCSGPodIndex(pclq *grovecorev1alpha1.PodClique, podIndex int) (*int, error) {
+	if pclq.Labels[apicommon.LabelPodCliqueScalingGroup] == "" {
+		return nil, nil
+	}
+
+	offsetValue, ok := pclq.Labels[apicommon.LabelPodCliqueScalingGroupPodIndexOffset]
+	if !ok {
+		return nil, fmt.Errorf("PodClique is missing required label %q", apicommon.LabelPodCliqueScalingGroupPodIndexOffset)
```
oleg-kushniriov: PodCliques that belong to a PCSG and were created before the operator upgrade will never have the `grove.io/podcliquescalinggroup-pod-index-offset` label, so they will not be able to create new pods: under the default RollingRecreate strategy nothing backfills the label on existing PodCliques, and this persists until the PodClique is recreated.


I suggest running Scale and Churn tests on this PR for the operator-upgrade scenario (deploy the previous operator version, create PCSG workloads, upgrade to this PR, then drive pod churn) to verify replacement pods are still created for pre-upgrade PodCliques.

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -333,6 +334,21 @@ func (r _resource) buildResource(logger logr.Logger, ss *syncSnapshot, pcsgRepli
...
 	r.addEnvironmentVariablesToPodContainerSpecs(pclq, pcsgTemplateNumPods)
+	pcsgPodIndexOffset, err := componentutils.GetPCSGPodIndexOffset(pcs, pcsg.Spec.CliqueNames, pclqTemplateSpec.Name)
+	if err != nil {
+		return groveerr.WrapError(err,
+			errCodeBuildPodClique,
+			component.OperationSync,
+			fmt.Sprintf("Error computing PodCliqueScalingGroup pod index offset for PodClique: %v", pclqObjectKey),
+		)
+	}
+	if pclqExists && currentPCSGPodIndexOffset != "" {
```
oleg-kushniriov: The `grove.io/podcliquescalinggroup-pod-index-offset` value is computed once at PodClique creation and retained for its lifetime, but a member PodClique's spec.replicas can still be increased (scale subresource, and preserved on update). Scaling up a member PodClique on an existing PCSG replica then gives its new pods `GROVE_PCSG_POD_INDEX` values that duplicate those of the next clique in cliqueNames order. This breaks the index's core contract of being unique within a PCSG replica: any workload that derives its rank or identity from it ends up with two pods claiming the same value, and nothing detects or logs the collision.

oleg-kushniriov (commented):
Recommend running ScaleUp/Down and Churn tests before and after the operator upgrade, validating the new label values - especially that `GROVE_PCSG_POD_INDEX` stays unique and stable within each PCSG replica.

operator/internal/controller/podclique/components/pod/pod.go
```diff
@@ -201,6 +211,25 @@ func (r _resource) buildResource(pcs *grovecorev1alpha1.PodCliqueSet, pclq *grov
 	return nil
 }
 
+// getPCSGPodIndex returns the pod's zero-based index within its PodCliqueScalingGroup replica.
+// Standalone PodCliques return nil because they have no group-wide index.
+func getPCSGPodIndex(pclq *grovecorev1alpha1.PodClique, podIndex int) (*int, error) {
```
oleg-kushniriov, line 237: This PR introduces an upgrade-sensitive dependency - pod creation for PCSG-member PodCliques now requires the `grove.io/podcliquescalinggroup-pod-index-offset` label, which pre-upgrade PodCliques don't have - but the existing upgrade e2e (`TestUpgradeFromLatestGitHubRelease`) can't exercise it: its workload has no PodCliqueScalingGroup, and it never churns pods of pre-upgrade PodCliques after the upgrade. Upgrade e2e coverage for this path should be a merge requirement.

Please extend the upgrade e2e (or add a case) with a PodCliqueScalingGroup in the workload, and after the upgrade: delete a pod of a pre-upgrade member PodClique and assert the replacement is created; validate the grove.io/podcliquescalinggroup-pod-index-offset label values and the uniqueness of GROVE_PCSG_POD_INDEX within each PCSG replica - including after ScaleUp/Down and churn.

operator/internal/controller/podclique/components/pod/pod.go
sttts: Addressed in fd02e18. Existing PCSG-member PodCliques are now backfilled in place with an internal offset annotation computed from the live member spec.replicas. The PodClique controller waits and retries while that annotation is absent; its update then triggers reconciliation. Existing Pods receive the final group-wide index label without recreation, and replacement Pods are created from the backfilled live offset. The upgrade E2E now verifies unchanged UIDs and successful churn of a pre-upgrade member Pod.

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
sttts: Addressed in fd02e18. The offset is no longer computed once and retained. The PCSG controller reconciles offsets from the current live member spec.replicas, patches affected member PodCliques, and the PodClique controller updates final index labels on existing Pods in place; newly created Pods use the same current offset. The upgrade E2E scales an earlier member down and up and checks the exact contiguous index set after each operation. Pod metadata converges eventually because the Kubernetes patches are not atomic; the documentation also notes that containers consuming a projected environment variable must restart after an earlier-member scale.

operator/internal/controller/podclique/components/pod/pod.go
sttts: Addressed in fd02e18. The upgrade workload now contains a PCSG. Before upgrading, the test scales an earlier member with the previous operator. After upgrading it verifies live-offset backfill, exact final Pod index labels, and unchanged Pod UIDs; it then deletes a pre-upgrade Pod, verifies its indexed replacement, and exercises member scale-down and scale-up before continuing the existing PCS-scale and init-container checks. Following the updated design, the public contract under test is the final grove.io/podcliquescalinggroup-pod-index Pod label; the offset remains an internal non-propagated annotation and no new automatic environment variable is introduced.

sttts: Follow-up in 835708e: the implementation now follows Groves existing Pod-index pattern completely. GROVE_PCSG_POD_INDEX is automatically injected by the Pod controller through a Downward API fieldRef to the final grove.io/podcliquescalinggroup-pod-index label. Injecting at Pod creation also ensures that replacement Pods from unchanged pre-upgrade PodCliques receive the new variable. This supersedes my earlier statement that no new automatic environment variable would be introduced; only the offset remains internal and non-propagated.

oleg-kushniriov (dismissed):
LGTM

shayasoolin (dismissed):

shayasoolin (dismissed):

oleg-kushniriov (dismissed):

shayasoolin (approved):

danbar2 (approved):

PR #759, opened 2026-08-07
fix(podclique): exclude pods already being deleted from scale-in candidates

### LINKED ISSUE #758: Scale-in can delete healthy Pods because deletion candidates include already-terminating Pods
### What happened?

Scaling a PodClique down (e.g. `kubectl scale pclq <name> --replicas=1` on a clique with 3 replicas) can end up deleting **all** existing Pods instead of only the excess ones, leaving the clique with zero serving Pods until a replacement starts.

The first deletion round is correct. `runSyncFlow` computes:

```go
// operator/internal/controller/podclique/components/pod/syncflow.go:176
diff := len(sc.existingPCLQPods) + len(createExpectations) - int(sc.pclq.Spec.Replicas) - len(deleteExpectations)
```

For 3 → 1 this is `3 + 0 - 1 - 0 = 2`, and `deleteExcessPods(2)` deletes 2 Pods.

The problem is the candidate selection itself. `componentutils.GetPCLQPods` returns terminating Pods as well, and neither `selectExcessPodsToDelete` nor `DeletionSorter` filters them out:

```go
// operator/internal/controller/podclique/components/pod/syncflow.go:231-243
func selectExcessPodsToDelete(sc *syncContext, logger logr.Logger) []*corev1.Pod {
	var candidatePodsToDelete []*corev1.Pod
	if diff := len(sc.existingPCLQPods) - int(sc.pclq.Spec.Replicas); diff > 0 {
		sorter := DeletionSorter{Pods: sc.existingPCLQPods}
		sorter.ExpectedPodTemplateHash = sc.getExpectedPodTemplateHash()
		sort.Sort(sorter)
		candidatePodsToDelete = append(candidatePodsToDelete, sorter.Pods[:diff]...)
	}
	return candidatePodsToDelete
}
```

`DeletionSorter.Less` (`operator/internal/controller/podclique/components/pod/deletionsort.go:48-76`) has five criteria — unassigned/assigned, phase, readiness, template hash, creation timestamp — and **none of them mentions `DeletionTimestamp`**. During the grace period a terminating Pod is still `Running` and still `Ready`, so it ties with the healthy survivor and the sort order between them is arbitrary.

**The deletion budget is computed twice, with different formulas.** `diff` decides *how many* to
delete and subtracts the delete-expectations; `selectExcessPodsToDelete` decides *from whom* and does
not:

```go
// syncflow.go:176   how many  — subtracts deleteExpectations
diff := len(sc.existingPCLQPods) + len(createExpectations) - int(sc.pclq.Spec.Replicas) - len(deleteExpectations)

// syncflow.go:233   from whom — does NOT subtract deleteExpectations
if diff := len(sc.existingPCLQPods) - int(sc.pclq.Spec.Replicas); diff > 0 {

// syncflow.go:208
numPodsToSelectForDeletion := min(diff, len(candidatePodsToDelete))
```

The second expression keeps the budget open for as long as terminating Pods linger, which is what
converts any `diff` inflation into an actual extra deletion. A live log shows the gap directly:
`found excess pods for PodClique numExcessPods=5` (that is the second expression) alongside
`Deleted excess pods diff=1 noOfPodsDeleted=1`.

This is the only deletion path in the operator that lacks such a filter. The sibling rolling-update path in the same package does exactly the right thing:

```go
// operator/internal/controller/podclique/components/pod/rollingupdate.go:174-177
func (r _resource) hasPodDeletionBeenTriggered(sc *syncContext, pod *corev1.Pod) bool {
	return k8sutils.IsResourceTerminating(pod.ObjectMeta) ||
		r.expectationsStore.HasDeleteExpectation(sc.pclqExpectationsStoreKey, pod.GetUID())
}
```

Two known ways a second deletion pass runs over the stale list and consumes the survivor:

1. **Rolling update in the same reconcile.** `runSyncFlow` does not return after `deleteExcessPods`; it falls straight through to `processPendingUpdates` using the same `sc.existingPCLQPods` captured once in `prepareSyncFlow` (`syncflow.go:94`):

   ```go
   // operator/internal/controller/podclique/components/pod/syncflow.go:148-158
   } else if diff > 0 {
       if err := r.deleteExcessPods(sc, logger, diff); err != nil { ... }
   }
   if componentutils.IsAutoUpdateStrategy(sc.pcs) && componentutils.IsPCLQAutoUpdateInProgress(sc.pclq) {
       if err := r.processPendingUpdates(logger, sc); err != nil { ... }
   }
   ```

2. **Operator restart mid scale-down**, which wipes the in-memory delete expectations so `diff` is recomputed as 2 again — see the companion issue on `SyncExpectations` restart recovery.

### Confirmed reproduction (grove-operator @ b5621d4)

Reproduced deterministically on a 2-node cluster running a build of `main` (`gitVersion=v0.1.0-dev`, `gitCommit=b5621d44c7957cd0eae366453f631dfd4d7df82a`, `gitTreeState=clean`).

Setup — a single standalone PodClique, no `autoScalingConfig`, no `startsAfter`, `minAvailable: 1`, `terminationGracePeriodSeconds: 300`, container `busybox` running `sh -c "sleep infinity"` (PID 1 is `sh`, which ignores SIGTERM, so Pods stay `Running`+`Ready` for the whole grace period). Created fresh with `replicas: 6`, so all six Pods share one `creationTimestamp`:

```
$ kubectl get pods -l grove.io/podclique=repro-0-worker \
    -o jsonpath='{range .items[*]}{.metadata.creationTimestamp}{"\n"}{end}' | sort -u
2026-08-05T07:20:40Z          # one distinct value for all 6
```

Steps:

1. `kubectl scale pclq repro-0-worker --replicas=1` — Grove correctly deletes exactly 5. All five stay `Running`+`Ready` with a `deletionTimestamp` of `07:26:23Z`. Survivor: `repro-0-worker-9zxgn`.
2. Within the grace window, restart the operator (`kubectl delete pod` on the operator Pod). This is the real-world equivalent of an OOM kill, upgrade, or leader handover.
3. The new operator reconciles ~1s after startup:

```
07:21:54.930Z  found excess pods for PodClique   numExcessPods: 5
07:21:54.932Z  Deleted Pod   repro-0-worker-9zxgn    <-- the HEALTHY survivor, deleted first
07:21:54.933Z  Deleted Pod   repro-0-worker-crplf    (already terminating)
07:21:54.934Z  Deleted Pod   repro-0-worker-kdc57    (already terminating)
07:21:54.935Z  Deleted Pod   repro-0-worker-tqxx4    (already terminating)
07:21:54.935Z  Deleted Pod   repro-0-worker-sfmqf    (already terminating)
07:21:54.935Z  Deleted excess pods   diff: 5, noOfPodsDeleted: 5
```

Outcome — all six original Pods carry a `deletionTimestamp` and one replacement is created. Exactly the reported "all N deleted, then one created":

```
NAME                   IDX   HOSTNAME           READY   DELETING
repro-0-worker-9zxgn   2     repro-0-worker-2   True    2026-08-05T07:26:54Z
repro-0-worker-crplf   0     repro-0-worker-0   True    2026-08-05T07:26:23Z
repro-0-worker-kdc57   3     repro-0-worker-3   True    2026-08-05T07:26:23Z
repro-0-worker-sfmqf   5     repro-0-worker-5   True    2026-08-05T07:26:23Z
repro-0-worker-tqxx4   1     repro-0-worker-1   True    2026-08-05T07:26:23Z
repro-0-worker-wkwbn   4     repro-0-worker-4   True    2026-08-05T07:26:23Z
repro-0-worker-t2fht   0     repro-0-worker-0   True    <none>              <-- replacement
```

The `numExcessPods: 5` line is the smoking gun: at that moment only **one** Pod exceeded the desired count of 1, but `diff` was recomputed as `6 existing + 0 createExp - 1 replicas - 0 deleteExp = 5` because the delete expectations did not survive the restart (see the companion `SyncExpectations` issue — the log line reads `created expectations`, not `raised expectations`, which only happens when `GetExpectations` returns `exists == false`).

### Reproduction requires Pods of the same age — criterion 5 masks the defect otherwise

An earlier attempt at this reproduction failed twice in a row for a reason worth recording, because it defines the blast radius.

If the sample is produced by scaling an existing PodClique back up, the survivor of the first pass is strictly **older** than its replacements. `DeletionSorter.Less` criterion 5 (`deletionsort.go:71-75`) is `s.Pods[i].CreationTimestamp.After(s.Pods[j].CreationTimestamp.Time)` — newer Pods sort first — so the oldest Pod is deterministically pushed to the back of the candidate list and survives. In two consecutive runs set up that way, the second deletion pass picked exactly the already-terminating Pods and spared the healthy one.

So the defect is exposed only when the surviving Pod is not older than the doomed ones, which is the case whenever Pods are batch-created (`metav1.Time` is second-granular, so a batch shares one timestamp):

| Scenario | Pod ages | Criterion 5 | Exposed? |
|---|---|---|---|
| Scale-in shortly after initial deploy | all identical | ties | yes |
| Scale-in after a single-step scale-up | all identical | ties | yes |
| Scale-in after a rolling update or failure replacement | survivor older | protects oldest | masked |
| Scale-in after a staged scale-up | mixed | partial | depends which Pod survives |

The exposed rows are the common operational path, which matches the intermittent-but-recurring reports.

### Collateral defects observed in the same run

Two further problems were captured in the post-reproduction state and are worth separate issues:

- **Duplicate Pod ordinal and hostname.** The replacement `repro-0-worker-t2fht` was assigned index `0`, which the still-terminating `repro-0-worker-crplf` also holds. Both are live and both have `spec.hostname: repro-0-worker-0`, so the DNS name `repro-0-worker-0.<svc>.<ns>.svc.cluster.local` is ambiguous. Cause: `index.GetAvailableIndices` counts only *active* Pods as occupying an ordinal, while `componentutils.GetPCLQPods` returns terminating Pods too.
- **Internally inconsistent PodClique status.** `status.replicas=1` while `status.readyReplicas=7` and `status.scheduledReplicas=7`. `mutateReplicas` derives `Replicas` excluding terminating Pods, but `ReadyReplicas`/`ScheduledReplicas` come from `CategorizePodsByConditionType` buckets that include them.
- **Service impact is worse than a plain gap.** `HeadlessServiceConfig.PublishNotReadyAddresses` defaults to `true`, so all six doomed Pods remained in the Service endpoints after being marked for deletion. Clients are not failing to resolve a backend; they are being routed to Pods that are about to be SIGKILLed.

### Why this reproduces only intermittently

Four independent sources of nondeterminism stack, which is why the same `kubectl scale` succeeds most of the time:

1. **A second deletion pass has to happen at all.** The first pass is correct and the expectations store then holds `diff` at 0. Only trigger (1) or (2) above produces a second pass, and both are conditional.
2. **The terminating Pods must still report `Ready`.** `DeletionSorter.Less` criterion 3 (`deletionsort.go:62-64`) puts not-Ready before Ready. Once kubelet flips the terminating Pods to `Ready=False`, they sort *ahead* of the survivor and the bug is masked. The exposure window is therefore from `deletionTimestamp` being set until readiness flips — which depends on how fast the container stops passing its readiness probe after SIGTERM, i.e. on the workload, not on Grove.
3. **All five sort criteria tie in the common case.** Criterion 1 (unassigned/assigned) ties — all Pods are scheduled. Criterion 2 (phase) ties — terminating Pods stay `Running` until containers exit. Criterion 3 ties inside the window above. Criterion 4 (template hash) ties — under trigger (1) all Pods still carry the old hash. Criterion 5 compares `CreationTimestamp`, but `metav1.Time` has **second** granularity, so Pods created in the same scale-up batch usually have byte-identical timestamps and `After()` returns false in both directions.
4. **With everything tied, the outcome is a coin flip.** `syncflow.go:239` uses `sort.Sort`, which is **not stable**, so the resulting order depends entirely on the input order. That input order is `sc.existingPCLQPods`, which comes from `componentutils.GetPCLQPods` → `cl.List` against the controller-runtime informer cache (`operator/internal/controller/common/component/utils/pod.go:33-55`) with no ordering applied. The cache iterates a Go map, so the order genuinely varies between calls. For 3 Pods choosing 2 deletion slots, the healthy survivor is hit roughly 2 times out of 3 — but only when steps 1–3 all line up.

A useful fingerprint when triaging: if the PodClique comes back with the **scaled** replica count (1), the PodClique object survived and this is pod-level over-deletion. Every path that deletes the PodClique object itself (PCS/PCSG gang termination) recreates it from the PCS template with the original count, because `buildResource` only preserves live replicas when the object already exists:

```go
// operator/internal/controller/podcliqueset/components/podclique/podclique.go:310-317
if pclqExists {
	currentPCLQReplicas := pclq.Spec.Replicas
	pclq.Spec = pclqTemplateSpec.Spec
	pclq.Spec.Replicas = currentPCLQReplicas
} else {
	pclq.Spec = pclqTemplateSpec.Spec
}
```

Scaling from 3 to 1 should delete exactly 2 Pods and keep 1 Pod serving throughout. Pods that are already terminating, or that already have a delete expectation recorded, should never be counted as deletion candidates.

Suggested fix (either is sufficient):

- Filter `hasPodDeletionBeenTriggered(sc, pod)` out of `selectExcessPodsToDelete`, mirroring `computeUpdateWork` at `rollingupdate.go:145-148`; or
- Add a highest-priority criterion to `DeletionSorter.Less` so that `DeletionTimestamp != nil` sorts first (already-doomed Pods absorb the deletion slots harmlessly).

### Telling the inflation causes apart in the log

All causes surface here as a **second** `Deleted excess pods` line. Which cause it was is readable
from the neighbouring lines:

| Evidence | Cause |
|---|---|
| second pass logs `created expectations` and `diff` equals the full excess (N-1) | **#3** — store entry was absent |
| second pass logs `raised expectations` and `diff` exceeds the true excess by exactly 1 | **#7** — unlocked map read, or the TOCTOU |
| `--previous` container log ends in `fatal error: concurrent map iteration and map write` | **#7** loud form, which then produced **#3** |
| only **two** `Deleted Pod` lines and the third Pod vanished with no Grove log at all | not Grove — external eviction / preemption / human |

With `-v=4` the whole computation is printed verbatim by the `synced expectations` line
(`syncflow.go:178-184`): `existingPCLPodNames`, `createExpectations`, `deleteExpectations`, `diff`.

### Environment

- Kubernetes version: any
- Grove version: `main` @ `b5621d4`
- Scheduler details: reproducible with the default `kube` backend; not scheduler-dependent
- Cloud provider or hardware configuration: n/a
- Anything else that is relevant: default `RollingRecreate` update strategy; `terminationGracePeriodSeconds` defaults to 30s (`operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go`), which is the window in which the stale list is observable

kangclzjc:
## What type of PR is this?
/kind bug

## What this PR does / why we need it:

`componentutils.GetPCLQPods` returns terminating Pods, and a Pod keeps reporting `Running` and `Ready` for the whole of its `terminationGracePeriodSeconds`. Two things followed from that on the scale-in path:

- `selectExcessPodsToDelete` computed the excess as `len(sc.existingPCLQPods) - spec.Replicas`, counting Pods that were already on their way out.
- `DeletionSorter.Less` had no `deletionTimestamp` criterion, so for batch-created Pods **all five** existing criteria tie — same node assignment, same `Running` phase, same `Ready` status, same pod-template-hash, and identical `creationTimestamp` because `metav1.Time` is second-granular. `sort.Sort` then performs zero swaps and the resulting order is whatever the informer cache's Go-map `List()` happened to return.

The net effect is that whenever the deletion budget exceeds the true excess, it can be spent on a Pod that is still serving instead of on one that is already terminating.

The rolling update path in the same package already gets this right — `computeUpdateWork` skips Pods via `hasPodDeletionBeenTriggered` (`rollingupdate.go`). The scale-in path did not.

### Changes

- `selectExcessPodsToDelete` becomes a method on `_resource` and filters out Pods that are terminating or already carry a delete expectation, so the excess count is computed over live Pods only. Reusing `hasPodDeletionBeenTriggered` keeps both paths on one definition of "already being deleted".
- Filtering into a fresh slice also stops `sort.Sort` from reordering `sc.existingPCLQPods` in place. Later steps of the same sync flow (`checkAndRemovePodSchedulingGates`, `computeUpdateWork`) read that slice.
- `DeletionSorter.Less` gains a `deletionTimestamp` criterion ahead of the existing five, as defence in depth: a caller that does hand it terminating Pods now spends the budget on Pods that are already doomed.

### Tests

- `deletionsort_test.go` (new): the terminating Pod sorts first regardless of input position; when every Pod is terminating the remaining criteria still decide.
- `syncflow_test.go`: table test for `selectExcessPodsToDelete` covering no-terminating-Pods, all-excess-terminating, mixed, and the delete-expectation-recorded case, asserting that no selected Pod is one already being deleted.

`go test ./internal/controller/podclique/...` and `golangci-lint run` on the package are clean.

Fixes #758

The behaviour change is confined to which Pods are *eligible* for deletion; the `diff` computation in `syncExpectationsAndComputeDifference` is untouched. On a healthy scale-in with no terminating Pods the selection is identical to before.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podclique/components/pod/deletionsort.go
+++ b/operator/internal/controller/podclique/components/pod/deletionsort.go
@@ -44,6 +44,15 @@ var podPhaseToOrdinal = map[corev1.PodPhase]int{corev1.PodPending: 0, corev1.Pod
 // Code partially adapted from https://github.com/kubernetes/kubernetes/blob/5a450884b127f7b8e477d48cf3967a2a5eca9126/pkg/controller/controller_utils.go#L702
 // Only 4 conditions have been taken as is and used here.
 func (s DeletionSorter) Less(i, j int) bool {
+	// 0. Already terminating < not terminating
+	// A Pod that has been marked for deletion keeps reporting Running and Ready until its containers
+	// actually stop, so none of the criteria below can distinguish it from a healthy Pod. Sorting it
+	// first ensures that any surplus deletion budget is spent on Pods that are already on their way
+	// out rather than on Pods that are still serving.
+	if iTerminating, jTerminating := s.Pods[i].DeletionTimestamp != nil, s.Pods[j].DeletionTimestamp != nil; iTerminating != jTerminating {
+		return iTerminating
+	}
+
 	// 1. Unassigned < assigned
 	// If only one of the pods is unassigned, the unassigned one is smaller
 	if s.Pods[i].Spec.NodeName != s.Pods[j].Spec.NodeName && (len(s.Pods[i].Spec.NodeName) == 0 || len(s.Pods[j].Spec.NodeName) == 0) {
--- a/operator/internal/controller/podclique/components/pod/deletionsort.go
+++ b/operator/internal/controller/podclique/components/pod/deletionsort.go
@@ -16,6 +16,7 @@ package pod
 
 import (
 	apicommon "github.com/ai-dynamo/grove/operator/api/common"
+	k8sutils "github.com/ai-dynamo/grove/operator/internal/utils/kubernetes"
 
 	corev1 "k8s.io/api/core/v1"
 )
@@ -49,8 +50,8 @@ func (s DeletionSorter) Less(i, j int) bool {
 	// actually stop, so none of the criteria below can distinguish it from a healthy Pod. Sorting it
 	// first ensures that any surplus deletion budget is spent on Pods that are already on their way
 	// out rather than on Pods that are still serving.
-	if iTerminating, jTerminating := s.Pods[i].DeletionTimestamp != nil, s.Pods[j].DeletionTimestamp != nil; iTerminating != jTerminating {
-		return iTerminating
+	if isPodTerminating(s.Pods[i]) != isPodTerminating(s.Pods[j]) {
+		return isPodTerminating(s.Pods[i])
 	}
 
 	// 1. Unassigned < assigned
@@ -80,6 +81,11 @@ func (s DeletionSorter) Less(i, j int) bool {
 		return s.Pods[i].CreationTimestamp.IsZero()
 	}
 	return s.Pods[i].CreationTimestamp.After(s.Pods[j].CreationTimestamp.Time)
+}
+
+// isPodTerminating checks if a pod has already been marked for deletion.
+func isPodTerminating(pod *corev1.Pod) bool {
+	return k8sutils.IsResourceTerminating(pod.ObjectMeta)
 }
 
 // isPodReady checks if a pod is ready by looking for the PodReady condition with status True
--- a/operator/internal/controller/podclique/components/pod/deletionsort.go
+++ b/operator/internal/controller/podclique/components/pod/deletionsort.go
@@ -46,10 +46,14 @@ var podPhaseToOrdinal = map[corev1.PodPhase]int{corev1.PodPending: 0, corev1.Pod
 // Only 4 conditions have been taken as is and used here.
 func (s DeletionSorter) Less(i, j int) bool {
 	// 0. Already terminating < not terminating
-	// A Pod that has been marked for deletion keeps reporting Running and Ready until its containers
-	// actually stop, so none of the criteria below can distinguish it from a healthy Pod. Sorting it
-	// first ensures that any surplus deletion budget is spent on Pods that are already on their way
-	// out rather than on Pods that are still serving.
+	//
+	// DEFENSIVE: the scale-in caller (selectExcessPodsToDelete) filters terminating Pods out
+	// before sorting, so in that path this criterion never fires. It is kept because
+	// DeletionSorter is exported along with its Pods field and a future caller may hand it an
+	// unfiltered list. A Pod that has been marked for deletion keeps reporting Running and Ready
+	// until its containers actually stop, so none of the criteria below can distinguish it from a
+	// healthy Pod — without this rule such a caller would spend its deletion budget on a Pod that
+	// is still serving.
 	if isPodTerminating(s.Pods[i]) != isPodTerminating(s.Pods[j]) {
 		return isPodTerminating(s.Pods[i])
 	}
--- a/operator/internal/controller/podclique/components/pod/syncflow.go
+++ b/operator/internal/controller/podclique/components/pod/syncflow.go
@@ -203,7 +203,7 @@ func getTerminatingAndNonTerminatingPodUIDs(existingPCLQPods []*corev1.Pod) (ter
 // The deletion of Pods are done in batches of increasing size. This is done to prevent burst of load
 // on the kube-apiserver. It will fail fast in case there is an
 func (r _resource) deleteExcessPods(sc *syncContext, logger logr.Logger, diff int) error {
-	candidatePodsToDelete := selectExcessPodsToDelete(sc, logger)
+	candidatePodsToDelete := r.selectExcessPodsToDelete(sc, logger)
 	numPodsToSelectForDeletion := min(diff, len(candidatePodsToDelete))
 	selectedPodsToDelete := candidatePodsToDelete[:numPodsToSelectForDeletion]
 
@@ -226,19 +226,36 @@ func (r _resource) deleteExcessPods(sc *syncContext, logger logr.Logger, diff in
 	return nil
 }
 
-// selectExcessPodsToDelete identifies excess pods for deletion using DeletionSorter for prioritization
-func selectExcessPodsToDelete(sc *syncContext, logger logr.Logger) []*corev1.Pod {
-	var candidatePodsToDelete []*corev1.Pod
-	if diff := len(sc.existingPCLQPods) - int(sc.pclq.Spec.Replicas); diff > 0 {
-		logger.Info("found excess pods for PodClique", "numExcessPods", diff)
-		sorter := DeletionSorter{
-			Pods: sc.existingPCLQPods,
+// selectExcessPodsToDelete identifies excess pods for deletion using DeletionSorter for prioritization.
+//
+// Pods whose deletion has already been triggered are excluded from the candidate set. GetPCLQPods
+// returns terminating Pods as well, and a Pod stays Running and Ready for the whole of its
+// terminationGracePeriodSeconds, so counting it as excess spends the deletion budget on a Pod that is
+// already on its way out - and, since DeletionSorter cannot tell it apart from a healthy Pod, can
+// select a Pod that is still serving instead. The rolling update path applies the same rule via
+// hasPodDeletionBeenTriggered (see computeUpdateWork in rollingupdate.go).
+//
+// Filtering into a fresh slice also keeps sort.Sort from reordering sc.existingPCLQPods in place,
+// which later steps of the same sync flow still read.
+func (r _resource) selectExcessPodsToDelete(sc *syncContext, logger logr.Logger) []*corev1.Pod {
+	livePods := make([]*corev1.Pod, 0, len(sc.existingPCLQPods))
+	for _, pod := range sc.existingPCLQPods {
+		if r.hasPodDeletionBeenTriggered(sc, pod) {
+			continue
 		}
-		sorter.ExpectedPodTemplateHash = sc.getExpectedPodTemplateHash()
-		sort.Sort(sorter)
-		candidatePodsToDelete = append(candidatePodsToDelete, sorter.Pods[:diff]...)
+		livePods = append(livePods, pod)
 	}
-	return candidatePodsToDelete
+	numExcessPods := len(livePods) - int(sc.pclq.Spec.Replicas)
+	if numExcessPods <= 0 {
+		return nil
+	}
+	logger.Info("found excess pods for PodClique", "numExcessPods", numExcessPods)
+	sorter := DeletionSorter{
+		Pods:                    livePods,
+		ExpectedPodTemplateHash: sc.getExpectedPodTemplateHash(),
+	}
+	sort.Sort(sorter)
+	return sorter.Pods[:numExcessPods]
 }
 
 func (sc *syncContext) getExpectedPodTemplateHash() string {
--- a/operator/internal/controller/podclique/components/pod/deletionsort_test.go
+++ b/operator/internal/controller/podclique/components/pod/deletionsort_test.go
@@ -61,10 +61,13 @@ func podNames(pods []*corev1.Pod) []string {
 	return names
 }
 
-// TestDeletionSorterPrefersTerminatingPods asserts the criterion that keeps a surplus deletion budget
-// away from Pods that are still serving. A Pod inside its termination grace period keeps reporting
-// Running and Ready, so without an explicit deletionTimestamp criterion it ties with a healthy Pod on
-// every other rule and the resulting order is whatever the caller happened to pass in.
+// TestDeletionSorterPrefersTerminatingPods pins the defensive criterion 0.
+//
+// The scale-in caller (selectExcessPodsToDelete) filters terminating Pods out before sorting, so this
+// criterion does not fire on that path. It exists for callers that hand DeletionSorter an unfiltered
+// list — the type and its Pods field are both exported. A Pod inside its termination grace period
+// keeps reporting Running and Ready, so without an explicit deletionTimestamp rule it ties with a
+// healthy Pod on every other criterion and the resulting order is whatever the caller passed in.
 func TestDeletionSorterPrefersTerminatingPods(t *testing.T) {
 	createdAt := metav1.NewTime(time.Now())
 
--- a/operator/internal/controller/podclique/components/pod/deletionsort_test.go
+++ b/operator/internal/controller/podclique/components/pod/deletionsort_test.go
@@ -1,0 +1,111 @@
+// /*
+// Copyright 2025 The Grove Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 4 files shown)


operator/internal/controller/podclique/components/pod/deletionsort.go
```diff
@@ -44,6 +46,15 @@ var podPhaseToOrdinal = map[corev1.PodPhase]int{corev1.PodPending: 0, corev1.Pod
 // Code partially adapted from https://github.com/kubernetes/kubernetes/blob/5a450884b127f7b8e477d48cf3967a2a5eca9126/pkg/controller/controller_utils.go#L702
 // Only 4 conditions have been taken as is and used here.
 func (s DeletionSorter) Less(i, j int) bool {
+	// 0. Already terminating < not terminating
+	// A Pod that has been marked for deletion keeps reporting Running and Ready until its containers
+	// actually stop, so none of the criteria below can distinguish it from a healthy Pod. Sorting it
+	// first ensures that any surplus deletion budget is spent on Pods that are already on their way
+	// out rather than on Pods that are still serving.
+	if iTerminating, jTerminating := s.Pods[i].DeletionTimestamp != nil, s.Pods[j].DeletionTimestamp != nil; iTerminating != jTerminating {
+		return iTerminating
```
shayasoolin: Why is it always `iTerminating` that's being returned? What if `iTerminating == nil` but `jTerminating != nil`? If I'm that's possible, `jTerminating` should be returned. Or am I missing something?

kangclzjc: Good catch on the readability — the logic was correct but the way I wrote it invited exactly this question, so I've rewritten it. 

sort.Interface.Less(i, j) answers "should i sort before j", so one bool covers both directions. The guard means we only reach the return when exactly one of the two is terminating:
  | `i` terminating | `j` terminating | returns | effect |
  | --- | --- | --- | --- |
  | yes | no | `true` | `i` sorts first ✓ |
  | no | yes | `false` | `j` sorts first ✓ |

  Returning jTerminating instead would invert the second row.

shayasoolin (dismissed):

yankay:
Thanks for the fix. I reproduced the scale-in issue locally: after scaling from 6 to 1 and restarting the operator, the base revision deleted the last healthy Pod. This PR preserved it in repeated runs, kept scale-to-zero working, and eliminated a simulated 12-second availability gap. LGTM :)

operator/internal/controller/podclique/components/pod/syncflow.go
```diff
@@ -226,19 +227,36 @@ func (r _resource) deleteExcessPods(sc *syncContext, logger logr.Logger, diff in
...
+// Pods whose deletion has already been triggered are excluded from the candidate set. GetPCLQPods
+// returns terminating Pods as well, and a Pod stays Running and Ready for the whole of its
+// terminationGracePeriodSeconds, so counting it as excess spends the deletion budget on a Pod that is
+// already on its way out - and, since DeletionSorter cannot tell it apart from a healthy Pod, can
+// select a Pod that is still serving instead. The rolling update path applies the same rule via
+// hasPodDeletionBeenTriggered (see computeUpdateWork in rollingupdate.go).
+//
+// Filtering into a fresh slice also keeps sort.Sort from reordering sc.existingPCLQPods in place,
+// which later steps of the same sync flow still read.
+func (r _resource) selectExcessPodsToDelete(sc *syncContext, logger logr.Logger) []*corev1.Pod {
```
oleg-kushniriov, line 240: Refresh the branch before merge; this PR is based before later fixes in syncflow.go 

Current `main` preserves third-party scheduling gates by removing only the Grove PodGang gate and keeps a `podNamesUpdatedInPCLQPodGangSet` lookup. 
The PR branch head still has the older implementation that clears `p.Spec.SchedulingGates = nil`.


This is risky enough to require a branch refresh before approval

operator/internal/controller/podclique/components/pod/deletionsort.go
```diff
@@ -44,6 +47,15 @@ var podPhaseToOrdinal = map[corev1.PodPhase]int{corev1.PodPending: 0, corev1.Pod
 // Code partially adapted from https://github.com/kubernetes/kubernetes/blob/5a450884b127f7b8e477d48cf3967a2a5eca9126/pkg/controller/controller_utils.go#L702
 // Only 4 conditions have been taken as is and used here.
 func (s DeletionSorter) Less(i, j int) bool {
+	// 0. Already terminating < not terminating
```
oleg-kushniriov, line 48: `selectExcessPodsToDelete` now filters terminating/delete-expected pods before sorting, so the new terminating-pod comparator is unused by scale-in. 
That makes the contract confusing: the caller excludes these pods, while the sorter prefers them.

Proposed wording:
Consider dropping the terminating-pod criterion and its dedicated sorter test, or document it as defensive behavior for future direct sorter callers.

operator/internal/controller/podclique/components/pod/syncflow.go
kangclzjc: You were right, thanks for catching this. Rebased onto `main` and force-pushed

operator/internal/controller/podclique/components/pod/deletionsort.go
kangclzjc: Agreed that the contract read as contradictory. I went with your second option — keeping the criterion but documenting it as defensive — rather than dropping it, so a future caller can hand it an unfiltered list and would silently get the pre-fix behaviour back.

oleg-kushniriov (approved):

shayasoolin (approved):

PR #774, opened 2026-08-13
feat: add scale dashboard overview filters

shayasoolin:
Refs #550 

## Summary
- add an all-tests total-runtime overview with per-test colored series
- add header date-range filtering with clear/reset behavior
- retain selected-test detailed charts below

## Validation
- `node --check operator/hack/scale-dashboard/app.js`
- `git diff --check`
- verified date filtering against the current 60-run scale-test-history dataset

Before:
image

After:
image

### DIFF
--- a/operator/hack/scale-dashboard/app.js
+++ b/operator/hack/scale-dashboard/app.js
@@ -49,6 +49,7 @@ const state = {
   testName: "",
   startDate: "",
   endDate: "",
+  hiddenOverviewTests: new Set(),
 };
 
 const els = {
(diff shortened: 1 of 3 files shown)


yankay:
Thanks for putting this together; it looks great! Would you consider making the legend entries clickable so each series can be shown or hidden? SoakChurn stretches the Y-axis, making the other trends harder to distinguish. Hiding it would let the chart rescale and make comparisons clearer. :)

Prototype: https://github.com/yankay/grove/commit/a98b39a0f03db0fa05c3e948c98d85c3b086830a

a581d758-eac4-4b69-b208-11ca61b6053a

Thanks :)

shayasoolin:
@yankay 
> Thanks for putting this together; it looks great! Would you consider making the legend entries clickable so each series can be shown or hidden? SoakChurn stretches the Y-axis, making the other trends harder to distinguish. Hiding it would let the chart rescale and make comparisons clearer. :)

Yes! Great feedback, done. 🙏🏼

Ronkahn21 (approved):

oleg-kushniriov (approved):

PR #649, opened 2026-06-06
GREP-393 implementation: Coherent Rolling Updates for PodCliqueSet

unmarshall:
## What type of PR is this?
/kind feature
/kind api

### Summary

This PR implements [GREP-393: Coherent Rolling Updates](https://github.com/ai-dynamo/grove/pull/640) for `PodCliqueSet`.

For motivation, goals/non-goals, user stories, and conceptual design, see #640. This PR description focuses on what's implemented and how to review it.

 The detailed design with file paths, function names, invariants, and the state-machine walkthrough is attached as a reviewer artifact (`coherent-updates-implementation-plan.md`) rather than committed to the repo.

  ### What's in this PR

  #### New API surface

  - **`PodGangMap` CRD** (`operator/api/core/v1alpha1/podgangmap.go`) — namespaced, one per PCS replica. Holds the desired PodGang composition: per-entry `PodCliqueSetGenerationHash`, standalone-PodClique pod
   counts, PCSG replica indices, and `DependsOn` for cross-PodGang scheduling order.
  - **`CoherentStrategy`** constant on `PodCliqueSetUpdateStrategy.Type` (kubebuilder default = `Coherent`).
  - **`PodCliqueSetReplicaUpdateProgress.InFlightPodGangs`** + `ErrorMessage` fields.
  - **`PodClique.Status.PodGangMapping`** (`map[string]int32`) and **`PodCliqueScalingGroup.Status.PodGangMapping`** (`map[string][]int32`) — decision-driven, authored by the child reconcilers; PodGangMap is a steady-state follower of these.
  - **`PodGangConditionTypeAvailable`** on the scheduler API. Set True after `MinReplicas` is released and all min-replica pods are ready. The orchestrator advances iterations on this condition.
  - **New labels**: `grove.io/podcliqueset-replica-index` and `grove.io/podcliqueset-generation-hash` are stamped on every PodGang at creation.

  #### Controller and component changes

  - **PodGangMap component** (`operator/internal/controller/podcliqueset/components/podgangmap/`) — desired-state author. Three regimes: fresh PCS bootstrap → MVU entries from spec; existing PCS pre-dating Coherent → reconstruction from live PodGangs; coherent update in progress → mixed old-hash and new-hash entries. Gates on `InFlightPodGangs` empty and on the prior MPG reaching `Available` before authoring the next iteration.
[...]

PR #778, opened 2026-08-15
Coherent Update PreReq PR#1: Migration of existing PodGang names to use epoch based names

### LINKED ISSUE #777: PodGang migration + PodGangMap + epoch-based naming

Introduce `PodGangMap` CRD which segregates the set of PodGangs into Anchor/Tail/ScaleOut roles/categories. An entry is identified by an `Epoch` which is guaranteed to be unique across the PodGangMap resource. Each such entry captures information like (PCS generation hash, Anchor information, number of pods for each PCLQ that belongs to this Epoch and replica indices for each PCSG that belong to this epoch. 

* During coherent update this will capture the step/sub-step drain of old PodGangEnty's and creation of new PodGangEntry's. In steady
* During steady state this will be synced with all scale-out/in that happen for every standalone PCLQ and PCSG resources.

This resource is created per PCS replica and serves as a single source of truth for the PodGang component which is responsible for reconciling (creating/updating/deleting) PodGang resources.

### Why is this needed?

We are harmonizing the PodGang name scheme. This PR introduces the first batch of changes that are required to realise coherent-update. It ensures that all PodGang names (existing and new) use a single scheme. This will also help ease the complexity of next PRs as they do not have to handle both legacy and new naming scheme for the PodGang resources.

> NOTE: It is mandatory that all consumers should update to the Grove release that has this PR. This ensures that the migrator code will run and migrate all legacy PodGang to the new naming scheme.

unmarshall:
#### What type of PR is this?

/kind feature
/kind api

##### Summary

PR #649 is a large PR and the decision is to break this PR into smaller PRs. This will ensure that the changes required for coherent-updates are introduced in batches, each building on top of another. This PR is the first batch of changes.

Introduces the **PodGangMap (PGM)** as the single source of truth to materialise PodGangs, and moves PodGang naming from the legacy scheme to an **epoch-based** scheme. A migration component brings existing (legacy) PodCliqueSets onto the new scheme in place, without recreating pods.

##### What changes this PR brings in?

**PodGangMap (new CRD + component)**
  - New `PodGangMap` CRD, one per PodCliqueSet replica, whose entries record each scheduling batch (epoch), its role (Anchor / Tail / ScaleOut), the standalone PodCliques and PodCliqueScalingGroup replica indices it carries, and its scheduling dependencies (`DependsOn`).
  - A spec-driven PodGangMap component bootstraps and reconciles the map.

**Epoch-based PodGang naming**
  - Anchor PodGang: `<pcs>-<replica>-<epoch>`
  - Non-Anchor PodGang: `<pcs>-<replica>-<epoch>-<pcsg>-<index>`.
  - The PodGang component derives all PodGangs (names, labels, and scheduling-gate dependencies) from the PodGangMap rather than from parsed names.
  - PodGangs carry `grove.io/epoch` and `grove.io/podgang-role` labels. The legacy `grove.io/base-podgang` label is dropped from the new scheme.

**Migration of legacy PodCliqueSets**
  - A new `PodGangMigrator` component (a sync group that runs after the PodGangMap component and before the PodClique/PodGang/PodCliqueScalingGroup components) rewrites the `grove.io/podgang` label on every PodClique and Pod of a legacy replica to the epoch-based name derived from the PodGangMap, and drops `grove.io/base-podgang`. It creates and deletes no PodGang. The PodGang component creates the epoch-named PodGangs and deletes the legacy ones once the Pods carry the new label.
  - A `PodGangMigrationInProgress` condition is introduced. It is set before the controller manager starts on every legacy PodCliqueSet, gates the PodCliqueScalingGroup and PodClique reconcilers so scaling cannot interleave with migration. The migration component clears it once every replica is migrated. New PodCliqueSets are already created with the new scheme and are never gated.

##### Why the migrator is a component (not a pre-start batch)?
Making it a component provides the following benefits:
* We re-use the PodGangMap component's ability to create PodGangMap resources for existing PCS.
* An existing Ability (component architecture) to auto-requeue on error is re-used by the migrator component.
* The migrator component only updates the PCLQ and Pod labels and additionally leverages the PodGang component to create new PodGangs (based on epoch-based scheme) and delete excess PodGangs (legacy name based PodGangs will become excess).

> NOTE: Migrator code (pre-controller-manager + component) is a short lived code and is introduced to automate the migration of existing PodGangs. Once all PodCliqueSet resources have been migrated then we can disable and eventually remove this component.

Fixes #777

```release-note
- Introduces a new PodGangMap custom resource. This resource will serve as a single source of truth to materialise PodGang resources. 
- A new naming scheme has been introduced for PodGangs.
- A migrator component is introduced which will auto-migrate all existing (now legacy) PodGang resources to the new scheme. Without disrupting the running workloads it will also patch labels on existing PCLQ and Pods to reflect the new PodGang and related labels.
```

```docs
API docs have been generated for the new PodGangMap resource.
```


operator/internal/podgangmigrator/gate.go
```diff
@@ -0,0 +1,97 @@
...
+		meta.SetStatusCondition(&pcs.Status.Conditions, metav1.Condition{
+			Type:    apiconstants.ConditionTypePodGangMigrationInProgress,
+			Status:  metav1.ConditionTrue,
+			Reason:  "LegacyPodGangsPendingMigration",
+			Message: "PodCliqueSet predates the epoch-based PodGang scheme and is pending migration",
+		})
+		if err := cl.Status().Update(ctx, pcs); err != nil {
+			return fmt.Errorf("failed to set PodGang migration gate on PodCliqueSet %v: %w", client.ObjectKeyFromObject(pcs), err)
+		}
+		logger.Info("Set PodGang migration gate on legacy PodCliqueSet", "podCliqueSet", client.ObjectKeyFromObject(pcs))
```
renormalize: If a status update for a single PodCliqueSet fails (for any reason), we simply return here and the process exits. Perhaps we could implement a retry for the status update with a limited number of retries, before we return an error here that causes the process to exit? Relying on a container restart to re-attempt a resource status update is very heavy handed.

`k8s.io/client-go/util/retry` provides `retry.OnError()` which we could possibly make use of here.

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,121 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName={pgm}
+
+// PodGangMap is the desired-state mapping between PodGangs and their constituent
+// PodClique and PodCliqueScalingGroup pod counts for a single PodCliqueSet replica.
+// One PodGangMap resource exists per PodCliqueSet replica, named <pcs-name>-<pcs-replica-index>.
+type PodGangMap struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Spec defines the desired PodGang-to-pod-count mapping for this PodCliqueSet replica.
```
renormalize: The specification talks generally, not in specifics.

```suggestion
	// Spec defines the desired PodGang-to-pod-count mapping for a PodCliqueSet replica.
```

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
```diff
@@ -0,0 +1,335 @@
...
+	var anchorEpoch string
+	for i := range entries {
+		if entries[i].Role == grovecorev1alpha1.PodGangEntryRoleScaleOut && entries[i].PodCliqueSetGenerationHash == currentHash {
+			return entries
+		}
+		if entries[i].Role == grovecorev1alpha1.PodGangEntryRoleAnchor && entries[i].PodCliqueSetGenerationHash == currentHash && entries[i].AnchorIndex == 0 {
+			anchorEpoch = entries[i].Epoch
+		}
+	}
+	scaleOut := newPodGangEntry(scaleOutEpoch, currentHash, []string{anchorEpoch})
```
yankay, line 322: Thanks for adding this path. Could we guard against an empty `anchorEpoch` here? With `OnDelete`, the anchor can retain the old generation hash, producing `DependsOn: [""]`; the pod controller then queries `<pcs>-<replica>-` and keeps scale-out Pods scheduling-gated indefinitely. I reproduced this with the full sync flow and a default-scheduler cluster. Resolving the effective anchor or requeueing here, plus a regression test, should prevent this.

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,121 @@
...
+	Entries []PodGangEntry `json:"entries"`
+}
+
+// PodGangEntryRole classifies the role a PodGangMap entry plays in a PodCliqueSet replica.
+// +kubebuilder:validation:Enum={Anchor,Tail,ScaleOut}
+type PodGangEntryRole string
+
+const (
+	// PodGangEntryRoleAnchor marks the entry that carries the MinAvailable replicas.
+	// It holds every standalone PodClique and each PodCliqueScalingGroup's MinAvailable replicas.
```
renormalize, line 65:
```suggestion
	// It holds all pods of every standalone PodClique, and MinAvailable replicas of each PodCliqueScalingGroup.
```

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,121 @@
...
+	PodGangEntryRoleTail PodGangEntryRole = "Tail"
+	// PodGangEntryRoleScaleOut marks the entry that holds PodCliqueScalingGroup replicas added by a
+	// steady-state scale-out beyond the template. It materializes into one PodGang per replica index.
+	// It is created on the first scale-out and removed once it holds no replicas.
+	PodGangEntryRoleScaleOut PodGangEntryRole = "ScaleOut"
+)
+
+// PodGangEntry describes one scheduling batch, identified by its epoch, that materializes into one
+// or more PodGangs. An Anchor entry materializes into a single anchor PodGang; a Tail or ScaleOut
+// entry materializes into one PodGang per (PodCliqueScalingGroup, replica index) it carries.
```
renormalize, line 82:
```suggestion
// entry materializes into one PodGang per {PodCliqueScalingGroup, replica index} it carries.
```

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,121 @@
...
+type PodGangEntry struct {
+	// Epoch is the identity of this entry and the group of PodGangs materialized from it. It serves
+	// two purposes. First, identity, it is unique across entries within a PodGangMap and is the
+	// listMapKey, and every PodGang materialized from this entry carries it as the grove.io/epoch
+	// label so those PodGangs are grouped by it. Second, ordering, DependsOn references epochs, and
+	// comparing epochs orders entries so scheduling dependencies can be expressed and the most recent
+	// anchor found. The value is a monotonic unix-nano integer used only as a distinct, orderable key.
+	// It is not interpreted as a wall-clock time.
+	Epoch string `json:"epoch"`
```
renormalize, line 94:
```suggestion
	// Epoch is the identity of this entry and the group of PodGangs materialized from it. It serves
	// two purposes. First, identity, which is unique across entries within a PodGangMap and is
	// `listMapKey`. Every PodGang materialized from this entry carries it as the `grove.io/epoch`
	// label so those PodGangs are grouped by it. Second, ordering, where `DependsOn` references epochs, and
	// comparing epochs orders entries so scheduling dependencies can be expressed and the most recent
	// anchor found. The value is a monotonic unix-nano integer used only as a distinct, orderable key.
	// It is not interpreted as a wall-clock time.
	Epoch string `json:"epoch"`
```

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,121 @@
...
+	// PodCliqueSetGenerationHash is the PodCliqueSet generation hash that pods in this PodGang
+	// must match. Used by PodClique and PodCliqueScalingGroup reconcilers to create pods at the
+	// correct spec version and to distinguish old pods from new pods during a coherent update.
+	PodCliqueSetGenerationHash string `json:"podCliqueSetGenerationHash"`
+	// Role classifies this entry. See PodGangEntryRole for the meaning of each value. Role is the
+	// durable marker that lets the entry structure be reconstructed from the entries alone, so no
+	// separate record needs to be persisted.
```
renormalize: My understanding from this comment is that the EntryRole allows grove to understand the type of PodGang, and the number of PodGangs that need to be created for the correpsonding entry.

I do not think `lets the entry structure be reconstructed from the entries alone` is the clearest when someone is reading the API for the first time. Can we rephrase/simplify this?

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,121 @@
...
+	Role PodGangEntryRole `json:"role"`
+	// AnchorIndex is the index of an anchor entry within its generation hash. It is set only on
+	// entries whose Role is Anchor and is 0 for every other entry. The index starts at 0 for each
+	// generation hash and increments for each additional anchor of the same hash. It forms the last
+	// segment of the anchor PodGang name.
+	// +optional
+	AnchorIndex int32 `json:"anchorIndex,omitempty"`
```
renormalize: Wouldn't it be easier to make this a pointer? If the entry is **actually** an anchor, we have an index, values starting from 0.

For non-anchor entries, there would be *no need* to define that the anchor index is 0, and does not mean anything. In the currently defined way, index 0 could mean two different things; and making this a pointer simply eliminates that altogether.

operator/api/common/labels.go
```diff
@@ -35,6 +35,18 @@ const (
 	LabelBasePodGang = "grove.io/base-podgang"
 	// LabelPodCliqueSetReplicaIndex is a key for a label that sets the replica index of a PodCliqueSet.
 	LabelPodCliqueSetReplicaIndex = "grove.io/podcliqueset-replica-index"
+	// LabelPodCliqueSetGenerationHash is set on PodGang resources to record the PodCliqueSet generation
+	// hash they were created for.
+	LabelPodCliqueSetGenerationHash = "grove.io/podcliqueset-generation-hash"
+	// LabelEpoch is set on PodGang resources to record the batch in which the PodGang was created.
+	// The value is a monotonic unix-nano integer used as a distinct orderable key. PodGangs created
+	// together share the same epoch. Subsequent batches carry strictly greater values. It is consumed
+	// by the pod-component scheduling-gate-removal logic via the PodGangEntry DependsOn.
```
renormalize: We should not be mentioning implementation details in API documentation. We mention the "pod-component" which could very well not exist in the future. Perhaps this could be reconstructed to only refer to other API fields like `DependsOn`.


```suggestion
	// LabelEpoch is set on PodGang resources to record the batch in which the PodGang was created.
	// The value is a monotonic unix-nano integer used as a distinct orderable key. PodGangs created
	// together share the same epoch. Subsequent batches carry strictly greater values.
```

renormalize (changes requested):
Thanks for the PR @unmarshall. This is the first set of comments I have in mind.

Will post the comments I have on implementation soon. Thanks!

operator/internal/controller/podcliqueset/register.go
```diff
@@ -161,7 +161,10 @@ func podCliqueScalingGroupPredicate() predicate.Predicate {
 			if !okOld || !okNew {
 				return false
 			}
-			return hasMinAvailableBreachedConditionChanged(oldPCSG.Status.Conditions, newPCSG.Status.Conditions) ||
+			// A bare Spec change (e.g. an external Replicas bump for scale-out) does not write PCSG
+			// status, so it must be observed via the generation change to enqueue the PodCliqueSet.
```
renormalize: I assume we need this for PGM updation when PCSG scales? Can that be documented here too? If I am wrong, could you please correct me and note down why this was introduced. The current predicates are somewhat opaque to new developers. We've not clearly documented the reasons behind each filter, so going forward we should attempt to document better.

operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
```diff
@@ -0,0 +1,250 @@
...
+			return nil, groveerr.New(errCodeGroupPodGangsByReplica, component.OperationSync,
+				fmt.Sprintf("PodGang %s has label %s but no %s", existingPodGangs[i].Name, apicommon.LabelEpoch, apicommon.LabelPodCliqueSetReplicaIndex))
+		}
+		pcsReplicaIndex, err := strconv.Atoi(replicaIndexLabel)
+		if err != nil {
+			return nil, groveerr.New(errCodeGroupPodGangsByReplica, component.OperationSync,
+				fmt.Sprintf("%s label on PodGang %s is not a valid integer: %q", apicommon.LabelPodCliqueSetReplicaIndex, existingPodGangs[i].Name, replicaIndexLabel))
+		}
+		podGangsByReplica[pcsReplicaIndex] = append(podGangsByReplica[pcsReplicaIndex], existingPodGangs[i])
+	}
```
renormalize: Nit:

```suggestion
	for _, existingPodGang := range existingPodGangs {
		labels := existingPodGang.Labels
		if _, hasEpoch := labels[apicommon.LabelEpoch]; !hasEpoch {
			continue
		}
		// An epoch-scheme PodGang always carries the replica-index label, so a missing or malformed
		// value is a coding error.
		replicaIndexLabel, ok := labels[apicommon.LabelPodCliqueSetReplicaIndex]
		if !ok {
			return nil, groveerr.New(errCodeGroupPodGangsByReplica, component.OperationSync,
				fmt.Sprintf("PodGang %s has label %s but no %s", existingPodGang.Name, apicommon.LabelEpoch, apicommon.LabelPodCliqueSetReplicaIndex))
		}
		pcsReplicaIndex, err := strconv.Atoi(replicaIndexLabel)
		if err != nil {
			return nil, groveerr.New(errCodeGroupPodGangsByReplica, component.OperationSync,
				fmt.Sprintf("%s label on PodGang %s is not a valid integer: %q", apicommon.LabelPodCliqueSetReplicaIndex, existingPodGang.Name, replicaIndexLabel))
		}
		podGangsByReplica[pcsReplicaIndex] = append(podGangsByReplica[pcsReplicaIndex], existingPodGang)
	}
```

operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
```diff
@@ -0,0 +1,250 @@
...
+	for pcsReplicaIndex := range int(syncSnap.pcs.Spec.Replicas) {
+		pgm, pgmExists := syncSnap.existingPGMByReplica[pcsReplicaIndex]
+
+		if pgmExists && len(pgm.Spec.Entries) == 0 {
+			return groveerr.New(
+				groveerr.ErrCodeRequeueAfter,
+				component.OperationSync,
+				fmt.Sprintf("PodGangMap for replica %d of PodCliqueSet %v has no entries, requeuing", pcsReplicaIndex, client.ObjectKeyFromObject(syncSnap.pcs)),
+			)
+		}
```
renormalize, line 187: I don't think any stale read would ever cause the PGM to have no entries. When the PGM is created, there are two possibilities:
- The cache is not up to date, and therefore the object itself is not present.
- The cache is up to date, and therefore the object is present, and has the entries that it was CREATED with.

It would never be the case that we create an object with a certain spec, but when fetched, the object is returned without its contents. The state of the object you're reading would be something that was never present in the API Server, which violates K8s principles.

Given this, it would not make sense to re-queue here. Instead, wouldn't we want to just recreate the PGM by reconstructing it?

operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
```diff
@@ -0,0 +1,250 @@
...
+		pgmName := apicommon.GeneratePodGangMapName(apicommon.ResourceNameReplica{Name: syncSnap.pcs.Name, Replica: pcsReplicaIndex})
+		if err = r.createOrPatchPodGangMapSpec(ctx, syncSnap.pcs, pgmName, pcsReplicaIndex, entries); err != nil {
+			return err
+		}
+	}
+	return r.deleteOrphanedPodGangMaps(ctx, syncSnap)
+}
+
+// createOrPatchPodGangMapSpec creates or patches the named PodGangMap with the given entries.
+func (r _resource) createOrPatchPodGangMapSpec(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet, pgmName string, pcsReplicaIndex int, entries []grovecorev1alpha1.PodGangEntry) error {
```
renormalize:
```suggestion
// createOrPatchPodGangMap creates or patches the named PodGangMap with the given entries.
func (r _resource) createOrPatchPodGangMap(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet, pgmName string, pcsReplicaIndex int, entries []grovecorev1alpha1.PodGangEntry) error {
```

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
unmarshall: @yankay Thanks for catching this issue.
PGM spec represents a desired state and from that perspective if PCS gets updated, its generation has is also updated in its Status. This should then be reflected for all entries in the PGM for (rolling-recreate and on-delete update strategies) since in both of these strategies there is not going to be any churn of PodGangs. In coherent-update this is different - during coherent update you will have both old and new generation anchor entries.

For rolling-recreate we advance the PCS generation has via `advanceEntriesGenerationHash`. IMO for both on-delete and rolling-recreate the entries remain the same - they do not change and therefore it would make sense to also advance the PCS generation has for all entries when update strategy is on-delete. This should then resolve the issue, right? 

WDYT?

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
```diff
@@ -0,0 +1,335 @@
...
+		if !hasEpoch || !hasRole {
+			continue
+		}
+		epoch, err := strconv.ParseInt(epochStr, 10, 64)
+		if err != nil {
+			continue
+		}
+		if _, seen := byRole[grovecorev1alpha1.PodGangEntryRole(role)]; !seen {
+			byRole[grovecorev1alpha1.PodGangEntryRole(role)] = epoch
+		}
```
renormalize, line 82: If we take an example of an Anchor PG, the only case where we might have multiple anchor entries, is during an update (correct me if I am wrong). Wouldn't it be just better to pick the anchor with the larger numerical value while constructing this, instead of a simple overwrite like this? So in general, instead of a simple overwrite, we can do a comparitive check and pick the larger epoch, as it is guaranteed to be monotonically increasing during the operator's lifecycle.

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
```diff
@@ -0,0 +1,335 @@
...
+// (PodCliqueScalingGroup, index). It returns false when no PodCliqueScalingGroup has replicas above
+// MinAvailable.
+func buildBootstrapTailEntry(pcs *grovecorev1alpha1.PodCliqueSet, epoch, anchorEpoch string) (grovecorev1alpha1.PodGangEntry, bool) {
+	pcsgReplicaIndices := make(map[string][]int32)
+	for _, pcsgConfig := range pcs.Spec.Template.PodCliqueScalingGroupConfigs {
+		replicas := *pcsgConfig.Replicas
+		minAvailable := *pcsgConfig.MinAvailable
+		if replicas <= minAvailable {
+			continue
+		}
```
renormalize: `replicas < minAvailable` is not possible.

```suggestion
		if replicas == minAvailable {
			continue
		}
```

operator/internal/controller/common/component/utils/podcliqueset.go
```diff
@@ -152,3 +152,54 @@ func GetExpectedStandAlonePCLQFQNsPerPCSReplica(pcs *grovecorev1alpha1.PodClique
...
+	result := make(map[string]int32)
+	for _, pcsgConfig := range pcs.Spec.Template.PodCliqueScalingGroupConfigs {
+		result[pcsgConfig.Name] = *pcsgConfig.Replicas
+	}
+	return result
+}
+
+// IsPCSReplicaInCurrentlyUpdating reports whether the given PCS replica index is currently under
+// update, that is it appears in Status.UpdateProgress.CurrentlyUpdating with UpdateEndedAt unset.
+func IsPCSReplicaInCurrentlyUpdating(pcs *grovecorev1alpha1.PodCliqueSet, pcsReplicaIndex int) bool {
```
renormalize: Nit: Other update check utils are named as `<resource>UpdateInProgress`:


```suggestion
func IsPCSReplicaUpdateInProgress(pcs *grovecorev1alpha1.PodCliqueSet, pcsReplicaIndex int) bool {
```

operator/internal/controller/common/component/utils/podcliqueset.go
```diff
@@ -152,3 +152,54 @@ func GetExpectedStandAlonePCLQFQNsPerPCSReplica(pcs *grovecorev1alpha1.PodClique
...
+	if pcs.Status.UpdateProgress == nil {
+		return false
+	}
+	for i := range pcs.Status.UpdateProgress.CurrentlyUpdating {
+		p := &pcs.Status.UpdateProgress.CurrentlyUpdating[i]
+		if int(p.ReplicaIndex) == pcsReplicaIndex && p.UpdateEndedAt == nil {
+			return true
+		}
+	}
```
renormalize: Simplifying:

```suggestion
	for _, currentlyUpdating := range pcs.Status.UpdateProgress.CurrentlyUpdating {
		if int(currentlyUpdating.ReplicaIndex) == pcsReplicaIndex && currentlyUpdating.UpdateEndedAt == nil {
			return true
		}
	}
```

operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
```diff
@@ -0,0 +1,250 @@
...
+			continue
+		}
+		if err := r.client.Delete(ctx, &pgm); err != nil {
+			return groveerr.WrapError(err,
+				errCodeDeletePodGangMaps,
+				component.OperationSync,
+				fmt.Sprintf("Error deleting orphaned PodGangMap %s for PodCliqueSet: %v", pgm.Name, client.ObjectKeyFromObject(syncSnap.pcs)),
+			)
+		}
+		syncSnap.logger.Info("Deleted orphaned PodGangMap", "name", pgm.Name)
```
renormalize:
```suggestion
		syncSnap.logger.Info("Deleted PodGangMap of a scaled-in PCS replica", "name", pgm.Name)
```

renormalize (changes requested):
A few more comments.

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
yankay: Hi @unmarshall , thanks for the explanation. I agree with this fix. Since the failure mode is subtle, would you mind adding a small regression test for the OnDelete update-then-scale-out path as well? Just my thought, and I really appreciate the clarification. :)

operator/api/core/v1alpha1/podgangmap.go
unmarshall: Changed in 5891f968ad73421b320222ff24d1756159eab704

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
```diff
@@ -0,0 +1,230 @@
...
+// migratePodGangLabels rewrites the PodGang-scheme labels on obj to the new scheme. It does the following:
+//   - Set grove.io/podgang to targetPodGangName.
+//   - Drop the legacy grove.io/base-podgang label.
+//
+// It patches only on divergence and touches labels only, never spec, so no Pod is recreated.
+func (r _resource) migratePodGangLabels(ctx context.Context, obj client.Object, targetPodGangName string) error {
+	labels := obj.GetLabels()
+	// The operator is the sole writer of these labels and always stamps the default label set, so an
+	// empty label map is an unexpected internal state that should never happen. Fail loudly rather than
+	// operate on a nil map.
```
renormalize: Nit: comment unnecessarily verbose. The string in the returned error conveys the same.


```suggestion
```

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return "", groveerr.WrapError(err, errCodeResolvePodGangName, component.OperationSync,
+			fmt.Sprintf("failed to resolve PodGang name for %q", objMeta.Name))
+	}
+	return podGangName, nil
+}
+
+// migratePodsOfPodClique rewrites grove.io/podgang to targetPodGangName and drops grove.io/base-podgang
+// on every non-terminating Pod owned by the named PodClique.
+func (r _resource) migratePodsOfPodClique(ctx context.Context, namespace, pclqName, targetPodGangName string) error {
```
renormalize, line 180: I was under the impression that pod migration would happen by the PCLQ controller, but for that to happen, we'd need to ungate the PCLQs. I'm assuming we are doing this here instead because PG component populates podreferences in PCS controller and we can't coordinate this action with the PCLQ controller's patching the PG label on the pods?

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
```diff
@@ -0,0 +1,230 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// resolveTargetPodGangName returns the epoch-based PodGang name the given PodClique or Pod must carry.
+// A standalone object, one without the grove.io/podcliquescalinggroup label, belongs to the anchor
+// PodGang. A PodCliqueScalingGroup-owned object is resolved through the PodGangMap entry that owns its
+// replica index.
```
renormalize:
```suggestion
// resolveTargetPodGangName returns the epoch-based PodGang name the given PodClique or Pod must carry.
// A standalone PodClique (and its pods) belongs to the anchor PodGang.
// A PodCliqueScalingGroup-owned object is resolved through the PodGangMap entry that owns its
// replica index.
```

operator/internal/controller/common/component/utils/podgangmap.go
```diff
@@ -0,0 +1,147 @@
...
+
+// AnchorPodGangEpoch returns the epoch of the AnchorIndex 0 anchor entry of the PodGangMap. Standalone
+// PodCliques always belong to this entry. It returns an error when no such anchor entry exists, a
+// contract violation that must be re-queued.
+func AnchorPodGangEpoch(pgm *grovecorev1alpha1.PodGangMap) (string, error) {
+	for i := range pgm.Spec.Entries {
+		entry := &pgm.Spec.Entries[i]
+		if entry.Role == grovecorev1alpha1.PodGangEntryRoleAnchor && entry.AnchorIndex == 0 {
+			return entry.Epoch, nil
+		}
```
renormalize: Can we just add a comment documenting that this will change in the future when updates are implemented, which causes more than one anchor to exist? Right now only the `0` index's epoch is returned.

operator/internal/controller/podclique/components/pod/deletionsort.go
```diff
@@ -38,7 +38,7 @@ func (s DeletionSorter) Swap(i, j int) {
 }
 
 // podPhaseToOrdinal maps pod phases to deletion priority order (lower values are deleted first)
-var podPhaseToOrdinal = map[corev1.PodPhase]int{corev1.PodPending: 0, corev1.PodUnknown: 1, corev1.PodRunning: 2}
+var podPhaseToOrdinal = map[corev1.PodPhase]int{corev1.PodPending: 0, corev1.PodRunning: 2}
```
renormalize, line 42: Why was this modified? Could you explain since I do not understand.

operator/internal/podgangmigrator/gate.go
unmarshall: Fixed in 86ad79baaa1eabb727eee313a43ffcb06fc78723

operator/api/core/v1alpha1/podgangmap.go
unmarshall: This might be accurate now, but once coherent updates comes in then it changes since post coherent update there can be more an one MPG (Anchor based entries) and all pods of a PCLQ will not belong to a single anchor. So i would keep it as is.

operator/api/core/v1alpha1/podgangmap.go
unmarshall: did not get what is wrong with using `()` and why do you prefer `{}`?

operator/internal/controller/common/component/utils/podcliqueset.go
unmarshall: You typically use
```go
for _, v := range s {
  // when you only read v and do not update it
  // v is cheap to copy
}
```
Here `CurrentlyUpdating` is a small struct which is cheap to copy and we are not updating it. So making this change here is fine.

operator/internal/controller/podclique/components/pod/deletionsort.go
unmarshall: `PodUnknown` has been deprecated and is not being set since 2015. See https://github.com/kubernetes/kubernetes/blob/e81f39c0e03ce8ed8e2660c9147b391edd9e262b/staging/src/k8s.io/api/core/v1/types.go#L3733

There is no point keeping it as its just noise.

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
unmarshall: In the current PR there is no coherent update, so there can be at most 1 anchor, 1 tail and 1 scale-out entry ever.

When coherent update comes in then this will change since during and after coherent update, there can be more than 1 anchor, more than 1 tail and 1 scale-out entry. So in that case even taking the max will not really work and will be wrong.

This PR deliberately keeps simple and does not over complicate when there is no need to. Somethings will have to be changed in the subsequent PR.

operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
unmarshall: You are right that it cannot be empty if the PGM exists. In this case it should still return an error. Perhaps not a Requeue After because this is not expected. There is no point in reconstructing it since we do not know what emptied it in the first place. If that is a coding error that is doing that you end up recreating it and that code flow does the opposite of emptying it.

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
unmarshall: For the PodGang component to work correctly you need to have the Pods already with corrected labels.
In the previous approach we wanted to do eveything before a single reconciler ran, but that meant duplicating a LOT of code across components.  This was not something that was desirable.

operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
unmarshall: what is the change here, just formatting on to 2 lines?

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
shayasoolin: Note that this also occurs during normal multi-replica `RollingRecreate`, not only `OnDelete`: the PCS current hash advances globally, but only the selected replica’s PGM entries advance. An unselected PGM therefore has no current-hash Anchor. If its old `ScaleOut` is empty, it is replaced by a new-hash `ScaleOut` with `DependsOn:[""]`; selecting that replica later advances hashes without repairing the dependency.

operator/internal/controller/podcliqueset/components/podgang/podgang.go
```diff
@@ -211,13 +203,18 @@ func emptyPodGang(objKey client.ObjectKey) *groveschedulerv1alpha1.PodGang {
...
 }
 
-// getLabels constructs labels for a PodGang resource.
-func getLabels(pcsName string) map[string]string {
+func (r _resource) buildLabels(pcs *grovecorev1alpha1.PodCliqueSet, pgi *podGangInfo, schedulerName string) map[string]string {
+	pgLabels := map[string]string{
+		apicommon.LabelComponentKey:               apicommon.LabelComponentNamePodGang,
+		apicommon.LabelPodCliqueSetReplicaIndex:   strconv.Itoa(pgi.pcsReplicaIndex),
+		apicommon.LabelSchedulerName:              schedulerName,
+		apicommon.LabelPodCliqueSetGenerationHash: *pcs.Status.CurrentGenerationHash,
```
shayasoolin, line 211: This uses the PCS-wide current hash rather than the generation of the PGM entry being materialized. During RollingRecreate, an unselected replica can still have an old-generation entry, and coherent update will eventually have mixed generations in one PGM. Since the PodGang’s epoch and role come from the entry, could its generation label come from that entry too, with a test where the PCS and entry hashes differ?
Unless we do not support upgrading the operator while a PCS update takes place?

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
```diff
@@ -0,0 +1,336 @@
...
+		return epoch
+	}
+	return defaultEpoch
+}
+
+// buildBootstrapAnchorEntry returns the anchor entry carrying every standalone PodClique's full
+// Replicas count and every PodCliqueScalingGroup's MinAvailable replicas (PCSG indices
+// [0, MinAvailable)). DependsOn is nil.
+func buildBootstrapAnchorEntry(pcs *grovecorev1alpha1.PodCliqueSet, epoch string) grovecorev1alpha1.PodGangEntry {
+	entry := newPodGangEntry(epoch, *pcs.Status.CurrentGenerationHash, nil)
```
shayasoolin, line 99: During an upgrade with no existing PGM, this assigns the PCS-wide current hash to every replica. If a RollingRecreate is already in progress, unselected replicas are still on the old target generation; the existing-PGM path preserves that distinction, but bootstrap loses it and cannot restore it later.
Could bootstrap determine the effective generation per PCS replica? If migration during an active update is unsupported, let's document it. If it does, please also add a test that covers it.

operator/cmd/main.go
```diff
@@ -106,6 +107,13 @@ func main() {
 		handleErrorAndExit(err, cli.ExitErrSynchronizeTopology)
 	}
 
+	// Close the PodGang migration gate on legacy PodCliqueSets before the controllers start, so that
```
shayasoolin, line 110: Does Grove support operator upgrades with leader election disabled? If so, we may run into a case of old operator running in parallel to the new one, but ignoring the new condition below.

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
```diff
@@ -0,0 +1,230 @@
...
+func (r _resource) Sync(ctx context.Context, logger logr.Logger, pcs *grovecorev1alpha1.PodCliqueSet) error {
+	if !podgangmigrator.IsMigrationInProgress(pcs) {
+		return nil
+	}
+	for pcsReplicaIndex := range int(pcs.Spec.Replicas) {
+		if err := r.migrateReplica(ctx, pcs, pcsReplicaIndex); err != nil {
+			return err
+		}
+	}
+	return r.clearMigrationGate(ctx, logger, pcs)
```
shayasoolin, line 101: This clears the migration condition after relabelling, before the later PodGang component has successfully replaced the legacy PodGangs. If that phase fails, the PCLQ/PCSG controllers resume after their timed retry; an active rollout can then delete an old Pod while its replacement remains gated because the epoch PodGang does not exist. Could the handoff remain gated until the expected replacement PodGang objects have been created (not necessarily Initialized) and add a failure/retry test across the migrator and PodGang phases?

operator/internal/controller/podcliqueset/register.go
shayasoolin, line 68: `PodGangMap` and `PodGang` are desired resources managed by this controller, but neither is watched here. If either is deleted, the PCS will not be reconciled until some unrelated event occurs, so the recovery logic may never run in a quiet cluster. Could we watch/own these resources and add a controller-level test verifying that deletion triggers recreation?

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,123 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	// Items is the list of PodGangMap resources.
+	Items []PodGangMap `json:"items"`
+}
+
+// PodGangMapSpec defines the desired PodGang composition for a PodCliqueSet replica.
+type PodGangMapSpec struct {
+	// PodCliqueSetReplicaIndex is the index of the PodCliqueSet replica this map belongs to.
+	PodCliqueSetReplicaIndex int32 `json:"podCliqueSetReplicaIndex"`
+	// Entries is the ordered list of desired PodGangs for this PodCliqueSet replica.
```
shayasoolin: This is not always one entry per PodGang: an Anchor entry produces one PG, while Tail and ScaleOut entries can produce one PG per PCSG replica.

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,123 @@
...
+	// It holds every standalone PodClique and each PodCliqueScalingGroup's MinAvailable replicas.
+	// It materializes into a single PodGang.
+	PodGangEntryRoleAnchor PodGangEntryRole = "Anchor"
+	// PodGangEntryRoleTail marks a non-anchor entry that holds a PodCliqueScalingGroup's replica
+	// indices above MinAvailable, as declared by the template. It materializes into one PodGang per
+	// replica index.
+	PodGangEntryRoleTail PodGangEntryRole = "Tail"
+	// PodGangEntryRoleScaleOut marks the entry that holds PodCliqueScalingGroup replicas added by a
+	// steady-state scale-out beyond the template. It materializes into one PodGang per replica index.
+	// It is created on the first scale-out and removed once it holds no replicas.
```
shayasoolin: This does not match the current behavior. An empty ScaleOut entry is created during bootstrap whenever the PCS has a PCSG, and the current-generation entry is kept even when empty. Can we update the comment?

operator/api/core/v1alpha1/podgangmap.go
```diff
@@ -0,0 +1,123 @@
...
+	// must match. Used by PodClique and PodCliqueScalingGroup reconcilers to create pods at the
+	// correct spec version and to distinguish old pods from new pods during a coherent update.
+	PodCliqueSetGenerationHash string `json:"podCliqueSetGenerationHash"`
+	// Role classifies this entry as anchor, tail or scale-out.
+	// See PodGangEntryRole for the meaning of each value.
+	Role PodGangEntryRole `json:"role"`
+	// AnchorIndex is the index of an anchor entry within its generation hash. It is non-nil only on
+	// entries whose Role is Anchor, and nil otherwise. For an Anchor entry the index starts at 0 for
+	// each generation hash and increments for each additional anchor of the same hash. It forms the
+	// last segment of the anchor PodGang name.
```
shayasoolin: I do not see AnchorIndex used in the Anchor PodGang name; the current name ends with the epoch. Is this describing future behavior? If so, can we document the current behavior here?

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
```diff
@@ -346,6 +350,24 @@ func (r _resource) buildResource(logger logr.Logger, pcs *grovecorev1alpha1.PodC
...
 
+// resolvePodGangName returns the PodGang name for a PodCliqueScalingGroup replica by reading its
+// epoch from the PodGangMap. The replica belongs to the anchor PodGang when its index is below the
+// PodCliqueScalingGroup's MinAvailable, otherwise to a non-anchor (Tail or ScaleOut) PodGang.
+func resolvePodGangName(pgm *grovecorev1alpha1.PodGangMap, rnr apicommon.ResourceNameReplica, pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pcsgReplicaIndex int32) (string, error) {
+	pcsgConfigName, err := apicommon.ExtractScalingGroupNameFromPCSGFQN(pcsg.Name, rnr)
+	if err != nil {
+		return "", err
+	}
+	epoch, err := componentutils.EpochForPCSGReplica(pgm, pcsgConfigName, pcsgReplicaIndex)
```
shayasoolin: Can we use PodGangNameForPCSGReplica here instead of resolving the epoch and then inferring the role again from MinAvailable? The PGM entry already contains the role, and the shared helper uses it. Both match the current layout, but this local rule may diverge when coherent update reshuffles replica
indices or adds more Anchors. It would also avoid duplicating the resolver logic and its tests.

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
unmarshall: Please see the fix 4a19c5a743265c1b060deb825947cee0f99e0f1b
It fixes for both Rolling-Recreate and On-Delete update strategies.

operator/cmd/main.go
unmarshall: If you are running HA (more than one replica) and you have no leader election, this means you are running active-active setup. Grove does not support that. In fact no k8s controller i know of will usually support this. Its always active-passive - unless you have workload sharding implemented.

operator/internal/controller/podcliqueset/components/podgang/podgang.go
unmarshall: Please see 4a19c5a743265c1b060deb825947cee0f99e0f1b

operator/internal/controller/podcliqueset/components/podgangmap/steadystate.go
unmarshall: Change done in 4a19c5a743265c1b060deb825947cee0f99e0f1b converges all paths to one strategy - which is to update the PCS generation hash across all PGMs (one per PCS replica) together and not iteratively.

operator/api/core/v1alpha1/podgangmap.go
unmarshall: There was back and forth on this. Now i have updated the comment.

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
unmarshall: This is a good catch. I think it is sufficient to check that the new-epoch-based PodGangs are created for all PCS replicas. Once that is done then the gate (via condition) can be lifted.

unmarshall: Change made in 8bcfd5d247453f713ae9e436024183614aedc643

operator/internal/controller/podcliqueset/register.go
unmarshall: For `PodGang` resource: this was perhaps an existing gap. You are right that in an otherwise quite cluster if someone deletes a PodGang resource then in the absence of any signal the PodGang remains deleted. This perhaps has not created a problem for us so far is due to following reasons:

1. No one deletes operator managed resources like PodGang.
2. If a PodGang is deleted after associated Pods have been gang-scheduled then there is no impact on the running pods. Some unrelated event will eventually create the PodGang. But if the PodGang resource is deleted before Pods get scheduled, that is when it could create an issue. I am not sure if we have ever run into this case yet and perhaps there is another event that is getting enqueued into the PCS reconciler (which i cannot explain right now).

For `PodGangMap` - it perhaps become important that we either prevent deletion of PGM (authorizer webhook) or allow it to be reconstructed from existing PGs, live PCSG and PCLQs and PCS. 

So today we already have an `authorizer` webhook that protects managed resources. I have now added PGM to it. But this is disabled by default. I do not recollect now but there were some issues with it and it was kept disabled. Perhaps we need to revive this as this is a nice to have thing.

In addition 6677b968f9f94b5b5c9b723950ebf846d422c0b1 commits add what you requested for (delete-only event predicate) for PGM and PG to the PCS reconciler.

operator/api/core/v1alpha1/podgangmap.go
unmarshall: yes that is right. We went back and forth on the PodGangMapSpec umpteen number of times and this was left behind. Fixed in 6677b968f9f94b5b5c9b723950ebf846d422c0b1

operator/api/core/v1alpha1/podgangmap.go
unmarshall: Each anchor entry only carries a single anchor PG (MPG). So epoch disambiguates it. I think it was part of the name in an earlier design-iteration because we were using PCS generation hash and not the epoch as part of the anchor PG name. To disambiguate we needed AnchorIndex.  But that is no longer the case as far as anchor PG name is concerned.
It has been corrected: See 63e085e764d126ca21f240b32eae7f5d65b0861f

operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go
unmarshall: changed in 882a28158ccde4923adab1d22cf8dc5857047bf6

operator/api/core/v1alpha1/podgangmap.go
shayasoolin: I think my comment still holds, can you check the doc again?

unmarshall: Addressed in 65c9d8d0fe6aea08d0b36ee0f1698d5c3c8a2836

shayasoolin (dismissed):

renormalize (approved):
Thanks for the PR!

Reminder to clean up the commit message while merging 🙏.

shayasoolin (approved):

PR #783, opened 2026-08-18
[Fix #782]: key generation-hash expectation by PCS UID

### LINKED ISSUE #782: PodCliqueSet can get stuck after it is deleted and recreated with the same name
# PodCliqueSet can get stuck after it is deleted and recreated with the same name

## What happened?

I created a `PodCliqueSet`, updated it, deleted it, and then created a new `PodCliqueSet` with the same name.

The new PCS stayed stuck:

- no `PodCliqueScalingGroup` or `PodClique` was created;
- the PCS reported `PCSGs=0/100` and `PCLQs=0/0` every five seconds;
- `status.currentGenerationHash` stayed empty.

The old and new PCS had different UIDs, so the events for the old PCS were not caused by the new one.

The operator did not report an error. It kept retrying before the part that creates child resources.

## Workload used

```yaml
apiVersion: grove.io/v1alpha1
kind: PodCliqueSet
metadata:
  name: mypcs
spec:
  replicas: 100
  updateStrategy:
    type: RollingRecreate
  template:
    cliqueStartupType: CliqueStartupTypeExplicit
    terminationDelay: 4h0m0s
    cliques:
      - name: worker
        labels:
          # Bump this value to trigger RollingRecreate.
          version: "1"
        spec:
          roleName: worker
          replicas: 1
          minAvailable: 1
          podSpec:
            terminationGracePeriodSeconds: 0
            containers:
              - name: main
                image: busybox:1.36
                command:
                  - /bin/sh
                  - -c
                  - sleep $(( $(od -An -N1 -tu1 /dev/urandom | tr -d " ") % 5 ));
                    touch /tmp/ready; while true; do sleep 30; done
                startupProbe:
                  exec:
                    command: ["/bin/sh", "-c", "test -f /tmp/ready"]
                  initialDelaySeconds: 0
                  periodSeconds: 1
                  failureThreshold: 600
                # This small delay helps widen the timing window.
                readinessProbe:
                  exec:
                    command: ["/bin/sh", "-c", "usleep 10000; test -f /tmp/ready"]
                  initialDelaySeconds: 0
                  periodSeconds: 1
                  failureThreshold: 600
                resources:
                  requests:
                    cpu: 10m
                    memory: 1Mi
    podCliqueScalingGroups:
      - name: unit
        cliqueNames:
          - worker
        minAvailable: 1
        replicas: 1
```

## Suspected cause

The operator keeps an in-memory value for the PCS generation hash, using only namespace and name as the key.

If a PCS is deleted while a template update is still being processed, that value may remain in memory. A new PCS with the same name then looks like it is waiting for an old status update. The operator retries forever and never creates its children.

This is timing-sensitive. I have not yet reproduced it reliably with a short create/delete/recreate sequence.

## Expected behavior

A newly created PCS should create its PCSGs and PCLQs, even when a deleted PCS previously used the same namespace and name.

## Suggested fix

Clear the saved generation-hash value when a PCS is deleted, or include the PCS UID in its key.

## Environment

- **Grove operator:** `grove-operator:v0.1.0-alpha.12-rc1` (the exact build where this happened)
- Kubernetes: `v1.35.5`
- Scheduler: KAI `v0.14.0`

unmarshall:
#### What type of PR is this?

/kind bug

The in-memory generation-hash expectation was keyed by namespace/name and was never removed when a PodCliqueSet was deleted. A PodCliqueSet deleted while an expectation was still cached left a stale entry, and a new PodCliqueSet created with the same name (a different UID, an empty CurrentGenerationHash) matched that entry, so processGenerationHashChange requeued forever and never created its children.

Fix: In-memory expectation store for PCS generation now suffixes the PCS UID (`namespace/name/UID`)  so a recreated PodCliqueSet does not inherit an entry for the previously deployed PCS (with the same name + namespace). In addition in the deletion flow of the PCS, entry for the PCS is removed.

Fixes #782 

```release-note
Fix:  Stale in-memory generation-hash expectation that could block reconciliation of a same-named PodCliqueSet after delete-and-recreate
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podcliqueset/reconciledelete.go
+++ b/operator/internal/controller/podcliqueset/reconciledelete.go
@@ -52,6 +52,8 @@ func (r *Reconciler) triggerDeletionFlow(ctx context.Context, logger logr.Logger
 		}
 	}
 	logger.Info("PodCliqueSet finalizer removed; Kubernetes garbage collector will cascade-delete owned resources")
+	// Remove the in-memory generation-hash expectation so the map does not retain an entry for a deleted PodCliqueSet.
+	r.pcsGenerationHashExpectations.Delete(pcsGenerationHashKey(pcs))
 	return ctrlcommon.DoNotRequeue()
 }
 
--- a/operator/internal/controller/podcliqueset/reconcilespec.go
+++ b/operator/internal/controller/podcliqueset/reconcilespec.go
@@ -32,7 +32,6 @@ import (
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/tools/cache"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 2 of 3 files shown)


shayasoolin (dismissed):

shayasoolin:
LGTM, thanks for the quick fix!

renormalize (commented):
We only found this bug because the same `pcsObjectKey` was used for both the old and the new PCS in the expectation store as the key. If we were appending the UID from before, we would have never have found that the expectations were actually never getting cleared correctly, and the map would have just monotonically increased in size with time.

I would say that the changes in `reconcile_delete.go` are sufficient to fix this bug. If anything, keeping just the pcsObjectName as the key for the expectation store will help surface similar bugs in the future.

What are your thoughts on this?

unmarshall:
> We only found this bug because the same `pcsObjectKey` was used for both the old and the new PCS in the expectation store as the key. If we were appending the UID from before, we would have never have found that the expectations were actually never getting cleared correctly, and the map would have just monotonically increased in size with time.
> 
> I would say that the changes in `reconcile_delete.go` are sufficient to fix this bug. If anything, keeping just the pcsObjectName as the key for the expectation store will help surface similar bugs in the future.
> 
> What are your thoughts on this?

The semantically correct key to use in an expectation store should ensure that it always points to a specific resource for which it is intended. So just `namespace/name` is not going to cut it. This ensures that we never re-use the same key across re-deployment of the same resource (in this case PCS). On top of that we have the clean-out logic that is part of the deletion flow.

shayasoolin:
> We only found this bug because the same `pcsObjectKey` was used for both the old and the new PCS in the expectation store as the key. If we were appending the UID from before, we would have never have found that the expectations were actually never getting cleared correctly, and the map would have just monotonically increased in size with time.
> 
> I would say that the changes in `reconcile_delete.go` are sufficient to fix this bug. If anything, keeping just the pcsObjectName as the key for the expectation store will help surface similar bugs in the future.
> 
> What are your thoughts on this?

I hear you. Just thinking that relying on finding such a subtle bug just to understand we're not clearing the memory correctly, sounds risky to me. In other words, I prefer to have robust code that is resilient to the more critical bug of stuck PCS, and manage the risk of memory increase in some other way. For example, storing both PCS name and UID, and when deleting an entry by UID, make sure to cleanup any other entries of the same PCS name.

renormalize (dismissed):

renormalize (approved):

shayasoolin (approved):

PR #784, opened 2026-08-20
fix #775: Prevent stuck MinAvailableBreached on PCLQ resources

### LINKED ISSUE #775: PodCliques MinAvailableBreached=True is not reconciled
### What happened?

For PodCliques with replicas=minUnavailable we regularly observe the situation, when after the deploy PodClique is stuck in MinAvailableBreached=True, even though all pods are in the ready state.
This keeps termination timer armed which eventually leads to clique restart.

Here is the sketch of `PodCliqueSet` we are using.

During re-deploy 1-2 `PodCliques` are stuck in MinAvailableBreached=True.

```
kind: PodCliqueSet
metadata:
  labels:
    ...
  name: ...
spec:
  replicas: 10
  template:
    cliqueStartupType: CliqueStartupTypeExplicit
    cliques:
      - annotations:
          ...
        labels:
          ...
        name: decode
        spec:
          minAvailable: 1
          podSpec:
            ...
          replicas: 1
          roleName: decode
      - annotations:
          ...
        labels:
          ...
        name: prefill
        spec:
          minAvailable: 2
          podSpec:
            ....
          replicas: 2
          roleName: prefill
    podCliqueScalingGroups:
      - cliqueNames:
          - decode
          - prefill
        minAvailable: 1
        name: unit
        replicas: 1
  updateStrategy:
    type: RollingRecreate

```

I expect MinAvailableBreached reconciled to False once all pods are ready.

### Environment

- Kubernetes version: v1.35.2+rke2r
- Grove version: grove-operator:v0.1.0-alpha.12-rc1
- Scheduler details: KAI v0.14.0

unmarshall:
#### What type of PR is this?

/kind bug

Issue description: https://github.com/ai-dynamo/grove/issues/775#issuecomment-5344927136
In this PR we implemented `change-2` and `change-3` as describe at https://github.com/ai-dynamo/grove/issues/775#issuecomment-5351694208

Fixes #775 

```release-note
Fixed a bug where a PodClique could remain stuck with MinAvailableBreached=True after its pods became Ready, causing unnecessary gang termination. Status updates now use optimistic locking so a write based on a stale cache is retried against current state instead of silently persisting or being skipped. (#775)
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podclique/reconciler.go
+++ b/operator/internal/controller/podclique/reconciler.go
@@ -20,6 +20,7 @@ import (
 	"github.com/ai-dynamo/grove/operator/api/common/constants"
 	configv1alpha1 "github.com/ai-dynamo/grove/operator/api/config/v1alpha1"
 	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
+	internalconstants "github.com/ai-dynamo/grove/operator/internal/constants"
 	ctrlcommon "github.com/ai-dynamo/grove/operator/internal/controller/common"
 	"github.com/ai-dynamo/grove/operator/internal/controller/common/component"
 	componentutils "github.com/ai-dynamo/grove/operator/internal/controller/common/component/utils"
@@ -82,9 +83,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	reconcileSpecFlowResult := r.reconcileSpec(ctx, logger, pclq)
-	if statusReconcileResult := r.reconcileStatus(ctx, logger, pclq); ctrlcommon.ShortCircuitReconcileFlow(statusReconcileResult) {
-		return statusReconcileResult.Result()
-	}
+	statusReconcileResult := r.reconcileStatus(ctx, logger, pclq)
+	reconcileResult := ctrlcommon.MergeStepResults(reconcileSpecFlowResult, statusReconcileResult)
 
-	return reconcileSpecFlowResult.Result()
+	result, err := reconcileResult.Result()
+	// Neither flow requested a requeue, so schedule the periodic status resync as a backstop that
+	// recomputes a status left stale by a lost update. This is applied here rather than in
+	// reconcileStatus so it cannot mask a shorter requeue from the spec flow, which drives
+	// rolling-update progression. A longer resync overriding that shorter requeue would delay the
+	// update by a full interval.
+	// See https://github.com/ai-dynamo/grove/issues/775.
+	if err == nil && result.RequeueAfter == 0 {
+		result.RequeueAfter = internalconstants.PodCliqueStatusResyncInterval
+	}
+	return result, err
 }
--- a/operator/internal/controller/podclique/reconciler.go
+++ b/operator/internal/controller/podclique/reconciler.go
@@ -16,11 +16,12 @@ package podclique
 
 import (
 	"context"
+	"time"
 
-	"github.com/ai-dynamo/grove/operator/api/common/constants"
+	apicommonconstants "github.com/ai-dynamo/grove/operator/api/common/constants"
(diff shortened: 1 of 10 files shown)


Ronkahn21:
Change 2 looks good to me, but I think we should discuss Change 1 separately and understand its overall performance impact before moving forward. Removing the guard makes every PodClique reconcile send a status PATCH, which could significantly affect controller and API-server load at scale. I’d like us to validate this with scale data first, as the overall impact could be substantial.

I added more details in [this issue comment](https://github.com/ai-dynamo/grove/issues/775#issuecomment-5353885574).

unmarshall:
>Removing the guard makes every PodClique reconcile send a status PATCH even when nothing changed. This cost is multiplied across every PodClique created under each PodCliqueSet and PodCliqueScalingGroup, potentially creating substantial controller and API-server load at scale.

It is true that this will send calls to the API server which will just drop the requests after checking it against the stored state in etcd. So yes there is cost. But there is no fear of change in `ResourceVersion` as is currently mentioned in the comment above this change.

>My suggestion is to keep the equality guard and add status-aware predicates at all three levels—PCS, PCSG, and PCLQ. Each controller could enqueue on status changes relevant to it, using targeted condition transitions where possible, such as MinAvailableBreached, or broader status changes if necessary.

I generally do not favor writing event predicates based on specific conditions. Over a period of time this set increases and your event predicate becomes complex. I generally categorise this into a hack category rather than a cleaner fix.

My simple proposal then would be to then have optimistic locking + periodic sync for status updates. That will resolve the issue completely.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -83,20 +84,17 @@ func (r *Reconciler) reconcileStatus(ctx context.Context, logger logr.Logger, pc
...
+		if err = r.client.Status().Patch(ctx, pcsg, patchObj); err != nil {
+			if apierrors.IsConflict(err) {
+				return ctrlcommon.ReconcileAfter(internalconstants.ComponentSyncRetryInterval, fmt.Sprintf("409-conflict when updating PodCliqueScalingGroup status, re-queueing: %v", pcsgObjectKey))
+			}
+			logger.Error(err, "failed to update PodCliqueScalingGroup status")
+			return ctrlcommon.ReconcileWithErrors("failed to update the status with label selector and replicas", err)
+		}
 	}
 
 	return ctrlcommon.ContinueReconcile()
```
Ronkahn21, line 100: should we requeaue after as well

unmarshall: I think it is not required here. We have requeue at the PCLQ level. Any status change in PCLQ will enqueue an event that will trigger a PCSG reconcile.

Ronkahn21 (dismissed):

operator/internal/controller/podclique/reconciler.go
```diff
@@ -82,9 +83,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	return reconcileSpecFlowResult.Result()
+	result, err := reconcileResult.Result()
+	// Neither flow requested a requeue, so schedule the periodic status resync as a backstop that
+	// recomputes a status left stale by a lost update. This is applied here rather than in
+	// reconcileStatus so it cannot mask a shorter requeue from the spec flow, which drives
+	// rolling-update progression. A longer resync overriding that shorter requeue would delay the
+	// update by a full interval.
+	// See https://github.com/ai-dynamo/grove/issues/775.
+	if err == nil && result.RequeueAfter == 0 {
+		result.RequeueAfter = internalconstants.PodCliqueStatusResyncInterval
```
sanjaychatterjee: Maybe make it to be the minimum if `PodCliqueStatusResyncInterval` and `TerminationDelay` set for the PCS.

sanjaychatterjee (approved):

shayasoolin (approved):

Issue #315, opened 2026-01-14
Rolling Update Deadlock in PodClique Controller


gflarity:
## Summary

Rolling updates can get permanently stuck due to a race condition between when `UpdateEndedAt` is set and when pods are counted. The test `Test_RU9_RollingUpdateAllPodCliques` fails intermittently with a 4-minute timeout waiting for the rolling update to complete.

## What Should Happen

1. **Update in progress:** Old pods are deleted, new pods are created.
   - `CurrentPodTemplateHash` = OLD hash
   - `RollingUpdateProgress.PodTemplateHash` = NEW hash
   - `UpdateEndedAt` = nil → `mutateUpdatedReplica()` counts against NEW hash ✓
   - `UpdatedReplicas` increases as new pods become Ready

2. **All new pods become Ready:**
   - `UpdatedReplicas` = 2, `Replicas` = 2
   - All old pods have been deleted

3. **`markRollingUpdateEnd()` sets `UpdateEndedAt`:**
   - `UpdateEndedAt` is now set → `IsPCLQUpdateInProgress()` returns `false`
   - At this moment: `UpdatedReplicas` = 2, `Replicas` = 2 ✓

4. **Next status reconciliation — hash switch happens safely:**
   
   **`mutateCurrentHashes()` runs first:**
   - `IsPCLQUpdateInProgress()`? → NO (because `UpdateEndedAt` is set)
   - `UpdatedReplicas == Replicas`? → YES (2 == 2) ✓
   - **Updates** `CurrentPodTemplateHash` to the NEW hash
   
   **`mutateUpdatedReplica()` runs next:**
   - `IsPCLQUpdateInProgress()`? → NO (because `UpdateEndedAt` is set)
   - Uses `CurrentPodTemplateHash` (now the **NEW** hash!)
   - Counts pods matching NEW hash → **2 pods match** ✓
   - `UpdatedReplicas` = 2

5. **Update completes successfully.**

## What Happened (The Bug)

1. **Update in progress:** Old pods are deleted, new pods are created.
   - `CurrentPodTemplateHash` = OLD hash
   - `RollingUpdateProgress.PodTemplateHash` = NEW hash
   - `UpdateEndedAt` = nil → `mutateUpdatedReplica()` counts against NEW hash ✓
   - `UpdatedReplicas` = 1 (one new pod is Ready)

2. **Last old pod deleted, but one new pod is still becoming Ready:**
   - `markRollingUpdateEnd()` is called (no old pods remain)
[...]

PR #785, opened 2026-08-20
fix: wait for updated pods before ending rollouts

### LINKED ISSUE #786: RollingRecreate can complete before all updated pods are Ready
### What happened?

During a `RollingRecreate` update, Grove eagerly deletes old-template pods that are Pending, unhealthy, still starting, or otherwise not Ready.

Those old non-ready pods do not pass through the ready-pod selection flow used for healthy old replicas. After deleting them, `processPendingUpdates()` can find no remaining old-pod work and call `markRollingUpdateEnd()` before all desired replacement pods with the target template hash are Ready.

A minimal example is:

```yaml
spec:
  replicas: 2
  minAvailable: 1
```

After a template update:

```text
target-template pod 1: Ready
target-template pod 2: crash-looping
```

Grove can set `status.updateProgress.updateEndedAt` even though only one of the two desired target-template pods is Ready.

This makes the rollout appear complete while the new generation is still starting or failing. Higher-level controllers can observe an ended update with incomplete readiness, and Grove no longer considers the PodClique to have an active automatic rollout.

Related history:

- #315 described both this premature completion and a subsequent `UpdatedReplicas`/current-hash deadlock.
- #671 fixed the old hash-counting deadlock by calculating target-hash `UpdatedReplicas` before advancing current hashes.
- The premature `UpdateEndedAt` behavior remains on current `main`.

Reproduction at the controller level:

1. Create a PodClique using `RollingRecreate` with `replicas: 2` and `minAvailable: 1`.
2. Trigger a pod-template update.
3. Allow old non-ready pods to be deleted eagerly.
4. Make one target-template replacement Ready and leave the other target-template replacement not Ready or crash-looping.
5. Observe that `processPendingUpdates()` can set `UpdateEndedAt` before both desired replacements are Ready.

A proposed fix and regression test are available in #785.

`RollingRecreate` should keep `status.updateProgress.updateEndedAt` unset until every desired pod with the target template hash is Ready:

```text
ready target-template pods >= spec.replicas
```

For the example above:

```text
1 Ready, 2 desired → update remains in progress
2 Ready, 2 desired → update may end
```

`minAvailable` should remain the availability floor used to determine whether Grove may disrupt the next healthy old pod. It should not define rollout completion when fewer than all desired replicas are Ready.

### Environment

- Kubernetes version: Controller-logic issue; reproduced with the controller-runtime fake client
- Grove version: `main` at `fcc3b3bbbc5e6a2a797cd080bdbc6983b1ccec24`
- Scheduler details: Scheduler-independent
- Cloud provider or hardware configuration: Provider-independent
- Tools that you are using Grove together with: Go unit tests
- Related pull request: #785
- Related historical issues/changes: #315 and #671

alexeldeib:
#### What type of PR is this?

/kind bug

`RollingRecreate` eagerly deletes old-template pods that are Pending, unhealthy, still starting, or otherwise not Ready. These pods bypass the normal ready-pod selection flow used for healthy old replicas.

After deleting those non-ready pods, `processPendingUpdates()` can currently find no remaining old-pod work and set `UpdateEndedAt` before all replacement pods with the target template hash are Ready. This makes the rollout appear complete while the new generation may still be starting or crash-looping.

This PR keeps `UpdateProgress` open until every desired target-template pod is Ready:

```text
ready target-template pods >= spec.replicas
```

The completion threshold intentionally uses `spec.replicas`, not `minAvailable`. `minAvailable` is the availability floor used to decide whether Grove may disrupt the next healthy old pod; it does not mean that a rollout with only a subset of its desired replicas Ready has completed.

This addresses the remaining premature-completion behavior described in #315.

The original `UpdatedReplicas`/current-hash self-lock from #315 has since been addressed by #671, which calculates target-hash `UpdatedReplicas` before advancing the current hashes.

Regression coverage includes a two-replica PodClique with `minAvailable: 1`, where one target-template pod is Ready and the other is crash-looping. The update remains open until both desired replicas are Ready.

closes #786

The relevant failure sequence is:

```text
old non-ready pods are deleted eagerly
    → they do not enter ready-pod selection tracking
    → no old-pod update work remains
    → UpdateEndedAt is set before replacements are Ready
```

The new behavior is:

```text
old non-ready pods are deleted eagerly
    → wait for all desired target-template pods to become Ready
    → set UpdateEndedAt
```

This is intentionally stricter than the earlier proposed `newReady >= minAvailable` check. For example, with `replicas: 2` and `minAvailable: 1`, one Ready pod plus one crash-looping pod must not be considered a completed rollout.

Tests run:

```text
go test ./internal/controller/podclique/components/pod
go test ./internal/controller/podclique/...
```

Both passed.

AI assistance was used to investigate the production failure, review the relevant Grove history, and prepare the initial patch and tests. The submitter reviewed the resulting behavior and every changed line.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podclique/components/pod/rollingupdate.go
+++ b/operator/internal/controller/podclique/components/pod/rollingupdate.go
@@ -129,7 +129,22 @@ func (r _resource) processPendingUpdates(logger logr.Logger, sc *syncContext) er
 		)
 	}
 
-	// If the control comes here, then mark the end of update.
+	// Old non-ready pods are deleted eagerly and do not go through the
+	// ready-pod selection flow above. Wait for every desired replacement to
+	// become Ready before closing the update, otherwise UpdateEndedAt can be
+	// set while the new generation is still starting or crash-looping.
+	if int32(len(updateWork.newTemplateHashReadyPods)) < pclq.Spec.Replicas {
+		return groveerr.New(
+			groveerr.ErrCodeContinueReconcileAndRequeue,
+			component.OperationSync,
+			fmt.Sprintf(
+				"waiting for updated pods to become ready before marking update end: %d ready, need %d",
+				len(updateWork.newTemplateHashReadyPods),
+				pclq.Spec.Replicas,
+			),
+		)
+	}
+
 	return r.markRollingUpdateEnd(sc.ctx, logger, pclq)
 }
 
--- a/operator/internal/controller/podclique/components/pod/rollingupdate_test.go
+++ b/operator/internal/controller/podclique/components/pod/rollingupdate_test.go
@@ -17,16 +17,21 @@ limitations under the License.
 package pod
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
 	"github.com/ai-dynamo/grove/operator/api/common"
+	grovecorev1alpha1 "github.com/ai-dynamo/grove/operator/api/core/v1alpha1"
 	"github.com/ai-dynamo/grove/operator/internal/expect"
+	testutils "github.com/ai-dynamo/grove/operator/test/utils"
 
 	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 )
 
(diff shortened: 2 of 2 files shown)


operator/internal/controller/podclique/components/pod/rollingupdate.go
```diff
@@ -129,7 +129,22 @@ func (r _resource) processPendingUpdates(logger logr.Logger, sc *syncContext) er
 		)
 	}
 
-	// If the control comes here, then mark the end of update.
+	// Old non-ready pods are deleted eagerly and do not go through the
+	// ready-pod selection flow above. Wait for every desired replacement to
+	// become Ready before closing the update, otherwise UpdateEndedAt can be
+	// set while the new generation is still starting or crash-looping.
+	if int32(len(updateWork.newTemplateHashReadyPods)) < pclq.Spec.Replicas {
```
oleg-kushniriov, line 136: With this gate, a permanently crash-looping new generation keeps UpdateProgress open indefinitely - and while an update is in progress, `MinAvailableBreached` is forced to Unknown (podclique/reconcilestatus.go:219), so PCSG per-replica recycling and PCS gang termination stay suppressed the whole time. Pre-PR the premature completion (wrongly) re-armed them. I think indefinite-open is the correct terminal state - recycling won't fix a bad image, and a corrected template resets the update via the new generation hash - but there's no progress-deadline concept, so this should be an explicit decision. 

Consider emitting a Warning event (or a RolloutStalled-style signal) when this branch persists, since users otherwise just see Unknown conditions and an update that never ends.

oleg-kushniriov, line 136: newTemplateHashReadyPods is built in computeUpdateWork with only an `IsPodReady` check - unlike the old-hash branch, it doesn't skip terminating pods. A new-hash pod that's Ready but already terminating (e.g. manually deleted, or racing deleteExcessPods) satisfies this gate and can close the update a step early. 

Suggest filtering IsResourceTerminating/delete-expectation pods out of newTemplateHashReadyPods - that would also tighten the pre-existing count check in isCurrentPodUpdateComplete.

oleg-kushniriov (commented):
1. Rebase

2. Is indefinite-open-update the intended terminal state for a never-ready new generation, given it holds MinAvailableBreached=Unknown and suppresses all breach-driven remediation? Should there be an observable stall signal?
3. Should terminating pods be excluded from the completion count?

operator/internal/controller/podclique/components/pod/rollingupdate.go
unmarshall: If you look at upstream k8s `Deployment` then there is `MaxUnavailable` which makes the update safer. So in both Deployment and STS the updates are indefinitely stuck. However, STS and Deployment do not have a side effect of disabled gang termination which PCS has currently. So we will have to carefully consider being in the update-blocked state indefinitely.

On second thoughts, it's fine for the update to get stuck if new pods are not coming up and becoming ready. But it needs the protection of `MaxUnavailable` so that we preserve availability which is not the case today.

unmarshall: Deployment has `ProgressDeadlineSeconds` which has a default value of 600s. If that exceeds then a condition (`Progressing`) is set to false with a reason `ProgressDeadlineExceeded` and `kubectl rollout status deploy/<deployment-name>` will exit with a non-zero status and it prints `exceeded its progress deadline`.

PR #790, opened 2026-08-22
fix: use durable health history for gang termination

### LINKED ISSUE #789: Gang termination can misclassify delayed initial failures as regressions
### What happened?

The historical-health heuristic added in #614 uses:

```text
LastTransitionTime > CreationTimestamp + 5s
```

This is a reasonable approximation for continuously active workloads, where the initial status is usually reconciled shortly after creation. However, it assumes that a later condition transition implies a previously observed healthy state.

Scale-from-zero makes this timing assumption easier to violate. An object may remain idle beyond five seconds and then fail its first activation. Its initial unhealthy condition can then be interpreted as a regression even though no healthy state was ever observed. Zero replicas are not the root cause; a delayed initial reconciliation can expose the same issue independently.

Because gang termination deletes and recreates resources, the impact can be significant. A local fault-injection reproduction observed:

- 6 PCS replica deletions within 88 seconds.
- 8 healthy `frontend` Pods terminated.
- Recreated `decode` Pods remained Pending.
- A repeated delete -> recreate -> fail cycle.

At PCS level, this can turn one component waiting for capacity into an outage of the entire PCS replica.

### Related context

- [GREP-0677: Zero-replica gang membership](https://github.com/ai-dynamo/grove/pull/686), which proposes suppressing gang termination while a component is activating.
- [Scale-to-zero design discussion](https://github.com/ai-dynamo/grove/issues/677).
- [Original scale-to-zero report](https://github.com/ai-dynamo/grove/issues/676), where zero-scaled workers affected a healthy frontend.
- [PR #614](https://github.com/ai-dynamo/grove/pull/614), which added the current heuristic and includes [discussion of repeated recreation for an initially unschedulable workload](https://github.com/ai-dynamo/grove/pull/614#issuecomment-4476216658).

It would be safer to record historical health only after Grove observes and persists a positive state. Gang termination could remain suppressed during initial activation.

### Environment

- Grove: `fcc3b3bbbc5e6a2a797cd080bdbc6983b1ccec24`
- Kubernetes: Kind v1.35.5
- Scheduler: default scheduler

yankay:
#### What type of PR is this?

/kind bug

Replaces timestamp-based health inference with a durable
`HealthyStateObserved=True` condition. It is recorded when a PCLQ is scheduled
or a PCSG reaches `AvailableReplicas >= MinAvailable`.

This avoids treating a delayed first unhealthy reconcile as a regression.
Existing gang-termination retry behavior remains unchanged; #798 tracks that
separately.

Fixes #789

`LastScheduled` remains the upgrade fallback for legacy PCLQs. Already-breached
PCSGs and PCLQs without durable history are not backfilled. Tests cover upgrade
migration and transient non-breached status without actual availability.

```release-note
action required: Legacy PCSGs already below MinAvailable during upgrade, and
PCLQs without LastScheduled, remain excluded from automatic gang termination
until they become healthy. PCLQs with LastScheduled remain eligible.
```

```docs
GREP-0677 (#686)
```

### DIFF
--- a/operator/api/common/constants/constants.go
+++ b/operator/api/common/constants/constants.go
@@ -118,18 +118,19 @@ const (
 	// ConditionTypePodCliqueScheduled indicates that the PodClique has been successfully scheduled.
 	// This condition is set to true when number of scheduled pods in the PodClique is greater than or equal to PodCliqueSpec.MinAvailable.
 	ConditionTypePodCliqueScheduled = "PodCliqueScheduled"
-	// ConditionTypeHealthyStateObserved is a monotonic signal set after a component reaches a
-	// genuinely scheduled or available state. Gang termination uses it to distinguish a regression
-	// from initial startup.
+	// ConditionTypeHealthyStateObserved records lifecycle history for gang termination,
+	// not current health. Once True, it is never cleared and persists across generations.
+	//
+	// Semantics:
+	//   - PodClique: PodCliqueScheduled was True at least once; readiness is not required.
+	//   - PodCliqueScalingGroup: AvailableReplicas reached MinAvailable at least once,
+	//     based on the readiness of its constituent PodCliques.
 	ConditionTypeHealthyStateObserved = "HealthyStateObserved"
 	// ConditionTypeGangTerminationInProgress indicates that PCS-level gang termination has fired for this
 	// PodCliqueScalingGroup and is still in flight. It is set on the PCSG when the PCS-level handler deletes
-	// the PodCliques of the whole PCS replica, and is cleared when the PCSG's MinAvailableBreached transitions
-	// back to False (recovered). While it is True, further PCS-level gang termination for the PCS replica is
-	// suppressed — at most one fire per breach episode, regardless of how long the workload stays below
-	// MinAvailable. This is a PCS-level-only mechanism: the PCSG-replica-scoped recycle path does not use this
-	// flag and instead breaks its own re-fire loop via WasPCLQEverScheduled, since a freshly recreated
-	// PodClique has never been scheduled and is therefore excluded from the breached set.
+	// the PodCliques of the whole PCS replica, and is cleared only when AvailableReplicas >= MinAvailable.
+	// While True, it suppresses repeated PCS-level firing within the same breach episode. The PCSG-replica
+	// recycle path instead uses WasPCLQEverScheduled, since a recreated PodClique has no scheduling history.
 	// Its only Reason is ConditionReasonGangTerminationActive.
 	ConditionTypeGangTerminationInProgress = "GangTerminationInProgress"
 	// ConditionTopologyLevelsUnavailable indicates that the required topology levels defined on a PodCliqueSet for topology-aware scheduling are no longer available.
@@ -159,8 +160,7 @@ const (
 	ConditionReasonInsufficientAvailablePCSGReplicas = "InsufficientAvailablePodCliqueScalingGroupReplicas"
 	// ConditionReasonSufficientAvailablePCSGReplicas indicates that the number of ready replicas in the PodCliqueScalingGroup is greater than or equal to the PodCliqueScalingGroupSpec.MinAvailable.
 	ConditionReasonSufficientAvailablePCSGReplicas = "SufficientAvailablePodCliqueScalingGroupReplicas"
-	// ConditionReasonHealthyStateObserved is used when a component first reaches a genuinely
-	// scheduled or available state.
+	// ConditionReasonHealthyStateObserved indicates the first observed scheduled or available state.
 	ConditionReasonHealthyStateObserved = "HealthyStateObserved"
 	// ConditionReasonUpdateInProgress indicates that the resource is undergoing rolling update.
 	ConditionReasonUpdateInProgress = "UpdateInProgress"
--- a/operator/api/common/constants/constants.go
+++ b/operator/api/common/constants/constants.go
@@ -118,6 +118,10 @@ const (
 	// ConditionTypePodCliqueScheduled indicates that the PodClique has been successfully scheduled.
 	// This condition is set to true when number of scheduled pods in the PodClique is greater than or equal to PodCliqueSpec.MinAvailable.
 	ConditionTypePodCliqueScheduled = "PodCliqueScheduled"
+	// ConditionTypeHealthyStateObserved is a monotonic signal set after a component reaches a
+	// genuinely scheduled or available state. Gang termination uses it to distinguish a regression
+	// from initial startup.
+	ConditionTypeHealthyStateObserved = "HealthyStateObserved"
 	// ConditionTypeGangTerminationInProgress indicates that PCS-level gang termination has fired for this
 	// PodCliqueScalingGroup and is still in flight. It is set on the PCSG when the PCS-level handler deletes
 	// the PodCliques of the whole PCS replica, and is cleared when the PCSG's MinAvailableBreached transitions
@@ -155,6 +159,9 @@ const (
 	ConditionReasonInsufficientAvailablePCSGReplicas = "InsufficientAvailablePodCliqueScalingGroupReplicas"
 	// ConditionReasonSufficientAvailablePCSGReplicas indicates that the number of ready replicas in the PodCliqueScalingGroup is greater than or equal to the PodCliqueScalingGroupSpec.MinAvailable.
 	ConditionReasonSufficientAvailablePCSGReplicas = "SufficientAvailablePodCliqueScalingGroupReplicas"
+	// ConditionReasonHealthyStateObserved is used when a component first reaches a genuinely
+	// scheduled or available state.
+	ConditionReasonHealthyStateObserved = "HealthyStateObserved"
 	// ConditionReasonUpdateInProgress indicates that the resource is undergoing rolling update.
 	ConditionReasonUpdateInProgress = "UpdateInProgress"
 	// ConditionReasonGangTerminationActive is the (only) Reason paired with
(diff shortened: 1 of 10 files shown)


operator/internal/controller/common/component/utils/podclique.go
```diff
@@ -84,52 +84,31 @@ func GroupPCLQsByPCSReplicaIndex(pclqs []grovecorev1alpha1.PodClique) map[string
...
-	if cond == nil {
-		return false
-	}
-	if cond.Status == metav1.ConditionFalse {
-		return true
+	return meta.IsStatusConditionTrue(pcsg.Status.Conditions, constants.ConditionTypeHealthyStateObserved)
+}
+
+// MarkHealthyStateObserved records genuine health once and never clears it.
+func MarkHealthyStateObserved(conditions *[]metav1.Condition, observedGeneration int64, message string) {
```
oleg-kushniriov: Objects already regressed at upgrade time (healthy in the past, breached now) never get this condition and are permanently excluded from gang termination - which is exactly their recovery path. 

Consider a one-shot backfill using the old heuristic on first reconcile, or at minimum a release-note warning plus a documented manual remediation.

operator/internal/controller/common/component/utils/podclique.go
```diff
@@ -84,52 +84,31 @@ func GroupPCLQsByPCSReplicaIndex(pclqs []grovecorev1alpha1.PodClique) map[string
...
+	if meta.IsStatusConditionTrue(*conditions, constants.ConditionTypeHealthyStateObserved) {
+		return
 	}
-	return cond.LastTransitionTime.After(pcsg.CreationTimestamp.Add(InitialScheduleGrace))
+	meta.SetStatusCondition(conditions, metav1.Condition{
+		Type:               constants.ConditionTypeHealthyStateObserved,
+		Status:             metav1.ConditionTrue,
+		Reason:             constants.ConditionReasonHealthyStateObserved,
+		Message:            message,
+		ObservedGeneration: observedGeneration,
```
oleg-kushniriov: Because of the early return above, this stays at the first-observed generation forever and nothing reads it. Either drop it, or document that it's intentionally "generation at first observation" - otherwise it reads as a stale-condition bug. 

Related: should health be re-proven per generation after a rolling update to a never-schedulable spec?

operator/api/common/constants/constants.go
```diff
@@ -118,6 +118,10 @@ const (
 	// ConditionTypePodCliqueScheduled indicates that the PodClique has been successfully scheduled.
 	// This condition is set to true when number of scheduled pods in the PodClique is greater than or equal to PodCliqueSpec.MinAvailable.
 	ConditionTypePodCliqueScheduled = "PodCliqueScheduled"
+	// ConditionTypeHealthyStateObserved is a monotonic signal set after a component reaches a
```
oleg-kushniriov: On a `PCLQ` this means "was scheduled once" (no readiness required) 
On a `PCSG` it means "was genuinely available once" (ReadyReplicas-based). 

Both mirror their old gates, but the shared name blurs it - worth spelling out both semantics here.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -233,6 +233,19 @@ func mutateMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1al
 			"reason", newCondition.Reason)
 		meta.SetStatusCondition(&pcsg.Status.Conditions, newCondition)
 	}
+	minAvailable := int32(1)
```
oleg-kushniriov: This re-implements the nil-default-to-1 logic from `computeMinAvailableBreachedCondition` (lines ~268–273), minus its legacy-CRD log line. 

Extract a small helper so the two can't drift.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -233,6 +233,19 @@ func mutateMinAvailableBreachedCondition(logger logr.Logger, pcsg *grovecorev1al
...
+	if pcsg.Spec.MinAvailable != nil {
+		minAvailable = *pcsg.Spec.MinAvailable
+	}
+	if pcsg.Spec.Replicas > 0 &&
+		newCondition.Status == metav1.ConditionFalse &&
+		pcsg.Status.AvailableReplicas >= minAvailable {
```
oleg-kushniriov: Nice - this closes the transient-False poisoning window (children exist but haven't reported status). 
Note the fix is scoped to never-healthy objects: a `PCSG` that was genuinely healthy keeps the durable flag, and the transient False during recreation still clears `GangTerminationInProgress` (line 250), so a permanently-unschedulable regression still recycles every TerminationDelay. Worth stating in the PR description that this is intentional.

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -276,6 +276,13 @@ func mutatePodCliqueScheduledCondition(pclq *grovecorev1alpha1.PodClique) {
 	if k8sutils.HasConditionChanged(pclq.Status.Conditions, newCondition) {
 		meta.SetStatusCondition(&pclq.Status.Conditions, newCondition)
 	}
+	if pclq.Spec.Replicas > 0 && newCondition.Status == metav1.ConditionTrue {
```
oleg-kushniriov, line 281: With the webhook enforcing `MinAvailable >= 1`, `PodCliqueScheduled=True `already implies ≥1 scheduled pod, so Replicas > 0 is defense-in-depth against a vacuous True (e.g. scale-subresource paths). 

A one-line comment saying so would keep someone from "simplifying" it away.

operator/internal/controller/common/component/utils/podclique.go
yankay: Good point. I dropped `ObservedGeneration` and documented that this is a lifetime history signal that persists across generations. Health is intentionally not re-proven after rolling updates, since a previously healthy workload should remain eligible for gang termination if the new generation regresses.

operator/api/common/constants/constants.go
yankay: Thank you, agreed. I updated the API comment to spell out both semantics: a PCLQ was scheduled at least once, while a PCSG genuinely reached `AvailableReplicas >= MinAvailable` based on readiness.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
yankay: Thank you, good catch. I extracted `effectiveMinAvailable` for both call sites and kept the legacy-CRD log where it was.

operator/internal/controller/podclique/reconcilestatus.go
yankay: Thank you, agreed. I added a comment explaining that `Replicas > 0` is defense-in-depth for scale-subresource edge paths.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
yankay: Hi @oleg-kushniriov,

Thanks for catching this. The current revision implements **B** by clearing [`[GangTerminationInProgress](https://github.com/ai-dynamo/grove/blob/3538aec4a4af3ada9e518a557cdf93895dc63cac/operator/internal/controller/podcliquescalinggroup/reconcilestatus.go#L234-L248)`](https://github.com/ai-dynamo/grove/blob/3538aec4a4af3ada9e518a557cdf93895dc63cac/operator/internal/controller/podcliquescalinggroup/reconcilestatus.go#L234-L248) only after genuine recovery:

- **A:** Preserve repeated recycling and document it as intentional.
- **B:** Recycle once per breach episode and re-arm after genuine recovery.

Since B changes behavior beyond the core health-history fix, do you think it belongs in this PR, or should it move to a follow-up issue/PR?

operator/internal/controller/common/component/utils/podclique.go
yankay: Thanks for highlighting this. Healthy existing objects now record HealthyStateObserved=True on their next reconciliation.
Already-regressed objects are not backfilled because the old timestamp heuristic is unreliable. They remain excluded from gang termination, preventing their existing PodCliques from being mistakenly deleted. I documented this limitation and the manual remediation in the release note.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
oleg-kushniriov: Follow-up, please - this PR should stay scoped to the health-history fix. B changes re-fire semantics (sticky flag, no automatic escape for partial-allocation deadlocks on non-gang backends) and deserves its own tests and sign-off. Restore clear-on-any-False here; with durable HealthyStateObserved it's no longer poisonous.

operator/internal/controller/common/component/utils/podclique.go
yankay: Thanks again for raising this. The action-required release note now documents the upgrade limitation and the manual remediation, using the same PCS-replica selector as the controller gang-termination delete. The tests retain coverage for both healthy legacy objects and already-regressed objects without durable history. Would you mind taking another look? :)

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
yankay: Thanks for clarifying. I switched this PR to option A, restoring clear-on-any-`MinAvailableBreached=False`, and opened #798 to discuss and test option B separately. Would you mind taking another look? :)

yankay:
Hi @unmarshall and @danbar2, would you mind taking a look when you have a chance? Thank you! :)

PR #792, opened 2026-08-25
Coherent Update PreReq PR#2: Enhance PodGangStatus with LastReady and LastScheduled fields.

### LINKED ISSUE #791: Enhance PodGangStatus with LastReady and LastScheduled

Introduce 2 new fields in `PodGangStatus` - `LastScheduled` and `LastReady`. Ensure that the PodGang component correctly sets this fields along with the conditions.

### Why is this needed?

**Background**
Today PodGangStatus exposes the Scheduled and Ready conditions (a condition tells you the current state) but not when each was first achieved. A metav1.Condition does carry LastTransitionTime, but it flips on every transition in either direction, so it cannot answer "has this PodGang ever been scheduled/ready, and when did that first happen" once the condition regresses to False.

**Why the fields are required?**
Coherent Rolling Updates roll a PodCliqueSet replica in bounded steps. Each step's PodGangs must not be scheduled until the PodGangs of the step they depend on have been placed. The orchestrator and the PodClique scheduling-gate-removal logic need a durable signal, set once and never cleared, of "this dependency has been scheduled at least once" to sequence the steps correctly.

- Ordering across steps. A later step's PodGang is gated on its dependency's PodGangs reporting LastScheduled != nil. This must stay true even if the dependency's Scheduled condition later flips False under transient pod loss. A condition alone would drop the signal on regression and stall or reorder the rollout.
- Progress gating. The orchestrator advances a sub-step only once the current epoch's PodGangs report LastReady != nil. It needs a durable "reached Ready" marker, not the live condition, because live readiness fluctuates as pods churn.

unmarshall:
#### What type of PR is this?

/kind feature
/kind api

Currently PodGang component only sets `Initialized` condition on the `PodGang` resource.  The `Scheduled` and `Ready` condition types were defined in the scheduler API but never set by the operator. This PR enhances the component to set all three conditions and adds two durable timestamp fields namely `LastScheduled` and `LastReady` to PodGangStatus.

Fixes #791

```release-note
Introduces `LastScheduled` and `LastReady` timestamp fields to `PodGangStatus`. Two additional conditions - `Scheduled` and `Ready` conditions are set on a PodGang resource. The timestamps are changed when the Scheduled and Ready conditions transitions to True (never cleared once set).
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podcliqueset/components/podgang/syncflow.go
+++ b/operator/internal/controller/podcliqueset/components/podgang/syncflow.go
@@ -446,9 +446,10 @@ func (r _resource) deleteExcessPodGangs(ctx context.Context, ss *syncState) erro
 	return nil
 }
 
-// createOrUpdatePodGangs creates or updates all expected PodGangs.
-// PodGangs are created with empty podReferences, Initialized=False.
-// Once all pods are created, PodReferences are populated and the PodGang is marked as Initialized=True.
+// createOrUpdatePodGangs creates or updates all expected PodGangs. PodGangs are created with empty
+// podReferences and Initialized set to False. Once all pods are created the PodReferences are
+// populated and the PodGang is marked Initialized True. The Scheduled and Ready conditions are
+// reconciled from live pod observation on every pass.
 func (r _resource) createOrUpdatePodGangs(ctx context.Context, ss *syncState) syncFlowResult {
 	result := syncFlowResult{}
 	for _, expectedPG := range ss.expectedPodGangs {
@@ -464,16 +465,18 @@ func (r _resource) createOrUpdatePodGangs(ctx context.Context, ss *syncState) sy
 			result.recordPodGangCreation(expectedPG.fqn)
 		}
 
-		// Verify all pods are created before proceeding
-		if err := r.verifyAllPodsCreated(ss, expectedPG); err != nil {
+		// verifyAllPodsCreated reports whether every constituent pod has been created and associated.
+		// Its result gates only the Initialized latch. The Scheduled and Ready conditions are live and
+		// are reconciled on every pass regardless, so a regression is reflected instead of frozen.
+		allPodsCreatedErr := r.verifyAllPodsCreated(ss, expectedPG)
+		if allPodsCreatedErr != nil {
 			ss.logger.Info("Not all pods are created or associated to the PodGang yet", "PodGangName", expectedPG.fqn)
-			result.recordError(err)
-			continue
+			result.recordError(allPodsCreatedErr)
 		}
 
-		// Reconcile the PodGang status conditions (Initialized, Scheduled, Ready) and their
-		// timestamps from live pod observation.
-		if err := r.reconcilePodGangStatus(ctx, ss, expectedPG); err != nil {
+		// Reconcile the PodGang Scheduled and Ready conditions and their timestamps from live pod
+		// observation. Initialized is set to True only once all pods are created, and is never reset.
+		if err := r.reconcilePodGangStatus(ctx, ss, expectedPG, allPodsCreatedErr == nil); err != nil {
 			ss.logger.Error(err, "failed to reconcile PodGang status", "PodGangName", expectedPG.fqn)
 			result.recordError(err)
 			continue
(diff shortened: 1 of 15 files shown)


sanjaychatterjee (dismissed):

sanjaychatterjee (dismissed):

scheduler/api/core/v1alpha1/podgang.go
```diff
@@ -184,6 +190,14 @@ type PodGangStatus struct {
...
 	// Conditions is a list of conditions that describe the current state of the PodGang.
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// LastScheduled is the time at which the Scheduled condition most recently transitioned to True.
+	// It is nil until the PodGang is first scheduled and is never reset to nil once set. On a later
+	// transition of the Scheduled condition back to True after a regression, it advances to the new time.
+	LastScheduled *metav1.Time `json:"lastScheduled,omitempty"`
+	// LastReady is the time at which the Ready condition most recently transitioned to True. It is nil
+	// until the PodGang is first ready and is never reset to nil once set. On a later transition of the
+	// Ready condition back to True after a regression, it advances to the new time.
+	LastReady *metav1.Time `json:"lastReady,omitempty"`
```
renormalize, line 203: nit: Following community guidelines to include these tags with fields making intent explicit.

```suggestion
	// LastScheduled is the time at which the Scheduled condition most recently transitioned to True.
	// It is nil until the PodGang is first scheduled and is never reset to nil once set. On a later
	// transition of the Scheduled condition back to True after a regression, it advances to the new time.
        // +optional
	LastScheduled *metav1.Time `json:"lastScheduled,omitempty"`
	// LastReady is the time at which the Ready condition most recently transitioned to True. It is nil
	// until the PodGang is first ready and is never reset to nil once set. On a later transition of the
	// Ready condition back to True after a regression, it advances to the new time.
        // +optional
	LastReady *metav1.Time `json:"lastReady,omitempty"`
```

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
```diff
@@ -468,19 +471,144 @@ func (r _resource) createOrUpdatePodGangs(ctx context.Context, ss *syncState) sy
...
+		}
+		if readyCount < pclq.minAvailable {
+			return false
+		}
+	}
+	return true
+}
+
 // patchPodGangInitializedStatus patches the Initialized condition with the given status.
 func (r _resource) patchPodGangInitializedStatus(ctx context.Context, ss *syncState, podGangName string, status metav1.ConditionStatus, reason, message string) error {
```
renormalize, line 620: This is being used in `createOrUpdatePodGang` where it invokes `setOrUpdateInitializedCondition`, which is similar to other helper functions introduced in this PR. Can this be modified so we don't have duplicate functionality spread across different files?

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
```diff
@@ -468,19 +471,144 @@ func (r _resource) createOrUpdatePodGangs(ctx context.Context, ss *syncState) sy
...
-				result.recordError(err)
-				continue
-			}
+		// Reconcile the PodGang status conditions (Initialized, Scheduled, Ready) and their
+		// timestamps from live pod observation.
+		if err := r.reconcilePodGangStatus(ctx, ss, expectedPG); err != nil {
+			ss.logger.Error(err, "failed to reconcile PodGang status", "PodGangName", expectedPG.fqn)
+			result.recordError(err)
+			continue
 		}
```
renormalize: Consider the case where in a previous reconciliation, all pods were created and the PodGang was set `Initialized=True`. In the current reconciliation, all pods are not present. This will case the `createOrUpdatePodGangs` function to short circuit at 471, since `verifyAllPodsCreated` will obviously return `false`. This causes the PodGang to not move away from the `Initialized=True` state.

Similarly, all other conditions set in `reconcilePodGangStatus` are not updated accordingly, if all pods are not present, i.e. the Scheduled condition, and the Ready condition.

These conditions *have* to be updated even when all pods are not present to actually represent the current state, as you've mentioned in the issue that these conditions are updated all the time, and therefore will not provide information about whether it has ever happened at least once.

renormalize (changes requested):

renormalize (commented):
The kube-scheduler backend does not sync the `PodGangStatus.Phase`. Would that lie in the scope of this PR, or another?

unmarshall:
>The kube-scheduler backend does not sync the PodGangStatus.Phase. Would that lie in the scope of this PR, or another?

The scope of this PR is clear. It only introduces `LastScheduled` and `LastReady` and changes around it. It does not touch anything else. `Phase` has been there since the begining and has not been set yet. We will revisit the API later on to either remove the fields that are not set/used or set them properly.

yankay:
HI @unmarshall 

Thanks for adding these conditions! https://github.com/ai-dynamo/grove/pull/792/changes#diff-c2be4fca55854714d846e5f971b65bc3887ce4d11ee165aa7345c98a862b63d6R765 says `LastScheduled` will be used for scheduling-gate removal, but a small local experiment showed that both `PodScheduled=False→True` and `ScheduledReplicas=0→1` are filtered by the current predicates. Could this delay `LastScheduled` updates and gate removal, or is the predicate wiring planned for a follow-up PR? :)

unmarshall:
> HI @unmarshall
> 
> Thanks for adding these conditions! https://github.com/ai-dynamo/grove/pull/792/changes#diff-c2be4fca55854714d846e5f971b65bc3887ce4d11ee165aa7345c98a862b63d6R765 says `LastScheduled` will be used for scheduling-gate removal, but a small local experiment showed that both `PodScheduled=False→True` and `ScheduledReplicas=0→1` are filtered by the current predicates. Could this delay `LastScheduled` updates and gate removal, or is the predicate wiring planned for a follow-up PR? :)

Originally i wanted to keep the wiring in the next PR, but the change was small, so i have now included in this PR itself to make things complete and also allow anyone to test/verify that the fields and conditions are set properly.
Addressed in bb69794c952251cfa9e40cbdd96a40988bcd2cf7

scheduler/api/core/v1alpha1/podgang.go
unmarshall: addressed in bb69794c952251cfa9e40cbdd96a40988bcd2cf7

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
unmarshall: addressed in bb69794c952251cfa9e40cbdd96a40988bcd2cf7

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
unmarshall: `Initialized` condition was never `LIVE` (you can check the main branch as well). So this PR keeps that as-is.
Rest is fixed in bb69794c952251cfa9e40cbdd96a40988bcd2cf7

shayasoolin (dismissed):
LGTM

shayasoolin (approved):

renormalize (approved):

Issue #787, opened 2026-08-21
Add a statically-configurable `external` scheduler backend


joeltg:
### What you would like to be added?

To fully realize the requirements for the scheduler backend framework presented in [GREP-375](https://github.com/ai-dynamo/grove/blob/main/docs/proposals/375-scheduler-backend-framework/README.md#story-1-third-party-scheduler-integration), we need a way for users to register and use a custom scheduler without code changes in Grove itself.

To do this, we need to stop treating the scheduler profile name (the current `SchedulerName` enum) as a closed set. The simplest way to accomplish this while avoiding dynamic linking / registration etc is to introduce a binary profile *type* to distinguish the built-in profiles from "external" ones. External schedulers then interface with workloads purely through PodGang/PodGroup CRs, without any runtime integration code in the Grove operator.

A sketch of the code changes might look like this:

```go
// api/config/v1alpha1/types.go
const (
    SchedulerProfileTypeBuiltIn  SchedulerProfileType = "BuiltIn"
    SchedulerProfileTypeExternal SchedulerProfileType = "External"
)

type SchedulerProfile struct {
    Name SchedulerName `json:"name"`
    // +kubebuilder:default=BuiltIn
    // +kubebuilder:validation:Enum=BuiltIn;External
    Type SchedulerProfileType `json:"type,omitempty"`
    Config *runtime.RawExtension `json:"config,omitempty"`
}
```

```go
// api/config/validation/validation.go — replace the flat SupportedSchedulerNames check
if p.Type == configv1alpha1.SchedulerProfileTypeExternal {
    // name lands in Pod.Spec.SchedulerName, so it just has to be a valid DNS subdomain
    for _, msg := range apivalidation.NameIsDNSSubdomain(string(p.Name), false) {
            allErrs = append(allErrs, field.Invalid(idxPath.Child("name"), p.Name, msg))
    }
} else if !slices.Contains(configv1alpha1.SupportedSchedulerNames, p.Name) {
    allErrs = append(allErrs, field.NotSupported(...))  // unchanged
}
```

```go
// registry/registry.go — in newSchedulerBackend, before the switch
if p.Type == configv1alpha1.SchedulerProfileTypeExternal {
    return external.New(p), nil
}
```

Plus `internal/scheduler/external/backend.go` — essentially `lpx/backend.go` with its `ValidatePodCliqueSet` driven by declared capabilities instead of hardcoded:

```go
type externalConfig struct {
[...]

PR #794, opened 2026-08-26
GREP-787: External scheduler integration

joeltg:
#### What type of PR is this?

/kind documentation

Adds GREP-787, proposing a `type` field on `SchedulerProfile` so that a scheduler which consumes Grove's `PodGang` resources directly can be registered through operator configuration alone, with no backend package in-tree.

The proposal asks for no stability commitment for `PodGang` and proposes no changes to it. External registration is scoped as alpha with no compatibility guarantee.

Refs #787

Written in response to the request on #787 for a GREP on the integration requirements. The validation and topology questions raised there are covered under *Workload validation model* and *Topology*.

A reference implementation exists and is deliberately not part of this PR. Happy to open it separately if the direction is agreed.

```release-note
NONE
```

```docs
docs/proposals/787-external-scheduler-integration/README.md
```

### DIFF
--- a/docs/proposals/787-external-scheduler-integration/README.md
+++ b/docs/proposals/787-external-scheduler-integration/README.md
@@ -1,0 +1,597 @@
+# GREP-787: External Scheduler Integration
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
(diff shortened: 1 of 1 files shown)


shayasoolin:
Hi @joeltg, thanks. Would you like to present the motivation and proposed solution in our weekly "Grove arch & design" meeting? Taking place on Mondays (8:00AM PST) and Thursdays (8:30AM PST).
It can be easier to discuss and get the attention of relevant stakeholders. We can also keep it offline if you prefer.

PR #799, opened 2026-08-28
Coherent Update PreReq PR#3: Support 1:N standalone PodClique to PodGang mapping

### LINKED ISSUE #797: Add support for 1:N PCLQ <-> PodGang mapping for standalone PCLQs

Pod component (and the PodGangMap component) needs to be enhanced to handle 1:N PodClique to PodGang mapping (where N >= 1). This is a prerequisite for coherent update.

### Why is this needed?

Today the operator assumes a standalone PodClique's pods always belong to a single PodGang. This holds during initial deployment and under the `RollingRecreate`/`OnDelete` update strategies. Once `coherent` update strategy is introduced, a standalone PodClique's pods will be distributed across multple anchor (MPG) PodGangs.

unmarshall:
#### What type of PR is this?

/kind enhancement
/kind api

Enhances the Pod component and the PodGangMap component so a standalone PodClique's pods can be distributed across one or more anchor PodGangs (1:N, where N >= 1), instead of assuming a single PodGang. This is a prerequisite for the coherent update strategy, where a standalone PodClique's pods spread across multiple anchor (MPG) PodGangs.

The 1:1 case (initial deployment, RollingRecreate, OnDelete) is preserved. The PodGangMap remains the single source of truth for per-PodGang composition. No new status fields are introduced.

Fixes #797 

**Changes**
-----------------------------------
**Standalone pod distribution (PodClique reconciler, pod component)**
  - Derives the desired per-anchor pod count from the PodGangMap entries, groups live pods by their PodGang label, computes the per-PodGang delta, and creates or deletes pods per PodGang.
  - Repairs pods missing the `grove.io/podgang label`. If there is just a single anchor PodGang, it adds the missing label in-place and requeues. If PodGangMap has multiple anchors (upcoming coherent update use case) then it will delete the pods that do not have the podgang label since it cannot find out which PodGang to stamp. In the next reconcile the Pod will get created with the correct PodGang.
  - Requeues while the PodGangMap has not yet absorbed a replica change under multiple anchors.
  - The scheudling-gate-removal dependency test now uses the PodGang LastScheduled timestamp (AllPodGangsAtEpochEverScheduled) rather than ScheduledReplicas >= MinReplicas.

**PodGangMap authoring (PodCliqueSet reconciler, PodGangMap component)**
  - Standalone PodClique counts are authored across current-generation anchors highest-index first, with scale-in draining from the highest anchor down.

**Gang-termination "ever scheduled" check**
  - Adds PodCliqueStatus.LastScheduled, set the first time the PodCliqueScheduled condition becomes True and advanced on each later transition to True.
  - WasPCLQEverScheduled now reads LastScheduled instead of inferring "ever scheduled" from the PodCliqueScheduled condition's LastTransitionTime against a fixed grace window. The old heuristic misclassified a PodClique that scheduled and then dropped below its scheduled count within the grace window as never-scheduled, which wrongly suppressed gang termination for a genuine regression.

**Per-PodGang pod expectations**
The pod expectations are now keyed per PodGang instead of per PodClique. A PodClique whose pods span multiple anchor PodGangs tracks pending creates and deletes separately for each PodGang, so a create or delete on one PodGang does not affect the counts of another.

**PodGangMap watch on the PodClique and PodCliqueScalingGroup controllers**
Both controllers now watch the PodGangMap. A coherent-update sub-step can move a standalone PodClique or a PodCliqueScalingGroup replica index between PodGangMap entries without changing the total. No other watch observes that move, so without this the redistribution waited for the periodic resync. The PodClique controller reacts to standalone PodClique placement changes; the PodCliqueScalingGroup controller reacts to replica-index placement changes.

> NOTE:  Status.PodGangMapping on PodCliqueStatus and PodCliqueScalingGroupStatus is intentionally not introduced. The PodGangMap component is the single writer and authors composition from child Spec.Replicas and live pods. GREP-393 still documents PodGangMapping and directionality flip. This is a design simplification and after the coherent update core engine is introduced in the next PR, then we will test this e2e to ascertain that this works properly. Then we will update the GREP as well.
 
> NOTE: Releasing the standalone PodGang MinReplicas constraint to 0 after scheduling is NOT part of this PR. It is only needed for the multi-anchor (1:N) coherent-update case and is deferred to the coherent-update engine PR, because a released MinReplicas of 0 sums into the KAI PodGroup top-level MinMember, which current released KAI versions reject (minMember must be >= 1). However this restriction has since been relaxed in the main branch. Once that change is part of a release we can then reconsider introducing it.

Adds PodCliqueStatus.LastScheduled, recording when a PodClique was first scheduled, used to arm gang-termination reliably.

```release-note
* Standalone PodClique pods can now be distributed across multiple anchor PodGangs (1:N mapping), a prerequisite for coherent updates. The 1:1 mapping used by initial deployment, RollingRecreate, and OnDelete is unchanged.
* Adds PodCliqueStatus.LastScheduled, recording when a PodClique was first scheduled, used to arm gang-termination reliably.
```

```docs
NONE
```


operator/internal/controller/podcliqueset/components/podgang/syncflow.go
yankay: Thanks for adding the per-PodGang replica count here. Would it make sense to also update `getPodsPendingCreationOrAssociation` to use this partitioned count and inspect only the pods assigned to the current PodGang? In a focused 1:2 reproduction, both PodGangs reported `pending=1` and remained `Initialized=False` because each treated the sibling anchor’s pod as a mismatch. Would you mind adding a regression test for this case as well?

shayasoolin (commented):
Review comments from the coherent-update prerequisite walkthrough.

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
```diff
@@ -0,0 +1,275 @@
...
+	// Wait until it has absorbed the change. A single anchor has nothing to decide.
+	if len(desiredCountByPodGang) > 1 && sumCounts(desiredCountByPodGang) != ss.pclq.Spec.Replicas {
+		return groveerr.New(groveerr.ErrCodeRequeueAfter, component.OperationSync,
+			fmt.Sprintf("PodGangMap has not yet absorbed the replica change for standalone PodClique %v, re-queueing", client.ObjectKeyFromObject(ss.pclq)))
+	}
+
+	r.expectationsStore.SyncExpectations(ss.pclqExpectationsStoreKey, nonTerminatingPodUIDs, terminatingPodUIDs)
+
+	livePodsByPodGang := groupPodsByPodGang(nonTerminatingPods)
+	countDeltaByPodGang := computeCountDeltaByPodGang(desiredCountByPodGang, liveCountByPodGang(livePodsByPodGang))
```
shayasoolin: We sync the expectations here, but `computeCountDeltaByPodGang` only uses desired and live counts; it does not account for the remaining create or delete expectations. If the informer cache is stale, a second reconcile can create the deficit again or delete another Pod while a previous deletion is still pending. This also affects the current 1:1 standalone path. Can we account for outstanding expectations before applying more changes, and add stale create/delete cache tests?

operator/internal/controller/podclique/reconcilestatus.go
```diff
@@ -278,6 +279,18 @@ func mutatePodCliqueScheduledCondition(pclq *grovecorev1alpha1.PodClique) {
...
 }
 
+// mutateLastScheduled advances Status.LastScheduled to now when the PodCliqueScheduled condition
+// transitions to True in this reconcile. It compares the condition before the mutators ran against
+// the condition now, so it re-stamps only on a fresh transition to True and not on every reconcile
+// while the PodClique stays scheduled. It is never reset to nil once set.
+func mutateLastScheduled(pclq *grovecorev1alpha1.PodClique, originalStatus *grovecorev1alpha1.PodCliqueStatus) {
+	scheduledNow := meta.IsStatusConditionTrue(pclq.Status.Conditions, constants.ConditionTypePodCliqueScheduled)
+	scheduledBefore := meta.IsStatusConditionTrue(originalStatus.Conditions, constants.ConditionTypePodCliqueScheduled)
+	if scheduledNow && !scheduledBefore {
```
shayasoolin: On upgrade, an existing PCLQ can already have `PodCliqueScheduled=True` while the new `LastScheduled` field is nil. This condition sees `true -> true` and never backfills it, so `WasPCLQEverScheduled` can later treat a previously running PCLQ as never scheduled and skip gang termination. Can we also set the field when `scheduledNow && LastScheduled == nil`, and add an upgrade test for this case?

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
```diff
@@ -233,6 +233,7 @@ func buildStandalonePCLQInfosForAnchorEntry(ss *syncState, pcsReplicaIndex int,
 			fqn:          pclqFQN,
 			replicas:     desiredPCLQReplicas,
```
shayasoolin, line 241: `getPodsPendingCreationOrAssociation` still assumes that all Pods of the PCLQ belong to this PodGang. With the new 1:N mapping, `pclq.Spec.Replicas` is the total across all PodGangs, and Pods labelled for another expected PodGang are valid. For example, a 6+6 split makes each PodGang count the other six Pods as pending association, so it never becomes Initialized. Can this check use this PodGang's `pclqInfo.replicas` and `associatedPodNames`, with a split-PCLQ test?

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
```diff
@@ -0,0 +1,275 @@
...
+	orderedPodGangNames := lo.Keys(countDeltaByPodGang)
+	sort.Strings(orderedPodGangNames)
+
+	createTasks, err := r.buildPerPodGangCreationTasks(logger, ss, countDeltaByPodGang, orderedPodGangNames)
+	if err != nil {
+		return err
+	}
+	deleteTasks := r.buildPerPodGangDeletionTasks(logger, ss, countDeltaByPodGang, orderedPodGangNames, livePodsByPodGang)
+
+	if len(createTasks) > 0 {
```
shayasoolin: A redistribution can have both positive and negative PodGang deltas, but this runs all creates before the deletes. GREP-393 says coherent update replaces in place: old Pods are taken down before replacements are created. Scheduling gates do not fully preserve that invariant, because a later reconcile can ungate the new Pods while the old Pods are still terminating. Can we delete first, wait for that state to be observed, and then create the replacements, with a redistribution test?

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
```diff
@@ -233,6 +233,7 @@ func buildStandalonePCLQInfosForAnchorEntry(ss *syncState, pcsReplicaIndex int,
...
 			minAvailable: *cliqueTemplate.Spec.MinAvailable,
```
shayasoolin, line 242: The PGM scale-in path can leave `clique: 0` on an Anchor that remains for other components. This code treats the present key as a PodGroup with zero PodReferences but the clique's positive MinReplicas, so that PodGang can never become Scheduled or Ready. The Pod consumer already skips zero counts. Can we also skip them here, or remove zero keys in the writer, with a test where another component keeps the Anchor alive?

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
```diff
@@ -0,0 +1,275 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// reconcileStandalonePCLQDistribution reconciles a standalone PodClique's pods across the anchor
+// PodGangs the PodGangMap assigns them to. It first repairs any non-terminating pod missing the
+// grove.io/podgang label, then reconciles per PodGang against the PodGangMap counts.
+func (r _resource) reconcileStandalonePCLQDistribution(ctx context.Context, logger logr.Logger, ss *syncSnapshot) error {
+	desiredCountByPodGang := buildDesiredCountByPodGang(ss)
```
shayasoolin, line 44: The Pod component now consumes PGM distribution changes, but this controller does not watch PodGangMap. A same-total redistribution may not trigger any existing watch: new PodGang creates are ignored, and PCS update-progress events only trigger when the selected replica changes. The PCLQ can then wait for the periodic resync, normally up to three minutes per sub-step. Can we watch PGM updates and enqueue the affected standalone PCLQs?

operator/internal/controller/common/component/utils/podgangmap.go
```diff
@@ -124,7 +124,7 @@ func podGangEntryForPCSGReplica(pgm *grovecorev1alpha1.PodGangMap, pcsgName stri
 }
 
 // AnchorPodGangEpoch returns the epoch of the AnchorIndex 0 anchor entry of the PodGangMap. Standalone
-// PodCliques always belong to this entry. It returns an error when no such anchor entry exists, a
+// PodCliques always belong to this entrnrry. It returns an error when no such anchor entry exists, a
```
shayasoolin: This comment is outdated now that a standalone PCLQ can span multiple Anchors. Can we describe this helper as limited to the 1:1 callers that still need Anchor 0? Also, `entrnrry` is a typo.

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
unmarshall: We are already listening for a PG resources. So the question that you are raising is for `same-total redistribution` - that is only going to happen during coherent updates and when that happens new PGs will be created which inturn creates newer events that gets into the PCLQ reconciler.
Can you tell me the case where a `same-total redistribution` will happen outside the case which covers in-progress coherent-update?

operator/internal/controller/podclique/components/pod/syncflow.go
```diff
@@ -73,62 +76,71 @@ func (r _resource) prepareSyncFlow(ctx context.Context, logger logr.Logger, pclq
 		return nil, err
 	}
 
-	// get the associated PodGang name.
-	ss.associatedPodGangName, err = r.getAssociatedPodGangName(pclq.ObjectMeta)
+	ss.cliqueName, err = utils.GetPodCliqueNameFromPodCliqueFQN(pclq.ObjectMeta)
```
renormalize: Nit: Just a suggestion for clarity's sake. Using `pclqTemplateName` and `pclqFQN` make things 100% unambigous, in my opinion.

```suggestion
	ss.cliqueTemplateName, err = utils.GetPodCliqueNameFromPodCliqueFQN(pclq.ObjectMeta)
```

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
unmarshall: Fixed in 3ec7be8cc6ac51a39a7abf21fc9d35806140c343

operator/internal/controller/podclique/reconcilestatus.go
unmarshall: fixed in 3ec7be8cc6ac51a39a7abf21fc9d35806140c343

operator/internal/controller/common/component/utils/podgangmap.go
unmarshall: fixed in 3ec7be8cc6ac51a39a7abf21fc9d35806140c343

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
unmarshall: Fixed in https://github.com/ai-dynamo/grove/commit/3ec7be8cc6ac51a39a7abf21fc9d35806140c343

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
unmarshall: Fixed in https://github.com/ai-dynamo/grove/commit/3ec7be8cc6ac51a39a7abf21fc9d35806140c343

operator/internal/controller/podclique/components/pod/syncflow.go
```diff
@@ -503,7 +550,7 @@ func (r _resource) createPods(ctx context.Context, logger logr.Logger, ss *syncS
 		// Get the available Pod host name index. This ensures that we fill the holes in the indices if there are any when creating
 		// new pods.
 		podHostNameIndex := availableIndices[i]
-		createTasks = append(createTasks, r.createPodCreationTask(logger, ss.pcs, ss.pclq, ss.associatedPodGangName, ss.pclqExpectationsStoreKey, i, podHostNameIndex))
+		createTasks = append(createTasks, r.createPodCreationTask(logger, ss.pcs, ss.pclq, ss.pcsgReplicaPodGangName, ss.pclqExpectationsStoreKey, i, podHostNameIndex))
```
renormalize: Since `createPods` is only run for PCSG PCLQ pods, can we document this in the function comment, and perhaps rename this so it is explicit that this is only run for PCSG PCLQs.

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
unmarshall: Fixed in 50a52c60c4dd666551c6be6c5f2097bb530b60df

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
unmarshall: Fixed in 31e582214d13cf1c80b4fc78666c7f11e4bf19fd
`applyCountDeltaByPodGang` now runs the deletes before the creates, so old Pods are taken down before their replacements are created. We don't block on the informer cache reflecting the deletions. Deletes run concurrently and, on error, return and requeue before any create happens. On success the delete expectations are recorded, so the next reconciliation won't re-issue them against a stale cache. Combined with the scheduling gates on the newly created Pods, this preserves the replace-in-place ordering. Added a redistribution test covering a PodGang with a negative delta alongside one with a positive delta.

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
shayasoolin: Thanks. I think a terminating Pod is now counted twice as gone. `reconcileLivePodCountWithExpectations` counts only `group.nonTerminating`, while `SyncExpectations` keeps the terminating Pod in the delete expectations that are subtracted. For example, desired=2 with 2 non-terminating Pods and 1 terminating Pod gives an effective count of 2 - 1 = 1, so Grove creates an extra Pod. The current pending-delete test keeps the Pod non-terminating, so it misses this case. Can we cover the terminating case and fix the effective count?

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
shayasoolin: I see that the Delete requests now run first, but replacements are still created in the same reconcile, before the old Pods are observed as gone. Gate removal also does not check whether the old Pods have finished terminating. Is a successful Delete request intended to satisfy the GREP’s “taken down” guarantee, even if old and new Pods can still overlap during the termination grace period?

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
shayasoolin: The case I’m concerned about is exactly a coherent-update sub-step. For example, the PGM changes from A:6 to A:3, B:3, with the same total. The PCLQ must reconcile to delete three Pods from A and create three for B.

Creating PodGang B does not currently trigger that reconcile: `podGangPredicate.CreateFunc` returns `false`, and the mapper also skips PodGroups without PodReferences. B cannot get those references or become initialized until the PCLQ creates its Pods, so the later Initialized event cannot bootstrap the flow either.

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
unmarshall: If the delete call is made which means that the Pods are now terminating. If we wait for pods to removed also from informer cache then we will further elongate the time it takes for the coherent to finish its update. Remember coherent update is anyways going to be slower as compared to rolling-recreate.

So we need to a call on this.
@sanjaychatterjee  what do you think?

shayasoolin: I agree that relying on the delete call is sufficient. If there are sufficient resources, update will semi-surge while the old pods are terminating, which I'm okay with. If there aren't, the new pods will be pending till the old ones are gone. SGTM

operator/internal/controller/podcliqueset/components/podgang/syncflow.go
yankay: Checked, looks good. Thanks! :)

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
unmarshall: Both controllers now watch PodGangMap. The PodClique controller fires when the standalone PodClique placement across entries changes and enqueues the affected standalone PodCliques. The PodCliqueScalingGroup controller has the same gap for a replica-index redistribution, so it now watches PodGangMap too, fires when the replica-index placement changes, and enqueues the affected PodCliqueScalingGroups. Both fire on create so a reconstructed PodGangMap is processed. Placement is keyed by epoch joined with the clique or config name, so a same-total move between entries counts as a change. Added mapper and predicate unit tests for both.

Fixed in 86c9aea33380ed0389fe4f2e2e8dd799d8d42006

operator/internal/controller/podclique/components/pod/standalonepclqdistribution.go
unmarshall: Addressed in 4fae044f803334bad6befd9e87d384a6e50d2b20

shayasoolin (approved):

yankay:
LGTM

sanjaychatterjee (approved):

Issue #796, opened 2026-08-27
PodGangMap recovery can lose externally scaled replicas when PCS template is stale


for-zero-one:
## Problem

Dynamo can scale Grove child resources directly through their `/scale` subresource:

- standalone components: `PodClique.spec.replicas`
- grouped components: `PodCliqueScalingGroup.spec.replicas`

The corresponding `PodCliqueSet` template replicas are intentionally not updated. This allows direct external scaling, including HPA-driven scaling, without having a parent-template reconciliation overwrite the child’s live replica value.

After [#778](https://github.com/ai-dynamo/grove/pull/778), `PodGangMap` is the source of truth for materialising `PodGang` resources. In the steady-state path, the map follows live child replica values. However, when a PodGangMap is absent, its bootstrap path derives entries from the PCS template.

Therefore, a missing/recreated PodGangMap can describe a different gang layout from the existing live PCLQ/PCSG resources.

Related: #777, #778, #649.

## Reproduction

1. Create a PCS whose PCSG template has:

   ```yaml
   replicas: 1
   minAvailable: 1
   ```

2. Scale the generated PCSG directly from `1` to `2`.

3. Wait for the existing PodGangMap to converge. It contains PCSG replica indices `0` and `1`, with index `1` in a ScaleOut entry.

4. Delete the PodGangMap for the PCS replica:

   ```sh
   kubectl delete podgangmap <pcs-name>-0
   ```

5. Wait for the PCS controller to recreate the PodGangMap.

## Observed behavior

The recreated PodGangMap is built from the PCS template, which still has `replicas: 1`.

Consequently, the new map contains only PCSG index `0` and omits the existing live PCSG index `1`.

Since the PodGang reconciler derives its expected PodGang set from PodGangMap entries, the scaled-out replica no longer has a corresponding desired PodGang. Its Pods can remain scheduling-gated while waiting for a PodGang that is absent from the map.

## Related standalone PodClique case

The same stale-template state affects recovery after deletion of a standalone PCLQ:

1. PCS template declares a standalone PCLQ with `replicas: 1`.
2. Dynamo scales the generated PCLQ directly to `4`.
3. The PCLQ is deleted.
4. The PCS controller recreates it using the PCS template, with `replicas: 1`.

[...]

PR #800, opened 2026-08-31
fix: preserve live replicas when rebuilding PodGangMaps

yankay:
#### What type of PR is this?

/kind bug

PodGangMap-only deletion normally self-heals on a subsequent reconcile. However, bootstrap reconstruction can temporarily omit externally scaled live replicas, causing ScaleOut PodGangs to be unnecessarily deleted and recreated.

This change reconciles bootstrap entries with live PodClique and PodCliqueScalingGroup replica counts before creating the PodGangMap, making recovery correct in the first reconcile and avoiding transient PodGang churn.

Includes focused unit tests and Kind verification.

Issue: #796

This is recovery-path hardening. Running Pods were not disrupted during verification. Recovery after deletion of the live child resource itself remains out of scope.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podcliqueset/components/podgangmap/podgangmap_test.go
+++ b/operator/internal/controller/podcliqueset/components/podgangmap/podgangmap_test.go
@@ -74,6 +74,51 @@ func TestSyncBootstrapsPodGangMapForFreshReplica(t *testing.T) {
 	assert.Equal(t, []string{anchor.Epoch}, scaleOut.DependsOn)
 }
 
+func TestSyncReconcilesBootstrapEntriesWithLivePCSG(t *testing.T) {
+	pcs := testutils.NewPodCliqueSetBuilder("pcs", "default", "uid").
+		WithReplicas(1).
+		WithScalingGroupConfig("sg", []string{"c"}, 1, 1).
+		WithPodCliqueSetGenerationHash(ptr.To("hash1")).
+		Build()
+	pcsg := testutils.NewPodCliqueScalingGroupBuilder("pcs-0-sg", "default", "pcs", 0).
+		WithReplicas(2).
+		Build()
+
+	cl := testutils.CreateDefaultFakeClient([]client.Object{pcs, pcsg})
+	operator := New(cl, groveclientscheme.Scheme, clocktesting.NewFakeClock(time.Unix(0, 1000)))
+
+	err := operator.Sync(context.Background(), logr.Discard(), pcs)
+	require.NoError(t, err)
+
+	pgm := getPodGangMap(t, cl, "pcs-0")
+	anchor := testutils.EntryByRole(pgm.Spec.Entries, grovecorev1alpha1.PodGangEntryRoleAnchor)
+	assert.Equal(t, []int32{0}, anchor.PCSGReplicaIndices["sg"])
+
+	scaleOut := testutils.EntryByRole(pgm.Spec.Entries, grovecorev1alpha1.PodGangEntryRoleScaleOut)
+	assert.Equal(t, []int32{1}, scaleOut.PCSGReplicaIndices["sg"])
+}
+
+func TestSyncReconcilesBootstrapEntriesWithLiveStandalonePodClique(t *testing.T) {
+	pcs := testutils.NewPodCliqueSetBuilder("pcs", "default", "uid").
+		WithReplicas(1).
+		WithStandaloneCliqueReplicas("clq-a", 1).
+		WithPodCliqueSetGenerationHash(ptr.To("hash1")).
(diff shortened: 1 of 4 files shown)


unmarshall:
Thanks for the PR!
As mentioned in the https://github.com/ai-dynamo/grove/issues/796#issuecomment-5475106532 I would like to understand the case where PGM cannot recover. I was unable to reproduce that. If you have been able to reproduce the issue that was reported then please list down the steps so that i can verify it at my end

yankay:
> Thanks for the PR! As mentioned in the [#796 (comment)](https://github.com/ai-dynamo/grove/issues/796#issuecomment-5475106532) I would like to understand the case where PGM cannot recover. I was unable to reproduce that. If you have been able to reproduce the issue that was reported then please list down the steps so that i can verify it at my end

HI @unmarshall 
Thanks for checking. I could not reproduce a permanently unrecoverable PodGangMap either. I only observed transient PodGang churn that self-healed during the subsequent reconcile. PR #800 is therefore recovery-path hardening that avoids this unnecessary churn.

unmarshall:
Hi @yankay 
>Thanks for checking. I could not reproduce a permanently unrecoverable PodGangMap either. I only observed transient PodGang churn that self-healed during the subsequent reconcile. PR https://github.com/ai-dynamo/grove/pull/800 is therefore recovery-path hardening that avoids this unnecessary churn.

Thanks for confirming. So both of us are not able to reproduce it.  Yes, for the PodGang churn we can solve that, but we need to properly harmonize it. Since this is a relatively new code, the fix that you propose will work but its not very clean.

So the flow should be:

```go

func (r _resource) runSyncFlow(ctx context.Context, syncSnap *syncSnapshot) error {
  for pcsReplicaIndex := range int(syncSnap.pcs.Spec.Replicas) {
       pgm, pgmExists := syncSnap.existingPGMByReplica[pcsReplicaIndex]

       if pgmExists && len(pgm.Spec.Entries) == 0 {
            return groveerr.New(
                 errCodePodGangMapNoEntries,
                 component.OperationSync,
                 fmt.Sprintf("PodGangMap %s for replica %d of PodCliqueSet %v has no entries, this is not expected. A live PodGangMap at least has an anchor entry", pgm.Na
  client.ObjectKeyFromObject(syncSnap.pcs)),
             )
        }

        entries, err := reconcileEntries(r.clk, syncSnap.pcs, pcsReplicaIndex, pgm,
             syncSnap.existingPodGangsByReplica[pcsReplicaIndex],
             syncSnap.existingStandalonePCLQsByReplica[pcsReplicaIndex],
             syncSnap.existingPCSGsByReplica[pcsReplicaIndex])
         if err != nil {
               return err
         }

         pgmName := apicommon.GeneratePodGangMapName(apicommon.ResourceNameReplica{Name: syncSnap.pcs.Name, Replica: pcsReplicaIndex})
         if err = r.createOrPatchPodGangMap(ctx, syncSnap.pcs, pgmName, pcsReplicaIndex, entries); err != nil {
             return err
         }
  }
  return r.deleteOrphanedPodGangMaps(ctx, syncSnap)
}
```

And so on and so forth...

unmarshall:
I can directly push changes to your PR (if i have access) as typing changes would mean i am coding in comments or you can take a shot at it as well since you have started this PR. Since there are no two paths now i would like to do some small refactoring and make the code clean with the new flow.

yankay:
Thanks for the guidance. I refactored the flow as suggested. Please take another look when you have some bandwidth. :)

unmarshall:
Ok, there are some other problems that I see that also needs to be resolved. For instance, PodGang always gets the latest PCS generation hash. This will not really work for coherent. I need to correct that as well. This is coming into the reconstruction path of PGM.

This needs a bit of thinking. So you either give me write access to your fork (so that i can push commits to this PR) or i create my own PR. This fix is a bit bigger than i imagined.

yankay:
> Ok, there are some other problems that I see that also needs to be resolved. For instance, PodGang always gets the latest PCS generation hash. This will not really work for coherent. I need to correct that as well. This is coming into the reconstruction path of PGM.
> 
> This needs a bit of thinking. So you either give me write access to your fork (so that i can push commits to this PR) or i create my own PR. This fix is a bit bigger than i imagined.

Thanks! It looks like the design has evolved quite a bit from the original issue we were trying to fix, and I’m honestly starting to lose track of some of the details here 😅

Please feel free to create your own PR — that might be the cleanest way forward, and it’ll probably be easier to review as well. Thanks for digging into this!

unmarshall:
@yankay as discussed with you closing this PR now in favor of #802

PR #802, opened 2026-09-01
fix: avoid PodGang churn when PodGangMap is reconstructed

### LINKED ISSUE #796: PodGangMap recovery can lose externally scaled replicas when PCS template is stale

unmarshall:
#### What type of PR is this?

/kind bug

When a PodGangMap is missing, its entries are bootstrapped from the PodCliqueSet template and the epochs of existing PodGangs, but the live child replica counts were ignored. A PodClique or PodCliqueScalingGroup scaled externally beyond its template (for example by an HPA or DGD through the `/scale` subresource) holds the
scaled count only on its own spec. So a rebuilt PodGangMap briefly omitted the scaled-out entries until a later reconcile repaired them from the live children, churning the scaled-out PodGang in the meantime.

This PR authors the entries through a single `reconcileEntries` path that bootstraps from the template when no PodGangMap exists and otherwise starts from the existing entries, then converges both against the live child replicas before the PodGangMap is written. The ScaleOut entry is ensured before the replica-index diff so a reconstructed scale-out index has a place to land, and the scale-out epoch is adopted from the existing PodGangs so the scale-out PodGang keeps its name. The reconstruction therefore reflects the scaled counts up front and avoids
the churn.

It also makes a PodGang carry the generation hash of the PodGangMap entry it is materialized from, rather than the PodCliqueSet's current generation hash, so a PodGang is labelled with the generation it belongs to. This matters during a coherent update when entries span more than one generation.

Fixes #796

This supersedes #800.

The change is limited to PodGangMap reconstruction. Recovery of the externally scaled count after deleting the child PodClique or PodCliqueScalingGroup resource itself is out of scope, since the scaled count lives only on the child's spec and is not recoverable once the resource is gone.

Reconstruction of a PodGangMap deleted during a coherent update (multiple anchor entries across generations, with per-entry anchor index and DependsOn that are held only on the PodGangMap) is also out of scope and deferred to a later PR which introduces the core coherent update engine.

**Additional change (refactoring)**
e2e scale helpers now scale through the resource's `/scale` subresource, the same path an external autoscaler uses. New PodGangMap reconstruction e2e tests cover the scaled PodCliqueScalingGroup, scaled standalone PodClique, fresh bootstrap, and scale-in cases.

Verified with:
- `go test ./internal/controller/podcliqueset/... ./internal/controller/common/component/utils/ -count=1`
- `make --directory=operator run-e2e TEST_PATTERN=Test_PGMR` (all 4 pass on a live k3d + KAI cluster)

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
+++ b/operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
@@ -111,7 +111,7 @@ func (r _resource) getExistingPCSGsByReplica(ctx context.Context, pcs *grovecore
 
 // getExistingPGMByReplica fetches all PodGangMaps for the PCS and groups them by PCS replica index.
 func (r _resource) getExistingPGMByReplica(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) (map[int]*grovecorev1alpha1.PodGangMap, error) {
-	existingPGMs, err := componentutils.ListPodGangMapsForPCS(ctx, r.client, client.ObjectKeyFromObject(pcs))
+	existingPGMs, err := componentutils.ListPodGangMapsForPCS(ctx, r.client, pcs.ObjectMeta)
 	if err != nil {
 		return nil, groveerr.WrapError(err,
 			errCodeListPodGangMaps,
--- a/operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
+++ b/operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
@@ -69,7 +69,7 @@ func (r _resource) takeSnapshot(ctx context.Context, logger logr.Logger, pcs *gr
 
 // getExistingStandalonePCLQsByReplica fetches all standalone PodCliques for the PCS and groups them by PCS replica index.
 func (r _resource) getExistingStandalonePCLQsByReplica(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) (map[int][]grovecorev1alpha1.PodClique, error) {
-	existingStandalonePCLQs, err := componentutils.GetPodCliquesWithParentPCS(ctx, r.client, client.ObjectKeyFromObject(pcs))
+	existingStandalonePCLQs, err := componentutils.GetPodCliquesWithParentPCS(ctx, r.client, pcs.ObjectMeta)
 	if err != nil {
 		return nil, groveerr.WrapError(err,
 			errCodeListPCLQs,
@@ -90,7 +90,7 @@ func (r _resource) getExistingStandalonePCLQsByReplica(ctx context.Context, pcs 
 
 // getExistingPCSGsByReplica fetches all PodCliqueScalingGroups for the PCS and groups them by PCS replica index.
 func (r _resource) getExistingPCSGsByReplica(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) (map[int][]grovecorev1alpha1.PodCliqueScalingGroup, error) {
-	existingPCSGs, err := componentutils.GetPCSGsForPCS(ctx, r.client, client.ObjectKeyFromObject(pcs))
+	existingPCSGs, err := componentutils.GetPCSGsForPCS(ctx, r.client, pcs.ObjectMeta)
 	if err != nil {
 		return nil, groveerr.WrapError(err,
 			errCodeListPCSGs,
--- a/operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
+++ b/operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
@@ -38,7 +38,7 @@ type syncSnapshot struct {
 	pcs                              *grovecorev1alpha1.PodCliqueSet
 	existingStandalonePCLQsByReplica map[int][]grovecorev1alpha1.PodClique
 	existingPCSGsByReplica           map[int][]grovecorev1alpha1.PodCliqueScalingGroup
-	existingPGMByReplica             map[int]grovecorev1alpha1.PodGangMap
+	existingPGMByReplica             map[int]*grovecorev1alpha1.PodGangMap
 	existingPodGangsByReplica        map[int][]groveschedulerv1alpha1.PodGang
 }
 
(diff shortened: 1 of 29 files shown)


yankay:
Thanks for adding this reconstruction support. I had two small questions that might be worth checking:

1. Could we add `^Test_PGMR` to both E2E matrices so the new tests run automatically in CI?
2. Could we filter reconstruction inputs with `metav1.IsControlledBy`? I reproduced a same-name PCS recreation where old-UID children caused stale replica counts and epochs to enter the new PodGangMap.

I may have missed some intended behavior here, so please feel free to correct me. Thank you :)

unmarshall:
>Could we add ^Test_PGMR to both E2E matrices so the new tests run automatically in CI?

Addressed in b8484f63b7c51445bec55772caf72930ce8f6823

unmarshall:
>Could we filter reconstruction inputs with metav1.IsControlledBy? I reproduced a same-name PCS recreation where old-UID children caused stale replica counts and epochs to enter the new PodGangMap.

Addressed in d7c3b8bd8cd2eb1243495266c8635f63633fb27c

operator/internal/controller/podcliqueset/components/podgangmap/syncflow.go
```diff
@@ -110,7 +110,7 @@ func (r _resource) getExistingPCSGsByReplica(ctx context.Context, pcs *grovecore
 }
 
 // getExistingPGMByReplica fetches all PodGangMaps for the PCS and groups them by PCS replica index.
-func (r _resource) getExistingPGMByReplica(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) (map[int]grovecorev1alpha1.PodGangMap, error) {
+func (r _resource) getExistingPGMByReplica(ctx context.Context, pcs *grovecorev1alpha1.PodCliqueSet) (map[int]*grovecorev1alpha1.PodGangMap, error) {
 	existingPGMs, err := componentutils.ListPodGangMapsForPCS(ctx, r.client, client.ObjectKeyFromObject(pcs))
```
shayasoolin: `ListPodGangMapsForPCS` still filters only by labels, unlike the other child-resource lists added in this PR.

If a PodCliqueSet is deleted and recreated with the same name before GC removes its old PodGangMap, this sync can pick up that stale PodGangMap. The later `CreateOrPatch` then fails because the old PodGangMap is controlled by the previous PCS UID, blocking reconstruction until GC removes it.

Please filter PodGangMaps by controller ownership as well, using the current PCS UID (metav1.IsControlledBy), consistent with PodGangs, PCSGs, and standalone PodCliques.

unmarshall: Addressed in cc5b23f64b2b45735ec7a55ab73a88adf0ba7b30

shayasoolin: I may be missing something, but I think this fix is incomplete. This filter prevents us from using the old PodGangMap while building the desired state, but it does not remove the old PodGangMap from Kubernetes. The next CreateOrPatch still looks up the same name, finds the old object, and fails because it is owned by the previous PCS UID. We need to handle that name collision explicitly: wait for or delete the old PodGangMap before creating the new one.

AFAIU this kind of issue might be relevant to any descendant resource of a PCS, that gets a pattern-based name without some random uuid/hash. Isn't it?

unmarshall: When a PCS is deleted then it cascade deletes all resources. So previous PGM will also get deleted, right?

unmarshall: Under background cascade deletion (the default), DELETE on the owner returns immediately. The owner object is gone, and the GC controller deletes the children asynchronously afterward. So for any controller whose children have deterministic names derived from the owner (name/namespace, no random suffix), a delete-then-recreate-same-name within the GC lag window hits exactly this. The new owner's reconcile tries to create a child at a name still occupied by the old owner's not-yet-collected child. So this issue is not introduces as part of this PR. But it self-heals as explained below.

You're right that the filter doesn't remove the stale object and the subsequent CreateOrPatch collides on the name. But in `buildResource` we invoke `controllerutil.SetControllerReference`. This will return `AlreadyOwnedError` because the lingering object is owned by the previous PCS UID. Due to this it will cause a requeue with error and it will eventually self-heal.

shayasoolin (approved):

sanjaychatterjee (approved):

yankay:
> > Could we add ^Test_PGMR to both E2E matrices so the new tests run automatically in CI?
> 
> Addressed in [b8484f6](https://github.com/ai-dynamo/grove/commit/b8484f63b7c51445bec55772caf72930ce8f6823)

Thanks It's also LGTM

PR #735, opened 2026-07-25
feat: aggregate KAI PodGroups by PodCliqueSet replica

enoodle:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Implements the aggregate KAI PodGroup representation described by GREP-525.

- Creates one KAI PodGroup per PodCliqueSet replica instead of one per Grove PodGang.
- Represents Base and Scaled PodGangs through a hierarchical KAI subgroup tree.
- Maps topology constraints according to GREP-244, including ClusterTopologyBinding resolution.
- Points newly created Pods at the aggregate PodGroup and migrates existing Pods from legacy per-PodGang PodGroups.
- Verifies Pod migration before deleting legacy PodGroups.
- Reconciles scale-down through a PodGang finalizer and garbage-collects aggregate PodGroups through PodCliqueSet ownership.
- Preserves KAI runtime-managed PodGroup fields during updates.
- Keeps synchronization failures retryable and reports them through logs and Warning Events on the triggering PodGang.

#### Which issue(s) this PR fixes:

Fixes #525 
Relates to #648 - Fixing it for KAI

#### Does this PR introduce a API change?

```release-note
NONE
```

#### Additional documentation e.g., enhancement proposals, usage docs, etc.:

```docs
docs/proposals/525-KAI-Scheduler-Backend/README.md
docs/proposals/244-topology-aware-scheduling/README.md
```

PR #814, opened 2026-09-04
feat(lpx): support KAI backend for non-LPX pod scheduling with LPX backend

### LINKED ISSUE #813: Support KAI-Scheduler fallback for non-LPX pod scheduling through LPX backend

LPX scheduler backend must be able to defer scheduling of non-LPX pods to another scheduler backend.

### Why is this needed?

LPX scheduler can only schedule LPX workloads. In order to support PodCliqueSets with mixed GPU and LPX workloads, the LPX backend must be able to defer pod grouping to a fallback scheduler. Currently, the only supported fallback is KAI-scheduler. When using the LPX scheduler backend, there is no guarantee of gang scheduling at this time.

steved:
#### What type of PR is this?

/kind feature

LPX scheduler does not support scheduling non-LPX pods. For a given PodCliqueSet that contains both LPX and GPU workers, this PR allows us to pass-through non-LPX pods to KAI-Scheduler.

Fixes #813

```release-note
NONE
```

```docs
N/A
```


operator/internal/scheduler/registry/registry.go
```diff
@@ -104,7 +105,14 @@ func newSchedulerBackend(cl, directClient client.Client, scheme *runtime.Scheme,
 	case configv1alpha1.SchedulerNameVolcano:
 		b = volcano.New(cl, scheme, rec, p)
 	case configv1alpha1.SchedulerNameLPX:
-		b = lpx.New(p)
+		var fallbackBackend scheduler.Backend
+
+		kaiProfileIndex := slices.IndexFunc(cfg.Profiles, func(p configv1alpha1.SchedulerProfile) bool { return p.Name == configv1alpha1.SchedulerNameKai })
```
sanjaychatterjee: Instead of hardcoding to KAI, why not use the `Config` in the LPX SchedulerProfile to setup the backend configuration. E.g.
```
// LPXSchedulerConfiguration defines the configuration for the lpx-scheduler backend.
type LPXSchedulerConfiguration struct {
	// FallbackProfileName is the scheduler profile used to schedule pods that do not
	// request LPX resources. It must name another profile in scheduler.profiles.
	// +optional
	FallbackProfileName string `json:"fallbackProfileName,omitempty"`
}
```
This will make the configuration easy via Helm charts. E.g.
```
scheduler:
  profiles:
    - name: default-scheduler
    - name: kai-scheduler
    - name: lpx-scheduler
      config:
        fallbackProfileName: kai-scheduler
```

operator/internal/scheduler/kai/backend.go
sanjaychatterjee, line 235: Will this work when the PodGang reference here is the fallback PodGang?

operator/internal/scheduler/lpx/backend.go
```diff
@@ -27,60 +29,136 @@ import (
...
+			return b.fallbackBackend.ValidatePodCliqueSet(ctx, pcs)
+		}
+	}
+
 	return nil
 }
+
+// fallbackPodGang constructs a new PodGang object without any LPX pod groups.
+// It is then passed to the SyncPodGang function of the fallback backend.
+func (b *schedulerBackend) fallbackPodGang(ctx context.Context, podGang *groveschedulerv1alpha1.PodGang) (*groveschedulerv1alpha1.PodGang, error) {
```
sanjaychatterjee, line 156: I think creating a temporary PodGang copy and mutating it for the KAI backend might have some side-effects such as [this](https://github.com/ai-dynamo/grove/pull/814/changes#r3962578585). Instead, I would suggest that LPX can pass the original PodGang reference to the KAI backend along with some additional info that specifies which Podgroups to select from the Grove PodGang, maybe call it a PodGroup selector. Another option is to make the Grove PodGang API more amenable for multi-scheduler setup where the PodGroup can include a new field called schedulerName, and each Grove scheduler backend only selects the PodGroups that are assigned to it. 
Need to hear from @unmarshall @enoodle @alaypatel07 @kangclzjc  on their thoughts.

operator/internal/scheduler/registry/registry.go
steved: Since it isn't actually configurable I figured it was overkill and added complexity to the init, but I've added ab579ff6. I wasn't sure if the intention was to parse the scheduler configs in `New()` (no error returned) or `Init()` so I went with `Init` for now.

operator/internal/scheduler/kai/backend.go
steved: At the moment, yes, it's a deep-copied object just with the configuration pared down.

operator/internal/scheduler/lpx/backend.go
steved: Are these the two options, more-or-less? Is there a preferable one?

Modify `SyncPodGang`:
```go
// SyncPodGang synchronizes (creates/updates) scheduler-specific resources for a PodGang
// reacting to a creation or update of a PodGang resource. If pod groups are specified, the 
// scheduler backend should only sync the selected names.
SyncPodGang(ctx context.Context, podGang *groveschedulerv1alpha1.PodGang, podGroups []string) error
```

Or:
```go
// PodGroup defines a set of pods in a PodGang that share the same PodTemplateSpec.
type PodGroup struct {
	// SchedulerName is the name of the scheduler backend that should manage this pod group.
	// If unspecified, the default scheduler backend profile will be used.
	SchedulerName string `json:"schedulerName"`
}
```

yankay:
Could a PCS update that adds or removes an LPU request leave the KAI PodGroup stale? The PodGang may remain unchanged, so I wonder what triggers reconciliation. Please correct me if I’m missing something.

enoodle (commented):
Overall looks great.
I think that there is a hidden assumption over all behind this PR that no one will update LPU pods to be GPUs or the other way around once the PCS exists, but if someone for any reason will do it then I think that things will break, so maybe we should have some validation that it doesn't happen?
I am not seeing that grove supports this kind of validation hook at allows to inspect the update, but maybe it should be added.

operator/internal/scheduler/lpx/backend.go
```diff
@@ -27,60 +30,156 @@ import (
...
+	if b.profile.Config == nil {
+		return nil
+	}
+
+	if err := json.Unmarshal(b.profile.Config.Raw, &config); err != nil {
+		return fmt.Errorf("%s: invalid config: %w", b.name, err)
+	}
+
+	if config.FallbackProfileName != "" {
+		if config.FallbackProfileName != string(configv1alpha1.SchedulerNameKai) {
```
unmarshall, line 78: what is the point of the configuration `FallbackProfileName` when you anyways hardcode `KAI` when comparing?

operator/internal/scheduler/lpx/backend.go
```diff
@@ -27,60 +30,156 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type schedulerBackend struct {
+	client          client.Client
+	name            string
+	profile         configv1alpha1.SchedulerProfile
+	fallbackBackend scheduler.Backend
```
unmarshall, line 37: I am not sure this should live here. It should be the job of the dispatcher to do the fan-out and it should not be done from within a specific backend.

operator/api/config/v1alpha1/types.go
```diff
@@ -114,6 +114,14 @@ type KubeSchedulerConfig struct {
 	GangScheduling bool `json:"gangScheduling,omitempty"`
 }
 
+// LPXSchedulerConfiguration defines the configuration for the lpx-scheduler backend.
+type LPXSchedulerConfiguration struct {
+	// FallbackProfileName is the scheduler profile used to schedule pods that do not
+	// request LPX resources. It must name another profile in scheduler.profiles.
+	// +optional
+	FallbackProfileName string `json:"fallbackProfileName,omitempty"`
```
unmarshall, line 122: "Fallback" implies a retry/degradation path. What it actually does is partition podgroups by scheduler. LPX backend handles LPX pods, the other scheduler backend handles the rest.
So perhaps a cleaner approach would be to get this inside the PodGroup itself.

operator/internal/scheduler/types.go
```diff
@@ -38,7 +38,7 @@ type Backend interface {
 
 	// Init provides a hook to initialize/setup one-time scheduler resources,
 	// called at the startup of grove operator.
-	Init(directClient client.Client) error
+	Init(directClient client.Client, backends map[string]Backend) error
```
unmarshall, line 41: passing backends via `Init` is not right. 
First: you should try and avoid injecting peer-backend wiring in the Backend interface. As stated elsewhere this should be done in a higher layer, not here.
Second: If you really want access to all supported backends then use the `New` constructor function and not the `Init` method.

operator/internal/scheduler/lpx/backend.go
```diff
@@ -27,60 +30,156 @@ import (
...
-) error {
-	return nil
+// SyncPodGang creates no new resources of LPX pod groups.
+// Any non-LPX pod groups are passed to the fallback backend SyncPodGang.
+func (b *schedulerBackend) SyncPodGang(ctx context.Context, podGang *groveschedulerv1alpha1.PodGang) error {
+	if podGang == nil {
+		return fmt.Errorf("podGang is nil")
+	}
+
+	if b.fallbackBackend == nil {
```
unmarshall, line 100: is there going to be no case where only LPX scheduler is used and no other backend scheduler is ever required?

operator/internal/scheduler/lpx/backend.go
```diff
@@ -27,60 +30,156 @@ import (
...
 }
 
-func (b *schedulerBackend) SyncPodGang(
-	_ context.Context,
-	_ *groveschedulerv1alpha1.PodGang,
-) error {
-	return nil
+// SyncPodGang creates no new resources of LPX pod groups.
+// Any non-LPX pod groups are passed to the fallback backend SyncPodGang.
+func (b *schedulerBackend) SyncPodGang(ctx context.Context, podGang *groveschedulerv1alpha1.PodGang) error {
```
unmarshall, line 95: The only purpose you have put something in the SyncPodGang is to just call KAI. This is a design smell which points to the fact that the this should not be done in the LPX backend itself.

operator/internal/scheduler/lpx/backend.go
```diff
@@ -27,60 +30,156 @@ import (
...
+	if b.fallbackBackend == nil {
+		return nil
+	}
+
+	newPodGang, err := b.fallbackPodGang(ctx, podGang)
+	if err != nil {
+		return err
+	}
+
+	return b.fallbackBackend.SyncPodGang(ctx, newPodGang)
```
unmarshall, line 109: If you look at the currently badly named podgang controller, then you will realise that it will call SyncPodGang on the backend scheduler on create/update/delete of the PodGang. When the PodGang is created then it does not have Pod references. As and when the Pods are created this list of pod references are updated. So this will generate a lot of events. On each such creation you will do a DeepCopy + N client.Gets on PCLQ -> build a partial PG which is then given to KAI.

operator/internal/scheduler/lpx/backend.go
```diff
@@ -27,60 +30,156 @@ import (
...
+			fallback.Spec.PodGroups = append(fallback.Spec.PodGroups, group)
+		}
+	}
+
+	return fallback, nil
+}
+
+// usesLPX determines whether a given pod spec is requesting LPX support by
+// checking the resource requests and limits for lpu.nvidia.com/lpu or nvidia.com/lpu.
+func usesLPX(podSpec corev1.PodSpec) bool {
```
unmarshall, line 177: is there a possibility that `InitContainers` also request LPU resources? You do not check for init containers currently.

operator/internal/scheduler/lpx/backend.go
unmarshall: We can add scheduler name to PodGroup. However lets be careful because once this gets added to the API, then we are now saying that we support multiple schedulers per PCS. To even do gang scheduling the two or more collaborating schedulers will now have to synchronize to ensure that gang semantics are not broken. You then need to have a barrier across schedulers to ensure that each participating scheduler can find sufficient resources for the pods in a PodGang that it manages. Only when that is done can scheduling gates be lifted.

enoodle: The "virtual" PodGang can be created on KAI's end or the LPX end - But it will miss those updates. I am not sure how the original PodGang should get those updates anyway from the KAI side.

In #735 I am trying to add an ability to KAI's scheduler backend to reload the whole PCS to build a unified PodGroup from all the pod gangs. This will break with this behavior but could work if we add `SchedulerName` to each PodGroup - So this is my preferred path here. 

This multi scheduler setup will lead to breaking the gang scheduler guarantee though unless the two scheduler find a way to communicate information between them.

sanjaychatterjee: I agree we need a long-term plan here. If `schedulerName` in the PodGroup API is the preferred option then we are committing to the multi-scheduler path which means we need to have a viable solution for gang-scheduling. I am not comfortable going ahead in that direction without a proper GREP. So, for the short-term, I think we should unblock LPX with a shortcut where the logic can preferably stay within the LPX backend. Let me setup a short sync so we are all aligned on the short term solution. @steved @enoodle @unmarshall

operator/internal/scheduler/lpx/backend.go
steved: theoretically, but not currently possible.

operator/internal/scheduler/lpx/backend.go
steved: @enoodle let me know if there's a good way to indicate the LPX podgroups for https://github.com/ai-dynamo/grove/pull/735. we could theoretically stick the selected podgroups in an annotation, but that could potentially exceed the limit. otherwise, I think the best way might be to move the `fallbackPodGang()` behavior to the KAI backend.

enoodle: @steved I suggest that we proceed here against what is currently in main and in 735 I will find a way to work with this - If needed we can add the annotation there.

sanjaychatterjee (approved):

enoodle (approved):

PR #816, opened 2026-09-08
Enable `EnableReadYourWritesConsistency` in `sigs.k8s.io/controller-runtime@v0.25.0`

renormalize:

https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.25.0 introduced the `EnableReadYourWritesConsistency` feature, which ensures all writes are reflected in subsequent reads from the cache backed client that grove-operator **heavily** uses.

Due to the extremely large number of events that grove-operator deals with overall, we have seen time and again that the cached client does not read with the latest writes reflected, and is the source of various defensive programmed code paths.

We rely on re-queues to help the cache catch up to the actual state of the cluster. This feature will reduce the requeues, thereby reducing the total number of events that grove-operator has to deal with.

Although the feature is marked experimental by the SIG, I would highly recommend we as a project adopt this feature as it is a solution to a source of many problems we face/have faced in the past. Any observations we have can be contributed upstream to the SIG. If maintainers are not too confident, we can make this feature configurable in Grove's `OperatorConfiguration`.

Fixes #

Downstream users like dynamo-operator will be forced to upgrade their `k8s.io/*` dependencies to `0.37.0` once we merge this. Please do not merge until we have an estimate timeline on their own upgrade schedule.

```release-note
Upgrade to `sigs.k8s.io/controller-runtime@v0.25.0` and enable `EnableReadYourWritesConsistency`
```

```docs

```


renormalize:
@julienmancuso is there a timeline on dynamo-operator's upgrade to `k8s.io/*` dependencies to `0.37.0`? Thanks!

renormalize:
In a discussion with @unmarshall and @sanjaychatterjee, it was made clear that experimental features such as this should not be included *even as opt-in features* in projects running in production, such as grove.

I will revisit this once this feature becomes GA, so all controller code using expectations can be simplified **significantly**.

Issue #807, opened 2026-09-03
Core coherent update engine for PodCliqueSet


unmarshall:
### What you would like to be added?

This issue builds on top of #777 , #791 and #797 and introduces the core coherent update engine:
* API changes: 
  * PCS `status.updateProgress` is enhanced to track coherent update progress.
  * `MaxUnavailable` is introduced as an additional update configuration handle.
* A central orchestrator (in `PodCliqueSetReplica` component) is introduced as coherent is a multi-step process.
* PodGangMap component is enhanced with the logic to compute and record next-(sub)step of coherent updates.
* PCLQ and PCSG webhooks are introduced to prevent scaling operations for the PCS replica under update (only applicable for coherent update strategy)

### Why is this needed?

Read [GREP-393](https://github.com/ai-dynamo/grove/tree/coherent-updates/docs/proposals/393-coherent-rolling-updates) to know about motivation, its goals and the detailed proposal.

PR #817, opened 2026-09-09
feat: Coherent Update PreReq PR#4: Per-Component MaxUnavailable and ProgressDeadline for RollingRecreate updates

### LINKED ISSUE #741: Add maxUnavailable safeguards for PodClique rollouts

Add an opt-in `rollingUpdate.maxUnavailable` field to the PodClique and PodCliqueScalingGroup configuration in `PodCliqueSet`.

The field should bound controller-initiated rollout and scale-in deletions:

- For a standalone PodClique, the budget is measured in Pods.
- For a PodCliqueScalingGroup, the budget is measured in complete logical replicas, where all member PodCliques at the same replica index form one unit.
- Rollout and scale-in operations share the same availability budget.
- Deletion expectations consume the budget immediately, before informer caches observe the deletion.
- When the field is unset, existing behavior remains unchanged.

The policy controls concurrency only and does not impose ordinal or Pod-index rollout ordering.

The initial implementation is proposed in #740.

### Why is this needed?

During rolling updates, Grove can delete multiple old Pending or unavailable Pods/replicas before the existing ready-replica update state machine runs. Scale-in also uses a separate deletion path. Continuous configuration updates, existing unavailable replicas, or overlapping rollout and scale-in operations can therefore cause multiple replicas to be recreated at once.

For serving workloads, this reduces the number of Ready replicas and concentrates requests on the remaining capacity, which can destabilize latency and throughput. A shared `maxUnavailable` budget allows users to limit disruption—for example, `maxUnavailable: 1` ensures that Grove starts no additional controller-driven deletion until the current Pod or logical PCSG replica is available again.

### LINKED ISSUE #786: RollingRecreate can complete before all updated pods are Ready

### LINKED ISSUE #815: Introduce per-component MaxUnavailable and ProgressDeadline for rolling updates

IIntroduce a per-component rolling-update disruption budget and progress deadline for Grove rolling updates. The configuration is strategy-agnostic and is a prerequisite for Coherent updates (GREP-393), which will consume the same fields rather than defining their own. This change implements it for RollingRecreate, which is what exists on main today. The Coherent strategy reuses the same fields in the follow-up #807 PR.

**What it includes:**

 - Add a rollingUpdate configuration to each standalone PodClique (PodCliqueTemplateSpec) and each PodCliqueScalingGroup (PodCliqueScalingGroupConfig), carrying:
     - maxUnavailable: the maximum number of pods (standalone PodClique) or complete logical replicas (PodCliqueScalingGroup) that may be unavailable at any moment while that component is being rolled.
       It is measured against the current desired count.
     - progressDeadline (optional duration): if a rolling component makes no progress within this duration, its update is reported as not progressing. Absence means no deadline.
- Make the RollingRecreate rollout readiness-aware. It disrupts at most maxUnavailable of a component at a time and does not consider the update complete until the desired replacements are Ready. This
   also fixes the premature-completion behaviour reported in #786.
- Add an UpdateInProgress condition on PodClique, PodCliqueScalingGroup, and PodCliqueSet:
     - True: an update is currently in progress.
     - False: no update is currently in progress.
     - Unknown (reason ProgressDeadlineExceeded): an update is in progress but a component has exceeded its progressDeadline. It is Unknown because Grove cannot determine the cause, which may be a
       deadline that is too tight while pods load model weights, a slow network, a lack of scheduler capacity, or something else.
     - At the PodCliqueSet level the condition aggregates over the components that are actually rolling, and its message reports how many are stuck per kind so an operator can drill down.
- Scale-in remains outside maxUnavailable. A scale-in changes the desired count and the reconciler simply converges to it under the existing minAvailable floor.

**Scope notes:**

 - The rollingUpdate fields are shared and strategy-agnostic, but the behavior in this change applies to RollingRecreate. Coherent reuses the same fields in the follow-up PR.
 - Components whose template did not change are updated by a generation-hash bump with no pod downtime and are not paced by maxUnavailable.
 - When a rolling update is stuck, gang termination stays suspended for the stuck component by design. Recreating the gang cannot fix a bad image or a broken external dependency, so resolution is
   operator-driven, either a corrected update or fixing the dependency, after which the update resumes and the condition clears.

### Why is this needed?

Today a RollingRecreate update gives operators no control over how much of a component is disrupted at once, and it does not wait for replacements to become Ready before treating the rollout as done (see #786). For serving workloads this can concentrate traffic on the remaining Ready replicas, or declare a rollout complete while the new generation is still starting or crash-looping.

 A per-component maxUnavailable budget lets operators bound worst-case disruption during an update and trade rollout speed against preserved serving capacity. A per-component progressDeadline combined with the UpdateInProgress condition gives an observable, drill-downable signal when a rollout is not making progress, so operators can inspect and act. Making the rollout readiness-aware fixes premature-completion bug in #786.

unmarshall:
#### What type of PR is this?

/kind api
/kind feature

Adds configurable, budget-based control over RollingRecreate updates for PodClique and PodCliqueScalingGroup, makes rollout completion readiness-aware, and establishes the shared update primitives that the upcoming Coherent update strategy will build on.

This is a prerequisite for Coherent updates: the RollingUpdateConfiguration (`MaxUnavailable` and `ProgressDeadline`) and the `UpdateInProgress` condition introduced here are deliberately strategy-neutral and will be reused by the Coherent flow (which lands in a follow-up PR). Introducing and hardening them under RollingRecreate first keeps that follow-up focused on the coherent orchestration itself.

**API changes**

- `RollingUpdateConfiguration` (new type) — a per-component rolling-update configuration, shared by RollingRecreate today and Coherent later.
 - `MaxUnavailable` (*int32) — how many units (pods for a standalone PodClique, replicas for a PodCliqueScalingGroup) may be unavailable at once during an update, measured against the component's desired count.
 - `ProgressDeadline` (*metav1.Duration) — how long a rollout may make no progress before it is flagged ProgressDeadlineExceeded; unset disables the deadline.
 - `RollingUpdate` on PodCliqueTemplateSpec — configures a standalone PodClique's rollout.
 - `RollingUpdate` on PodCliqueScalingGroupConfig — configures a PodCliqueScalingGroup's rollout; the group owns the budget for its member cliques.
 
**Webhook changes**
 - Defaulting — under RollingRecreate, maxUnavailable defaults to 1 when unset. Defaulting never clears a consumer-set rollingUpdate; removing it when switching to OnDelete is the consumer's responsibility.
 - Validation — rollingUpdate is rejected under OnDelete and on PodCliqueScalingGroup member cliques. maxUnavailable must be greater than 0 and must not exceed the component's replicas.

**Rollout**

- Both the standalone PodClique and the PodCliqueScalingGroup paths now disrupt units against a stateless budget derived from MaxUnavailable, replacing the previous name-blind one-at-a-time step counter.
 - A rollout only completes once the new-hash units are actually Ready, which fixes premature completion (issue #786).
 - The PodCliqueScalingGroup rollout records a delete-expectation for each deleted replica, so the MaxUnavailable budget is respected even though member-PodClique readiness is eventually consistent. Every deletion is counted against the budget, so a replica is never replaced based on a stale not-Ready status read.

**Observability**

 - A new `UpdateInProgress` condition on PodClique, PodCliqueScalingGroup, and (aggregated) PodCliqueSet, reported as True while rolling, False when idle, and Unknown with reason `ProgressDeadlineExceeded` when a rollout stalls past its `ProgressDeadline`.
 - A reason-only `Update` printer column on all three kinds.

> NOTE: The change also carries a few preparatory refactors (relocating component utilities to a neutral package, moving the PodClique FQN name helpers, and switching to k8s.io/apimachinery sets.Set)

**Reverting the VUPG1 upgrade e2e test**

What #755 added: Test_VUPG1 was extended to directly scale a PodClique that belongs to a PodCliqueScalingGroup (the bootstrap member) out and in, and to assert on the resulting PCSG pod indices. The upgrade.yaml workload was also changed to wrap bootstrap/worker in a PodCliqueScalingGroup.

There is a gap in expectations and reality. Direct scaling of PCLQ that belong to a PCSG has never been tested since this was not meant to be supported. Unfortunately its also not been prevented. So when #755 landed it did not really pay attention to the fact that once you start scaling individual PCLQs that belong to a PCSG then `DeletionSorter` comes into play and that chooses pods to delete based on set of criterion. On scale-in, the excess-pod selection does not consider the PCSG pod index, so it can delete a pod whose removal leaves a gap in the PCSG pod indices that never converges. The test therefore times out waiting for the indices to converge. This is a pre-existing operator behavior that is independent of the changes in this PR — Test_VUPG1 fails intermittently on main (roughly 2 out of 5 local runs) with the same "PCSG pod indices did not converge" error.

A separate issue #818 has now been created to track and fix the issue brought in by #755.

Fixes #786
Fixes #741 
Fixes #815

 - Foundation for Coherent updates: RollingUpdateConfiguration and UpdateInProgress are intentionally shared building blocks. The Coherent strategy will consume the same MaxUnavailable budget helper (ComputeAllowedBudget), the same ProgressDeadline bookkeeping, and the same condition, so please review these as cross-strategy primitives rather than RollingRecreate-only.
 - MaxUnavailable is the sole disruption bound for RollingRecreate, matching Kubernetes Deployment semantics.  
 - ProgressDeadline is opt-out (nil disables the deadline). A stalled rollout keeps gang termination suspended by design so the workload is not torn down while an operator investigates.
 - e2e coverage drives the scenarios with KWOK Stage fixtures (delayed readiness and a simulated crash loop) under a new e2e/k8s/kwok helper package, since KWOK otherwise marks pods Ready immediately.

```release-note
API removal (action required):
  - Removed `status.readyPodsSelectedToUpdate` from PodClique and `status.readyReplicaIndicesSelectedToUpdate` from PodCliqueScalingGroup, along with the `PodsSelectedToUpdate` and `PodCliqueScalingGroupReplicaUpdateProgress` types. The rolling update now derives its disruption budget from live Pod and replica state each reconcile and no longer tracks a selection cursor, so these fields conveyed no information. They are removed outright rather than left permanently unset, so consumers that read them must drop that usage (a Go client will get a compile error against the new API module).

Adds rolling-update configuration and rollout progress reporting to the Grove API.
   Spec additions (all +optional):
   - New type `RollingUpdateConfiguration`.
   - `maxUnavailable` (*int32): maximum pods (standalone PodClique) or replicas (PodCliqueScalingGroup) that may be unavailable during an update, measured against the component's desired count. Defaults to 1 under RollingRecreate and must be greater than 0 and must not exceed the component's replicas. Valid only under the RollingRecreate strategy.
   - `progressDeadline` (*metav1.Duration): duration a rolling update may make no progress before it is flagged as ProgressDeadlineExceeded. Opt-out; unset (nil) disables the deadline.
   - `PodCliqueTemplateSpec.rollingUpdate` (*RollingUpdateConfiguration): rolling-update config for a standalone PodClique. A PCSG member PodCliqueTemplateSpec must not set it.
   - `PodCliqueScalingGroupConfig.rollingUpdate` (*RollingUpdateConfiguration): rolling-update config for a PodCliqueScalingGroup, which owns the budget for its member cliques.

Status additions (all +optional):
  - `lastProgressedAt` (*metav1.Time) on the PodClique and PodCliqueScalingGroup update-progress status, recording when the rolling update last advanced.
  - New `UpdateInProgress` condition on PodClique, PodCliqueScalingGroup, and PodCliqueSet, with reasons `Progressing`, `NoActiveUpdate`, and `ProgressDeadlineExceeded`. It is True while rolling,
 False when idle, and Unknown when a rollout exceeds its progress deadline.

CRD additions:
  - An `Update` printer column (shows the UpdateInProgress reason) on PodClique, PodCliqueScalingGroup, and PodCliqueSet.

Behavioral change: rolling updates now complete only after replacement pods become Ready.
```

```docs
NONE
```


unmarshall:
All API changes will be captured in updates to [GREP-393](https://github.com/ai-dynamo/grove/tree/0be004cff387eccdd476ff3566caa9e5f0e27c55/docs/proposals/393-coherent-rolling-updates) as part of PR for #807

sanjaychatterjee (dismissed):

operator/api/common/constants/constants.go
```diff
@@ -142,6 +142,10 @@ const (
 	// and cleared once every replica has migrated. While it is True the PodCliqueScalingGroup and
 	// PodClique reconcilers requeue without acting, so scaling does not interleave with the migration.
 	ConditionTypePodGangMigrationInProgress = "PodGangMigrationInProgress"
+	// ConditionTypeUpdateInProgress indicates whether a rolling update of the component is currently in progress.
```
renormalize, line 145: Strategy neutral terminology according to the PR description.

```suggestion
	// ConditionTypeUpdateInProgress indicates whether an update of the component is currently in progress.
```

operator/api/common/constants/constants.go
```diff
@@ -142,6 +142,10 @@ const (
...
+	// True means an update is progressing, False means no update is active, and Unknown means the update has not
+	// made progress within its ProgressDeadline and warrants operator inspection.
```
renormalize: Additional phrasing that matches the rest of this file.

```suggestion
	// made progress within its configured ProgressDeadline and warrants operator inspection.
```

operator/api/common/constants/constants.go
```diff
@@ -170,6 +174,13 @@ const (
 	ConditionReasonGangTerminationActive = "GangTerminationActive"
 	// ConditionReasonClusterTopologyNotFound indicates that the ClusterTopologyBinding resource required for topology-aware scheduling was not found.
 	ConditionReasonClusterTopologyNotFound = "ClusterTopologyNotFound"
+	// ConditionReasonProgressing indicates that a rolling update of the component is in progress and advancing.
```
renormalize, line 177:
```suggestion
	// ConditionReasonProgressing indicates that an update of the component is in progress and advancing.
```

operator/api/common/constants/constants.go
```diff
@@ -170,6 +174,13 @@ const (
...
+	ConditionReasonProgressing = "Progressing"
+	// ConditionReasonNoActiveUpdate indicates that no rolling update of the component is currently in progress.
```
renormalize, line 179:
```suggestion
	// ConditionReasonNoActiveUpdate indicates that no update of the component is currently in progress.
```

operator/api/common/constants/constants.go
```diff
@@ -170,6 +174,13 @@ const (
...
+	ConditionReasonNoActiveUpdate = "NoActiveUpdate"
+	// ConditionReasonProgressDeadlineExceeded indicates that a rolling update has not made progress within its
```
renormalize, line 181:
```suggestion
	// ConditionReasonProgressDeadlineExceeded indicates that an update has not made progress within its
```

operator/api/core/v1alpha1/podclique.go
```diff
@@ -174,11 +179,18 @@ type PodCliqueUpdateProgress struct {
 	// ReadyPodsSelectedToUpdate captures the pod names of ready Pods that are either currently being updated or have
 	// been previously updated. This field is only set for auto update strategies where Grove orchestrates Pod deletions.
 	// For the OnDelete strategy this field is not set, because Pod replacement is initiated by user-driven Pod deletions.
+	//
+	// Deprecated: the rolling update is now driven by an availability budget computed from live Pod state on each
+	// reconcile and no longer tracks a selection cursor. This field is no longer set and will be removed in a future
+	// release. Do not depend on it.
```
renormalize: A status field never being set, when it previously was, is equivalent to removing it from the API altogether.

If we are to actually follow patterns used by the community to deprecate API fields, we need a few releases for consumers to adapt to this change; which you are proposing we not follow in this PR. We faced a somewhat similar dilemma in #438, and the codeowners were strongly for the introduction of the new field, and removal of the (still populated) deprecated field a few releases later.

However, in this case, it makes no sense to try and adapt the logic we will follow henceforth just to populate this field. If we're breaking the deprecation conventions anyway, might as well remove this field from the API and be done with it.

Consumers who did read this field during updates will now run into `nil` errors if we keep it in the API. Removing it from the API forces them to reconsider their code that reads this, when they update their Grove API dependency.

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -172,6 +173,39 @@ type PodCliqueSetReplicaUpdateProgress struct {
...
+// the OnDelete strategy, where defaulting clears it and the PodCliqueSet validating webhook rejects
+// it if set.
+type RollingUpdateConfiguration struct {
+	// MaxUnavailable is the maximum number of pods (for a standalone PodClique) or
+	// PodCliqueScalingGroup replicas (for a PCSG) that may be unavailable at any moment during an
+	// update of this component, measured against the component's desired count.
+	//
+	// Defaulting:
+	//   - RollingRecreate: defaults to 1.
+	//   - OnDelete: the whole RollingUpdate is cleared by defaulting, because it does not apply to
```
renormalize:
```suggestion
	//   - OnDelete: the whole RollingUpdateConfiguration is cleared by defaulting, because it does not apply to
```

operator/api/core/v1alpha1/podcliqueset.go
```diff
@@ -172,6 +173,39 @@ type PodCliqueSetReplicaUpdateProgress struct {
...
+	//     OnDelete. This also lets a transition away from RollingRecreate drop the stale configuration.
+	//
+	// Validation:
+	//   - When set, must be greater than 0.
+	//   - OnDelete: RollingUpdate must not be set. The PodCliqueSet validating webhook rejects a
```
renormalize, line 196: I am confused. If the defaulting webhook (which runs first) clears the `RollingUpdateConfiguration`, it would never be the case where the validating webhook rejects the PCS when the update strategy is `OnDelete`: the defaulting always corrects it so the validating webhook approves the specification.

Consider the case where a consumer changes the update strategy to `OnDelete`, which then causes the defaulting webhook to unset this field. Now, the consumer has lost access to the `RollingUpdateConfiguration` in the specficiation, especially if they are not tracking specifications through version control.

In my opinion, it should be the responsibility of the validating webhook to reject it when the update strategy is set to `OnDelete`, and defaulting does nothing. Removing this field from the specification must only be done by the consumer. Defaulting webhook should only fill in "holes" in the specification that might be required by the controller, not remove fields that the consumer has actively populated.

operator/api/core/v1alpha1/scalinggroup.go
```diff
@@ -139,11 +144,18 @@ type PodCliqueScalingGroupUpdateProgress struct {
 	// chosen at a time to update, once it is updated and becomes ready, the next ready replica is chosen for update.
 	// This field is only set for auto update strategies where Grove orchestrates Pod deletions.
 	// For OnDelete strategy this field is not set, because Pod replacement is initiated by user-driven Pod deletions.
+	//
+	// Deprecated: the rolling update is now driven by an availability budget computed from live replica state on each
+	// reconcile and no longer tracks a selection cursor. This field is no longer set and will be removed in a future
+	// release. Do not depend on it.
```
renormalize: https://github.com/ai-dynamo/grove/pull/817#discussion_r3977215060

renormalize (commented):
Comments on the proposed API changes. Comments on the implementation will follow.

operator/api/common/constants/constants.go
unmarshall: This is not accurate, it must be a rolling update. For on-delete this does not hold.

operator/api/common/constants/constants.go
unmarshall: This is not accurate, it must be a rolling update. For on-delete this does not hold.

operator/api/common/constants/constants.go
unmarshall: This is not accurate, it must be a rolling update. For on-delete this does not hold.

operator/api/common/constants/constants.go
unmarshall: This is not accurate, it must be a rolling update. For on-delete this does not hold.

operator/api/core/v1alpha1/podclique.go
unmarshall: > Consumers who did read this field during updates will now run into nil errors if we keep it in the API.

Why? This is a pointer field. This automatically means that it can be nil and in-fact will be nil. So the consumers can run into nil pointer dereference if they have incorrectly assumed that this field is always going to have non-nil value during update.

I am not sure if anyone really depends upon it. This was more meant to have a persistent state of current/completed pods so that the reconciler can know where it was in case it crashed. I will check if we can still populate this field for the time being.

operator/api/common/constants/constants.go
renormalize: There was a slight confusion from my side, which has now been clarified. A "Rolling update" could refer to both `Coherent` and `RollingRecreate`.

We used the "auto update" terminology before where the orchestration is performed by the operator.

Have we defined what a rolling update is anywhere in the API? Can you point me to that if I have missed it? If not, I think it is important we define this terminology somewhere in the API. We already mention the upcoming Coherent update in the API documentation in this PR, so might as well define this too.

According to our current understanding, would rolling update always imply a grove-operator orchestrated update? If yes, can we get rid of the auto update terminology since we have two phrases: "Rolling Update" and "Auto Update" that refer to the same thing.

operator/api/common/constants/constants.go
renormalize: https://github.com/ai-dynamo/grove/pull/817#discussion_r3977742923

operator/api/common/constants/constants.go
renormalize: https://github.com/ai-dynamo/grove/pull/817#discussion_r3977742923

operator/api/common/constants/constants.go
renormalize: https://github.com/ai-dynamo/grove/pull/817#discussion_r3977742923

operator/api/core/v1alpha1/podclique.go
renormalize: > Why? This is a pointer field. This automatically means that it can be nil and in-fact will be nil. So the consumers can run into nil pointer dereference if they have incorrectly assumed that this field is always going to have non-nil value during update.

Sure, `nil` errors will not occur with the appropriate `nil` checks, but if consumers have logic based on the fact that this field is populated when they initiate an update, their flows will break. 

Do I think anyone's using this? Probably not.

Are we sure? For the most part, but I am not 100% positive about it.

> I am not sure if anyone really depends upon it. 

I don't think anyone would be using it either at the moment, but it is a part of the public API, and we need to follow the deprecation policy for all public API fields.

operator/api/common/constants/constants.go
unmarshall: Rolling update is a standard term that is well understood. On-Delete is not rolling update because it is user controlled and not orchestrated as you have rightly pointed to in your comment. Rolling update has variants - rollout of updates via recreates or rollout of updates via coherent strategy.

I have now harmonised the vocabulary and removed `auto-updates` as that is not a standard term that is used in the community.

operator/api/common/constants/constants.go
unmarshall: responded : https://github.com/ai-dynamo/grove/pull/817#discussion_r3977913450

operator/api/common/constants/constants.go
unmarshall: responded : https://github.com/ai-dynamo/grove/pull/817#discussion_r3977913450

operator/api/common/constants/constants.go
unmarshall: responded : https://github.com/ai-dynamo/grove/pull/817#discussion_r3977913450

operator/internal/utils/component/rollingupdate.go
```diff
@@ -0,0 +1,49 @@
...
+const DefaultRollingRecreateMaxUnavailable int32 = 1
+
+// EffectiveMaxUnavailable returns the MaxUnavailable to use while rolling a component under
+// RollingRecreate, tolerating a nil rollingUpdate or a nil MaxUnavailable by applying
+// DefaultRollingRecreateMaxUnavailable.
+//
+// NOTE: this exists only for PodCliqueSets created before the RollingUpdate fields existed, which
+// reconcile without re-admission and carry a nil MaxUnavailable. Once every PodCliqueSet has been
+// re-admitted and carries a MaxUnavailable populated by the defaulting webhook, this accessor is no
+// longer required and should be removed in favor of reading the field directly.
```
renormalize, line 33: `RollingUpdate` fields are only read when the PCS in actively being updated. Even if there are a bunch of events that arrive and make grove-operator reconcile the PCS, these fields are never looked at since an update has not yet been triggered. When the update is eventually triggered, the PCS goes through defaulting and validating webhooks, which cause these fields to be populated.

At operator startup, all events are fed to it so it "catches up" too.

So I am not 100% sure if this is needed.

We can just keep this in if you choose to, I don't have strong opinions on this.

operator/api/core/v1alpha1/podcliqueset.go
unmarshall: The reason that it was introduced was that the user can switch from `RollingRecreate` to `OnDelete` update strategy. For `RollingRecreate` defaulting webhook defaults the `MaxUnavailable` to 1. So when the user switches to `OnDelete` it also must now reset/remove the RollingUpdateConfiguration that the user did not explicitly set. 

But perhaps that was a weak argument.

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
```diff
@@ -97,19 +120,41 @@ func defaultPodCliqueScalingGroupConfigs(scalingGroupConfigs []grovecorev1alpha1
...
+
 // defaultPodSpec adds defaults to PodSpec.
 func defaultPodSpec(spec *corev1.PodSpec) *corev1.PodSpec {
 	defaultedPodSpec := spec.DeepCopy()
 	if utils.IsEmptyStringType(defaultedPodSpec.RestartPolicy) {
 		defaultedPodSpec.RestartPolicy = corev1.RestartPolicyAlways
 	}
 	if defaultedPodSpec.TerminationGracePeriodSeconds == nil {
-		defaultedPodSpec.TerminationGracePeriodSeconds = ptr.To[int64](30)
+		defaultedPodSpec.TerminationGracePeriodSeconds = ptr.To(defaultTerminationGracePeriodSeconds)
```
renormalize, line 160: The default `TerminationGracePeriodSeconds` set by kubernetes for any pod is 30 seconds. Also, the default restart policy set by Kubernetes is `Always`. Seems like we can get rid of this function altogether.

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
```diff
@@ -97,19 +120,41 @@ func defaultPodCliqueScalingGroupConfigs(scalingGroupConfigs []grovecorev1alpha1
...
+	}
+	defaulted.MaxUnavailable = ptr.To(componentutils.DefaultRollingRecreateMaxUnavailable)
+	return defaulted
+}
+
 // defaultPodSpec adds defaults to PodSpec.
 func defaultPodSpec(spec *corev1.PodSpec) *corev1.PodSpec {
 	defaultedPodSpec := spec.DeepCopy()
 	if utils.IsEmptyStringType(defaultedPodSpec.RestartPolicy) {
 		defaultedPodSpec.RestartPolicy = corev1.RestartPolicyAlways
```
renormalize: This is also redundant. The default policy is `Always`. We can get rid of this function altogether.

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
```diff
@@ -97,19 +120,41 @@ func defaultPodCliqueScalingGroupConfigs(scalingGroupConfigs []grovecorev1alpha1
...
 	return defaultedScalingGroupConfigs
 }
 
+// defaultRollingUpdateConfiguration returns the RollingUpdateConfiguration normalized for the active
+// update strategy. RollingUpdate only applies to RollingRecreate, so it is cleared for any other
+// strategy, which keeps a transition away from RollingRecreate from leaving a stale configuration that
+// the validating webhook would then reject. For RollingRecreate an existing MaxUnavailable is
+// preserved and a missing one is defaulted to 1. ProgressDeadline is never defaulted, a nil value opts
+// out of the deadline.
+func defaultRollingUpdateConfiguration(existing *grovecorev1alpha1.RollingUpdateConfiguration, updateStrategy grovecorev1alpha1.UpdateStrategyType) *grovecorev1alpha1.RollingUpdateConfiguration {
```
renormalize, line 135: This function and its side-effects are already being discussed here: https://github.com/ai-dynamo/grove/pull/817#discussion_r3977350701.

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
```diff
@@ -19,33 +19,52 @@ import (
...
+	defaultTerminationGracePeriodSeconds int64 = 30
 )
 
 // defaultPodCliqueSet adds defaults to a PodCliqueSet.
 func defaultPodCliqueSet(pcs *grovecorev1alpha1.PodCliqueSet) {
 	if utils.IsEmptyStringType(pcs.Namespace) {
-		pcs.Namespace = "default"
+		pcs.Namespace = metav1.NamespaceDefault
+	}
+	defaultUpdateStrategy(pcs)
```
renormalize: This function should be invoked in `defaultPodCliqueSetSpec`, to mirror the tree structure of the type.

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
```diff
@@ -486,13 +499,37 @@ func validateCliqueDependencies(cliques []*grovecorev1alpha1.PodCliqueTemplateSp
...
+	if v.updateStrategyType() == grovecorev1alpha1.OnDeleteStrategy {
+		if rollingUpdate != nil {
+			allErrs = append(allErrs, field.Forbidden(fldPath, "rollingUpdate must not be set when the update strategy is OnDelete"))
+		}
+		return allErrs
+	}
+	if rollingUpdate == nil {
+		return allErrs
+	}
+	if rollingUpdate.MaxUnavailable != nil && *rollingUpdate.MaxUnavailable <= 0 {
```
renormalize: I would tend to think `MaxUnavailable` must also always be lesser than or equal to `Replicas`. I don't see such an explicit check here.

operator/api/core/v1alpha1/podcliqueset.go
unmarshall: addressed in 1650c8cdaa8bac3fc3e8346aae270a6164fac651

operator/api/core/v1alpha1/podcliqueset.go
unmarshall: addressed in 1650c8cdaa8bac3fc3e8346aae270a6164fac651

operator/api/common/constants/constants.go
unmarshall: addressed in 1650c8cdaa8bac3fc3e8346aae270a6164fac651

operator/api/common/constants/constants.go
unmarshall: addressed in 1650c8cdaa8bac3fc3e8346aae270a6164fac651

renormalize (commented):
Comments on webhooks.

operator/api/core/v1alpha1/podclique.go
unmarshall: Since the API is v1alpha i am not really inclined to keep these fields.
@sanjaychatterjee do you agree that we can just remove this field?

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
renormalize: Latest commit [1650c8c](https://github.com/ai-dynamo/grove/pull/817/commits/1650c8cdaa8bac3fc3e8346aae270a6164fac651) has addressed this already.

operator/api/core/v1alpha1/podclique.go
renormalize: I am a strong proponent for dropping it. It's a call you folks must make as maintainers of the project.

operator/internal/utils/component/rollingupdate.go
unmarshall: A PodCliqueSet created before the RollingUpdate field existed is stored with a nil RollingUpdate. Defaulting only runs on admission, so the stored object is not retroactively defaulted on a Grove upgrade.

>At operator startup, all events are fed to it so it "catches up" too.

This will not trigger defaulting + validating webhooks.

Consider the following scenario:
Older version of PCS that does not have `RollingUpdate` fields set. Now an update starts. In the middle of the update, Grove has been upgraded. This updates the CRDs, the webooks and the code that now looks at RollingUpdate fields. If it reads MaxUnavailable directly then it can result in nil-pointer dereference causing a panic.

So I would like to keep this.

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
unmarshall: this has already been addressed in https://github.com/ai-dynamo/grove/commit/1650c8cdaa8bac3fc3e8346aae270a6164fac651

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
unmarshall: This was an existing code. The only change that has been done is to use a constant instead of the hard coded value. You are right that these are defaulted by k8s apiserver. However there is one caveat, today a this is defaulted as part of PCS defaulting webhook, so these fields take part in the hash computation.

Since this is code that already existed i would not increase the scope of this PR further (its already increased due fixes to e2e tests). If you wish to take it up, then create an issue, analyse the impact and then you can remove it if its safe to do so.

renormalize: Okay - fair enough. The hash computation is a very valid argument. There's no further analysis required, since this is in the pod template in the PCS. Removing this will definitely change the hash.

operator/internal/utils/component/rollingupdate.go
renormalize: Okay, let's keep this.

operator/internal/webhook/admission/pcs/defaulting/podcliqueset.go
unmarshall: addressed in 3fdd2c4f13c5162335a062b862830fa42e3e978b

operator/internal/webhook/admission/pcs/validation/podcliqueset.go
unmarshall: addressed in 3fdd2c4f13c5162335a062b862830fa42e3e978b

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -461,3 +466,75 @@ func pruneStrayPCSGPCLQs(pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pclqsPer
...
+		return apicommon.GeneratePodCliqueScalingGroupName(apicommon.ResourceNameReplica{Name: pcs.Name, Replica: pcsReplicaIndex}, cfg.Name) == pcsg.Name
+	})
+	if !ok || pcsgConfig.RollingUpdate == nil {
+		return nil, nil
+	}
+	return pcsgConfig.RollingUpdate.ProgressDeadline, nil
+}
+
+// isPCSGAutoUpdateInProgress reports whether the PodCliqueScalingGroup is under an auto-orchestrated update.
+func isPCSGAutoUpdateInProgress(pcsg *grovecorev1alpha1.PodCliqueScalingGroup) bool {
```
renormalize: As discussed in https://github.com/ai-dynamo/grove/pull/817#discussion_r3977913450, could this helper be renamed accordingly?

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
```diff
@@ -461,3 +466,75 @@ func pruneStrayPCSGPCLQs(pcsg *grovecorev1alpha1.PodCliqueScalingGroup, pclqsPer
...
+// LastProgressedAt is cleared and the condition is False (NoActiveUpdate).
+func mutateUpdateInProgressCondition(pcsg *grovecorev1alpha1.PodCliqueScalingGroup, originalStatus *grovecorev1alpha1.PodCliqueScalingGroupStatus, progressDeadline *metav1.Duration) {
+	now := metav1.Now()
+	if isPCSGAutoUpdateInProgress(pcsg) {
+		if pcsg.Status.UpdateProgress.LastProgressedAt == nil || pcsg.Status.UpdatedReplicas > originalStatus.UpdatedReplicas {
+			pcsg.Status.UpdateProgress.LastProgressedAt = &now
+		}
+	} else if pcsg.Status.UpdateProgress != nil {
+		pcsg.Status.UpdateProgress.LastProgressedAt = nil
+	}
```
renormalize, line 504: Please correct me if I am wrong: we're setting this to `nil` here only because the update has already ended and it doesn't really matter when it was last progressed at, correct?

operator/internal/controller/podcliquescalinggroup/components/podclique/rollingupdate.go
```diff
@@ -48,97 +63,87 @@ const (
...
 	}
-	// always delete PCSG replicas that are either pending or unavailable
-	if err = r.deleteOldPendingAndUnavailableReplicas(ctx, logger, sc, work); err != nil {
-		return err
+
+	desiredNumReplicas := int(sc.pcsg.Spec.Replicas)
+
+	// Completion is readiness-aware. End the update only when the desired number of replicas are fully
+	// updated and Ready, so a rollout never completes while replacements are not yet available.
+	if uw.numUpdatedReadyReplicas >= desiredNumReplicas {
```
renormalize: Don't think `numupdatedReadyReplicas` would ever exceed `Spec.Replicas`

```suggestion
	if uw.numUpdatedReadyReplicas == desiredNumReplicas {
```

operator/internal/controller/podclique/components/pod/rollingupdate.go
```diff
@@ -35,134 +36,154 @@ import (
...
 	}
 
-	// Check if there is currently a pod that is selected for update and its update has not yet completed.
-	if isAnyReadyPodSelectedForUpdate(pclq) && !isCurrentPodUpdateComplete(ss, uw) {
+	desiredNumPods := int(ss.pclq.Spec.Replicas)
+
+	// Completion is readiness-aware. End the update only when no old-hash Pods remain and the desired
+	// number of new-hash Pods are Ready, so a rollout never completes while replacements are not yet
+	// available.
+	if uw.oldHashPodCount == 0 && uw.newReadyPodCount >= desiredNumPods {
```
renormalize: https://github.com/ai-dynamo/grove/pull/817#discussion_r3979546525

renormalize (dismissed):
Added a couple more nits, looks good otherwise. Please also check RU12 which is failing in CI.

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
unmarshall: If UpdateProgress != nil and the update has ended (UpdateEndedAt != nil) then it resets the progress marker for the next update. `LastProgressedAt` is only meaningful for the in-progress ProgressDeadline check (now - LastProgressedAt > deadline). If we do not clear it and there is a next update then the next update will start with a stale LastProgressedAt (from the previous rollout). The deadline check would then falsely report that progress deadline has exceeded. Clearing it guarantees the next update measures the deadline from its own first reconcile

operator/internal/controller/podcliquescalinggroup/expectations/pcsgexpectations.go
```diff
@@ -0,0 +1,87 @@
...
+	return store.ExpectDeletions(logger, key, uids...)
+}
+
+// HasPCSGReplicaDisruptionBeenTriggered reports whether the disruption of a PodCliqueScalingGroup
+// replica has already been triggered this rollout, by checking whether any member PodClique carries a
+// delete expectation under the PCSG-scoped key. Such a replica is treated as unavailable immediately,
+// without waiting for the eventually-consistent cache or member PodClique status to catch up.
+func HasPCSGReplicaDisruptionBeenTriggered(store *expect.ExpectationsStore, key string, members []grovecorev1alpha1.PodClique) bool {
+	return lo.SomeBy(members, func(pclq grovecorev1alpha1.PodClique) bool {
+		return store.HasDeleteExpectation(key, pclq.GetUID())
```
yankay, line 82: This introduces a reproducible data race: reconcile calls `HasDeleteExpectation` while the informer can concurrently call `ObserveDeletions`. The former reads `uidsToDelete` without `ExpectationsStore.mu`, while the latter mutates it under that lock. `go run -race` reports the concurrent read/write. Could we lock all expectation-set accesses and return copies from collection accessors?

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
unmarshall: addressed in d937e2521bfcd6e8a399e574163e854fe186c23e

operator/internal/controller/podcliquescalinggroup/components/podclique/rollingupdate.go
unmarshall: addressed in d937e2521bfcd6e8a399e574163e854fe186c23e

operator/internal/controller/podclique/components/pod/rollingupdate.go
unmarshall: addressed in d937e2521bfcd6e8a399e574163e854fe186c23e

operator/internal/controller/podcliquescalinggroup/expectations/pcsgexpectations.go
unmarshall: Good catch!
This was an existing bug. This PR only added more accessor functions.
Addressed now in 6277d9303473bf7ed15d020dc48400a5cb640dfa

operator/internal/controller/podcliquescalinggroup/reconciledelete.go
```diff
@@ -32,13 +33,30 @@ import (
 // Child PodCliques carry a controller owner reference to this PCSG, so finalizer
 // removal hands the cascade off to the Kubernetes garbage collector.
 func (r *Reconciler) triggerDeletionFlow(ctx context.Context, logger logr.Logger, pcsg *grovecorev1alpha1.PodCliqueScalingGroup) ctrlcommon.ReconcileStepResult {
```
unmarshall, line 35: @renormalize  found the following issue (reporting this on his behalf)

Delete-expectation entry leaks on PCSG deletion. triggerDeletionFlow (reconciledelete.go:34) still only removes the finalizer and hands off to the GC — no DeleteExpectations(key). The DeleteExpectations method exists (expectations.go:100) and the standalone path uses it, but the PCSG delete path does not. The empty ControlleeExpectations struct + key persists in the indexer after member cliques are GC'd. Unbounded leak across PCSG lifecycles. Fix: call DeleteExpectations with PCSGScopedExpectationsStoreKey(pcsg.ObjectMeta) in the deletion flow.

unmarshall: This has now been fixed in 392085e84f346d2f80f46d72914fe598dda39659

operator/internal/controller/podclique/components/pod/rollingupdate.go
```diff
@@ -35,134 +36,154 @@ import (
...
 	}
 
-	// Check if there is currently a pod that is selected for update and its update has not yet completed.
-	if isAnyReadyPodSelectedForUpdate(pclq) && !isCurrentPodUpdateComplete(ss, uw) {
+	desiredNumPods := int(ss.pclq.Spec.Replicas)
+
+	// Completion is readiness-aware. End the update only when no old-hash Pods remain and the desired
+	// number of new-hash Pods are Ready, so a rollout never completes while replacements are not yet
+	// available.
+	if uw.oldHashPodCount == 0 && uw.newReadyPodCount == desiredNumPods {
```
unmarshall: @renormalize reported this issue (mentioning this on his behalf)

Standalone rollout completes only when both hold:
```go
uw.oldHashPodCount == 0 && uw.newReadyPodCount == desiredNumPods
```
It is possible that `uw.newReadyPodCount == desiredNumPods` is true while the old pods are stuck in terminating. This would mean that the update will never be marked as done and as a consequence the gang termination functionality will not be re-enabled (even though there is no pending update work). Then new versioned pods are ready and serving.

**The PCSG reconciler does it differently:**
The PCSG path already does the robust thing — computePendingUpdateWork skips replicas where allPodCliquesTerminating (podcliquescalinggroup/components/podclique/rollingupdate.go:176) and completes on numUpdatedReadyReplicas == desiredNumReplicas. A fully-terminating old replica doesn't block PCSG completion. Only the standalone path blocks on terminating pods.

This check is therefore stricter than #786 and inconsistent with PCSG, and it turns any stuck-terminating old pod into an indefinite "in-progress + gang-termination-suspended" state.

unmarshall: Addressed in 180588154c9496a591c8293499c4f508bfcf1811

operator/api/core/v1alpha1/podclique.go
unmarshall: Had a discussion with @sanjaychatterjee we will drop this field.
Addressed in eee970bbd7fe8aa98169d1b6e0576c91771c1995

operator/internal/controller/podcliquescalinggroup/reconcilestatus.go
renormalize: Thanks for the clarification!

renormalize (approved):

sanjaychatterjee (approved):

PR #819, opened 2026-09-10
fix(podclique): preserve PCSG pod indices during scale-in

### LINKED ISSUE #818: Scaling in a PodClique that belongs to a PodCliqueScalingGroup leaves a gap in PCSG pod indices
### What happened?

Scaling a PodClique that belongs to a PodCliqueScalingGroup is an operation the operator accommodates today. The PodCliqueScalingGroup controller, when it reconciles an existing member PodClique, overwrites the PodClique spec from the template but intentionally keeps the PodClique's current .Spec.Replicas instead of resetting it to the template-derived value. See [here](https://github.com/ai-dynamo/grove/blob/0be004cff387eccdd476ff3566caa9e5f0e27c55/operator/internal/controller/podcliquescalinggroup/components/podclique/podclique.go#L339-L343). This is done so that any external scaling of the PCLQ (that belongs to the PCSG) is honoured. The user guide also documents this as a supported operation (docs/user-guide/01_core-concepts/03_pcsg_intro.md).

PR #755 added a PodCliqueScalingGroup-wide pod index on top of this already-supported member scaling (the `grove.io/podcliquescalinggroup-pod-index` label and GROVE_PCSG_POD_INDEX env var). It did not update the scale-in path to keep that index correct.

On scale-in, the excess pod to delete is chosen by `selectExcessPodsToDelete` → `DeletionSorter`. That sorter picks a pod using generic criteria (terminating, unassigned, phase, readiness, template hash, and finally newest creation time). It has no notion of the pod's index. As a result it can delete a pod that is not the last one by grove.io/podclique-pod-index. The remaining pods are then no longer numbered contiguously, so the PCSG pod index (offset + podclique-pod-index) has a gap and can collide with another member's index. The indices never converge, and the test times out.

Scaling in a PodClique that belongs to a PodCliqueScalingGroup should keep the PCSG pod indices contiguous. Because member scaling is already accommodated, the scale-in pod selection should be pod-index-aware (for example, delete the highest podclique-pod-index first, StatefulSet-style) so the surviving pods retain a gap-free index set. The user guide should also note the current scale-in limitation until this is fixed.

### Environment

- Kubernetes version: any (reproduced on the KWOK-based e2e cluster)
- Grove version: main including #755 
- Scheduler details: n/a
- Cloud provider or hardware configuration: n/a
- Tools that you are using Grove together with: n/a

git-jxj:
#### What type of PR is this?

/kind bug

Scaling in a PodCliqueScalingGroup member can delete a lower-index pod because it is Pending or was recreated more recently, leaving a gap in the surviving pod indices. Prefer the highest numeric PodClique pod index during PCSG member scale-in so survivors remain in the new replica range. Standalone PodClique deletion priorities remain unchanged.

The regression test reproduces scale-in from 3 to 1 and from 12 to 10 with a newer Pending pod at index 0. Before the fix it selects pod-0 instead of the highest ordinal; both cases pass after the fix. It also verifies that sorting does not reorder the sync snapshot.

Validation: `go test ./internal/controller/podclique/components/pod -count=1` and `KUBEBUILDER_ASSETS=<local Kubernetes 1.35.0 assets> go test ./...` from `operator/` passed.

Fixes #818

This fixes pod selection for scale-in from a contiguous index set. It does not renumber already-running pods with pre-existing gaps.

```release-note
NONE
```

```docs
NONE
```


git-jxj:
Current head [4aed6ac9](https://github.com/ai-dynamo/grove/commit/4aed6ac914ddd3781f60dd8092857da183907324) merges the latest `main` (`c6e18547`) to satisfy the strict up-to-date requirement. The merge was conflict-free and preserves both the PCSG scale-in pod-index fix and the upstream expectations/component-helper migration.

Post-merge local validation passed: `make validate` (generation, formatting, API docs, lint, and proposal TOC) and `make test` (71 packages with tests passed, including the operator’s `go test -race ./...`; 32 packages have no tests). The focused PCSG scale-in/deletion-order race regression also passed. Validation used Go 1.26.6 and cached Kubernetes 1.35.0 envtest assets, with no generated-file drift.

The new head’s remote checks are starting. Required E2E validation still needs the vetted `pull-request/` copy branch. Could a PR vetter authorize **4aed6ac914ddd3781f60dd8092857da183907324** through the repository’s `/ok to test` process so the relevant E2E jobs can run? This updates the earlier request for `ad299579`; no E2E pass is claimed for the new head yet.

Ronkahn21:
@git-jxj thanks addressing the issue,
I have some concern about the original issue, that I talking with @unmarshall about 
I will update you

git-jxj:
Thanks for the update. I’ll wait for the outcome of your discussion with @unmarshall and am happy to adjust the approach if the original issue calls for different scale-in semantics.

PR #822, opened 2026-09-10
Use typed `List` calls instead of `PartialObjectMetadata` for Grove resources

### LINKED ISSUE #821: Replace `PartialObjectMetadata` list calls with typed list calls for Grove resources

A lot of components of grove-operator make use of `PartialObjectMetadataList` in the implementations of the `GetExistingResourceNames` method implementation. I'd like these list calls to be replaced with typed `List` calls.

### Why is this needed?

Separately invoking this metadata only list creates another informer, which has its own cache. Since informers are already created by controller-runtime for grove resources, it would be better to rely on a singular cache, and not deal with potential inconsistencies between both these caches.

Although rare, I have observed reconciliation runs where these two caches were not in sync, causing erroneous logs, later fixed by subsequent reconciliations.

renormalize:
#### What type of PR is this?

/kind cleanup

`GetExistingResourceNames` across several components listed resources as `metav1.PartialObjectMetadataList`. For Grove's own custom resources (`PodClique`, `PodCliqueScalingGroup`, `PodGang`, `PodGangMap`), controller-runtime already instantiates typed informers because these types are watched by the operator. Reading them via `PartialObjectMetadata` routes through a *separate* metadata informer cache, which:

- creates a duplicate cache for a GVK the operator already watches typed
- can lag behind the typed cache, increasing the likelihood of inconsistent
  reads between the two caches.

This PR replaces those metadata-only list calls with typed `client.List` calls that read from the existing typed informer cache.

Changes:

- Make `FilterMapOwnedResourceNames` generic.
- Convert the `PodClique`, `PodCliqueScalingGroup`, `PodGang`, and `PodGangMap` list sites to typed lists.
- Update affected tests.

Fixes #821

```release-note
NONE
```

```docs
NONE
```


renormalize:
/ok to test [aea197a](https://github.com/ai-dynamo/grove/pull/822/commits/aea197aabd3a08ad3d4089f6d1221f422006f00e)

yankay:
I wonder if `GetExistingResourceNames` in the HPA component https://github.com/ai-dynamo/grove/blob/main/operator/internal/controller/podcliqueset/components/hpa/hpa.go#L62 might have a similar issue, since it uses `PartialObjectMetadataList` while `CreateOrPatch` reads typed HPA objects?

renormalize:
/ok to test [4ef168f](https://github.com/ai-dynamo/grove/pull/822/commits/4ef168f00229424c2807412f0e66f38a0a297603)

Ronkahn21:
I am wonder if it impact performance, mostly around serialize and deserialize for our large objects 
we should check if it possible to achieve partialObject using the same cache

renormalize:
@Ronkahn21 there wouldn't be any additional serialization and deserialization costs when we use the typed object instead of the `PartialObjectMetadata`. The informer cache that is local to the operator already has these resources as a part of the cache; they are deserialized anyway in the cache when the event is received. When the `Get` is performed, it is returned from the cache directly.

Ronkahn21 (dismissed):

operator/internal/utils/kubernetes/objectmeta.go
```diff
@@ -20,11 +20,16 @@ import (
...
-	return lo.FilterMap(candidateResources, func(objMeta metav1.PartialObjectMetadata, _ int) (string, bool) {
-		if metav1.IsControlledBy(&objMeta, &ownerObjMeta) {
-			return objMeta.Name, true
+// FilterMapOwnedResourceNames filters the candidate typed objects and returns the names of those that are controlled by the given owner object meta.
+// The type parameter is constrained so that *T implements client.Object, allowing List results (which are slices of values) to be passed directly.
+func FilterMapOwnedResourceNames[T any, PT interface {
+	*T
+	client.Object
+}](ownerObjMeta metav1.ObjectMeta, candidateResources []T) []string {
+	return lo.FilterMap(candidateResources, func(obj T, _ int) (string, bool) {
```
unmarshall: Nit: `FilterMap` passes each element by value, copying an entire resource (T) per element just to invoke `GetName` on the type. Perhaps you can switch to using an `index-loop` instead.
So your code becomes:
```go
   func FilterMapOwnedResourceNames[T any, PT interface {
      *T
      client.Object
   }](ownerObjMeta metav1.ObjectMeta, candidateResources []T) []string {
      names := make([]string, 0, len(candidateResources))
      for i := range candidateResources {
         obj := PT(&candidateResources[i]) // pointer into the slice, no copy
         if metav1.IsControlledBy(obj, &ownerObjMeta) {
            names = append(names, obj.GetName())
         }
      }
      return names
   }
```

unmarshall: It perhaps would be even easier to read if you defined PT separately:
```go
// ObjectPtr is a constraint interface for generic helpers that receive a slice of
// API objects by value (e.g. List().Items) but need to call pointer-receiver methods
// such as GetName or DeepCopyObject.
//
// A constraint interface may contain two kinds of elements, separated by ";" (or
// newlines). This one uses both:
//
//   - *T is a type element (generics-only syntax): it restricts the type set, requiring
//     PT to be *T.
//   - client.Object is an embedded method interface: it requires PT to implement those
//     methods.
//
// The elements are AND-combined, so PT's type set is exactly {*T} when *T implements
// client.Object, and empty (uninstantiable) otherwise.   
type ObjectPtr[T any] interface {
      *T
      client.Object
}

func FilterMapOwnedResourceNames[T any, PT ObjectPtr[T]](ownerObjMeta metav1.ObjectMeta, candidateResources []T) []string {
      names := make([]string, 0, len(candidateResources))
      for i := range candidateResources {
         obj := PT(&candidateResources[i]) // pointer into the slice, no copy
         if metav1.IsControlledBy(obj, &ownerObjMeta) {
            names = append(names, obj.GetName())
         }
      }
      return names
}
```

operator/internal/controller/podcliqueset/components/podgang/podgang.go
```diff
@@ -89,7 +88,7 @@ func (r _resource) GetExistingResourceNames(ctx context.Context, logger logr.Log
 			fmt.Sprintf("Error listing PodGang for PodCliqueSet: %v", k8sutils.GetObjectKeyFromObjectMeta(pcsObjMeta)),
 		)
 	}
-	return k8sutils.FilterMapOwnedResourceNames(pcsObjMeta, objMetaList.Items), nil
+	return k8sutils.FilterMapOwnedResourceNames(pcsObjMeta, podGangList.Items), nil
```
unmarshall, line 91: I agree with #822 (comment)
If we are doing this change in PCLQ, PCSG and PG components then we can do this change in all components instead, right?

renormalize: I wanted to restrict this PR to only the components that have duplicate informers, like PCS, PCSG, PCLQ. Will change it for all components.

renormalize: Addressed in [413a4ed](https://github.com/ai-dynamo/grove/pull/822/commits/413a4edf3e64289492f04ce1cd1e2ec94cf668bd).

operator/internal/utils/kubernetes/objectmeta.go
renormalize: Addressed in [413a4ed](https://github.com/ai-dynamo/grove/pull/822/commits/413a4edf3e64289492f04ce1cd1e2ec94cf668bd).

unmarshall (approved):

renormalize:
/ok to test [391d835](https://github.com/ai-dynamo/grove/pull/822/commits/391d835569d79db4262a90bd373ecce4de094603)

renormalize:
/ok to test [c9c0b27](https://github.com/ai-dynamo/grove/pull/822/commits/c9c0b27892ef3fe313e7439872e7013966f9e65f)

sanjaychatterjee (approved):

Issue #793, opened 2026-08-25
Engine-coordinated live resize of a PCSG member PodClique


julienmancuso:
## Summary

Define the Grove contract needed for an external workload controller to resize one `PodClique` inside a specific `PodCliqueScalingGroup` replica while a distributed engine remains live.

Grove already supports scaling a PCSG-owned PodClique through its Scale subresource. The missing piece is coordination around that cardinality change: initial gang admission, topology-local joining capacity, engine-authorized removal, and failure recovery without unnecessarily recreating the complete PCSG or PCS replica.

The motivating consumer is Dynamo-managed Elastic Expert Parallelism, but the contract should remain engine-neutral.

## Resource model

For the Dynamo use case:

- one PCSG replica represents one independent engine world;
- `PCSG.spec.replicas` controls the number of worlds;
- selected member PodCliques provide the physical replica allocations within each world;
- an external Dynamo Engine Group controller owns the engine resize transaction;
- Grove owns pod allocation, scheduling, topology placement, pod identity, and deletion.

Engine-active membership is deliberately not Grove state. Dynamo records desired, allocated, available, and engine-active replicas separately and calls the engine lifecycle API.

One logical replica allocation may contain one or more pods. Grove need not understand its engine meaning, but it must preserve the allocation boundary for admission, availability, and release.

## Why ordinary replica scaling is insufficient

Changing `PodClique.spec.replicas` expresses only a target cardinality. A live distributed engine also needs the following ordering.

### Scale up

1. Allocate the requested additional replica allocations.
2. Admit each complete joining allocation with the required gang and topology guarantees.
3. Report the exact allocated pod identities and readiness of each complete allocation.
4. Let the external controller start or prepare the new ranks and commit the engine topology.
5. Only then expose the ranks through the serving runtime.

### Scale down

1. The external controller selects and drains the departing ranks.
2. The engine commits a topology that no longer contains those ranks.
3. The external controller authorizes deletion of the corresponding exact replica allocations.
4. Grove deletes only the authorized pods, without selecting different victims or racing to replace them.

### Unexpected failure

1. Report the exact failed pod/rank capacity.
2. Allow the external controller to commit a survivor topology.
3. Avoid automatically recycling the whole world while this recovery is progressing.
4. Allocate and start replacement capacity through the backend-required restoration protocol, then expose enough state for the external controller to observe restoration of the missing membership.

## Concrete scenarios

[...]

PR #823, opened 2026-09-10
GREP-0793: Capacity Identity and Authorized Release

tzulingk:
## What this is

A GREP answering the **identity and release half** of #793, filed as a draft to get the shape agreed before it goes further.

#793 enumerates six required semantics. This proposal argues that one of them is a prerequisite for the other five and should be settled first: an external controller cannot express a survivor topology, a staged fleet operation, or a targeted recovery until it can *name* a piece of physical capacity and hand that name back to Grove as an instruction. The other five are listed as explicit Non-Goals with pointers to the issues they belong with, rather than folded in here.

The proposal is two additive, opt-in contracts:

- **Capacity slots** — a durable logical position that survives pod replacement, formalizing the pod-index labels Grove already assigns.
- **Authorized release** — an external controller nominates the exact pods a scale-in may delete, bound to immutable pod UIDs. Grove refuses stale authorizations and waits rather than substituting a different victim.

Default behavior is unchanged: `releasePolicy` defaults to `Heuristic`, which is today's `DeletionSorter` path.

## Two findings from reading the current code

These motivate the specifics, and both are worth confirming independently of whether this proposal's direction is accepted.

**1. `DeletionSorter` prefers the newest pod.** Rule 5 in [`deletionsort.go`](https://github.com/ai-dynamo/grove/blob/main/operator/internal/controller/podclique/components/pod/deletionsort.go) sorts by `CreationTimestamp.After`, so scale-in prefers the most recently created pod. For an elastic engine the newest member is typically the one that most recently finished joining the topology — so the current default is close to the worst available choice when the engine has drained a specific, older member.

**2. The PCSG-wide pod index is not stable under live inner resize.** `getPCSGPodIndexOffset` sums the *live* `spec.replicas` of preceding member cliques, so resizing an earlier clique shifts every later clique's offset, and `syncPCSGPodIndexLabels` then patches the `grove.io/podcliquescalinggroup-pod-index` label on pods that never moved.

The second one has a sharper consequence than renumbering. `GROVE_PCSG_POD_INDEX` is a downward-API `fieldRef` reading that same label, and env-var `fieldRef`s resolve at container start rather than tracking later patches. So after a sibling resize the running container and its own pod label report *different* indices, silently, until the container restarts.

This matters directly for #793, which nominates the PCSG-wide pod index as "a promising basis" for capacity identity — it is, but not in its current form. There's an e2e test in the Test Plan that should fail against `main` today.

## Open questions for maintainers

1. **Index stability fix** — freeze the offset at PCSG replica creation, or stop flattening and make the slot a `(cliqueName, podIndex)` pair? The GREP states the required property and deliberately leaves the mechanism to Grove.
2. **Authorization transport** — a status-acknowledged `spec` field is proposed as least invasive; a dedicated `PodCliqueReleaseAuthorization` resource is the alternative if authorization should have independent RBAC. #793 says the transport is Grove's call, so this is the main thing to push back on.
3. **Rolling updates** — the GREP takes the conservative line that `Authorized` release requires the `OnDelete` update strategy, since the update path also deletes pods. Is that too restrictive?
4. **Scope** — is splitting #793 this way the right call, or would maintainers rather see one proposal covering all six semantics?

## Notes

- `make verify-toc` passes.
- No code changes; documentation only.

cc @julienmancuso (author of #793), @unmarshall, @shayasoolin, @sanjaychatterjee


tzulingk:
@sttts @oleg-kushniriov — tagging you because this GREP proposes changing the PCSG-wide pod index you built and reviewed in #755, and because my first draft got that history wrong. Still a draft; I'd rather correct it before asking for real review.

What the first draft got wrong, and what `4b415a7` changes:

**It ranked a rejected design first.** The draft listed "freeze the offset at PCSG replica creation" as the preferred fix, with the collision risk as a trailing caveat. That is the design #755 originally shipped, and @oleg-kushniriov rejected it in review for giving new pods indices that duplicate the next clique's range — after which the offset became the live-sum computation that merged. Proposing it back without citing that exchange was careless. It is now written up as already-tried, with the review link, and the pair `(cliqueName, podIndex)` scheme is ranked first instead.

I also worked through a reserved-stride variant (offsets derived from a declared per-clique capacity rather than live counts) and it does not hold up, so it is documented and declined rather than proposed: there is no per-clique capacity field reachable from a PCSG today — the PCS webhook rejects `AutoScalingConfig` on scaling-group members — nothing would enforce the ceiling, so overflow is silent duplication rather than a validation error, and because the offset annotation is patched unconditionally with no feature gate, it would renumber every existing multi-clique PCSG at operator upgrade.

**It claimed existing consumers were "already broken."** That was false. The true position is stronger: `0be004c` landed after `v0.1.0-alpha.13`, `git tag --contains` is empty, so nothing reading this label has shipped. The scheme can be settled as a pre-release correction with no migration story.

**It framed a documented trade as an oversight.** Contiguity in `cliqueNames` order is stated in the user guide along with the remedy — restart affected containers. That remedy is exactly what a live resize cannot do, which is the actual ask here, and the GREP now says so rather than implying something was missed.

Two things I found while checking, offered as observations rather than asks:

1. **Duplicate indices are reachable on `main` without any resize.** Clique `a` has `replicas: 3` with pods at `{0,1,2}`; the pod at index 1 goes NotReady; `a` scales to 2; `DeletionSorter` has no index criterion and deletes index 1; survivors keep `{0,2}` while `offset(b)` recomputes to 2 — so `a`'s survivor and `b`'s first pod both carry `podcliquescalinggroup-pod-index=2`. Happy to open a separate issue if that is worth tracking on its own.

2. **A `spec` field on a generated PodClique would not survive.** Both PodClique reconcilers rebuild the child spec from the PodCliqueSet template each sync and preserve only `spec.replicas`, and under `OnDelete` — which this proposal mandates — that path always runs. That likely sinks the spec-field transport the draft proposed, so the GREP now leans toward a separate resource and defers the choice to you.

Also corrected: `observedGeneration` was misusing a reserved convention and is now `topologyGeneration`; added the `minAvailable` interaction, which arms gang termination of the whole PCSG replica after an authorized release and means this is not independently shippable; added the `PerReplica` ResourceClaim assumption that the occupied index set equals `[0, spec.replicas)`; and listed the existing unit and e2e tests this proposal inverts rather than fixes, so a red `upgrade_test` reads as the intended signal.

One question that decides whether this proposal is viable at all, raised in [Dependencies](https://github.com/ai-dynamo/grove/pull/823/files): GREP-0677 (#686) states that a PodClique owned by a PodCliqueScalingGroup must not be scaled independently. This GREP assumes the opposite for the non-zero case — resizing a member while the group stays at a fixed positive `replicas`. #793 asks the same question directly. If #686's rule is adopted as written, it forecloses the mechanism here, so I have listed resolving it as a prerequisite rather than an adjacency.

PR #827, opened 2026-09-15
fix(kai): resolve externally-managed topology name before scheduling

### LINKED ISSUE #828: kai: externally-managed topology name not resolved before scheduling
## Summary

When a `ClusterTopologyBinding` declares an externally-managed KAI topology via `Spec.SchedulerTopologyBindings[]` (i.e. its `TopologyReference` names a KAI `Topology` resource that differs from the `ClusterTopologyBinding`'s own `metadata.name`), the KAI scheduler backend sends the `ClusterTopologyBinding`'s own name to KAI instead of the resolved `TopologyReference`. KAI then looks for a `Topology` object under that (nonexistent) name and rejects the `PodGroup`.

## Details

`SchedulerTopologyBindings[].TopologyReference` (`operator/api/core/v1alpha1/clustertopologybinding.go`) exists specifically to map a `ClusterTopologyBinding` to an externally-managed backend topology resource whose name differs from the binding's own name. It's already consulted for drift-checking (`operator/internal/controller/clustertopology/reconciler.go` → `CheckTopologyDrift`), but was never consulted when building the actual scheduling request.

The KAI `PodGroup`-construction path (`operator/internal/controller/podcliqueset/components/podgang/podgang.go` → `operator/internal/scheduler/kai/backend.go`) took the `ClusterTopologyBinding`'s own name straight off the `PodGang`'s `grove.io/topology-name` annotation and wrote it directly as `kaischedulingv2alpha2.TopologyConstraint.Topology`, with no lookup against `SchedulerTopologyBindings`.

## Impact

Native topology scheduling cannot be configured end-to-end for any `ClusterTopologyBinding` whose KAI topology is externally managed under a different name than the binding itself.

## Fix

See #827, which adds a resolution step to the KAI backend's `PodGroup`-building path: it looks up the `ClusterTopologyBinding` referenced by the `PodGang`'s topology-name annotation and, if it declares an externally-managed `SchedulerTopologyBinding` for `kai-scheduler`, uses that entry's `TopologyReference` instead of the binding's own name.

estroz:
Fixes #828

## Summary

When a `ClusterTopologyBinding` declares an externally-managed KAI topology via `Spec.SchedulerTopologyBindings[]` (i.e. its `TopologyReference` names a KAI `Topology` resource that differs from the `ClusterTopologyBinding`'s own `metadata.name`), the KAI scheduler backend still sent the `ClusterTopologyBinding`'s own name to KAI when building the `PodGroup`'s `TopologyConstraint.Topology`. KAI then looked for a `Topology` object under that (nonexistent) name and rejected the `PodGroup`.

`TopologyReference` was already consulted for drift-checking (`CheckTopologyDrift`), but never when constructing the actual scheduling request sent to KAI.

## Fix

Add `resolveTopologyName` to the KAI scheduler backend (`operator/internal/scheduler/kai/backend.go`). Given the `ClusterTopologyBinding` name recorded on a `PodGang`'s topology-name annotation, it looks up the binding and, if it declares an externally-managed `SchedulerTopologyBinding` for `kai-scheduler`, uses that entry's `TopologyReference` instead. It falls back to the binding's own name when no such entry exists or the binding can't be found, preserving existing behavior for Grove-managed topologies.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full `operator` module)
- [x] Added `TestBackend_ResolveTopologyName` covering: empty binding name, externally-managed binding resolving to `TopologyReference`, Grove-managed binding falling back to its own name, binding with no entry for `kai-scheduler`, and a binding that doesn't exist
- [x] Added `TestBackend_SyncPodGang_ResolvesExternallyManagedTopologyName` covering the full `SyncPodGang` path, asserting the resulting KAI `PodGroup.Spec.TopologyConstraint.Topology` equals the resolved `TopologyReference`, not the binding's own name


operator/internal/scheduler/kai/backend.go
```diff
@@ -256,6 +260,31 @@ func getTopologyName(podGang *groveschedulerv1alpha1.PodGang) string {
...
+// KAI Topology resource may be named differently from the ClusterTopologyBinding itself. Otherwise, or
+// if the ClusterTopologyBinding cannot be found, bindingName is returned unchanged.
+func (b *schedulerBackend) resolveTopologyName(ctx context.Context, bindingName string) (string, error) {
+	if bindingName == "" {
+		return "", nil
+	}
+	ct := &grovecorev1alpha1.ClusterTopologyBinding{}
+	if err := b.client.Get(ctx, client.ObjectKey{Name: bindingName}, ct); err != nil {
+		if apierrors.IsNotFound(err) {
+			return bindingName, nil
```
sanjaychatterjee: Is it correct to silently ignore the missing ClusterTopologyBinding resource? cc @enoodle

sanjaychatterjee (commented):
Looks straightforward. @enoodle can you please review?

operator/internal/scheduler/kai/backend.go
enoodle: This should not pass validation (https://github.com/ai-dynamo/grove/blob/main/operator/internal/webhook/admission/pcs/validation/podcliqueset.go#L424-L434) but theoretically someone could have deleted the ClusterTopologyBinding between then and now. I think it should reject creating the resource so it will not schedule.

enoodle (commented):
Can you update / add a new test to operator/e2e/tests/topology_test.go for this case please?

operator/internal/scheduler/kai/backend.go
estroz: Good catch, this should fail loudly

estroz: Fixed

estroz:
Pushed updates addressing both review comments:

- `resolveTopologyName` now fails loudly (returns an error) instead of silently falling back when the `ClusterTopologyBinding` can't be found — the `PodGroup` won't be created/synced in that case.
- Added `Test_TAS24_ExternallyManagedTopologyNameResolution` in `operator/e2e/tests/topology_test.go`: creates an externally-managed KAI `Topology` under a different name than its `ClusterTopologyBinding`, deploys a workload referencing the binding, and verifies pods schedule successfully and the resulting KAI `PodGroup`'s `TopologyConstraint` carries the resolved `TopologyReference` rather than the binding's own name.

sanjaychatterjee (approved):

operator/internal/scheduler/kai/backend.go
```diff
@@ -256,6 +260,30 @@ func getTopologyName(podGang *groveschedulerv1alpha1.PodGang) string {
...
+// topology-name annotation. When the binding declares an externally-managed SchedulerTopologyBinding
+// for this backend, that binding's TopologyReference is returned instead, since the externally-managed
+// KAI Topology resource may be named differently from the ClusterTopologyBinding itself. Otherwise
+// bindingName is returned unchanged. The admission webhook requires the referenced ClusterTopologyBinding
+// to exist, but it could have been deleted since admission; treat that as an error rather than silently
+// scheduling against an unresolved (and possibly wrong) topology name.
+func (b *schedulerBackend) resolveTopologyName(ctx context.Context, bindingName string) (string, error) {
+	if bindingName == "" {
+		return "", nil
+	}
```
Ronkahn21, line 274: Its seem like it should consider an error either way can you move this check outside of this function

sanjaychatterjee: Wouldn't returning the error prevent the KAI PodGroup creation? If `bindingName` is empty, then topology is essentially ignored for scheduling and that would be expected behavior, right?

estroz: The new method doesn't change behavior. An empty `topologyName` is handled downstream by `toKAITopologyConstraint`

Ronkahn21 (approved):

enoodle (approved):

PR #831, opened 2026-09-17
Coherent update engine

### LINKED ISSUE #807: Core coherent update engine for PodCliqueSet

unmarshall:
#### What type of PR is this?

/kind feature
/kind api

Implements the core coherent update engine for `PodCliqueSet`. This PR builds on #778, #792, and #799.

**API Changes**

 - Makes Coherent the default update strategy. A PodCliqueSet that does not set spec.updateStrategy.type now defaults to Coherent (previously RollingRecreate).
 - Adds `MaxUnavailable` as an update configuration handle.
 - Adds coherent update progress status fields:
     - `PodCliqueSet.status.updateProgress.inScopeStandalonePodCliques`
     - `PodCliqueSet.status.updateProgress.inScopePodCliqueScalingGroups`
     - `PodCliqueSet.status.updateProgress.currentlyUpdating[].inFlightEpochs`
     - `PodCliqueSet.status.updateProgress.currentlyUpdating[].message`
     - `PodClique.status.updateProgress.updatedReadyReplicas`
- Drops the internal `AnchorIndex` field from `PodGangEntry`. Anchor order and the base anchor are derived from the globally monotonic epoch.

**Coherent update engine (three layers):**

1. Orchestrator (PodCliqueSetReplica component) — drives the coherent update as a multi-step process, one PCS replica at a time, and records progress.
 2. Core engine (PodGangMap component) — computes the current update state and advances the update by authoring the next sub-step, then re-converges the map to a single generation as in-scope content drains, so a subset update (only some components changed) does not leave it split across generations.
 3. Executor (PodClique / PodCliqueScalingGroup components) — recreates standalone PodClique pods and PCSG members onto the PodGang the PodGangMap has committed for the current generation.

**Scale guard:**
 - PCLQ and PCSG validating webhooks reject scaling operations while a coherent update is in progress on the owning PCS.
> NOTE: In future this constraint will be removed (in iterations).

Fixes #807 

- End-to-end coverage lives in operator/e2e/tests/update/coherentupdate_test.go (Test_CU1–Test_CU8), covering bootstrap layout, standalone/PCSG/subset updates, leftover steps, back-to-back updates, PCS-wide scale blocking across replicas, and operator-restart resume. The full suite was validated live on a k3d + KWOK cluster.
 - The AnchorIndex removal is internal to the PodGangMap (present in no resource name — names key on epoch), so it renames no resources.

```release-note
The `Coherent` update strategy is now available for `PodCliqueSet` and is now the default update strategy.  A PodCliqueSet that does not set `spec.updateStrategy.type` now rolls updates coherently (a Minimum Viable Unit at a time, one PCS replica at a time) instead of RollingRecreate. Set `spec.updateStrategy.type: RollingRecreate` explicitly to keep the previous behavior.
Adds a `MaxUnavailable` update configuration handle and new coherent progress status fields (`PodCliqueSet.status.updateProgress.inScopeStandalonePodCliques`,
`inScopePodCliqueScalingGroups`, `currentlyUpdating[].inFlightEpochs`, `currentlyUpdating[].message`, and `PodClique.status.updateProgress.updatedReadyReplicas`). Scaling of a `PodClique` or `PodCliqueScalingGroup` is rejected while a coherent update is in progress on the owning `PodCliqueSet`.
```

```docs
GREP-393 Coherent Rolling Updates, updated in this PR to reflect the implementation: docs/proposals/393-coherent-rolling-updates/README.md
```


operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
```diff
@@ -0,0 +1,493 @@
...
+// entry, so spreading them over floor(replicas/MinAvailable) anchors would create one PodGang per pod. With
+// PodCliqueScalingGroups in the MVU, the anchors are the proportional MVUs, bound by the component supplying
+// the fewest MinAvailable-sized slices, the min over components of floor(replicas/MinAvailable).
+func computeNumAnchorBearingSteps(liveReplicas map[string]int32, mvu *mvuTemplate) int32 {
+	if len(mvu.pcsgs) == 0 {
+		return 1
+	}
+	numAnchorBearingSteps := int32(math.MaxInt32)
+	for componentName, minAvailable := range lo.Assign(mvu.standalonePCLQs, mvu.pcsgs) {
+		numAnchorBearingSteps = min(numAnchorBearingSteps, liveReplicas[componentName]/minAvailable)
```
oleg-kushniriov, line 153: The defaulting webhook sets a standalone clique's `MinAvailable` to `Replicas` when unset (`pcs/defaulting/podcliqueset.go:95`). Here that gives `floor(R / R) = 1`, so any unconfigured standalone clique forces `numAnchorBearingSteps = 1` regardless of the scaling groups.

With the usual layout (standalone frontend, prefill/decode as PCSGs) and no explicit `minAvailable` on the frontend, the first MPG sub-step recreates every frontend pod at once and the old version has no frontend for the rest of the update. Since `Coherent` is now the default strategy, this is the out-of-the-box behaviour.

GREP-393 never mentions this default; every worked example sets `minAvailable`
to 1 or 2 explicitly.

Suggestion: do not let a standalone clique's `MinAvailable` bound the step count (derive its step unit differently, or take the step count from PCSGs and roll standalone cliques proportionally), or give `Coherent` a different default. Either way add a planner test for the default-config case.

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
```diff
@@ -0,0 +1,493 @@
...
+
+	// Beyond the MinAvailable that every anchor-bearing step reserves, a component's remaining replicas
+	// are its tail. tailPerStep spreads that tail evenly over the anchor-bearing steps, so each step
+	// rolls MinAvailable plus tailPerStep of the component, which is its anchorBearingStepTarget. Integer
+	// division rounds tailPerStep down, so leftover is what remains once every anchor-bearing step has
+	// rolled its target, drained afterward by the single leftover step.
+	anchorBearingStepTarget := make(map[string]int32, len(minAvailableByComponent))
+	leftover := make(map[string]int32, len(minAvailableByComponent))
+	for componentName, componentMinAvailable := range minAvailableByComponent {
+		tailPerStep := (liveReplicas[componentName] - numAnchorBearingSteps*componentMinAvailable) / numAnchorBearingSteps
```
oleg-kushniriov, line 172: `numAnchorBearingSteps` is `min(floor(R / M))` and can be 0 whenever an in-scope component has
`R < M`. `liveReplicas` is built only from children that exist (`coherentupdate.go:44`), so a missing child gives `R = 0`.


Suggestion: guard `numAnchorBearingSteps == 0`, surface a planner error in `CurrentlyUpdating[].Message`, and add a test with a component absent from `liveReplicas`.

operator/internal/controller/podcliqueset/components/podclique/podclique.go
```diff
@@ -290,80 +290,108 @@ func (r _resource) doCreateOrUpdate(ctx context.Context, logger logr.Logger, pcs
...
-	var dependentPclqNames []string
-	if dependentPclqNames, err = identifyFullyQualifiedStartupDependencyNames(pcs, pclq, pcsReplica, foundAtIndex); err != nil {
+
+	dependentPCLQNames, err := identifyFullyQualifiedStartupDependencyNames(pcs, pcsReplica, cliqueName, pclq)
+
+	if err != nil {
 		return err
 	}
-	pclq.Spec.StartsAfter = dependentPclqNames
+	pclq.Spec.StartsAfter = dependentPCLQNames
```
oleg-kushniriov, line 381: On `main` the spec was always reset from the template before this line, so `StartsAfter` held
clique names. With `preserveRevision` the existing spec is kept and `StartsAfter` already holds FQNs from the previous reconcile. `getExplicitStartupDependencies` then feeds those FQNs to `GenerateDependencyNamesForBasePodGang`, which prefixes again: `pcs-1-a` becomes `pcs-1-pcs-1-a`, then `pcs-1-pcs-1-pcs-1-a`, persisted by `CreateOrPatch`.

Any pod recreated on a not-under-update or already-updated replica during the update gets an
init container waiting on a PodClique that does not exist. `InOrder` is unaffected.

Suggestion: compute explicit dependencies from the template's `StartsAfter`, not the live
object's. Add a test that reconciles a preserved replica twice under `Explicit` and asserts
`StartsAfter` is stable.

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
```diff
@@ -0,0 +1,493 @@
...
+	anchorPCSGIndices := make(map[string][]int32, len(p.mvu.pcsgs))
+	for pcsgName, minAvailable := range p.mvu.pcsgs {
+		anchorPCSGIndices[pcsgName] = lo.RangeFrom(stepIndex*p.plan.anchorBearingStepTarget[pcsgName], int(minAvailable))
+	}
+	return &subStep{
+		epoch:                     newEpoch(p.clk),
+		dependsOn:                 dependsOn,
+		opensAnchor:               true,
+		anchorPCSGReplicaIndices:  anchorPCSGIndices,
+		drainStandalonePCLQCounts: p.mvu.standalonePCLQs,
```
oleg-kushniriov, line 405: The MPG sub-step drains `MinAvailable` of every component unconditionally. `p.mvu.standalonePCLQs` maps each standalone clique to its `MinAvailable` (built by`computeMVUTemplate` via `CoherentMinAvailableByComponent`), and `anchorPCSGIndices` is `RangeFrom(..., minAvailable)` per scaling group from `p.mvu.pcsgs`. The component that bounds the step count has `StepTarget == MinAvailable`, so its last old-version unit goes in the sub-step that *opens* the last step, while the others still have that step's tail plus `leftover` on the old version.

Example:
consider following : (`R` replicas, `M` `MinAvailable`, `U` `MaxUnavailable`, per component; `U` defaults to `M` under `Coherent`)
Trace for `R = 4/50/70`, `M = U = 1/3/4` : after 3 steps old is
`1/14/19`; opening step 4 drains `1/3/4` and old becomes `0/11/15`. From there through `0/8/11`,
`0/5/7`, `0/2/3`, `0/2/2` and the leftover sub-step the old version has no component 1. Every
transition is readiness-gated, so a slow tail keeps that unusable allocation alive indefinitely.

This follows GREP-393 (trailing TPG-only step, Case B), so the design needs amending as well.

Suggestion: make the last step a non-anchor step and drain to zero together. Rule per sub-step,
with `remaining[c]` the old-version count: while any `remaining[c] > U[c]`, drain `min(U[c],
remaining[c] - U[c])` of those components only; once every `remaining[c] <= U[c]`, one final
sub-step drains all of them to zero. The bounding component has `remaining == M <= U`, so it
waits for the final sub-step without a special case, and ties are covered. The last step then
opens no MPG: standalone pods subsume into the previous anchor and PCSG indices get tail gangs,
which is what `buildNonAnchorSubStep` already does for the leftover step, so `leftover` folds
into the last step and the trailing step goes away. Trace: `1/14/19` -> `1/11/15` -> `1/8/11` ->
`1/5/7` -> `1/2/3` -> `0/0/0`. Add a per-sub-step invariant test: for every component `old > 0`
unless every component's `old == 0`.

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+}
+
+// computeNumAnchorBearingSteps returns how many anchor-bearing steps a coherent update takes for one PCS
+// replica. A standalone-only MVU uses a single anchor because standalone pods can live only on an anchor
+// entry, so spreading them over floor(replicas/MinAvailable) anchors would create one PodGang per pod. With
+// PodCliqueScalingGroups in the MVU, the anchors are the proportional MVUs, bound by the component supplying
+// the fewest MinAvailable-sized slices, the min over components of floor(replicas/MinAvailable).
+func computeNumAnchorBearingSteps(liveReplicas map[string]int32, mvu *mvuTemplate) int32 {
+	if len(mvu.pcsgs) == 0 {
```
oleg-kushniriov, line 148: Returning 1 here makes the per-step target the whole replica count, so every component drains at its own `MaxUnavailable` pace. For standalone `4/50/70` with caps `1/3/4` the second sub-step already moves component 1 to 2 (requirement: hold at 1 until the others reach `6/8`), and after four sub-steps old is `0/38/54`.

The accepted GREP-393 on `main` specifies the opposite for exactly this case (Case E: `replicas
10, minAvailable 1` gives 10 anchor-bearing steps, "each is itself an anchor"). This PR's rewrite of the GREP removes Case E and adds "the plan collapses to a single anchor-bearing step", so the design change is made in the same PR as the code and is not called out in the description. The PodGang-count argument is about how many anchor entries exist, not about pacing.

Suggestion: keep the step plan uniform across component kinds. If one physical anchor is preferred for standalone-only scope, separate that from logical step pacing and subsume each step's pods into the single anchor. Call the design change out explicitly if it is intended. Test the unequal-replica scenario with and without a PCSG.

operator/internal/controller/podcliquescalinggroup/components/podclique/rollingupdate.go
```diff
@@ -104,7 +104,7 @@ func (r _resource) processPendingUpdates(ctx context.Context, logger logr.Logger
 	// bounded by this budget, so when a member PodClique's readiness status is stale (numReadyReplicas
 	// under-reported) the budget shrinks and the rollout waits, instead of replacing replicas that may
 	// actually be Ready.
-	effectiveMaxUnavailable := componentutils.EffectiveMaxUnavailable(rollingUpdateConfigForPCSG(sc))
+	effectiveMaxUnavailable := componentutils.EffectiveMaxUnavailable(rollingUpdateConfigForPCSG(sc), sc.pcs.Spec.UpdateStrategy.Type, *sc.pcsg.Spec.MinAvailable)
```
oleg-kushniriov: `main`'s defaulting webhook never populated `UpdateStrategy`, and the CRD default sits on `type`
inside an optional pointer, so existing objects carry `updateStrategy: nil` until re-admitted. `IsRollingUpdateStrategy` treats nil as RollingRecreate, so an in-flight RollingRecreate update lands here and panics on every reconcile of the updating replica until the user edits the PCS.
That re-admission then silently flips the object to `Coherent` (`pcs/defaulting/podcliqueset.go:60`), which is a different inconsistency: the defaulter maps nil to `Coherent`, the validator and the operator map nil to RollingRecreate.

Same deref in `podcliquescalinggroup/components/podclique/rollingupdate.go:107`.

Suggestion: one nil-tolerant strategy resolver shared by defaulter, validator and operator; decide and document what nil means for existing objects. The upgrade e2e dropped the PCSG from its workload in this PR, which removes coverage exactly here.

operator/internal/controller/podcliqueset/components/podcliquesetreplica/rollingupdate.go
```diff
@@ -208,52 +198,53 @@ func orderPCSReplicaInfo(pcs *grovecorev1alpha1.PodCliqueSet, minAvailableBreach
...
-// computeUpdateProgress calculates update completion for a PCS replica.
-func (pri *pcsReplicaInfo) computeUpdateProgress(pcs *grovecorev1alpha1.PodCliqueSet) {
-	updatedPCLQs := 0
-	for _, pclq := range pri.pclqs {
-		if isPCLQUpdateComplete(pcs, &pclq) {
-			updatedPCLQs++
+// isUpdateComplete reports whether every expected PodClique and PodCliqueScalingGroup of the replica has
+// converged to the current generation hash. A missing PodClique keeps the replica incomplete because the
+// count of converged PodCliques falls short of the expected count.
+func (pri *pcsReplicaInfo) isUpdateComplete(pcs *grovecorev1alpha1.PodCliqueSet) bool {
```
oleg-kushniriov, line 230: Consider following scenario: a coherent update is rolling replica 0. Replica 1 is still waiting on the old revision. A child on replica 1 has been below `MinAvailable` for longer than `TerminationDelay` (unrelated cause), so gang termination fires for replica 1. Suppression only covers children that are being updated, so replica 1 is not protected.

What happens next: replica 1's PodCliques are deleted and recreated. Recreated children get the new template (`preserveRevision` only applies to objects that already exist), so replica 1 ends up on the new revision without the coherent engine ever running for it. `isUpdateComplete` sees all children at the current hash and the orchestrator never selects replica 1.

Result: replica 1's PodGangMap keeps all entries at the old generation hash for good. Nothing advances it (`shouldAdvanceEntriesGenerationHash` returns false under `Coherent`). After the update, scale-out of that replica's scaling group fails with "no PodGangMap entry owns replica index ... and no ScaleOut entry exists", standalone scale-out is never reflected in the map, and the next update starts from a two-generation map.

A user deleting all PodCliques of a not-yet-updated replica during the update produces the same
result.

Suggestion: either suspend gang termination for every replica of the PCS while a coherent update is in progress, or (preferred, covers the manual case too) make replica completion require the PodGangMap to be single-generation at the current hash, so a rebuilt replica is still selected and its map is reconverged.

operator/internal/controller/podcliquescalinggroup/components/podclique/coherentupdate.go
```diff
@@ -0,0 +1,106 @@
...
+// replicaIndicesNotOnCommittedPodGang returns the replica indices whose member PodCliques are not on the
+// PodGang the PodGangMap has committed the replica to. A replica with no members is skipped, since
+// createExpectedPCLQs creates it directly on its committed PodGang.
+func replicaIndicesNotOnCommittedPodGang(ss *syncSnapshot) ([]string, error) {
+	rnr := apicommon.ResourceNameReplica{Name: ss.pcs.Name, Replica: ss.pcsReplicaIndex}
+	membersByReplicaIndex := componentutils.GroupPCLQsByPCSGReplicaIndex(ss.existingPCLQs)
+
+	var replicaIndicesToRecreate []string
+	for pcsgReplicaIndex := range int(ss.pcsg.Spec.Replicas) {
+		members := membersByReplicaIndex[strconv.Itoa(pcsgReplicaIndex)]
```
oleg-kushniriov, line 68: `members` is not filtered for `DeletionTimestamp`, and `createExpectedPCLQs` runs earlier in the same sync. Move replica 2 (members A, B): `DeleteAllOf` removes both; A finishes fast, B lingers with a long grace period and the old `grove.io/podgang` label. Next reconcile A is recreated on the committed gang, then this check sees `[A (new), B (terminating, old)]`, `allMembersOnPodGang` is false, and A is deleted again. Repeats until B is gone; the sub-step cannot become Ready meanwhile.

Suggestion: skip a replica while any member has a `DeletionTimestamp` (as `allPodCliquesTerminating` does in the RollingRecreate path), and do not recreate members of a replica that is still draining.

operator/charts/templates/pclq-validating-webhook-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+    clientConfig:
+      {{- if .Values.webhooks.caBundle }}
+      caBundle: {{ .Values.webhooks.caBundle }}
+      {{- end }}
+      service:
+        name: {{ required ".Values.service.name is required" .Values.service.name }}
+        namespace: {{ .Release.Namespace }}
+        path: /webhooks/validate-podclique
+        port: {{ required ".Values.config.server.webhooks.port" .Values.config.server.webhooks.port }}
+    failurePolicy: Fail
```
oleg-kushniriov, line 25: PodClique and PCSG `UPDATE`s are issued by the operator on every reconcile, including finalizer removal on delete. During `helm upgrade` the new webhook config lands before the new operator pod serves the path; the old pod returns 404 and every PCLQ/PCSG update, including a `kubectl delete pcs` cascade, is denied until the new pod is ready.

Suggestion: `failurePolicy: Ignore` for these two (the guard is advisory and a missed scale in that window is recoverable), or scope the rules to the scale subresource and exclude the operator service account via `matchConditions`.

operator/internal/controller/podcliqueset/components/podgangmap/coherentupdate.go
```diff
@@ -0,0 +1,233 @@
...
+// count for the current update, its live replicas minus MaxUnavailable. A standalone PodClique is measured
+// by its Ready Pods and a PodCliqueScalingGroup by its available replicas.
+func maxUnavailableBudgetSatisfied(standalonePCLQByComponent map[string]grovecorev1alpha1.PodClique, pcsgByComponent map[string]grovecorev1alpha1.PodCliqueScalingGroup, liveReplicas, maxUnavailableByComponent map[string]int32) bool {
+	for componentName, pclq := range standalonePCLQByComponent {
+		if pclq.Status.ReadyReplicas < liveReplicas[componentName]-maxUnavailableByComponent[componentName] {
+			return false
+		}
+	}
+	for componentName, pcsg := range pcsgByComponent {
+		if pcsg.Status.AvailableReplicas < liveReplicas[componentName]-maxUnavailableByComponent[componentName] {
```
oleg-kushniriov, line 228: The check is `available >= R - U` (`R` = `spec.replicas`, `U` = the component's `rollingUpdate.maxUnavailable`, `M` below = its `minAvailable`; `U` defaults to `M` under `Coherent`) and does not subtract what the next sub-step will drain. PCSG `R = 6`, `M = U = 2`, indices 4 and 5 unavailable for unrelated reasons: available is 4, the gate passes, the MPG sub-step drains healthy 0 and 1, and unavailability is 4 against a declared maximum of 2. Same after any completed batch when unrelated replicas fail. The API comment says "at any moment";
Deployments compute allowed deletions as `available - (R - U)`.

GREP-393 predicate 3 is written the same way, so the proposal needs the fix too.

Suggestion: `headroom = available - (R - U)`. Hold an MPG sub-step until `headroom >= M` for every component (the MPG is atomic); bound tail sub-steps by `min(U, headroom)`. Add a gate test with unrelated unavailability at the threshold.

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: That is a good point. If someone does not explicitly specify `MinAvailable` for a standalone PCLQ then it is currently defaulted to `Replicas`. This would mean that if there is an incompatible update done, then all `FrontEnd` pods will be updated to the newer version whereas the decode/prefill PCSG might get rolled in steps.

>Suggestion: do not let a standalone clique's MinAvailable bound the step count (derive its step unit differently, or take the step count from PCSGs and roll standalone cliques proportionally), or give Coherent a different default. Either way add a planner test for the default-config case.

This is not accurate. `MinAvailable` defines the number of pods that must be present/scheduled as part of a single PodGang (anchor/MPG). This is the entire definition of an anchor/MPG PodGang. During coherent updates there can be one or more anchor PodGangs. Each anchor **must** carry minAvailable replicas of each component. To know how many such anchor entries needs to be created as part of the update, you use minAvailable to naturally compute that number and that is quite accurate.

> GREP-393 never mentions this default

This default has been there for a very long time (much before coherent was envisioned). So GREP-393 should not state a fact that has been well established. It should state it if the default gets changed.

So as i see it - the main issue is what should be an acceptable default value of `MinAvailable` for standalone PCLQs? Should we revisit it? If we change the default the impact is: `Gang Scheduling`, `Gang Termination`. From my point of view we can revisit `MinAvailable` default and bring it line with that of PCSG (that is 1).

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: Nice find!

> Nit: So the term `liveReplicas` is badly named. If you see how its constructed it actually reads `spec.replicas`.  So now its renamed to `desiredReplicas`.

Given that scaling is blocked for all components across all PCS replicas this is stable.

**Issue-1**
While computing `desiredReplicas` in `buildCoherentUpdateEntries` we omit components that are not existing. This is a mistake. The component can be missing if someone externally deletes that PCLQ or PCSG (right not nothing prevents that). This can then result in missing desired-replica count for that component and that will lead to `R=0` case. For missing components we should take the desired replicas from PCS spec template instead. 

**Issue-2**
PCSG and PCLQ validating webhook needs to disallow changing spec.Replicas on PCSG/PCLQ to be scaled below minAvailable.

On the suggested numAnchorBearingSteps == 0 guard and CurrentlyUpdating[].Message: rather than detect and report the degenerate state at plan time, fixes to `Issue-1` and `Issue-2` remove the possibility. With missing components sourced from the template (>= MinAvailable) and present components kept >= MinAvailable by the webhook plus the existing PCS-template validation (for immutable minAvailable), numAnchorBearingSteps is always >= 1, so there is no zero case left to surface. 

Addressed in a34183e96647b9d4b22f79d5f501ec0f2bea3605 and ea4f2344f8e87eb7a6ca2dc231d7bab80f74d703

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -116,20 +107,21 @@ Components the user has not changed in this update event are not part of the MVU
...
-| **MPG** (Minimum-Viable PodGang) | New    | Initial deployment under `Coherent`, and again at the first sub-step of every MPG-bearing step of a coherent update. | The smallest PodGang that satisfies availability — `MinAvailable` replicas of every standalone `PodClique` and every `PodCliqueScalingGroup` in scope. Plays the same role BPG used to, but is generated fresh on each MPG-bearing step rather than persisted across updates. |
-| **TPG** (Tail PodGang)           | New    | During a coherent update, alongside an MPG. | PCSG replicas above `MinAvailable` that need to roll in lockstep with the MPG. Depends on the MPG via `DependsOn`, so the scheduler places it only after the MPG is up. |
+| Role | New or legacy | When created | What it holds |
+| ---- | ------------- | ------------ | ------------- |
+| Anchor | New | At bootstrap, and at the first sub-step of every anchor-bearing step of a coherent update. | The MinAvailable replicas of every in-scope component, the standalone PodClique pod counts and the MinAvailable replica indices of every scaling group. It is the smallest PodGang that satisfies availability. |
```
renormalize: Saying `MinAvailable replicas of every in-scope component`, and then immediately saying `standalone PodClique pod counts` is slightly confusing phrasing. The former contradicts the latter.

```suggestion
| Anchor | New | At bootstrap, and at the first sub-step of every anchor-bearing step of a coherent update. | All standalone PodClique pods, and the MinAvailable replica indices of every scaling group. It is the smallest PodGang that satisfies availability. |
```

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: @sanjaychatterjee as discussed with you, i will be changing the default minAvailable for standalone PCLQ to 1 in line with what we have for PCSG. The user will have to explicitly set it to a higher value.

operator/internal/controller/podcliqueset/components/podclique/podclique.go
unmarshall: Good catch. Addressed in cbd87d8811534032f1c947b6fc5c68d65c28fae2

`getExplicitStartupDependencies` (and `identifyFullyQualifiedStartupDependencyNames`) now reads the template's `StartsAfter` instead of the live object's.  Since `StartsAfter` is immutable, it remains unchanged once created, so using the template instead of existing PCLQ yields the same single-prefixed names every reconcile.

Also added the test you asked for: `TestBuildResource_ExplicitStartsAfterStaysSinglePrefixed` reconciles a preserved replica twice under `CliqueStartupTypeExplicit` and asserts that it stays single-prefixed. It fails on the old code and passes now.

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: @yankay for your GREP on scale-in to zero (https://github.com/ai-dynamo/grove/pull/686) which allows replicas to go below minAvailable (specifically 0 only) this webhook change will disallow it. So when you start your implementation you will have to make changes to the webhook accordingly.

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: Addressed in 8b0f6cad140d93c383203ce0c292663952b1ad9d

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: You raise a good point. The motivation of coherent is to ensure that new-compatible components are recreated in lock-step and that also applies to draining older-compatible members. While we focused on the new-compatible components we did not pay attention to older-compatible members. Your comment highlights that.

Building on your proposal:
```
# ============================================================
# Inputs (per in-scope component c in S)
# ============================================================
  R[c]    # desired replicas
  M[c]    # minAvailable   (>= 1)
  U[c]    # maxUnavailable (= M[c] under a Coherent update)

# old[c] = replicas of c still on the OLD revision (starts at R[c])
# "migrate n of c" = move n replicas of c from old revision to new, then
#  wait until the new pods are Ready (the readiness gate)

# ============================================================
# Step 1 — decide how many full-MVU anchors that can be created
# (an anchor carries M[c] of EVERY component)
# ============================================================
  # K is the total number of anchor-bearing steps that can be created (this is as per the GREP)
  K = min over c in S of floor(R[c] / M[c])
  remAfterK[c] = R[c] - K * M[c]   # note: remAfterK[c] == 0 exactly when R[c] is a multiple of M[c]
   
  # This condition checks after creating K anchor PGs, if every component has something left
  # then its safe to create K anchor PGs. If even one of the components is completely consumed
  # after creating all K anchor PGs, then you end up in a situation that at least 1 component is completely 
  # updated and you are left with other components that are older versioned.  This breaks coherence for
  # older version components.
  if  remAfterK[c] > 0  for every c in S:
       # no component is fully consumed by K anchors -> open all K, hold the remainders
       numAnchors = K 
   else:
       # some component has remAfterK[c] == 0 (its replicas divide evenly by M)
       if K == 1:
           # SPECIAL CASE: remAfterK==0 with K==1 means R[c] == M[c] for that component.
           # There is only one MVU's worth of it, so it cannot be held back at all.
           # We must still open the single anchor -> it flips early (unavoidable stranding). In
           # this case we cannot prevent stranded left-over older-version components that must
           # roll without the component(s) that have already been drained/updated.
           numAnchors = 1
       else:
           # K >= 2: we can afford to drop one anchor so that component keeps a full M.
           numAnchors = K - 1   # holds a full MVU -> no stranding

   # old replicas of each component still on the OLD revision after the anchors are open
   held[c] = R[c] - numAnchors * M[c]               # >= 0 for all c

# ============================================================
# Step 2 — open the anchors (each is one gang-scheduled MVU)
# ============================================================
   repeat numAnchors times:
       for each c in S:
           update M[c] of c      # M[c] <= U[c], within MaxUnavailable
   
   # after this loop, old[c] == held[c] for every c

# ============================================================
# Step 3 — drain-together tail: bring old[c] from held[c] down to 0
# ============================================================

  # 3a. shrink every component still above its U floor, never dipping below U
  while some c in S has old[c] > U[c]:
       for each c in S with old[c] > U[c]:
           d = min( U[c], old[c] - U[c] )           # stop exactly on the U floor
           update d of c
           old[c] -= d

 # 3b. final flip: every old[c] <= U[c] now -> drain them all together
   for each c in S:
       migrate old[c] of c                          # old[c] <= U[c], within MaxUnavailable
       old[c] = 0
```

The above algorithm does away with `tailPerStep` and `stepTarget` altogether but provides a much better coherence/ up-time for compatible component-sets (old and new).

@sanjaychatterjee have a look.

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: Please see https://github.com/ai-dynamo/grove/pull/831#discussion_r4092447015
Yes, there has been a change from the original GREP. Your comment is made on the line where MVU does not have any PCSGs that have been updated. This means that only standalone PCLQs have been updated. We have never created dedicated PodGangs for standalone PCLQs. They have been part of the anchor PodGangs (previously called BPG and are same as MPGs).

operator/internal/controller/podcliquescalinggroup/components/podclique/rollingupdate.go
unmarshall: Addressed in 81d2c37fcd99b4941c3b4719d8802cd625202653

docs/proposals/393-coherent-rolling-updates/README.md
```diff
@@ -257,10 +249,14 @@ The strategy type is a single PCS-wide choice — every component rolls under th
...
+- `RollingUpdate` must not be set at all when the update strategy is `OnDelete`, which does not consume it.
+- When set, `MaxUnavailable` must be greater than 0.
+- When set, `MaxUnavailable` must not be greater than the component's `Replicas`.
+- Under the `Coherent` strategy, `MaxUnavailable` must not be less than the component's `MinAvailable`, because the MVU sub-step takes down `MinAvailable` pods of the component at once. This check is specific to `Coherent`.
+- When set, `ProgressDeadline` must be greater than 0.
 
 **Strategy-flip safety.** If the user later switches `UpdateStrategy.Type` (e.g. `RollingRecreate` → `Coherent`) without explicitly setting `MaxUnavailable`, the previously-defaulted value (e.g. `1`) sticks — the defaulting webhook only defaults *unset* fields. The `MaxUnavailable < MinAvailable` rule closes this loop: under `Coherent`, the previously-defaulted `1` is below `MinAvailable` for any component with `MinAvailable >= 2`, and admission rejects the PCS until the operator sets an appropriate value. No strategy-conditional check at runtime is needed.
```
renormalize: `operator` is read as grove-operator in this context in my opinion. This is clearer.

```suggestion
**Strategy-flip safety.** If the user later switches `UpdateStrategy.Type` (e.g. `RollingRecreate` → `Coherent`) without explicitly setting `MaxUnavailable`, the previously-defaulted value (e.g. `1`) sticks — the defaulting webhook only defaults *unset* fields. The `MaxUnavailable < MinAvailable` rule closes this loop: under `Coherent`, the previously-defaulted `1` is below `MinAvailable` for any component with `MinAvailable >= 2`, and admission rejects the PCS until the user sets an appropriate value. No strategy-conditional check at runtime is needed.
```

operator/internal/controller/podcliquescalinggroup/components/podclique/coherentupdate.go
unmarshall: Addressed in 07f265a1cbd9505ba74893cafd48e5d3ac8fd827

operator/internal/controller/podcliqueset/components/podgangmap/substepplanner.go
unmarshall: After reworking the example that your provided with current algo:
```
M -> MinAvailable
U -> MaxUnavailable

F(Frontend): 4, M=1, U=1
P(Prefill): 50, M=3, U=3
D(Decode): 70, M=4, U=4

anchorSteps = min (4/1, 50/3, 70/4) = 4
tailPerStep[F] = 0
tailPerStep[P] = 9
tailPerStep[D] = 13

stepTarget[P] = 12
stepTarget[D] = 17

leftover[P] = 2
leftover[D] = 2
```

By the time we reach step 4, the only replicas that are left in old versions for P and D are 2 each. Majority of P and D are now updated and serving. The impact of not having any old Frontend in this case is very low.

So your idea of not exhausting the lowest replica component till end is fine but this has implications.

* This will significantly change the algo if its the algo that i have described in the [comment](https://github.com/ai-dynamo/grove/pull/831#discussion_r4092447015) above. Alternative that @sanjaychatterjee  suggested is to take the last anchor (in the above scenario that would be anchor-4) towards the end after all leftover steps are done and schedule it then. This will ensure that you always have minAvailable of all components in old version available till the end. This will be an optimisation over the current approach/algo and can be taken up after careful evaluation / review.
* None of this solves the case where there is only 1 anchor. We can change your case above where Frontend is replicas: 2, minAvailable: 2. This will just create a single anchor where both pods are included. Then you are left with a large number of Prefill/Decode which are old and there is no old Frontend pod available.

So i would recommend that we go with the current algo and after careful evaluation optimise this as a followup PR.

operator/charts/templates/pclq-validating-webhook-config.yaml
unmarshall: Using `failurePolicy: Ignore` is not the correct solution as it potentially allows scaling events to be allowed during coherent update during the time the kube-apiserver is unable to reach the webhook. This can happen even after Grove pod is fully up and ready. This is a constant risk which i am not at all comfortable with.

Second suggestion is to always trust Grove operator (using the SA that Grove uses). In general that is fine but also carries a risk that i am not willing to take. If there is a bug in the operator code and that is now allowed/ignored by the webhook then we have problems.

The time it takes to upgrade to the newer version of Grove which is having these webhooks is very less. Even if webhook cannot be reached during that time i think its completely fine. When Grove is up and running it will get all events anyways.

Also this is a one time upgrade from a version that does not have pcsg and pclq validation webhooks to a version that introduces them. Henceforth all versions of Grove will have it. So for a very short rejection of UPDATE events to PCLQ/PCSG I do not recommend any change.

PR #833, opened 2026-09-17
fix: support helm version with metadata suffix

### LINKED ISSUE #832: Helm chart version with metadata suffix fails to install with invalid labels
### What happened?

```
$ yq -i '.version = "0.1.0+digest"' Chart.yaml
$ helm template . | kubectl apply --dry-run=server -f -
Error from server (Invalid): error when creating "STDIN": ServiceAccount "grove-operator" is invalid: metadata.labels: Invalid value: "grove-charts-0.1.0+digest": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
...
```

[Flux](https://fluxcd.io/flux/components/helm/options/#feature-gates) appends a chart digest by default using the metadata (+) suffix.

Successful install.

### Environment

- Kubernetes version
- Grove version
- Scheduler details

steved:
#### What type of PR is this?

/kind bug

Installing the chart where the version has a metadata suffix will fail to install due to invalid labels:
```
Error from server (Invalid): error when creating "STDIN": ServiceAccount "grove-operator" is invalid: metadata.labels: Invalid value: "grove-charts-0.1.0+digest": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

Fixes #832

_N/A_

```release-note
NONE
```


sanjaychatterjee:
@renormalize @danbar2 can you please review?

operator/charts/templates/_helpers.tpl
```diff
@@ -84,7 +84,7 @@ grove-operator-cm-{{ include "operator.config.data" . | sha256sum | trunc 8 }}
 {{- end -}}
 
 {{- define "common.chart.labels" -}}
-chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+chart: "{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}"
```
yankay: Thanks

Roughly LGTM.
Non-blocking suggestion: using `trimAll "-_."` would also cover repeated trailing hyphens and truncation ending in `.` or `_`:

```gotemplate
chart: "{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimAll "-_." }}"
```

steved: ~~`version` is semver, so only `-` is [valid](https://semver.org/#spec-item-10) and `trimSuffix` will clear the repeated chars.~~ I was wrong, fixed!

operator/charts/templates/_helpers.tpl
```diff
@@ -84,7 +84,7 @@ grove-operator-cm-{{ include "operator.config.data" . | sha256sum | trunc 8 }}
 {{- end -}}
 
 {{- define "common.chart.labels" -}}
-chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+chart: "{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-_." }}"
```
renormalize: `trimSuffix` is incorrect.

It would fail for something like `0.1.0+digest-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-extrastuff` as the version. This branch outputs:

```
❯ helm template test operator/charts | grep -m1 'chart:'
    chart: "grove-charts-0.1.0-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-"
```

The correct approach is to use `trimAll` that accepts a character set, instead of a substring the way `trimSuffix` does.

```suggestion
chart: "{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimAll "-_." }}"
```

renormalize: I see @yankay had an identical comment which I hadn't noticed till now in https://github.com/ai-dynamo/grove/pull/833#discussion_r4057017661. Perhaps the fix was lost in the rebase.

steved: Err yep, fixed. Thanks.

renormalize (approved):

unmarshall (approved):

PR #839, opened 2026-09-18
fix(migrator): migrate scheduler PodGroup membership annotations on existing Pods

### LINKED ISSUE #837: [Bug] Existing Pods retain stale Volcano PodGroup annotations after PodGang migration
### Background

Our Kubernetes cluster uses Volcano as the scheduler backend for Grove workloads.

We upgraded Grove Operator from `v0.1.0-alpha.12` to `v0.1.0-alpha.13`. The workload already existed before the upgrade:

```text
Namespace: dynamo-test
PodCliqueSet: test-grove-rolling-update
Scheduler backend: Volcano
```

The PodCliqueSet contains three PodCliques:

```text
frontend
sglangprefill
sglangdecode
```

Before upgrading the operator, we manually set the migration gate to prevent workload controllers from changing existing Pods during the migration:

```bash
NOW=$(date -u +"%Y-%m-%dT%H:%M:%SZ")

kubectl -n dynamo-test patch podcliqueset test-grove-rolling-update \
  --subresource=status \
  --type=merge \
  -p "{
    \"status\": {
      \"conditions\": [{
        \"type\": \"PodGangMigrationInProgress\",
        \"status\": \"True\",
        \"reason\": \"ManualPreUpgradeGate\",
        \"message\": \"Manual gate before upgrading to alpha.13\",
        \"lastTransitionTime\": \"${NOW}\"
      }]
    }
  }"
```

The purpose of this gate was to allow Grove to migrate the legacy PodGang resources without recreating the existing running Pods. The manual condition itself does not change PodGang membership and does not appear to be the cause of the problem.

### Simplified failure flow

```mermaid
flowchart LR
    A["alpha.12<br/>Legacy Grove PodGang<br/>Legacy Volcano PodGroup"]
    B["Upgrade to alpha.13<br/>Migrate to epoch-based PodGang"]
    C["New Grove PodGang<br/>New Volcano PodGroup"]
    D["Existing Pods remain Running<br/>Grove label updated<br/>Volcano annotations remain old"]
    E["Later configuration change<br/>creates a replacement Pod"]
    F["New Pod points to new PodGroup<br/>Existing Pods still point to old PodGroup"]
    G["Volcano sees incomplete<br/>frontend/prefill/decode subgroups"]
    H["NotEnoughPodsOfTask<br/>Replacement Pod remains Pending"]

    A --> B --> C --> D --> E --> F --> G --> H
```

### Observed behavior

During the migration, the legacy Grove PodGang was replaced by an epoch-based PodGang:

```text
Old: test-grove-rolling-update-0
New: test-grove-rolling-update-0-1789712555413903410
```

The legacy Volcano PodGroup was also replaced because it was owned by the legacy Grove PodGang.

Existing Pods were not recreated, which was the expected behavior. However, only the Grove PodGang label was migrated:

```yaml
metadata:
  labels:
    grove.io/podgang: test-grove-rolling-update-0-1789712555413903410
  annotations:
    scheduling.k8s.io/group-name: test-grove-rolling-update-0
    scheduling.volcano.sh/group-name: test-grove-rolling-update-0
```

The existing Pods continued running because they had already been scheduled and bound to nodes, so the incomplete migration was not immediately visible.

Later, we changed the frontend container environment variables, which triggered a rolling update. The newly created replacement Pod correctly referenced the new epoch-based PodGroup:

```yaml
metadata:
  labels:
    grove.io/podgang: test-grove-rolling-update-0-1789712555413903410
  annotations:
    scheduling.k8s.io/group-name: test-grove-rolling-update-0-1789712555413903410
    scheduling.volcano.sh/group-name: test-grove-rolling-update-0-1789712555413903410
```

However, the replacement Pod remained Pending. Volcano scheduler reported:

```text
Job <dynamo-test/test-grove-rolling-update-0-1789712555413903410> is not valid,
reason: NotEnoughPodsOfTask,
message: Not enough valid subGroups of each task for gang-scheduling
```

The generated Volcano PodGroup required one subgroup for each PodClique:

```yaml
spec:
  minMember: 3
  subGroupPolicy:
    - name: test-grove-rolling-update-0-frontend
      subGroupSize: 1
      minSubGroups: 1
    - name: test-grove-rolling-update-0-sglangprefill
      subGroupSize: 1
      minSubGroups: 1
    - name: test-grove-rolling-update-0-sglangdecode
      subGroupSize: 1
      minSubGroups: 1
```

Volcano determines PodGroup membership from `scheduling.k8s.io/group-name`. Because existing Pods still referenced the legacy PodGroup, Volcano initially saw only the newly created frontend Pod in the new PodGroup:

```text
frontend subgroup: 1
prefill subgroup:  0
decode subgroup:   0
```

The new Job therefore failed `CheckSubJobValid()` before entering normal node resource allocation.

### Root cause analysis

The PodGang migration currently updates only Grove labels on existing Pods.

In `operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go`, the migration changes:

```go
labels[apicommon.LabelPodGang] = targetPodGangName
delete(labels, apicommon.LabelBasePodGang)
```

but it does not update scheduler-backend-specific metadata.

New Pods do not have this problem because the normal Pod creation path invokes `backend.PreparePod(pod)`. The Volcano backend then writes:

```go
pod.Annotations[volcanov1beta1.KubeGroupNameAnnotationKey] = podGangName
pod.Annotations[volcanov1beta1.VolcanoGroupNameAnnotationKey] = podGangName
```

As a result:

- Existing Pods retain the legacy Volcano PodGroup annotations.
- New Pods use the new epoch-based Volcano PodGroup annotations.
- The new Volcano PodGroup sees an incomplete set of subgroups.
- Subsequent rolling updates can become blocked by gang scheduling.

The migration completion check also verifies only that the new Grove PodGang exists. It does not verify that the scheduler backend resource exists or that existing Pods reference the new PodGroup.

### Workaround and validation

We manually updated both Volcano annotations on all existing Pods to the new epoch-based PodGroup name:

```text
scheduling.k8s.io/group-name
scheduling.volcano.sh/group-name
```

After this update, Volcano re-associated the existing Pods with the new Job. The PodGroup recovered to:

```text
phase: Running
running: 7
Scheduled: True
```

The previously Pending replacement Pod was then scheduled successfully.

## What did you expect to happen?

The migration should preserve existing running Pods without recreating them, while also migrating their scheduler-specific PodGroup membership.

After migration, the following fields should all reference the new epoch-based PodGang/PodGroup:

```text
grove.io/podgang
scheduling.k8s.io/group-name
scheduling.volcano.sh/group-name
```

A later rolling update should be able to create and schedule replacement Pods normally.

A possible fix would be to:

1. Add a scheduler-backend migration hook for existing Pods.
2. Update both Volcano group annotations during migration.
3. Do not consider a Pod migrated based only on `grove.io/podgang`.
4. Keep the migration gate until the new scheduler backend resource exists and all existing Pods have the expected membership metadata.
5. Delete the legacy PodGang only after membership migration succeeds.
6. Add an upgrade test that verifies both zero Pod recreation during migration and successful scheduling during a subsequent rolling update.

## Environment

```text
Kubernetes version: not specified
Grove Operator before upgrade: v0.1.0-alpha.12
Grove Operator after upgrade: v0.1.0-alpha.13
Scheduler backend: Volcano
Volcano scheduler version: v1.14.2
Namespace: dynamo-test
PodCliqueSet: test-grove-rolling-update
```

The upgrade itself may appear successful because existing Pods remain Running. However, the workload is left in an inconsistent scheduler state, and a later scaling or rolling-update operation can become blocked by Volcano gang scheduling.

unmarshall:
#### What type of PR is this?

/kind bug

When a PodCliqueSet is migrated to the epoch-based PodGang scheme, the migrator rewrote only the `grove.io/podgang` label on existing Pods. It left the scheduler-backend PodGroup membership annotations untouched, so they kept pointing at the legacy PodGroup:
 - Volcano: `scheduling.k8s.io/group-name`, `scheduling.volcano.sh/group-name`
 - KAI: `pod-group-name`

New Pods get these annotations from `backend.PreparePod` during creation, but migration does not recreate Pods, so existing Pods were left stale. The new epoch-based PodGroup then saw an incomplete set of members, and a later scaling or rolling update left replacement Pods Pending with Volcano reporting `NotEnoughPodsOfTask`.

This PR adds an optional `PodGangMembershipAnnotator` capability (defined in the podgangmigrator package, satisfied structurally by the Volcano and KAI backends). During migration the migrator resolves the Pod's backend from `spec.schedulerName` and, when the backend implements the capability, refreshes the membership annotations alongside the label in a single metadata-only patch. Spec is never touched, so Pods are not recreated. Backends without Pod-level membership metadata (kube, lpx) do not implement the interface and are skipped.

**Notes on scope:**
 - The completion gate is unchanged: since migration stamps label + membership atomically per Pod before the gate is cleared, a Pod is never considered migrated on the label alone. Verified via the migrator's cached client, which WaitForCacheSync guarantees is populated with the pre-existing legacy Pods.
 - MergeFrom (not optimistic locking) is used: these are single-writer metadata keys during a gated migration, and optimistic locking would only churn against benign kubelet status updates.

Unit tests cover the stale→epoch annotation rewrite, that existing and newly-prepared Pods resolve to the same PodGroup, and that a backend without membership annotations is skipped.

Fixes #837

- Interface `PodGangMembershipAnnotator` intentionally lives in the temporary `internal/podgangmigrator` package (not the stable `scheduler.Backend` interface), so the whole capability is deleted together with the migrator. Backends satisfy it structurally, without importing it.
- A gang-scheduler upgrade e2e (Test_VUPG3: legacy KAI workload → upgrade → assert Pods not recreated + membership migrated → scale up → gang schedules) is not in this PR. The upgrade e2e cluster currently runs with KAI disabled (--set scheduler.kai.enabled=false), so it would need infra enablement (drop that flag / depend on e2e-cluster-up with KAI + queues + a KAI workload YAML). I will be tracking this as a follow-up.

```release-note
NONE
```

```docs
NONE
```


sanjaychatterjee:
@for-zero-one can you please validate on your end?

for-zero-one:
> @for-zero-one can you please validate on your end?

@sanjaychatterjee Thanks @unmarshall for the quick fix! I've verified it locally and it works as expected.

operator/internal/podgangmigrator/membership.go
```diff
@@ -0,0 +1,31 @@
...
+// scheduler-specific annotations that express PodGang membership. The migrator applies these
+// annotations to an existing Pod so it joins the epoch-based PodGroup during migration without being
+// recreated.
+//
+// The normal Pod-creation path stamps this same membership metadata via Backend.PreparePod (for
+// example Volcano's scheduling.k8s.io/group-name and scheduling.volcano.sh/group-name annotations).
+// Migration does not recreate Pods, so it refreshes the annotations in place instead. Backends that
+// carry no PodGang membership annotations on the Pod (for example the plain kube backend) do not
+// implement this interface and are skipped.
+type PodGangMembershipAnnotator interface {
```
sanjaychatterjee, line 27: 1. Can we make this interface be part of scheduler backend types? Also, if it is a temporary one, please note it as such.
2. Can we rename it to be a more generic metadata specific interface name, e.g. `PodGangMetadata`? Membership annotations could be one of many other getters.

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
```diff
@@ -186,43 +217,81 @@ func (r _resource) migratePodsOfPodClique(ctx context.Context, namespace, pclqNa
...
+	annotator, ok := r.schedRegistry.GetOrDefault(pod.Spec.SchedulerName).(podgangmigrator.PodGangMembershipAnnotator)
+	if !ok {
+		return nil
+	}
+	return annotator.PodGangMembershipAnnotations(targetPodGangName)
+}
+
+// annotationsNeedUpdate reports whether existing lacks any key of desired or maps it to a different
+// value.
+func annotationsNeedUpdate(existing, desired map[string]string) bool {
```
sanjaychatterjee, line 286: Align function names. e.g. `podGangAnnotationsNeedUpdate`

operator/internal/podgangmigrator/membership.go
sanjaychatterjee: Addressed in reviewer comments.

sanjaychatterjee (approved):

operator/internal/controller/podcliqueset/components/podgangmigrator/podgangmigrator.go
unmarshall: annotationNeedsUpdate does not check for any specific annotation hence the more generic name. We have discussed this over a call so resolving this.

renormalize (approved):
thanks!

renormalize:
/ok to test [72faf1e](https://github.com/ai-dynamo/grove/pull/839/commits/72faf1e4ccbf788de4275824545a52977a545375)

